PR #19, opened 2019-04-17
add vendor packages into main package of drone-oss-08

AkiraNorthstar:
The same as the cncd - stuff ;-)


AkiraNorthstar:
Hi @laszlocph !
Are all builds currently fail?

laszlocph:
> Are all builds currently fail?
Some unit tests deadlock nondeterministically on master. Started tracking it here: https://github.com/laszlocph/drone-oss-08/issues/20

I reran the test for this PR. It's green now.

laszlocph:
And huge <3 for making the very first PR to this project.
:rocket: :medal_sports: :)

laszlocph:
I see that you are making further changes. My reasoning behind pulling in `cncd/*` was that I planned major work on it and felt cumbersome having such a central component in a separate package.

The ones that you pull in are not that critical right now in my opinion to be pulled in. They are vendored already in `vendor/` that's good enough imo. Once we plan making changes we can pull them in.

I can be of course convinced. Best if you describe the scope of the PR and the reason behind it.

AkiraNorthstar:
Ok, I'm trying to explain.

First: did not think that the made changes also land in your codebase - **SORRY!**
I have to get used to it, since I used local gitea and drone-0.8.x and build fedora packages for myself.

- Sooner or later, these dependencies will be adjusted so that they will no longer be compatible with version 0.8.xx.
- The tests under go are also simplified, so you do not have to test external "sources" separately for function ...

The just generated pull request of the vendor packages are really not critical, I admit - Personally, I'm still a bit cautious when it comes to drone.

Recent changes to the drone license show that features simply can not be used anymore (for example: resource limitation) ...
But that's another topic.

I hope I could explain my motives to you.

Regards

laszlocph:
Thanks for your message - I didn't want to put you off, and probably I was too quick on the trigger :)

> Sooner or later, these dependencies will be adjusted so that they will no longer be compatible with version 0.8.xx.

Very good point.

> Personally, I'm still a bit cautious when it comes to drone.

Yeah, I also forked all there is. Just in case. Plus these dependencies are part of the `vendor/` folder of this repo.

> Recent changes to the drone license show that features simply can not be used anymore (for example: resource limitation) ...
But that's another topic.

I think if you use Gitea you can use the real Drone Enterprise Edition for free - as per https://discourse.drone.io/t/licensing-and-subscription-faq/3839

> The just generated pull request of the vendor packages are really not critical
I will wait a bit and decide if we merge this in a few days, or later in smaller bits.

AkiraNorthstar:
> Thanks for your message - I didn't want to put you off, and probably I was too quick on the trigger :)

That's alright for me, it's your fork of drone and asking is better than banishing ;-)
> I think if you use Gitea you can use the real Drone Enterprise Edition for free - as per https://discourse.drone.io/t/licensing-and-subscription-faq/3839

Theoretically yes, but I have not managed to limit the CPU resources so far.
I depend on it because I also share other services on the build server like plex, nfs / smb (for satellite receivers).
In short: I use a pimped file server for building packages. ;-)

laszlocph:
I don't want to leave this open longer.. I don't feel this should be merged right now as it doesn't give much benefits at this point. Does packages are quite static.

Issue #62, opened 2019-07-29
Build Patch: drone-oss-08 -> woodpecker


AkiraNorthstar:
I hope this patch helps you to build woodpecker again.
Primarily I changed the import paths.
[drone-oss-08-to-woodpecker.patch.txt](https://github.com/laszlocph/woodpecker/files/3443249/drone-oss-08-to-woodpecker.patch.txt) or #63 

Regrads,
AkiraNorthstar

laszlocph:
#65 is meant to fix this

PR #63, opened 2019-07-29
Import paths: drone-oss-08 -> woodpecker

AkiraNorthstar:
#62


laszlocph:
Thanks for this!

It was missing renaming a couple of files, so I ended up working on my own branch and merged https://github.com/laszlocph/woodpecker/pull/65

Thanks again for the PR and the patience, I just got back from vacation seeing the mistake I made :)

PR #69, opened 2019-09-14
Status update

mhmxs:
Create separate Go file for build status updates


server/build.go
```diff
@@ -275,9 +268,6 @@ func PostApproval(c *gin.Context) {
 		c.String(500, "cannot decline a build with status %s", build.Status)
 		return
 	}
-	build.Status = model.StatusPending
```
laszlocph: :heavy_check_mark:

mhmxs: Was a bit strange that the code update some status, than did some other things not related to the status change and than updated the status in the store :)

PR #73, opened 2019-10-02
Proc status update tests

mhmxs:


server/procStatus_test.go
```diff
@@ -0,0 +1,285 @@
+// Copyright 2018 Drone.IO Inc.
```
laszlocph: Are you sure of this copyright? I think something like this https://github.com/laszlocph/woodpecker/pull/69/files#diff-bc0cbeb6dabbe3062df6c0c8e94d7eb3R1 is more accurate.

mhmxs: Thanks, nice catch. I changed it, so once the build is gree, you can merge.

PR #79, opened 2019-10-20
Migrate go modules

nickborysov:
Signed-off-by: Nikolay Borysov <niks.borisov@gmail.com>


laszlocph:
Hello Nikolay,

I really appreciate the PR. The changeset is kinda large, but most of it is the deletion of `vendor/` as I see.

More important is that the test on https://cloud.drone.io/laszlocph/woodpecker/456 is failing. Do you plan to fix those?

Could you tell me how you did the migration? I'm just curiouss as I had no clue how to approach it :)

Br,
Laszlo

nickborysov:
Hello Laszlo!

It's very easy to migrate to go modules. You just need to call `go mod init` and almost everything will be configured. Also you need to change your make file and configs for CI such us .drone.yml and go.yml, because variable `GO111MODULE` has value `off` at $GOPATH/src. I would like to use $GOPATH/pkg/mod for project at CI, because it is default directory for all your dependencies with go modules. That's all. 
Also you always can always get your depencies at vendor, if call `go mod vendor`

nickborysov:
And yes, I want to fix failed test on CI. I think I need to use oldest version of some dependencies to make it work properly

nickborysov:
@laszlocph I don't khow how to fix it. I will try again tomorrow. Do you know how to fix it?

laszlocph:
Hi @nickborysov thanks for all the pointers and the effort. I ended up fixing the issue in #80. Followed the steps that you provided and hit the exact same test failure as you had. I then found the git hash for a dependency that was vendored, and was captured badly in the vendor.json.

PR #107, opened 2019-11-15
Editorconfig

laszlocph:


kevinsimper:
This is pretty cool, I don't think every project has to fall for the lowest denominator like switch to spaces just because that is what everyone else use, but that said it does make it easier.

I am a big big fan of prettier so I can only applause this PR 👏👏It saves a lot of discussions

WIth prettier the need for editorconfig is not as big, but it helps in a project.

It all looks good ✅

laszlocph:
Appreciate the feedback, this is my first JS project :angel:

PR #117, opened 2020-05-18
Add basic GCP autoscaler

imduffy15:
This autoscaler takes in the name of an instance group on GCP.

Based on the build queue it will adjust the amount of instances required in the instance group.

All agent configuration should be done via a GCP instance template.


autoscaler/providers/google/provider.go
```diff
@@ -0,0 +1,130 @@
...
+	p.service, err = compute.New(client)
+	if err != nil {
+		logrus.WithError(err).Warn("Failed to create compute client")
+		return nil, err
+	}
+
+	return p, nil
+}
+
+func (p provider) SetCapacity(n int, minimumAge time.Duration) error {
```
laszlocph, line 77: Does this account for nodes that are running a build at the point of autoscaling? I fear nodes with running builds can be killed by the scaler.

imduffy15: It may be best to leave this one unmerged or to have it marked as beta.

Yes, there is potential for a running node to be killed. This isn't great but the agents do appear to have graceful shutdown implement. However, GCP will kill an ondemand instance in 90 seconds or a preemptive one in 30 seconds so even with graceful shutdown they could still be forcibly deleted. 

In the environment I'm running this in everything is on preemptive instances so instances can disappear at any moment. The cost effectiveness of this outweights a developer sometimes being annoyed and having to hit rebuild.

laszlocph: I have an autoscaler codebase unpublished, will try to merge the two. The mentioned problem is solved there, but this implementation also have merits.

imduffy15: Oh interesting! if you push it to a branch I'll be sure to check it out.

PR #128, opened 2020-08-03
Update server-setup.md

JensHeinrich:
Add a shared socket directory instead of using an unencrypted TCP connection

### DIFF
--- a/docs/docs/server-setup.md
+++ b/docs/docs/server-setup.md
@@ -334,25 +334,30 @@ spec:
             value: woodpecker.tools.svc.cluster.local:9000
           - name: DRONE_SECRET
             value: "xxx"
-          - name: DOCKER_HOST
-            value: tcp://localhost:2375
         resources:
           limits:
             cpu: 2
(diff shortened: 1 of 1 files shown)


laszlocph:
Could you explain your approach?

JensHeinrich:
The approach uses only a unix socket via a shared volume; the DinD container starts docker and puts the control socket onto a shared volume; from here the Woodpecker container can talk to the socket, as if they were running on the same host.

So it is basically what
```
    volumes:
```
does for in the `docker-compose.yml`

laszlocph:
> The approach uses only a unix socket via a shared volume; the DinD container starts docker and puts the control socket onto a shared volume; from here the Woodpecker container can talk to the socket, as if they were running on the same host.
> 
> So it is basically what
> 
> ```
>     volumes:
>       - /var/run/docker.sock:/var/run/docker.sock
> ```
> 
> does in the `docker-compose.yml`

Makes sense! Thank you

PR #148, opened 2020-11-23
feat: global secret

CrabHo:
- List
```
drone globalsecret ls
```
- Add
```
drone globalsecret add --name foo --value bar
```
- Info
```
drone globalsecret info --name foo
```
- Update
```
drone globalsecret update --name foo --event pull_request
```
- Remove
```
drone globalsecret rm --name foo
```

### DIFF
--- a/cli/drone/globalsecret/secret.go
+++ b/cli/drone/globalsecret/secret.go
@@ -1,0 +1,16 @@
+package globalsecret
+
+import "github.com/urfave/cli"
+
+// Command exports the secret command.
+var Command = cli.Command{
+	Name:  "globalsecret",
+	Usage: "manage global secrets",
+	Subcommands: []cli.Command{
+		globalSecretCreateCmd,
+		globalSecretDeleteCmd,
+		globalSecretUpdateCmd,
+		globalSecretInfoCmd,
+		globalSecretListCmd,
+	},
+}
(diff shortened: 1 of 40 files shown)


laszlocph:
High quality PR.

Need a couple of days to crosscheck it with https://woodpecker.laszlo.cloud/environment-variables/#global-environment-variables
How the two features can co-exist.

What do you think about the feature overlap of the two?

CrabHo:
env is for organization to setup non credentials data. e.g organization root domain, jira link or other data.

secret is for admin to set token like docker username and password.

in my opinion there are different purpose.

laszlocph:
Thank you for the contributions. I'm now processing the backlog of PRs and my goal is to merge this in the coming days.

laszlocph:
@CrabHo I see that this is building on top of the original Drone global secrets feature.

The problem here is that parts of that was not licensed with Apache2. I see that you pulled in the Apache code, but haven't got to the end in figuring out if all files are Apache or not.

It would help if you write down the approach you followed with the non Apache 2 code.

Tchoupinax:
Hello folks,

What's up with this feature? :)

Moreover, are global secret shared with all repositories or only scoped with an organization?

lafriks:
@qwerty287 I think yes, my PR fully replaces this one, except for drone cli part.

lafriks:
I can implement CLI later on in other PR as I did not want to make PR too big :)

PR #169, opened 2021-03-11
fix: steps of non multi pipline is not display

CrabHo:
### Issue
the steps of non multipipeline aren’t displayed because the div is hidden

### Solution
add a condition to decide the div css class

### DIFF
--- a/web/src/screens/repo/screens/build/components/procs.js
+++ b/web/src/screens/repo/screens/build/components/procs.js
@@ -18,7 +18,7 @@ const renderEnviron = data => {
 class ProcListHolder extends Component {
   constructor(props, context) {
     super(props, context);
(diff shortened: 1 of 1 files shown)


laszlocph:
Doh, I missed this bug. Obviously only checked the multipipeline case.

laszlocph:
I have to understand the logic in the PR. Will take a day or two. Thanks for the PR!!

CrabHo:
enhance:
- display status text with pipline and matrix name
- steps div is setup to default open when there is only one process (renderName is false)

laszlocph:
This fixes the problem. I will make a few adjustments after merge to hide the group header if it's a non-multi pipeline build.

PR #174, opened 2021-03-20
Add support for path-prefix condition

alexef:
Example:
```yaml
...
when:
  path: '*.md'
```
should match only builds in which the commit added/removed or modified files with the *.md extension

This works for new builds, but not for restarts of existing builds (as I didn't add a field to the `builds` table in meddler).

If you think this makes sense, can you advise how to deal with a ddl migration?

cc: @ksvital: this might simplify a bit our pipelines


alexef:
Hey @laszlocph, do you think these changes make sense? I was able to test it end-to-end, and it also follows what GA are doing: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

laszlocph:
> Hey @laszlocph, do you think these changes make sense? I was able to test it end-to-end, and it also follows what GA are doing: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

This generally makes sense.
I was meaning to test it and propose how to persist things in `builds` table in meddler, but so far I couldn't get here.
Let me come back to you within a week, hopefully with a proper review.

inverse:
@laszlocph did you have a chance to take a look and test out properly?

alexef:
My main concern about adding it to the `builds` table is that the change list can be quite large (assume someone decide to change 1000 files at once, all in the maximum available path).

In that case, I would suggest not storing any file, and just building everything (as if all files have changed).

model/build.go
```diff
@@ -16,38 +16,39 @@ package model
...
+	Avatar       string   `json:"author_avatar" meddler:"build_avatar"`
+	Email        string   `json:"author_email"  meddler:"build_email"`
+	Link         string   `json:"link_url"      meddler:"build_link"`
+	Signed       bool     `json:"signed"        meddler:"build_signed"`   // deprecate
+	Verified     bool     `json:"verified"      meddler:"build_verified"` // deprecate
+	Reviewer     string   `json:"reviewed_by"   meddler:"build_reviewer"`
+	Reviewed     int64    `json:"reviewed_at"   meddler:"build_reviewed"`
+	Procs        []*Proc  `json:"procs,omitempty" meddler:"-"`
+	Files        []*File  `json:"files,omitempty" meddler:"-"`
+	ChangedFiles []string `json:"changed_files,omitempty" meddler:"-"`
```
alexef, line 51: Can `meddler` deal with really long lists of long file names? I don't see limits for other fields, so I'm curious how this will work.

laszlocph: We will set the column type to postgres `TEXT` that deals with any length of text.

https://github.com/woodpecker-ci/woodpecker/blob/master/store/datastore/ddl/postgres/files/003_create_table_builds.sql as I see the practice is to use varchar(x) in the codebase, but we will use text type as all three engines (postgres, sqlite3, mysel) can support that.

alexef: pushed an attempt at that

alexef:
@laszlocph any chance I can re-trigger the build of this PR? what am I missing

laszlocph:
I have to enable the build manually. The repo is set to protected and which requires a manual approval. Partly against malicious PRs and crypto miners ;)

alexef:
Any chance we can merge this?

6543:
gitea also support it with v1.14.x -> https://gitea.com/gitea/go-sdk/src/branch/release/v0.14/gitea/repo_commit.go#L50

alexef:
@laszlocph I know you are carving out of your time to test and maintain woodpecker, but what can we do to support you and see this change merged? Having it in trunk would unblock a couple of our teams waiting for this feature that will simplify their pipelines by a lot.

docs/docs/pipeline.md
```diff
@@ -384,6 +384,28 @@ when:
   instance: stage.drone.company.com
 ```
 
+Execute a step only on commit with certain files added/removed/modified:
+
+**NOTE: Feature is only available for Github repositories.**
+
+```diff
+when:
+  path: src/*
```
laszlocph, line 393: Can you make sure the syntax in this doc is up to date?

I tried both from this docs, and neither of them worked :thinking:

store/datastore/ddl/mysql/ddl_gen.go
```diff
@@ -725,3 +729,11 @@ UPDATE repos SET repo_fallback='false'
 var updateTableSetRepoFallbackAgain = `
 UPDATE repos SET repo_fallback='false'
 `
+
+//
+// 025_add_builds_changed_files_column.sql
+//
+
+var addBuildsChangedfilesColumn = `
+ALTER TABLE builds ADD COLUMN changed_files TEXT
```
laszlocph, line 738: The column should have a default value of `[]`

Otherwise existing builds are not read from the DB, as meddler can't parse empty strings to json array.

laszlocph (dismissed):
Can you provide a full yaml example?

I couldn't get the path condition to work.

docs/docs/pipeline.md
alexef: let me get my local setup working again, and will provide a working example. I haven't touched it in a while 😓

store/datastore/ddl/mysql/ddl_gen.go
alexef: the problem is that "TEXT" columns cannot have a default value: https://stackoverflow.com/questions/3466872/why-cant-a-text-column-have-a-default-value-in-mysql

Should I maybe run an update all in the migration?

alexef: I updated the code like that.

docs/docs/pipeline.md
alexef:
```suggestion
  path: "src/*"
```
the wildcard must be escaped in YAML

docs/docs/pipeline.md
```diff
@@ -384,6 +384,28 @@ when:
...
+
+Execute a step only on commit excluding certain files added/removed/modified:
+
+
+**NOTE: Feature is only available for Github repositories.**
+
+```diff
+when:
+  path:
+    exclude: *.md
```
alexef, line 404:
```suggestion
    exclude: "*.md"
```

alexef:
@laszlocph I managed to revisit this PR, build and run it on our infra.

To answer your questions:
1. here is an example pipeline that implements and respects the when path conditions:

```yaml
# This file was generated by gygdev. Do not manually modify it
pipeline:
  docs:
    image: busybox
    commands:
    - 'true'
    when:
     path: '*.md'
  all:
    image: busybox
    commands:
    - 'true'
  folder:
    image: busybox
    commands:
    - 'true'
    when:
     path: 'subfolder/*'
  no-drone:
    image: busybox
    commands:
    - 'true'
    when:
     path: 
      exclude: .drone.yml
```
A few executions of this pipeline, correctly removing the steps not affected by the changes:
Screenshot 2021-06-19 at 15 12 24
Screenshot 2021-06-19 at 15 12 18
Screenshot 2021-06-19 at 15 12 12

2. Note that when using `*` I had to escape the paths

3. I couldn't set a default in the `changed_files` column. I'm using an `UPDATE` sql statement to prevent builds not showing up. However, for an "--allow-empty" empty commit, the default value meddler puts in is the string `"null"` instead of the string `"[]"`. Should I also use that? Is there a different way to tell meddler what to use?

Screenshot 2021-06-19 at 15 12 55


Please let me know if I should make additional changes!

tboerger:
Do we really want to add this single method of changed files directly to the core? On drone we delayed that because different people had different requirements. That's why extensions had been built.

6543:
> ... different people had different requirements...

What could you make different?

tboerger:
At least the matching code, but I directly know about two extensions:

* https://github.com/meltwater/drone-convert-pathschanged
* https://github.com/microadam/drone-config-changeset-conditional

laszlocph:
Thank you @alexef for the changes. I could get it work, and feels good to use this feature!

>     2\. However, for an "--allow-empty" empty commit, the default value meddler puts in is the string `"null"` instead of the string `"[]"`. Should I also use that?

I believe we have to set an empty array manually as a null will break meddler I think.
We have to make the  `--allow-empty` work  before we merge it, otherwise I think it is good to go.

laszlocph:
> Do we really want to add this single method of changed files directly to the core? On drone we delayed that because different people had different requirements. That's why extensions had been built.

We don't have a plugin system at place, nor process or momentum to work one out in and deliver a feature like this in a 3 months timeframe in my opinion.

That's why, I prefer having momentum and merging a quality and useful feature like this that cover the naive usecases well.

Maybe I got this wrong. Work effort wise one person can really architect it and implement it in lot less time than 3 months, but I don't see that person just yet.

Long term is a different question, maybe we will have a flourishing ecosystem and an architecture that allows the drone kind of pluggablity, but right now opinions, PRs merged and momentum could mean the way forward for the project.

6543:
please resolve conflicts

tboerger:
Beside this feature I will jump into the code and check how easy or complicated it would be to carve out an extension system similar to drone 1.0

alexef:
@tboerger agree that a generic extension system would be nicer, I would be happy to help out! For now, this unblocks us.

@laszlocph I updated the code following https://apoorvam.github.io/blog/2017/golang-json-marshal-slice-as-empty-array-not-null/ . Meddler seems to deal fine with both `[]` and `null` as values stored in the TEXT field.

tboerger:
Ok, what about mentioning within the docs that this feature is experimental in terms of getting removed from core when a plugin system have been introduced? That way anybody who enables it could be aware about some potential break with some update? That way we can keep the option to put it into a plugin instead of bloating the core :)

alexef:
Okay to mark it as experimental, but not sure if the plugin approach would be the best for this specific feature.

Keeping it in core means no additional calls are made to the GitHub API (as we get the changed files in the webhook payload Github Pushes to us, and we preserve them in the db). The plugin will fetch them every time: https://github.com/meltwater/drone-convert-pathschanged/blob/f44e0009fb8da1cb33b292484ebb527d45621a56/providers/github.go#L15

I see a plugin dealing with more complex path conditions, if it can work with data the core already has. Then the plugin will actually be generic, and not depend on the git backend the core deals with (@6543 mentioned this can also be implemented for gitea - while the two plugins only support Github and Bitbucket).

tboerger:
I think that's more about some implementation detail. I could also imagine that the whole webhook payload also gets sent to the extension, than ideally it won't have to call the github or gitea api.

alexef:
that's the thing. the webhook will be specific to the git provider. so the extension would have to implement logic based on which git provider the user is using, which is suboptimal

alexef:
@6543 conflicts resolved. Any chance we can move this forward? I see Laszlo is away atm.

6543:
let me test and if all works god to go :)

cncd/pipeline/pipeline/frontend/yaml/constraint.go
```diff
@@ -163,3 +171,72 @@ func (c *ConstraintMap) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		y, _ := yaml.Marshal(value)
+		return fmt.Errorf("Could not parse condition: %s", y)
+	}
+
+	return nil
+}
+
+func (c *ConstraintPath) Match(v []string, message string) bool {
+	// ignore file pattern matches if the commit message contains all
+	if strings.Contains(message, "[ALL]") {
```
6543, line 203: I'm not sure if we should go with this hard-coded value ...

alexef: I know what you say. should I go the `[SKIP CI]` way, make it a case insensitive regex? or you don't want it at all in this iteration?
https://github.com/woodpecker-ci/woodpecker/blob/bfc4aa805926d73730cc4c7c1f529993099ef2dd/server/hook.go#L43

6543:
@alexef th the code coment :point_up: 

what about a "ingore" or "ignore_message"?

```
...yaml
when:
    path: 
        exclude: .drone.yml
        ignore_message: "[ALL]*"
```

cncd/pipeline/pipeline/frontend/yaml/constraint.go
alexef: @6543 got your comment, here is the change: https://github.com/woodpecker-ci/woodpecker/pull/174/commits/7c95fb2ad7b8be67ba7947790bd51674c86f25c4

cncd/pipeline/pipeline/frontend/yaml/constraint.go
```diff
@@ -36,6 +38,12 @@ type (
 		Include map[string]string
 		Exclude map[string]string
 	}
+
+	ConstraintPath struct {
```
6543, line 42:
```suggestion
	// ConstraintPath defines a runtime constrain for paths
	ConstraintPath struct {
```

cncd/pipeline/pipeline/frontend/yaml/constraint.go
```diff
@@ -163,3 +172,74 @@ func (c *ConstraintMap) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+	if err1 != nil && err2 != nil {
+		y, _ := yaml.Marshal(value)
+		return fmt.Errorf("Could not parse condition: %s", y)
+	}
+
+	return nil
+}
+
+func (c *ConstraintPath) Match(v []string, message string) bool {
```
6543, line 204:
```suggestion
// Match returns true if file paths in string slice matches the include and not exclude patterns
//  or if commit message contains ignore message.
func (c *ConstraintPath) Match(v []string, message string) bool {
```

inverse:
I'm guessing these docs will also need to be updated https://github.com/woodpecker-ci/woodpecker/blob/master/docs/docs/pipeline.md#conditional-step-execution

6543 (approved):

6543:
@inverse already in place :)

PR #175, opened 2021-03-23
Using configured server URL instead of obtained from request

pboguslawski:
Woodpecker should use configured server URL when creating self
URL-s for VCS UI and not depend on URL read from request. In
case of internal calls like answers for webhooks from Gitea
on backend, address in URL may be internal and this causes
Gitea to display CI status for commit with intenal URL
to woodpecker not external one.

Author-Change-Id: IB#1107596


laszlocph (approved):

laszlocph:
I agree in principle, but need to test it before I merge.

6543:
please resolve conflicts

pboguslawski:
> please resolve conflicts

Done.

6543 (approved):

PR #182, opened 2021-03-29
Incomplete Gitea repository list fixed

pboguslawski:
Woodpecker may display incomplete list of Gitea repositores user has access to because Gitea forces paginated repo list reading. 

How to reproduce: create 35 test repos in Gitea and sync it to Woodpecker - only about 30 will be displayed.

This mod fixes it.

Related: https://gitea.com/gitea/go-sdk/issues/507
Author-Change-Id: IB#1107754

### DIFF
--- a/remote/gitea/gitea.go
+++ b/remote/gitea/gitea.go
@@ -1,4 +1,5 @@
 // Copyright 2018 Drone.IO Inc.
+// Copyright 2021 Informatyka Boguslawski sp. z o.o. sp.k., http://www.ib.pl/
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -11,6 +12,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+//
+// This file has been modified by Informatyka Boguslawski sp. z o.o.
 
 package gitea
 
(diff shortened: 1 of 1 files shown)


remote/gitea/gitea.go
```diff
@@ -229,15 +232,41 @@ func (c *client) Repos(u *model.User) ([]*model.Repo, error) {
 		return nil, err
 	}
 
-	all, _, err := client.ListMyRepos(gitea.ListReposOptions{})
-	if err != nil {
-		return repos, err
-	}
+	// Gitea SDK forces us to read repo list paginated.
```
6543, line 235: This should be marked as workaround so it can be rm if it got fixed upstream

adding a `// TODO: rm if ... got solved?

*this is just my personal suggestion* :)

pboguslawski:
> This should be marked as workaround so 

This mod was designed to workaround Giteas SDK problem but also to be compatible with paging requirement after Gitea is fixed so I don't see point in marking this as workaroud.

BTW: repo list paging in Woodpecker would be nice feature but its separate topic.

remote/gitea/gitea.go
6543: ok It look's like the whole setDefaults for ListOptions in the sdk do not work as they should ...

so you should have to itterate ... but I'll make a bugfix upstream ... just update to this version after it is patchend :)

pboguslawski: Didn't see any option in Gitea API for getting all repos in 1 shot so without Gitea API changes better to use iteration now. Not sure if Gitea should be brave to offer queries without limiting number of results to avoid overloads; I can see some preallocations for max possible number of results in

https://gitea.com/gitea/go-sdk/src/branch/master/gitea/repo.go#L134

and this may hurt to set limit to high. So looping in Woodpecker seem to be fine for now.

6543 (approved):

laszlocph (approved):

6543:
ok https://gitea.com/gitea/go-sdk/pulls/510 got merged ... se we can & should update the sdk ...

6543:
with #184 you dont need the `PageSize: 50` :)

pboguslawski:
> with #184 you dont need the `PageSize: 50` :)

Seems that code

https://gitea.com/gitea/go-sdk/src/branch/master/gitea/list_options.go#L35

does not touch PageSize if it's not set explicitly (=0) and Gitea called with

https://gitea-url/api/v1/user/repos?page=1&limit=0

returns hardcoded page limit 30 so I think its better to leave explicit paging request on Woodpecker side to avoid doubts and allow Woodpecker to fetch repos with max page size allowed by Gitea SDK which above code limits to 50.

Not sure but maybe there should be 

-	if o.PageSize < 0 || o.PageSize > maxPageSize {
+ 	if o.PageSize < 1 || o.PageSize > maxPageSize {

in

https://gitea.com/gitea/go-sdk/src/branch/master/gitea/list_options.go#L35

if 0 will not be defined in Gitea API for fetching all repos in one shot (now 0 does not return all repos). If Gitea will allow to fetch all repos in one shot using PageSize=0, o.PageSize < 0 should be lef untouched.

PR #201, opened 2021-05-25
Non blocking errors

laszlocph:
This PR is fixing a multi pipeline bug when a pipeline depends on two other pipelines.

Before https://github.com/laszlocph/woodpecker/compare/non-blocking-errors?expand=1#diff-40e8ea0681d92b72e347faac0930ec8a25143c4e31389d1039cf3a5c304a1dedR360 we only set the first dependency state, and if the second dependency has a non matching status, it is simply ignored.

The rest of the PR is just test cases and logging facilities.

### DIFF
--- a/cncd/queue/fifo.go
+++ b/cncd/queue/fifo.go
@@ -356,8 +356,8 @@ func (q *fifo) updateDepStatusInQueue(taskID string, status string) {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


alexef:
fyi: @costash this should fix the "deploy running if build failed" errors

PR #205, opened 2021-05-27
Filter main branch

alexef:
One thing I find myself often doing is scrolling through the builds in order to find the latest master build.

With this new button, it becomes easier.

By default, preserve the list of all builds from all branches. When clicked, show only main/master ones.
Screenshot 2021-05-27 at 06 42 06
Screenshot 2021-05-27 at 06 42 16


laszlocph (approved):

laszlocph:
One thing to note is that this lists only the fetched builds. Have you considered the paging behavior? Moving this to the api level?

I know that that is a larger task.. so probably we merge this and extend it to the api some point.

alexef:
Yes, it would be best implemented in the api/backend. But even now, with the "show more builds" button, this is usable.

PR #206, opened 2021-05-27
Docs URL configuration

alexef:
I believe this was lost in the migration, but some parts remained in the code.


laszlocph:
> I believe this was lost in the migration, but some parts remained in the code.

I'm not sure if I follow. What migration and what remained in the code?

The PR looks good (although I'm not sure why to put this in to a configuration value)

alexef:
This is the part that I found in the code: https://github.com/woodpecker-ci/woodpecker/blob/5604ff2a43b03289b70c0c1fe6adb42eedc81deb/server/web/template.go#L108 

Having it in a config value would allow us to link our internal docs in the UI, in the face of the devs :)

Issue #209, opened 2021-05-29
Phase out environment variables prefixed DRONE_, use WOODPECKER_


marians:
Currently (as of v0.13.0), every environment variable for configuring woodpecker seems to come in two flavours: One prefixed `DRONE_` and one prefixed `WOODPECKER_`.

The coexistence of these two versions of variables with identical meaning makes documentation, help text, and error messages hard to read (see my PR https://github.com/woodpecker-ci/woodpecker/pull/208 for an example).

To improve the user experience, I recommend to make a decision regarding when to drop the `DRONE_` spelling and communicate that decision clearly.

Since woodpecker is still in the v0.x.x major version realm, I would argue such a change could happen with any minor version bump.

laszlocph:
Let's do it.

marians:
How about using the occasion for these additional clean-ups?

- Remove `DATABASE_DRIVER` in favour of `WOODPECKER_DATABASE_DRIVER`
- Remove `DATABASE_CONFIG` in favour of `WOODPECKER_DATABASE_DATASOURCE`
- Remove `WOODPECKER_HOST`in favour of `WOODPECKER_SERVER_HOST`
- Remove `WOODPECKER_SECRET` in favour of `WOODPECKER_AGENT_SECRET`

IMO it would help to remove the guesswork and make the code more maintainable.

marians:
... and rename `XDG_CACHE_HOME` to `WOODPECKER_LETS_ENCRYPT_PATH`

tboerger:
It's best practice for many projects to prefix the env variables by the project name, so let's use WOODPECKER_.

The xdg env variable should keep the name as it is.

tboerger:
XDG_CACHE_HOME is a standard env variable name from freedesktop and it's only the base path where the let's encrypt certificates are stored.

PR #211, opened 2021-05-29
Use WOODPECKER_ env vars in docs

marians:
Assuming that the `DRONE_` prefixed environment variables should at some point be phased out (see https://github.com/woodpecker-ci/woodpecker/issues/209) this PR replaces them with the `WOODPECKER_*` versions.

This PR is made in the assumption that there is a replacement for every server config variable. I have also checked a few based on the output of the server's `--help` output.

Environment variables to be used in pipeline commands as in `docs/docs/environment-variables.md` have remained unchanged. (I am not certain that there are any `WOODPECKER_` variables yet in that context.)

### DIFF
--- a/docs/docs/environment-variables.md
+++ b/docs/docs/environment-variables.md
@@ -94,10 +94,10 @@ This is the reference list of all environment variables available to your build 
 
 ## Global environment variables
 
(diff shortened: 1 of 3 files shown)


laszlocph (approved):
:clap:

PR #223, opened 2021-06-09
Gitea: Allow User Login with Token

### LINKED ISSUE #195: 2fa Gitea
Should this work with 2fa and gitea? When I try to use my gitea account I get the following error

time="2021-05-03T16:00:47Z" level=error msg="cannot authenticate user. basic auth required"

6543:
close  #195

### DIFF
--- a/remote/gitea/gitea.go
+++ b/remote/gitea/gitea.go
@@ -139,8 +139,8 @@ func (c *client) Login(res http.ResponseWriter, req *http.Request) (*model.User,
 		return nil, nil
 	}
 
-	// Create Client with Basic Auth
-	client, err := c.newClientBasicAuth(username, password)
+	// Create Client with Basic Auth or Token
+	client, err := c.newClient(username, password)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


remote/gitea/gitea.go
```diff
@@ -139,8 +139,8 @@ func (c *client) Login(res http.ResponseWriter, req *http.Request) (*model.User,
 		return nil, nil
 	}
 
-	// Create Client with Basic Auth
-	client, err := c.newClientBasicAuth(username, password)
+	// Create Client with Basic Auth or Token
```
lunny, line 142: How about move if there is a password here?

6543: where? should I put a `move`? 

sorry - I think you have to explain your idea :/

6543: where? should I put a `move`? 

sorry - I think you have to explain your idea :/ @lunny

techknowlogick:
Rather than logging in with a password, the alternative of logging in with oauth is better.

techknowlogick:
Now that OAuth support has been added, this could probably be closed.

Issue #131, opened 2020-08-18
Multiple pipelines not yet implemented for Gitea remote


PeterAlabaster:
As mentioned in [this discussion](https://github.com/laszlocph/woodpecker/pull/32#issuecomment-503545809) is yet to be implemented for remotes other than GitHub

Multi-pipeline builds as documented [here](https://woodpecker.laszlo.cloud/multi-pipeline/) appears to not be implemented on the following:
- bitbucket
- gitea
- gogs
- coding
- gerrit
- gitlab

For reference, the first implementation in GitHub is [here](https://github.com/laszlocph/woodpecker/blob/master/remote/github/github.go#L242)

taz77:
The documentation didn't specifically state this and I was banging my head on the wall trying to get this to work with Bitbucket.

ygbillet:
@laszlocph I would like to give a try to implement this for Gitea. 
I'm unfamiliar with Gitea API (and Go SDK) and Woodpecker.

Regarding `func Dir(u *model.User, r *model.Repo, b *model.Build, f string)`, f would typically have value ".drone" ?

laszlocph:
> @laszlocph I would like to give a try to implement this for Gitea.

Amazing :clap: 

> I'm unfamiliar with Gitea API (and Go SDK) and Woodpecker.
> 
> Regarding `func Dir(u *model.User, r *model.Repo, b *model.Build, f string)`, f would typically have value ".drone" ?

I believe it's the config path override. Since being an override I think it's mostly an empty string. But best to see it on your test installation in the sqlite db.

![image](https://user-images.githubusercontent.com/4289031/121307709-8e543b80-c900-11eb-9fb9-53ccbb320ea6.png)

PR #225, opened 2021-06-09
Add mutli-pipeline to Gitea

ygbillet:
from https://woodpecker.laszlo.cloud/multi-pipeline/

Signed-off-by: Yves-Gaël BILLET <ygbillet@ifotec.com>


ygbillet:
A proposition for adding multi-pipeline to Gitea.

remote/gitea/gitea.go
```diff
@@ -295,7 +297,44 @@ func (c *client) File(u *model.User, r *model.Repo, b *model.Build, f string) ([
...
+			if err != nil {
+				errors += fmt.Sprintf("cannot get %s: %s", e.Path, err)
+			}
+
+			config := remote.FileMeta{
+				Name: e.Path,
+				Data: data,
+			}
+
+			configs = append(configs, &config)
```
6543, line 329:
```suggestion
			configs = append(configs, &remote.FileMeta{
				Name: e.Path,
				Data: data,
			})
```
just a code format suggestion ...

laszlocph:
This addresses #131 for Gitea

Anyone wants to test this?
CI passing, implementation looks sane.

speatzle:
This Works for me, though it took me a while to figure out i need to set the Pipeline Path to .drone (the folder) and that if i have the pipelines .drone/.drone.yaml and .drone/.test.yaml the .drone.yml pipeline will run but the ui only shows the .test.yml one for me, is this a bug you can confirm?

laszlocph:
> i need to set the Pipeline Path to .drone (the folder) and that if i have the pipelines .drone/.drone.yaml and .drone/.test.yaml

correct

> the .drone.yml pipeline will run but the ui only shows the .test.yml one for me, is this a bug you can confirm?

You mean you have the multipipes in `.drone/` and also a `.drone.yml` in root?
Hm, not sure how that should work. Needs to be checked and documented.

speatzle:
> You mean you have the multipipes in `.drone/` and also a `.drone.yml` in root?

No i have the pipelines `.drone/.drone.yaml` and `.drone/.test.yaml` 
The ui wouldn't show me the .drone/.drone.yaml one but it would still run.

speatzle:
@laszlocph can you test whether a .drone/.drone.yml pipeline is also invisible in the UI when using github multi-pipeline?

ygbillet:
We need to reflect change in Dir() at [gitea_oauth.go](https://github.com/ygbillet/woodpecker/blob/gitea-multi-pipeline/remote/gitea/gitea_oauth.go#L282).

But we will have a code duplication. 
@6543  : do you have any insight ?

6543:
At the moment we have lot of different code and differnt code styles in remote.

In the end I'd like to refactor it to have one Interface and a base implementation, each remote do extend&overwrite.

For example: giteas webhook response have its own defined struct in woodpecker and we use the SDK, sobwe should use the Structs of the SDK since its up to date and tested ...

ygbillet:
OK. 

Do we need more work for this PR (multipipeline to GITEA) ? 
We use it at work and it will be good not to have to maintain our own version.

speatzle:
> We use it at work and it will be good not to have to maintain our own version.

We have the same situation, would be nice to have it in the latest image. i can make a new issue for the invisible pipeline problem, i belive that is a general multipipeline issue anyway and not specific to gitea.

6543:
I can merge it once I had time to give it a final "manual live" test

6543 (approved):

6543:
code is fine, and test worked too - didn't had a look int this ui issue but created an issue -> #239

@ygbillet would be nice if you could provide more info or even better dig bit into it :)

ygbillet:
Thanks !

Regarding #239, i will give it a try (i'm not the one who found this bug)

UnlimitedCookies:
I guess the documentation now needs to be amended accordingly. Currently, it still states the following:

> NOTE: Feature is only available for Github repositories. Follow this issue to support further development.
https://woodpecker.laszlo.cloud/multi-pipeline/

6543:
@UnlimitedCookies ->  #254

PR #228, opened 2021-06-18
Enforce code format

6543:


.drone.sh
```diff
@@ -12,4 +12,4 @@ fi
 echo "Building $VERSION"
```
6543, line 13: @laszlocph do we use this script somewhere ? `rg` was not able to find a ref in this codebase

laszlocph: Not anymore

laszlocph (approved):

PR #235, opened 2021-06-28
Add support to gitea remote for path-prefix condition

6543:
follow up to #174 

respect file changed information on gitea's push hook's


alexef (approved):
This was fast!

speatzle:
Nice.
The gitea webhook gives back an array. Is there a reason why this only take the zeroth commit ?

6543:
yes since you normaly run only against 1 but we could expand this of course ... this will make it a bit complicated ...

-> if you add and then delete a complete file now it will notice it, if we do it right it get ignored ...

PR #236, opened 2021-06-28
update / remove drone dependencies

6543:
* github.com/drone/signal -> single file we can just include and maintain in main repo
* github.com/drone/envsubst -> still in active development, so just updated it
* github.com/drone/expr -> stale and repo is read only, migrated last version into woodpecker org


cmd/drone-agent/signal.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2017 Drone.IO Inc.
+//
+// This work is licensed under the terms of the MIT license.
```
6543, line 3:
```suggestion
// This file is licensed under the terms of the MIT license.
```

Issue #239, opened 2021-07-05
invisible pipeline ".drone.yaml" if ".drone" multi pipeline is used


6543:
if you use a folder to store pipelines in it, and call one of them `.drone.yaml` it wont show up in ui

ygbillet:
I investigate a bit this issue.

It seems to come from [web/src/screens/repo/screens/build/components/procs.js](https://github.com/woodpecker-ci/woodpecker/blob/c699ea5789c136903a72aee935de5f51f6728c82/web/src/screens/repo/screens/build/components/procs.js#L36).

I'm not so fluent in JavaScript, so i'm not sure how to proceed.

ygbillet:
How it works today :  if the name of "proc" (see JSON in browser) is drone, we assume we are not in multipipeline mode.

Two options : 

- Instead of checking names, we could check if there are multiple procs or only one. But i don't know how to do that (i never learn React).
- Maybe we shouldn't change anything and write in the doc not to use ".drone.yml" or "drone.yml" as the name of the file. The rationale behind it : name of a file describes a "pipeline". A pipeline "drone" is not very explicit, better use : tests, lint, build, publish, docker, doc, etc.

6543:
@anbraten this one;)

PR #248, opened 2021-07-19
Replace www-path with www-proxy option for development

anbraten:
By adding a new ENV variable called `WOODPECKER_WWW_PROXY` it is possible to serve a webinterface via a proxy configured by the `WOODPECKER_WWW_PROXY` value for development instead of serving the interface from the bundled code or from some folder location as the old `WOODPECKER_WWW` option allowed. Using a proxy allows developing the UI with hot-reloading.

image


cmd/drone-server/server.go
```diff
@@ -83,13 +84,31 @@ func server(c *cli.Context) error {
...
-	// we are switching from gin to httpservermux|treemux,
-	// so if this code looks strange, that is why.
-	tree := setupTree(c)
+	proxyWebUI := c.String("www-proxy")
+
+	var webUIServe func(w http.ResponseWriter, r *http.Request)
+
+	if proxyWebUI == "" {
+		// we are switching from gin to httpservermux|treemux,
+		// so if this code looks strange, that is why.
```
6543, line 93:
```suggestion
```

6543:
I'll have a closer look & test it once v0.14.0 got released - witch is only hold back by one issue

6543:
@anbraten if you have free cycles It would be nice if you could have a look at it - I wont have time until next week :/

anbraten:
> @anbraten if you have free cycles It would be nice if you could have a look at it - I wont have time until next week :/

@6543 Do you mean this PR or #239 for 0.14.0. I can have a look at #239 anyways.
I already tested this PR and it should work without any problems.

router/router.go
```diff
@@ -49,7 +48,8 @@ func Load(mux *httptreemux.ContextMux, middleware ...gin.HandlerFunc) http.Handl
 				session.User(c),
 			),
 		)
-		mux.ServeHTTP(c.Writer, req)
+
```
6543, line 51:
```suggestion
```

cmd/drone-server/server.go
```diff
@@ -89,13 +90,30 @@ func server(c *cli.Context) error {
 	store_ := setupStore(c)
 	setupEvilGlobals(c, store_, remote_)
 
-	// we are switching from gin to httpservermux|treemux,
```
6543, line 92: ... // we are switching from gin to httpservermux|treemux, ...

please do not remove - I think we should convert it into an TODO ...

anbraten: I don't think it is a TODO, it sound more like a we are switching context from gin to httpservermux|treemux to me.

6543: to be hosntest - it's guly and we only should use one router - so use gin for all and if it does not support all there are other drop in replacements

lets leave it out I'll create an issue!

6543 (approved):

PR #257, opened 2021-07-23
Improve release workflow

anbraten:
# Changes
- [x] build docker images with kaniko (non-root)


.goreleaser.yml
```diff
@@ -0,0 +1,92 @@
...
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      # - windows
+      # - darwin
+  - id: agent
+    main: ./cmd/drone-agent
+    binary: woodpecker-agent
+    env:
+      - CGO_ENABLED=0
```
6543, line 20:
```suggestion
      - CGO_ENABLED=1
```

anbraten: @6543 With `CGO_ENABLED=1` I only get the following error: `standard_init_linux.go:228: exec user process caused: no such file or directory`. In the current makefile CGO is enabled for the server, but not for the agent 🤔

6543: well server need the cgo since we have sqlite3 as option to store things

anbraten: changed server release to use cgo with ee8fe2f74b1c6afcff8236b000ed93f80c4d695d

.woodpecker/.release.yml
```diff
@@ -0,0 +1,15 @@
+clone:
+  git:
+    image: plugins/git:next
+
+pipeline:
+  release:
+    image: golang:1.10
```
6543, line 7:
```suggestion
    image: golang:1.16
```

.woodpecker/.release.yml
```diff
@@ -0,0 +1,15 @@
+clone:
+  git:
+    image: plugins/git:next
+
+pipeline:
+  release:
+    image: golang:1.16
+    secrets: [github_token, docker_username, docker_password]
+    commands:
+      curl -sL https://git.io/goreleaser | bash
```
inverse, line 10: Wouldn't having a pinned version make more sense?

anbraten: @inverse You're right. Just copied that code from the official docs from goreleaser, but version pinning should be preferred. What do you think about using the goreleaser docker image directly instead of a golang image?

6543 (commented):
& change makefile to use goreleaser too

docker/Dockerfile.agent
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.16
```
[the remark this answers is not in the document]
anbraten: changed by 7754133832492e3c1132fb686791e2688697ca88

docker/Dockerfile.agent.alpine
```diff
@@ -0,0 +1,12 @@
+FROM alpine:3.9.4
```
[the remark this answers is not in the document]
anbraten: changed by 7754133832492e3c1132fb686791e2688697ca88

docker/Dockerfile.server
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.16
```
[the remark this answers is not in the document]
anbraten: changed by 7754133832492e3c1132fb686791e2688697ca88

docker/Dockerfile.server
```diff
@@ -0,0 +1,15 @@
...
+
+FROM scratch
+EXPOSE 8000 9000 80 443
+
+ENV DATABASE_DRIVER=sqlite3
+ENV DATABASE_CONFIG=/var/lib/drone/drone.sqlite
+ENV GODEBUG=netdns=go
+ENV XDG_CACHE_HOME /var/lib/drone
+
+COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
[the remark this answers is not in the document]
anbraten: changed by 7754133832492e3c1132fb686791e2688697ca88

.woodpecker/.release.yml
```diff
@@ -0,0 +1,15 @@
...
+    image: plugins/git:next
+
+pipeline:
+  release:
+    image: ghcr.io/goreleaser/goreleaser:v0.174.1
+    secrets: [github_token, docker_username, docker_password]
+    commands:
+      - release
+    when:
+      event: tag
```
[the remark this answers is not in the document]
anbraten: @6543 I think goreleaser uses the tag name as version for tags and bumps the version if executed on pushes based on the conventional commits since the last version? Wouldn't that break the release flow?

If I understand you correctly, you want to release snapshots / pre-releases on master and the official releases on tags?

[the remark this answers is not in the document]
anbraten: What do you think about continuous delivery with semantic-versioning based on conventional commits? The tool [go-semantic-release](https://github.com/go-semantic-release/semantic-release) looks quite nice for that. We could then stay with the current way of building docker images, binaries via the makefile and build the packages (apk, deb, rpm, ...) with the goreleasers interal tool [nfpm](https://nfpm.goreleaser.com/). As quite some features of goreleaser are only available with the pro edition, this could be a nice and free solution.

docker/Dockerfile.agent
```diff
@@ -0,0 +1,14 @@
+# use golang image to copy ssl certs later
+FROM golang:1.16
+
+FROM scratch
+ENV GODEBUG=netdns=go
+EXPOSE 3000
+HEALTHCHECK CMD ["/bin/woodpecker-agent", "ping"]
+
+# copy certs from golang:1.16 image
+COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
UnlimitedCookies, line 10:
```suggestion
# use alpine image to copy ssl certs later
FROM alpine:3.14.1
RUN apk add -U --no-cache ca-certificates

FROM scratch
ENV GODEBUG=netdns=go
EXPOSE 3000
HEALTHCHECK CMD ["/bin/woodpecker-agent", "ping"]

# copy certs from golang:1.16 image
COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
```

Wouldn't it be more consistent to also use the alpine image here for the ca-certs?

.woodpecker/.release.yml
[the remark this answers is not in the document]
anbraten: Okay I changed the release flow a bit so it should create a `next` image for pushes to master and create the appropriate github release with binaries, checksums, changelog etc for tags / releases.

docker/Dockerfile.agent
anbraten: As the scratch image has nothing to do directly with the alpine image I would stick with getting it from the go image. This is additionally a quite common practice in many projects.

UnlimitedCookies: fair enough

UnlimitedCookies:
I would like to suggest adding a `:dev` tag to Dockerhub that always includes the latest pushes to the master branch. This is very similar to how Gitea currently handles their Dockerhub releases, and it makes a lot of sense imo.

Meanwhile, the `:latest` tag should be the latest released version.

I may have overlooked it, but I couldn't find any mention of this endeavor.

anbraten:
I added such a thing already. The version / tag is created by the makefile. For pushes to master the tag is `next` and the version `next-<commit-sha>`. For git tags it  creates stable releases to `latest` and the corresponding version number.

UnlimitedCookies:
@anbraten I've got some more questions: 
The release.yml is using the docker image `anbraten/kaniko`.  What does this image do? Is the source available? (couldn't find a repo on your GitHub account) And does it perhaps make sense to have this inside the woodpecker org?

anbraten:
@UnlimitedCookies The source of the image is in this repo: release/docker/Dockerfile.kaniko-builder. It makes sense to push it under the `woodpeckerci` name. Maybe I can get it upstream into kaniko as well.

anbraten:
Just created a PR GoogleContainerTools/kaniko#1748 for it at kaniko.

UnlimitedCookies:
@anbraten Thanks for clarifying! This would also be solved by #278

6543:
ok I think it's time to resolve conflicts :dagger:

PR #279, opened 2021-08-28
fix: filter pipeline config files

### LINKED ISSUE #271: Bug: multipipeline runs on more than .yml files
If I have a file called `.tests.yml.inactive` inside my multi-pipeline folder, it gets executed as well. IMO only specific files should be executed like `.yml`, `.yaml`, whatever else is possible to be used.

anbraten:
closes #271 

# Done
- filter pipeline config folders for `.yml` and `.yaml` files
- improve `fetchConfig` tests
- update remote mock and correct wrong folder name `mock` => `mocks` to match package name
- fix: return correct filename for fallback
- improve config loading by checking if folder or not before sending api call


6543:
@anbraten 

```
--- FAIL: TestFetch (0.02s)
    --- FAIL: TestFetch/Not_found_and_disabled_fallback (0.00s)
        configFetcher_test.go:164: error fetching config File not found
[0xc00028b140 0xc00028b3e0]
FAIL
```

PR #280, opened 2021-08-28
fix: use custom config path to sanitize build names

### LINKED ISSUE #270: Bug: multi-pipeline depends_on not working with custom config-path
If I use `.woodpecker` as config_path for my repo with two pipeline files:

```yml
# .test.yml

pipeline:
  test:
    image: golang:1.16
    group: test
    commands:
      - make formatcheck
``` 

```yml
# .release.yml

pipeline:
  release:
    image: ghcr.io/goreleaser/goreleaser:v0.174.1
    secrets: [github_token, docker_username, docker_password]
    commands:
      - goreleaser release
    when:
      event: [tag, push]
      branch: master

depends_on:
  - test
  # instead the following line would work
  # - "woodpecker/.test"
``` 

This comes from a bug inside the [procBuilder](https://github.com/woodpecker-ci/woodpecker/blob/2ff91e6a055c79d913dda6e7c2774b4615659fe0/server/procBuilder.go#L361-L366) not deleting the confiured config_path value.

If the name of the `depends_on` is not found, the job simply gets removed. For a better UX it could be a better solution to throw an error.

anbraten:
closes #270

### DIFF
--- a/server/hook.go
+++ b/server/hook.go
@@ -207,7 +207,7 @@ func PostHook(c *gin.Context) {
 
 	// persist the build config for historical correctness, restarts, etc
 	for _, remoteYamlConfig := range remoteYamlConfigs {
-		_, err := findOrPersistPipelineConfig(build, remoteYamlConfig)
+		_, err := findOrPersistPipelineConfig(repo, build, remoteYamlConfig)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


anbraten:
The docs should be updated to make clear how the name for `depends_on` in multi-pipelines is generated.

6543 (approved):

PR #281, opened 2021-08-28
Remove push, tag and deployment webhook filters

### LINKED ISSUE #243: woodpecker cli to support enabling tag based builds
When we enable a repo, deployment and push events gets automatically enabled. We have automation of onboarding new microservices, where a part includes creating repo and integrate woodpecker. We drone cli to enable repos at the same time we want to enable tag based events. Is there any other way I can do this, like an API call?

### LINKED ISSUE #265: Remove repository hooks option
# Problem
Currently there is a setting per repo to listen only to specific webhooks:

Screenshot from 2021-08-18 00-33-53

I don't see a reason to have this setting. If I am not interested in webhooks like deployment then my workflow file simply does not have a deployment trigger.

Having this setting makes things just complicated and duplicates settings IMO.

# Suggestion
Remove that setting and always listen to all 4 events with the webhook.

anbraten:
closes #265 
closes #243 

# Plan
- remove `AllowPush`, `AllowTag`,  `AllowDeploy` options.
- keep `AllowPullRequest` option and disable it by default


6543:
I would leave pull-request filters but the rest can be erased

anbraten:
Sounds reasonable.

UnlimitedCookies:
> I would leave pull-request filters but the rest can be erased
> 
> -> so you can act fast on external threads etc ... (if you miss-configured your yaml ...)

Can we maybe achieve this by issuing a warning when you enable a repo that has CI for PRs enabled, so that you have to confirm it? This confirmation will also need to show when you add a `when` for PRs later on. Meaning that Woodpecker stores a variable whether PRs should trigger a CI run or not. This would take the control from the VCS webhook filter to woodpecker itself. The webhook settings would stay untouched this way.

6543:
please resolve conflicts and fix sqlite migration :)

server/repo.go
```diff
@@ -42,8 +42,7 @@ func PostRepo(c *gin.Context) {
 
 	repo.IsActive = true
 	repo.UserID = user.ID
-	if !repo.AllowPush && !repo.AllowPull && !repo.AllowDeploy && !repo.AllowTag {
-		repo.AllowPush = true
+	if !repo.AllowPull {
```
6543, line 45: This invert c0nf - can you ad a smal explanation

anbraten: I will remove it. It has been used before if all checks are disabled they assumed the user hasn't specified it in the update post and took some default values as deactivating all webhooks would disable the whole repo.

6543 (approved):

6543:
> > I would leave pull-request filters but the rest can be erased
> > -> so you can act fast on external threads etc ... (if you miss-configured your yaml ...)
> 
> Can we maybe achieve this by issuing a warning when you enable a repo that has CI for PRs enabled, so that you have to confirm it? This confirmation will also need to show when you add a `when` for PRs later on. Meaning that Woodpecker stores a variable whether PRs should trigger a CI run or not. This would take the control from the VCS webhook filter to woodpecker itself. The webhook settings would stay untouched this way.

I'm against this. the woodpecker yaml file sould be the main source of configuration (so it is versionized ...)
And from woodpecker maintainer prespective: It's redundant code logic :)

6543 (approved):

UnlimitedCookies:
Well, I agree. Then we should remove the web hook filter completely. A
warning about the implications of PR events in the docs should be
sufficient.

PR #284, opened 2021-08-29
Use go's vendoring

6543:
* exec `go mod tidy && go mod vendor` update vendor folder after changes to **go.mod**

this way we can delete some go get steps on our CI and make it more resistant against external changes made by 3rd-party's ...

* [x] change .gitignore
* [x] add vendor folder
* [x] rm go get statements


Makefile
```diff
@@ -23,9 +23,11 @@ deps:
 	go get -u golang.org/x/net/context
 	go get -u golang.org/x/net/context/ctxhttp
 	go get -u github.com/golang/protobuf/proto
-	go get -u github.com/golang/protobuf/protoc-gen-go
+	go get -d github.com/golang/protobuf/protoc-gen-go
 	go get -d docker.io/go-docker
 	go get -d github.com/jackspirou/syscerts
```
anbraten, line 28: Why do we need `go get` here at all? Could we use `go mod download` instead? 

If `go get` is needed I would suggest to use `-d` to get reproducible dependencies.

6543:
> Would it be possible to download and cache the vendor folder in CI instead of having it inside the repo to keep the repo size smaller?

that's how it's done now - and that's why it sometimes fail unrelated & storing it on an external storage miss the point either ...

anbraten:
> that's how it's done now - and that's why it sometimes fail unrelated & storing it on an external storage miss the point either ...
 Ahh okay. Sorry, now I get the whole point of this PR. 🤦🏾

PR #298, opened 2021-09-11
Issue templates

anbraten:
This PR adds some issue templates using on Githubs issue forms.

[Demo](https://github.com/anbraten/woodpecker-demo-2/issues/new/choose)

After pressing on `New Issue` the user will get following screen:
Screenshot from 2021-09-11 08-24-30

The form for a bug and feature request will look like this:
Screenshot from 2021-09-11 08-25-49
Screenshot from 2021-09-11 08-25-23


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,8 @@
+blank_issues_enabled: false
```
6543, line 1:
```suggestion
blank_issues_enabled: true
```
let the option we do nudge user already ...

6543 (approved):
lgtm

PR #299, opened 2021-09-11
Change pipeline config path resolution

### LINKED ISSUE #133: Use .woodpecker.yml as default pipeline config file
As the current .drone.yml files are not supported by woodpecker, it seems to useful to switch to  supporting .woodpecker.yml as a default configuration source

anbraten:
# Config resolution
- pipeline-config setting empty / not specified (default): `.woodpecker/` => `.woodpecker.yml` => `.drone.yml`
- pipeline-config setting defined by user: try that file / folder and no fallback (if a user sets some special value that is normally done for some reason)

# Changes
- pipeline-config setting will be empty by default
- remove fallback setting for config loading (simplifies config)

---
closes #133


anbraten:
@6543 Do you know if I need to add a migration to remove the Fallback Flag from database?

6543:
If I find some time I'll look into this :)

6543 (approved):

PR #305, opened 2021-09-14
Add podman backend

### LINKED ISSUE #85: Support "native" podman as backend
Hello Laszlo!
Is it possible to integrate podman in woodpecker?

Podman does not run as a daemon (like docker : /var/run/docker.sock) but is fully compatible with the docker command line.

Another point is that podman can also mount secrets via [mount.conf](https://github.com/containers/libpod/blob/master/install.md#mountsconf), then you maybe have also solved [Global Secrets](https://github.com/laszlocph/woodpecker/issues/7).

Also cgroups v2 is native supported by podman.

Further information about podman:
- https://podman.io/
- https://github.com/containers/libpod

mfulz:
closes #85 

This is a first working podman backend, which can be used by calling the agent with --use-podman.

It is working on my environment fine for now, but would need broader testing.

This is a early version and cannot be called stable at all for now, but it makes sense to include it eary in my opinion.


6543:
A hint: We do use vendoring yoz should do:

```sh
go mod vendor
go mod tidy
```and commit result.

**Comment by 6543** [comment]
ok problem is - we do use very outdated dependencys ... - upgrade of them need to be in an own pull

**Comment by mfulz** [comment]
Yep I've tried it but there seem to be very outdated deps inside :)

**Comment by 6543** [comment]
It will take me some time but I'll do migrate them one by one ... - you can just merge master into from time to time to check if it will start working untill it do ;)

**Comment by mfulz** [comment]
> It will take me some time but I'll do migrate them one by one ... - you can just merge master into from time to time to check if it will start working untill it do ;)

Great I'lll work further on the implementation in between as well

**Comment by 6543** [comment]
@mfulz now it is building - I think you can move forward :)

File: agent/runner.go [hunk 1]
```diff
@@ -60,7 +60,7 @@ func NewRunner(workEngine rpc.Peer, f rpc.Filter, h string, state *State, backen
 	}
 }
 
-func (r *Runner) Run(ctx context.Context) error {
+func (r *Runner) Run(ctx context.Context, usePodman bool) error {
```
6543, line 63: instead of a pdman flag we should add a option struct

something like`type RunOptions struct {...}` in .../runner package

mfulz: Great will work on this during this feature

anbraten: @mfulz I merged master and adjusted the code a bit so it should now work with #463.

pat-s:
Is this in essence compatible with `next` or would it need a complete rewrite? (assuming it was based off from `0.15.x` initially).

6543:
It targets the main branch ... so its next what ever version next is going to be when it's finished :)

The interface for backends did not changed that mouch ... at all anyway

pat-s:
@mfulz Are you still using your interface or WP at all? Curious how close this is to be "production ready", also given how long this PR is already open :)

mfulz:
> @mfulz Are you still using your interface or WP at all? Curious how close this is to be "production ready", also given how long this PR is already open :)

Sorry for the silence atm this is stalled for me, as I'm loaded with too much other things, that cannot be postponed :(
I was just checking wp out and drafted the podman stuff and definitely go to proceed with this.

But when this will happen is impossible to tell atm. could be end of this year or next year - sorry for that

mfulz:
At least I can fix the errors for the checks if that would help.
But I cant access the details on what is failing for the test, can someone help me out here?

6543:
Well then just let it open as is ... open pulls per se dont hurt.

Others can base new pulls on it or this gets ready ...

pat-s:
Both the underlying WP types changed on a major level as well as the podman pkg (v3/v4). This needs a full rewrite, hence I am closing here.

PR #322, opened 2021-09-18
docs(status): rewrite

catdevnull:
status: changed doesn't exist in the current codebase and the
description was just confusing.


docs/docs/20-usage/20-pipeline-syntax.md
```diff
@@ -243,18 +243,7 @@ when:
...
-when:
-  status: changed
-```
-
-Woodpecker uses the container exit code to determine the success or failure status of a build. Non-zero exit codes fail the build and cause the pipeline to immediately exit.
-
-There are use cases for executing pipeline steps on failure, such as sending notifications for failed builds. Use the status constraint to override the default behavior and execute steps even when the build status is failure:
-
-Execute a step when the build is passing or failing:
+There are use cases for executing pipeline steps on failure, such as sending notifications for failed builds. Use the status constraint to override the default behavior and execute steps even when the build fails:
```
anbraten, line 246:
```suggestion
There are use cases for executing pipeline steps on failure, such as sending notifications for failed pipelines. Use the status constraint to execute steps even when the pipeline fails:
```

anbraten (approved):

PR #328, opened 2021-09-19
Fix for gitea token/secret length

zloidemon:


anbraten:
@zloidemon Could you explain that the error was and why `TEXT` hasn't worked for you?

6543:
can you describe what issue it solves - best case a smal how to reproduce

zloidemon:
@anbraten @6543 After update to latest version gitea and use oauth, gitea returns long token in my case it was >800 characters. DB schema of the table has varchar(500), I've updated it to varchar(1000).

anbraten:
> @anbraten @6543 After update to latest version gitea and use oauth, gitea returns long token in my case it was >800 characters. DB schema of the table has varchar(500), I updated it to varchar(1000) to solve the problem.

Ahh okay for mysql and postgres it is `Varchar(500)` and sqlite it uses `Text`. Thank you for the explanation.

anbraten (approved):

6543:
I'll add it ...

6543:
> I'll add it ...

ok togo does not allow noop ...

PR #330, opened 2021-09-20
Rename cmd agent and server folders and binaries

anbraten:
Renamed `cmd/drone-agent` to `cmd/agent` and `cmd/drone-server` to `cmd/server` and binaries to get rid of the drone name.


Makefile
```diff
@@ -91,5 +91,5 @@ release-cli:
 release: release-agent release-server
 
 install:
```
6543, line 93: I think we could delete install target!

anbraten: removed it and renamed binaries to match new folder path

anbraten:
> What is the /changelog binary you added? Is that a mistake?

Upps. That was by mistake 🙈

6543 (approved):

PR #337, opened 2021-09-21
Refactor: Reorganize into server/{api,grpc,shared} packages

cognifloyd:
Reorganize server/* so that the purpose is evident in the package structure.

- json/rest-ish APIs are in `server/api`
- grpc server (for agent communication) is in `server/grpc`
- `server.Config` is in `server/config.go` as it is used by both `server/api` and `server/grpc`
- misc utils (used by `server/api` and `server/grpc`) are in `server/shared`
- drop droneserver import


cognifloyd:
K. I'm satisfied with this now. :)

cognifloyd:
We could also move:
- `server/debug` to `server/api/debug`
- `server/metrics` to `server/api/metrics`

Does that make sense / would that be a good change?

cmd/server/server.go
```diff
@@ -40,7 +40,8 @@ import (
 	"github.com/woodpecker-ci/woodpecker/remote"
 	"github.com/woodpecker-ci/woodpecker/router"
 	"github.com/woodpecker-ci/woodpecker/router/middleware"
-	droneserver "github.com/woodpecker-ci/woodpecker/server"
+	woodpecker "github.com/woodpecker-ci/woodpecker/server"
```
anbraten, line 43: We should leave the import as `server` to match the way we import it in all other files.

server/helpers/procBuilder.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package server
+package helpers
```
anbraten, line 15: Most part of this package seem to be more than "just" helpers to me 🤔

cmd/server/server.go
cognifloyd: Hmm. I assumed this was imported as droneserver because it was in a server.go file. I'll have to give that a try.

server/helpers/procBuilder.go
cognifloyd: Yeah. They're used by both `server/api` and `server/grpc`. Some provide logic around updating models. Others have core logic, but they don't directly serve api or grpc, so I wanted to put them somewhere else to make navigating the API code more straight forward.

Do you have any suggestions for a better name or a better way to organize?

server/helpers/procBuilder.go
```diff
@@ -31,10 +31,11 @@ import (
 	"github.com/woodpecker-ci/woodpecker/cncd/pipeline/pipeline/frontend/yaml/matrix"
 	"github.com/woodpecker-ci/woodpecker/model"
 	"github.com/woodpecker-ci/woodpecker/remote"
+	"github.com/woodpecker-ci/woodpecker/server"
 )
 
-// Takes the hook data and the yaml and returns in internal data model
-type procBuilder struct {
+// ProcBuilder Takes the hook data and the yaml and returns in helpers data model
```
cognifloyd, line 37:
```suggestion
// ProcBuilder Takes the hook data and the yaml and returns in internal data model
```
Oops. I tried internal first then went with helpers, but renaming internal to helpers was a little too aggressive.

server/helpers/procBuilder.go
cognifloyd: K. I moved `server/helpers` to `server/shared`

cmd/server/server.go
```diff
@@ -40,7 +40,8 @@ import (
 	"github.com/woodpecker-ci/woodpecker/remote"
 	"github.com/woodpecker-ci/woodpecker/router"
 	"github.com/woodpecker-ci/woodpecker/router/middleware"
-	droneserver "github.com/woodpecker-ci/woodpecker/server"
+	"github.com/woodpecker-ci/woodpecker/server"
+	woodpeckerGrpcServer "github.com/woodpecker-ci/woodpecker/server/grpc"
```
anbraten, line 44: Could this be just `grpc`?

cmd/server/setup.go
```diff
@@ -35,7 +35,7 @@ import (
 	"github.com/woodpecker-ci/woodpecker/remote/gitlab"
 	"github.com/woodpecker-ci/woodpecker/remote/gitlab3"
 	"github.com/woodpecker-ci/woodpecker/remote/gogs"
-	droneserver "github.com/woodpecker-ci/woodpecker/server"
+	woodpecker "github.com/woodpecker-ci/woodpecker/server"
```
anbraten, line 38: This could be just `server`?

anbraten (commented):
just two more naming question

cmd/server/setup.go
cognifloyd: yep. done

cmd/server/server.go
cognifloyd: No. We're already importing `"google.golang.org/grpc"` so we need to disambiguate somehow.

cognifloyd: `grpcServer` was also not available since we're using that var. We could call it `grpcImplementation` or `realGrpcServer`.

6543: this soulds like we should at some point move `"google.golang.org/grpc"` into our own grpc moduel ??

for now please leave it - this would be more than just code move and split I guess

cognifloyd: Or we could play with the purpose of it. This grpc server is for communication between the woodpecker-server and woodpecker-agent. Maybe `agentGrpc`?

anbraten: We can just leave it for now. At some point we need to refactor the var names of this file containing drone anyways.

anbraten (approved):

PR #338, opened 2021-09-22
Refactor: Create agent/ package for backend agnostic logic

cognifloyd:
I'm reviewing what it takes to add a new backend (eg for kubernetes).

This is a little bit of the refactoring based on what I saw in the kube branch, but without any reference to kubernetes.
Unlike the kube branch, I called the package `agent/` instead of `runner/`. That fits the pattern of keeping server related stuff under `server/`, cli related stuff under `cli/`, and now agent-related stuff under `agent/`.

For reference, see these commits:
7f4a0c8 Factor into Runner
f7fe9ed Using Runner in server


agent/stats.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2018 Drone.IO Inc.
```
anbraten, line 1: Shouldn't this file be called `state`?

cognifloyd: I copied the name from those commits. state makes more sense, so I'll move it momentarily.

anbraten (approved):

agent/runner.go
```diff
@@ -0,0 +1,363 @@
...
+	"github.com/woodpecker-ci/woodpecker/cncd/pipeline/pipeline/multipart"
+	"github.com/woodpecker-ci/woodpecker/cncd/pipeline/pipeline/rpc"
+
+	"github.com/rs/zerolog/log"
+	"github.com/tevino/abool"
+)
+
+// NOTE we need to limit the size of the logs and files that we upload.
+// The maximum grpc payload size is 4194304. So until we implement streaming
+// for uploads, we need to set these limits below the maximum.
```
6543, line 39:
```suggestion
// TODO: Implement log streaming.
// Until now we need to limit the size of the logs and files that we upload.
// The maximum grpc payload size is 4194304. So we need to set these limits below the maximum.
```

6543 (approved):

PR #348, opened 2021-09-24
Cleanup Code

6543:
* Fix "Empty slice declaration using a literal"
* Fix "collides with imported package name"
* Remove unused code in pipeline
* Remove unused oauth2.providerAuthHeaderWorks()
* Add TODOs
* Format Code
* Cleanup doublestar import
* Migrate deprecated functions


server/store/datastore/init.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !cgo
```
6543, line 15: INFO: this is used by newer go versions - we keep the old one too to be backwards compatible

server/shared/configFetcher.go
```diff
@@ -26,6 +26,7 @@ func NewConfigFetcher(remote remote.Remote, user *model.User, repo *model.Repo,
 
 func (cf *configFetcher) Fetch() (files []*remote.FileMeta, err error) {
 	var file []byte
+	cf.repo.Config = strings.TrimSpace(cf.repo.Config)
```
anbraten, line 29: This should be done at in post func saving the string to the database.

anbraten (commented):
Could we introduce some kind of linter like golangci to automatically check most of your changes.

pipeline/frontend/yaml/constraint.go
```diff
@@ -5,7 +5,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	doublestar "github.com/bmatcuk/doublestar/v4"
+	"github.com/bmatcuk/doublestar/v4"
```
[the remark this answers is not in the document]
6543: yes go do understand versions

server/shared/configFetcher.go
6543: oh sorry did add int unintentional should be in a extra pull :) - will rm it

anbraten (approved):

PR #353, opened 2021-09-24
Improve ci tests

anbraten:


Makefile
```diff
@@ -39,19 +39,30 @@ vet:
...
+	(cd web/; yarn lesshint)
+	(cd web/; yarn lint --quiet)
+
 test-agent:
-	$(DOCKER_RUN) go test -race -timeout 30s github.com/woodpecker-ci/woodpecker/cmd/agent $(GO_PACKAGES)
+	$(DOCKER_RUN) go test -race -timeout 30s github.com/woodpecker-ci/woodpecker/cmd/agent github.com/woodpecker-ci/woodpecker/agent/...
 
 test-server:
-	$(DOCKER_RUN) go test -race -timeout 30s github.com/woodpecker-ci/woodpecker/cmd/server
+	$(DOCKER_RUN) go test -race -timeout 30s github.com/woodpecker-ci/woodpecker/cmd/server $(shell go list github.com/woodpecker-ci/woodpecker/server/... | grep -v 'store/')
```
6543, line 51:
```suggestion
	$(DOCKER_RUN) go test -race -timeout 30s github.com/woodpecker-ci/woodpecker/cmd/server $(shell go list github.com/woodpecker-ci/woodpecker/server/... | grep -v '/store')
```

6543:
`test` do run  but `main` dont on pull (and so there are no builds run - I'd like to see pulls build too - so we see if it fails)

anbraten:
> `test` do run but `main` dont on pull (and so there are no builds run - I'd like to see pulls build too - so we see if it fails)

Ahh I think that is because I added `depends_on: [ test ]` to main.yml, but the CI is still running on some quite old version which does not have #270 & #326.

6543:
yes from code perspective I was going to lgtm it but having a final look at the ci and found that issue

we need to upgrade our ci! cc @laszlocph

6543 (approved):

PR #354, opened 2021-09-24
If config is no dir assume its a folder

6543:
If config is no dir assume its a folder (user do not always add / suffix)


anbraten (approved):

6543:
Sorry did messed up the title :/

PR #357, opened 2021-09-25
Make configFetcher fallback work for gitea

### LINKED ISSUE #133: Use .woodpecker.yml as default pipeline config file

6543:
#299 who closed #133 did not take into account, that that gitea (and eventually) other forges do return 200 and empty string if file was not found - this make configFetcher more resilient

### DIFF
--- a/server/shared/configFetcher.go
+++ b/server/shared/configFetcher.go
@@ -1,6 +1,7 @@
 package shared
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
@@ -35,7 +36,7 @@ func (cf *configFetcher) Fetch() (files []*remote.FileMeta, err error) {
 				// either a file
 				if !strings.HasSuffix(config, "/") {
 					file, err = cf.remote_.File(cf.user, cf.repo, cf.build, config)
-					if err == nil {
+					if err == nil && len(file) != 0 {
 						return []*remote.FileMeta{{
 							Name: cf.repo.Config,
 							Data: file,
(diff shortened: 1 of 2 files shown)


server/shared/configFetcher.go
```diff
@@ -54,25 +55,29 @@ func (cf *configFetcher) Fetch() (files []*remote.FileMeta, err error) {
...
+				if err == nil && len(file) != 0 {
 					return []*remote.FileMeta{{
 						Name: ".drone.yml",
 						Data: file,
 					}}, nil
 				}
+
+				if err == nil && len(files) == 0 {
+					return nil, fmt.Errorf("ConfigFetcher: Fallback did not found config")
+				}
```
anbraten, line 80:
```suggestion
			  return nil, fmt.Errorf("ConfigFetcher: Fallback did not found config")
```

Shouldn't this be enough?

6543: well then we would drop the err if there is one

6543: if we simplify we otherwise would have to drop retry func

anbraten: > if we simplify we otherwise would have to drop retry func

Upps, we currently don't use the retry anyways as of `return nil, err` in line 83.

Adding some separate tests for empty directories and empty file contents would be nice.

server/shared/configFetcher_test.go
```diff
@@ -192,22 +247,11 @@ func TestFetch(t *testing.T) {
...
-				var receivedFileNames []string
-				for _, file := range files {
-					receivedFileNames = append(receivedFileNames, file.Name)
-				}
-				t.Fatal("expected some other pipeline files", tt.expectedFileNames, receivedFileNames)
+			matchingFiles := make([]string, len(files))
+			for i := range files {
+				matchingFiles[i] = files[i].Name
 			}
+			assert.ElementsMatch(t, tt.expectedFileNames, matchingFiles, "expected some other pipeline files")
```
anbraten, line 254: nice improvement 😉

server/shared/configFetcher.go
```diff
@@ -54,25 +55,30 @@ func (cf *configFetcher) Fetch() (files []*remote.FileMeta, err error) {
...
 			}
 
 			return nil, err
```
anbraten, line 84: I think this line is stopping the retry loop. Should we just remove the retry loop?

6543: can we leave it for now?

I like to add logging (more precice one!) - if we do have that we can drop it

anbraten: We can just leave it as it is now and add a TODO, saying the retry loop is inactive and could maybe be deleted in future.

6543, line 84:
```suggestion
			// TODO: retry loop is inactive and could maybe be fixed/deleted
			return nil, err
```

anbraten (approved):

PR #358, opened 2021-09-25
Refactor Gitlab Remote

### LINKED ISSUE #285: Refactor gitlab remote
- remove gitlab v3 api support as it has been removed with gitlab 11 https://docs.gitlab.com/ee/api/v3_to_v4.html
- use some library similar to github and gitea remotes ([go-gitlab](https://github.com/xanzy/go-gitlab) is great)
- support `Dir` function for multi-pipelines

6543:
close #285

* [x] Replace custom client
* [x] Update Docs
* [x] Test if it works
* [x] Update Tests


server/remote/gitlab/gitlab.go
```diff
@@ -17,33 +17,51 @@ package gitlab
...
 
-	"github.com/woodpecker-ci/woodpecker/server/remote/gitlab/client"
+	"github.com/xanzy/go-gitlab"
 )
 
-const DefaultScope = "api"
+const (
+	defaultScope  = "api"
+	perPage       = 100
+	statusContext = "ci/drone"
```
anbraten, line 39: Can we change this to `ci/woodpecker` or is it matched by Gitlab to this `SetDroneCIService`?

6543: Yes gitlab does not have a Service for woodpecker, we could use generic webhooks instead

6543: If desired, I would do a followup pull

anbraten: Okay, we need to test if we can just change it in a follow-up or if we need to add a Woodpecker-service to gitlab / use generic webhooks.

anbraten (approved):

PR #364, opened 2021-09-26
Use moby definitions for docker pipeline backend

6543:
* use moby as per https://github.com/docker/go-docker/issues/21
* migrate deprecated functions


pipeline/backend/docker/docker.go
```diff
@@ -8,10 +8,10 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/network"
 	"github.com/docker/docker/api/types/volume"
-	"github.com/docker/docker/client"
-	"github.com/docker/docker/pkg/jsonmessage"
-	"github.com/docker/docker/pkg/stdcopy"
-	"github.com/docker/docker/pkg/term"
+	"github.com/moby/moby/client"
+	"github.com/moby/moby/pkg/jsonmessage"
```
anbraten, line 12: Isn't moby providing types as well? Why do we need both libraries? Could you explain the difference I didn't really understand the difference.

6543: It's technically the same librarie - only how go do handle it differs

anbraten: Okay I see `github.com/docker/docker/api/types/volume` can not be changed to `github.com/moby/moby/api/types/volume`

anbraten (approved):

PR #366, opened 2021-09-27
Refactor: move model/ to server/model/

6543:


anbraten:
Shouldn't we try to replace drone-go types with models and therefore leave the models at some central point?

6543:
the drone-go is the web sdk and I think we should differ between database backend and rest api structs

anbraten (approved):

PR #371, opened 2021-09-28
Pass context.Context down

6543:
this will make it possible to cancel requests for top and trace processes ...

Guide: https://golangbyexample.com/using-context-in-golang-complete-guide/
Ref: https://pkg.go.dev/context


server/remote/remote.go
```diff
@@ -14,63 +14,61 @@
 
 package remote
 
-//go:generate mockery -name Remote -output mock -case=underscore
+//go:generate mockery -name Remote -output mocks -case=underscore
```
6543, line 17: fix go generate

server/remote/bitbucketserver/internal/client.go
```diff
@@ -233,9 +241,22 @@ func (c *Client) GetHooks(owner string, name string) (*HookSettings, error) {
 
 //TODO: make these as as general do with the action
 
+//Helper function to help create get
+func (c *Client) doGet(url string) (*http.Response, error) {
+	request, err := http.NewRequestWithContext(c.ctx, "PUT", url, nil)
```
anbraten, line 246:
```suggestion
	request, err := http.NewRequestWithContext(c.ctx, "GET", url, nil)
```

server/remote/coding/coding.go
```diff
@@ -145,7 +145,7 @@ func (c *Coding) Refresh(u *model.User) (bool, error) {
 }
 
 // Teams fetches a list of team memberships from the remote system.
-func (c *Coding) Teams(u *model.User) ([]*model.Team, error) {
+func (c *Coding) Teams(ctx context.Context, u *model.User) ([]*model.Team, error) {
 	// EMPTY: not implemented in Coding OAuth API
 	return nil, nil
```
anbraten, line 150:
```suggestion
	return nil, fmt.Errorf("Not implemented")
```

server/remote/gitlab3/gitlab_test.go
```diff
@@ -49,15 +49,15 @@ func Test_Gitlab(t *testing.T) {
 		g.Describe("AllProjects", func() {
 			g.It("Should return only non-archived projects is hidden", func() {
 				gitlab.HideArchives = true
-				_projects, err := gitlab.Repos(&user)
+				_projects, err := gitlab.Repos(nil, &user)
```
anbraten, line 52: Why don't you use `ctx.Background()` in this file?

server/shared/configFetcher.go
```diff
@@ -27,81 +29,90 @@ func NewConfigFetcher(remote remote.Remote, user *model.User, repo *model.Repo,
...
+	logrus.Tracef("ConfigFetch[%s]: user did not defined own config follow default procedure", cf.repo.FullName)
+	// no user defined config so try .woodpecker/*.yml -> .woodpecker.yml -> .drone.yml
+
+	// test .woodpecker/ folder
+	// if folder is not supported we will get a "Not implemented" error and continue
+	config = ".woodpecker"
+	files, err := cf.remote_.Dir(ctx, cf.user, cf.repo, cf.build, config)
+	logrus.Tracef("ConfigFetch[%s]: found %d files in '%s'", cf.repo.FullName, len(files), config)
+	files = filterPipelineFiles(files)
+	if err == nil && len(files) != 0 {
```
anbraten, line 86:
```suggestion
	files = filterPipelineFiles(files)
	if err == nil && len(files) != 0 {
          logrus.Tracef("ConfigFetch[%s]: found %d files in '%s'", cf.repo.FullName, len(files), config)
```

server/remote/gitlab3/gitlab_test.go
6543: because gitlab3 get deleted soon

server/remote/bitbucketserver/internal/client.go
6543: Good catch missed that ... (copy/paste issue)

6543:
for gitlab3 see  #358 it will be removed ...
for gitlab I did not pass it down since I would have to extend the custom client just to throw that change after #358 got merged

so if  #358 lands first I'll add it here else it's the other way arround

anbraten (approved):

PR #373, opened 2021-09-28
Remove unsupported pipeline file extension  "yaml" 

anbraten:
Docs, `sanitizePath` and json-schema do not mention / support `.yaml` (only `.yml`) files so they should simply be ignored.


6543:
I think we can add it to the docs instead ...

6543:
and in terms of speed we could use a precompiled regex Match

anbraten:
> I think we can add it to the docs instead ...

`woodpecker.yaml` or `drone.yaml` doesn't work as well, so I don't see a point in supporting it.

6543 (approved):

PR #379, opened 2021-09-29
rm web/dist/dist_gen.go as per gitignore

6543:
looking forward do #353 ... ;)


anbraten:
what do you think about #382 as an alternative?

Issue #234, opened 2021-06-28
move from meddler to xorm (switch orm)


6543:
move from **https://github.com/russross/meddler** to **https://xorm.io/**

-> get rid of sql file generation
-> make code easyer
-> add support for more databases (i.e. mssql, oracle, ...)
-> easier migrations

anbraten:
After working with the server code and the database now a bit I found quite some a lot of unused parts. Their are quite some columns not being used at all. Using a more advanced orm would probably help a lot with that problem.

catdevnull:
Wouldn't it be better to identify the issues currently and try to fix them directly before switching tools?

6543:
well the issue is: db is done manually by writing sql for each db ...

6543:
side note: https://github.com/bradrydzewski/togo is unmaintained ...

6543:
TODO:
 * [ ] migrate db-migration-framework
 * [ ] unify store things
 * [ ] dedub code

PR #382, opened 2021-09-29
Use go embed for web files and remove httptreemux

### LINKED ISSUE #308: [Refactor] we should only use one router framework
we only should use one router - so use gin for all and if it does not support all there are other drop in replacements

(currently gin & httpservermux|treemux are used)

_Originally posted by @6543 in https://github.com/woodpecker-ci/woodpecker/pull/248#discussion_r709457468_

anbraten:
closes #308 
- replace togo with go embed
- replace httptreemux with gin


6543:
tools.go can be deleted ...

anbraten:
> tools.go can be deleted ...

I think the ddl files for sql migrations still need togo.

6543:
yes you are right until we did #234

6543 (approved):

PR #394, opened 2021-10-02
Move envsubset into repo

anbraten:


6543:
You have to make sure, that files have license header - its MIT not apache as the rest

PR #403, opened 2021-10-04
Add changelog

anbraten:


6543:
I'm personally against this since adding changelog entry do mostly encounter manual polishing

PR #405, opened 2021-10-04
Release apk, deb, rpm bundles

anbraten:


6543:
let's move on but keep it in mind :)

PR #414, opened 2021-10-08
V0.14.2 fix

alexef:
Ran `go mod vendor` to pass the tests


6543:
thanks but  #415 will do it :)

PR #416, opened 2021-10-08
Fix various typos

luzpaz:
Found via `codespell -q 3 -S vendor -L pullrequest,pullrequests`


6543:
I wonder if we could add a "CI lint step using codespell"?

anbraten:
> I wonder if we could add a "CI lint step using codespell"?

Its absolutely doable. We have that at the company I work at and it is just awesome 🤩

PR #427, opened 2021-10-10
Replace DRONE_ with CI_ variables in pipeline steps

anbraten:
# Changes
Dropped support for `DRONE_*` environment variables in pipeline steps. Pipeline meta-data can be accessed with `CI_*` variables.
  - `CI_*` prefix replaces `DRONE_*`
  - `CI` value is now `woodpecker`
  - `DRONE=true` has been removed

# Follow ups
- add `CI_AGENT_*` variables
- improve agent (cli exec) arch detection
- improve tracer environment updating


pipeline/frontend/metadata.go
```diff
@@ -18,22 +19,23 @@ const (
...
-		Prev Build  `json:"prev,omitempty"`
-		Job  Job    `json:"job,omitempty"`
-		Sys  System `json:"sys,omitempty"`
+		ID    string `json:"id,omitempty"`
+		Repo  Repo   `json:"repo,omitempty"`
+		Curr  Build  `json:"curr,omitempty"`
+		Prev  Build  `json:"prev,omitempty"`
+		Job   Job    `json:"job,omitempty"`
+		Sys   System `json:"sys,omitempty"`
+		Agent Agent  `json:"agent,omitempty"`
```
6543, line 28: need own pull i would say

anbraten: reverted those changes for now

docs/docs/91-migrations.md
```diff
@@ -12,15 +12,36 @@ Some versions need some changes to the server configuration or the pipeline conf
...
-- From version `0.15.0` ongoing there will be three types of docker images: `latest`, `next` and `x.x.x` with an alpine variant for each type like `latest-alpine`.
-  If you used `latest` before to try pre-release features you should switch to `next` after this release.
+- Dropped support for `DRONE_*` environment variables in pipeline steps. Pipeline meta-data can be accessed with `CI_*` variables.
+  - `CI_*` prefix replaces `DRONE_*`
+  - `CI` value is now `woodpecker`
+  - `DRONE=true` has been removed
+  - Some variables got deprecated and will be removed in future versions. Please migrate to the new names. Same applies for `DRONE_` of them.
+    - CI_ARCH => use CI_SYSTEM_ARCH
+    - CI_COMMIT => CI_COMMIT_SHA
+    - CI_REMOTE_URL => use CI_REPO_REMOTE
```
6543, line 32: as i understand the pull we still have no DRONE_ fallback

that would make all existing plugins using them useless!

6543: we dont have to promote them and I would add them in a seperate func (even seperated in a own go file) but It should be there

pipeline/pipeline.go
```diff
@@ -127,6 +128,13 @@ func (r *Runtime) exec(proc *backend.Step) error {
 		}
 	}
 
+	// TODO: using DRONE_ will be deprecated with 0.15.0. remove fallback with following release
```
anbraten, line 131: @6543 This is now generating the `DRONE_` environment variables based on the `CI_` versions by simply replacing the prefixes. This way seemed to be much easier to maintain IMO.

anbraten: I placed it directly before executing the pipeline step, so every environment variable will be caught (so it works for cli this way as well).

6543: oh! - ok I'll review tomorow

agent/runner.go
```diff
@@ -276,26 +276,18 @@ func (r *Runner) Run(ctx context.Context) error {
 			state.Pipeline.Step.Environment = map[string]string{}
 		}
 
-		state.Pipeline.Step.Environment["DRONE_MACHINE"] = r.hostname
+		state.Pipeline.Step.Environment["CI_MACHINE"] = r.hostname
 		state.Pipeline.Step.Environment["CI_BUILD_STATUS"] = "success"
 		state.Pipeline.Step.Environment["CI_BUILD_STARTED"] = strconv.FormatInt(state.Pipeline.Time, 10)
 		state.Pipeline.Step.Environment["CI_BUILD_FINISHED"] = strconv.FormatInt(time.Now().Unix(), 10)
-		state.Pipeline.Step.Environment["DRONE_BUILD_STATUS"] = "success"
```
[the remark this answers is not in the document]
anbraten: FYI: From my research it seems that these `state....Environment[...]` variables are only used for internal state checks and have nothing to do with environment variables in the common sense (those changes are not applied to any containers as env vars). If I am correct we could refactor them to be a special state struct with variables like started, finished and status.

6543 (approved):
well yes move on :)

Issue #393, opened 2021-10-02
Breaking changes before releasing 0.15.0


anbraten:
To have less breaking changes in the future to not annoy users again and again, breaking changes should be part of the next version.

# Some ideas
- [x] replace `DRONE_` env vars with `WOODPECKER_` prefix #218
- [x] drop Gitea basic auth support #365
- [x] change config path resolution  (`.woodpecker/*.ymll` -> `.woodpecker.yml` -> `.drone.yml`) #299 
- [x] replace repository hooks setting with only allow PRs #281
- [ ] ~~migrate to cert-magic #360~~
- [x] remove `DRONE_` env vars passed to pipeline steps (only use `CI_`) #427 
- [x] rename default paths `/var/lib/drone` => `/var/lib/woodpecker` and `/drone` => `/woodpecker` #431 
- [x] rename default datasource: `/var/lib/drone/drone.sqlite` => `/var/lib/woodpecker/woodpecker.sqlite` #438 
- [x] rename prometheus namespace `drone` => `woodpecker` #439 
- pipeline config changes
  - [x] move plugin-settings in pipeline steps to some sub-setting like: `pipeline.[step].settings` and pass all KV options as `PLUGIN_` #464 
  - [ ] remove support for `root.pipeline.[step].commands` string values in favor of array lists
  - [x] ~~move `root.skip_clone` under `root.clone.skip`~~
  - [ ] ~~move `root.platform`, `root.branches`, `root.labels`,  `root.runs_on`,  `root.depends_on` to `root.when.xxx` to match step filters~~ #283 -> next release
  - [ ] ~~make `root.when` and `root.pipeline.[step].when` an array of the current with items using the current settings to allow ORs~~ non breaking & can move into next release
  - [ ] ~~remove `root.Cache` and functionality in favor of plugins?~~ - need more discussion

ivictbor:
Very good point, this will also allow woodpecker users to create some public docs / bloogs / tootorials that might be compatible in future, at least for 0.15.* according to https://semver.org/

Issue #410, opened 2021-10-06
make status context equal 


6543:
make context status equal for all forges

`woodpecker/${event}[/${pipeline}]`

PR #431, opened 2021-10-12
Change default paths to woodpecker

jolheiser:
Change default paths to `woodpecker` counterparts

This should check off some boxes on #393

### DIFF
--- a/.woodpecker/main.yml
+++ b/.woodpecker/main.yml
@@ -9,6 +9,7 @@ pipeline:
     commands:
       - make test
       - make vet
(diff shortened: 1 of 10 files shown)


anbraten:
Nice changes. It would probably best if you could split changes for path changes and Prometheus. Changing the gitlab-context could be done in combination with #410. Separate PRs should be preferred to get own changelog entries and a descriptive commit history. As those renaming changes are all breaking changes I think we should add migration notices to the docs as well. Feel free to ask if you need some help.

PR #440, opened 2021-10-13
Replace debug with log-level flag

jolheiser:
Adds log level support

### DIFF
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -58,12 +58,12 @@ func loop(c *cli.Context) error {
 		)
 	}
 
-	logLevel := zerolog.WarnLevel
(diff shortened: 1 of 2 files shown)


jolheiser:
Alright, I've removed the API endpoints and will add them in another PR. 
I've also changed the flags a bit to go in order.

The `debug` flag is different between agent/server, it's a `boolt` (bool flag defaulting to true) in the agent, but just a normal bool flag in the server.

The logging level starts out as `warn` (the previous default), moves to `debug` if the `debug` flag is `true` (and sends a deprecation warning if the flag was actually set), and finally passes through the new `log-level` flag, setting the global log level after those steps are complete.

jolheiser:
Yeah, fine with me. I wanted to limit the call site, but this is probably clearer. 🙂

PR #444, opened 2021-10-13
Add log level API

jolheiser:
A couple things to note:
* I added `zerolog` to the CLI so parsing (and possible error handling) can happen in the CLI instead of wasting the roundtrip to the server
   * Since the dep was added, I opted to change the output to use zerolog as well, let me know if you want it changed back
* I don't love the CLI implementation, willing to gather feedback. Particularly setting level with a flag, but I also didn't particularly like the alternative of checking for first arg either

It's a little awkward to think about log level because it can get conflated with "build logs" rather than "application logging".

### DIFF
--- a/cli/loglevel/loglevel.go
+++ b/cli/loglevel/loglevel.go
@@ -11,15 +11,10 @@ import (
 
 // Command exports the log-level command used to change the servers log-level.
 var Command = cli.Command{
-	Name:   "log-level",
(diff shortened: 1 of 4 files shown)


jolheiser:
1. Moved the route to `/api/log-level`.
2. Fixed error handling in the CLI.
3. Changed the CLI so that no args gets the log level, otherwise first arg sets the log level.

PR #448, opened 2021-10-15
add search to docs

anbraten:
Peek 2021-10-15 12-40


anbraten:
> do we need some cookie banner since algolia is an external service?

Good question. From my check it doesn't seem to use cookies, but local-storage to store old searches locally. From what I know local-storage does not need a banner as it isn't send to the server. So "only" request to their api will be done.

PR #449, opened 2021-10-15
Switch default git plugin

6543:


jolheiser:
This looks fine to me, just want to verify that we want to stick with the naming before we merge. 
Drone has a separate docker namespace for plugins, do we want the same?

6543:
well they just got lucky and did got the `plugins`... org - I can not create a second org for free .. so we should think it it is worth to pay docker hub to be able to get a second org & also what it's name should be

UnlimitedCookies:
I would argue that this is a "plugin" that is needed for the core functionality of woodpecker and therefore should be in the main organization. All other "non-essential" plugins should be in a second repo where the community is able to submit plugins and stuff (If that is feasible; maybe there's a way around this 1 free org restriction).

jolheiser:
To be fair, "plugins" are just docker images. There's nothing inherently special about them.

As this is a special case, I agree with it remaining in the normal namespace.
Any "official" plugins can probably live there.

PR #451, opened 2021-10-16
Add plugin marketplace (for official plugins)

### LINKED ISSUE #315: Own marketplace for plugins
Since it looks like drone do aboanden its marketpalce (did not even have ssl!) we should consider to provide a own one with more communety interaction (it should be easy to create and publish plugins ...)

tld: let the **community do** plugins 

_Originally posted by @anbraten in https://github.com/woodpecker-ci/woodpecker/pull/313#discussion_r711311964_

anbraten:
closes #315 

Peek 2021-10-16 18-22

List of Plugins is fetched on each build of the documentation.
Plugins are search by this query string: 
```
// search for repos in woodpecker-ci org with the topic: woodpecker-plugin including forks

org:woodpecker-ci topic:woodpecker-plugin fork:true
```
https://github.com/search?q=org%3Awoodpecker-ci+filename%3Adocs.md+WOODPECKER_PLUGIN_DOCS

You can find a sample documentation for a plugin here: https://github.com/woodpecker-ci/plugin-demo/blob/main/docs.md

# Preview
https://extra-small-jelly.surge.sh/plugins


jolheiser:
It would be nice if we could include community plugins, but with this it would need to exist in our organization, right?

anbraten:
> It would be nice if we could include community plugins, but with this it would need to exist in our organization, right?

I just tried it and it seems to be possible that we create a project in the woodpecker org (we could create an issue form for request for new plugins) and invite users to a signle repo to be either maintainers or even to be just a user with write permissions to it:

Screenshot from 2021-10-17 01-29-28

If that doesn't works in the future we could start using the `woodpecker-plugins` org I claimed.

6543:
I'm for adding plugins on request (new issue) to the main org and add already plugun-maintaining-people as maintainer to the repo or directly to org maintainer team if they are ok with it

6543:
That way it is clear what an "official plugin" is and what lives outside of our view ...

6543:
What about:

"if you manage a woodpecker-ci plugin and want to see it become official, just open an issue and apply for it. You wont loose write acces etc ... and can manage things as wanted, just some securety measurements are applyed, like a CI that has to be passes until you can merge to protected branch."

?

UnlimitedCookies:
You will not catch all cases of malicious intentions with a CI. 
Woodpecker maintainers could be considered trustworthy. Isn't it feasible that changes are proposed through pull requests and a maintainer of Woodpecker just has to accept it. I think that there won't be too many changes once a plugin is fully working.
Plugins should be located at the new `woodpecker-plugins` org so that this one is not getting cluttered.

6543:
of course CI wont catch a malicious request - but i think brain.sh can do the rest (to accept/reject a request) ;) ?

> new woodpecker-plugins org so that this one is not getting cluttered.

since we used repo pinning ... what's the point of cluttered?

anbraten:
I just added an issue form:

Screenshot from 2021-10-17 02-57-35

And linked it as last element in the plugin index:

Screenshot from 2021-10-17 03-21-51

UnlimitedCookies:
> of course CI wont catch a malicious request - but i think brain.sh can do the rest (to accept/reject a request) ;) ?

That's what I implied. :D  Having certain lint checks is certainly desirable.  It just sounded to me like the CI should check for security implications / malicious code. The PR process is definitely the only way to enforce this, as you described.

> since we used repo pinning ... what's the point of cluttered?

Right, it just seems nicer to have them separated. There's probably a reason that the drone guys thought that it should be separated. This also would enable making certain people plugin-only-maintainers and stuff like that later down the road.

jolheiser:
I still somewhat like the idea of allowing people to keep their own repo while still contributing a plugin.
Maybe sort of like an awesome list, we have a repo that people pr to add their plugin?
And regardless we need a notice that plugins outside our organization are not holding us liable. 

Allowing plugins  into our organization, even if someone else is a maintainer, seems like a potential liability to me.

jolheiser:
Considering that this is now for official plugins, I think we don't need to look for a magical `WOODPECKER_PLUGIN_DOCS` string in a `docs.md` file.

I do like the idea of `docs.md` for easily updating the docs without a PR to the index itself, however imo a topic would be easier to reason about. 
Even in our own org we can easily "toggle" plugins by changing their topic.

jolheiser:
> 
> 
> We need `fork: true` it only means include forks (not search only for forks). Tested it with plugin-demo (no fork) and plugin-s3, plugin-git (both forks from drone plugins) and seems to work correctly.

Ah, my bad. Seems like an ambiguous query on GHs part. 
https://docs.github.com/en/search-github/searching-on-github/searching-in-forks
`fork:true` _includes_ forks (like you've done
`fork:only` is for _only_ forks 🤦

6543:
@jolheiser @anbraten 
a. I did add logging back  - it was usefully while looking at this pull ...

b. if i understand it right, we only hit github once per doc-build, so that should be ok?

PR #455, opened 2021-10-18
Backport `TestPipelineName` for procBuilder

ctbur:
Backport #461 

Fixes a bug for `v0.14` where pipelines that use `depends_on` were not built in multi-pipeline builds. The problem is that pipelines names keep a leading `'/'` when the pipeline path does not have a trailing `'/'`.


ctbur:
Test is in https://github.com/woodpecker-ci/woodpecker/pull/461/files

PR #495, opened 2021-10-29
Embedding libcompose types for yaml parsing

6543:
since github.com/docker/libcompose deprecated and unmaintained (+archived).

and license is the same as woodpecker's, we can just copy stuff into woodpecker directly.
(we only use types of that project anyway)


pipeline/frontend/yaml/types/network.go
```diff
@@ -29,7 +28,7 @@ func (n Networks) MarshalYAML() (interface{}, error) {
 	return m, nil
 }
 
-// UnmarshalYAML implements the Unmarshaller interface.
+// UnmarshalYAML implements the Unmarshaler interface.
```
anbraten, line 31:
```suggestion
// UnmarshalYAML implements the Unmarshaller interface.
```

pipeline/frontend/yaml/types/command.go
```diff
@@ -1,17 +1,18 @@
...
 	"github.com/docker/docker/api/types/strslice"
+	// TODO: search for maintained
 	"github.com/flynn/go-shlex"
 )
 
 // Command represents a docker command, can be a string or an array of strings.
 type Command strslice.StrSlice
 
-// UnmarshalYAML implements the Unmarshaller interface.
+// UnmarshalYAML implements the Unmarshaler interface.
```
anbraten, line 15: I would keep the old spelling.
```suggestion
// UnmarshalYAML implements the Unmarshaller interface.
```

anbraten (approved):

6543:
@anbraten its not sow wording change, the interface is named that way: https://github.com/go-yaml/yaml/blob/496545a6307b/yaml.go#L36

PR #496, opened 2021-10-30
Fix bug where db file is posible moved outside of docker volume

6543:
https://github.com/woodpecker-ci/woodpecker/pull/494#discussion_r739373503 introduced a bug, where a migration function can remove the sqlite3 file outside of the mounted docker volume.
that would result in a data lose after a container recreate.

this fix it by only rename the file if in same folder else just use the old path as fallback and put warnings into the log

### DIFF
--- a/cmd/server/setup.go
+++ b/cmd/server/setup.go
@@ -97,7 +97,7 @@ func fallbackSqlite3File(path string) (string, error) {
 		return "", err
 	}
 	if err == nil {
-		return path, nil
+		return dockerDefaultPath, nil
 	}
 
 	// file is in new folder but not renamed
--- a/cmd/server/setup.go
+++ b/cmd/server/setup.go
@@ -86,7 +86,7 @@ func fallbackSqlite3File(path string) (string, error) {
 	}
 	if err == nil {
 		// rename in same folder should be fine as it should be same docker volume
-		log.Warn().Msgf("found sqlite3 file at '%s', rename to '%s'", standaloneOld, standaloneDefault)
+		log.Warn().Msgf("found sqlite3 file at '%s' and moved to '%s'", standaloneOld, standaloneDefault)
 		return standaloneDefault, os.Rename(standaloneOld, standaloneDefault)
 	}
 
@@ -106,14 +106,14 @@ func fallbackSqlite3File(path string) (string, error) {
 	}
 	if err == nil {
 		// rename in same folder should be fine as it should be same docker volume
-		log.Warn().Msgf("found sqlite3 file at '%s', rename to '%s'", dockerDefaultDir, dockerDefaultPath)
+		log.Warn().Msgf("found sqlite3 file at '%s' and moved to '%s'", dockerDefaultDir, dockerDefaultPath)
 		return dockerDefaultPath, os.Rename(dockerDefaultDir, dockerDefaultPath)
 	}
 
 	// file is still at old location
 	_, err = os.Stat(dockerOldPath)
 	if err == nil {
-		log.Error().Msgf("found sqlite3 file at deprecated path '%s', please migrate to '%s'", dockerOldPath, dockerDefaultPath)
+		log.Error().Msgf("found sqlite3 file at deprecated path '%s', please move it to '%s' and update your volume path if necessary", dockerOldPath, dockerDefaultPath)
 		return dockerOldPath, nil
 	}
 
--- a/cmd/server/setup.go
+++ b/cmd/server/setup.go
@@ -89,6 +89,15 @@ func fallbackSqlite3File(path string) (string, error) {
 		return dockerDefaultPath, nil
 	}
 
+	// file is at new default("woodpecker.sqlite")
+	_, err = os.Stat(standaloneDefault)
+	if err != nil && !os.IsNotExist(err) {
+		return "", err
+	}
+	if err == nil {
+		return standaloneDefault, nil
+	}
+
 	// woodpecker run in standalone mode, file is in same folder but not renamed
 	_, err = os.Stat(standaloneOld)
 	if err != nil && !os.IsNotExist(err) {
--- a/cmd/server/setup.go
+++ b/cmd/server/setup.go
@@ -72,7 +72,7 @@ func fallbackSqlite3File(path string) (string, error) {
 	const defaultDir = "/var/lib/woodpecker/drone.sqlite"
 	const oldPath = "/var/lib/drone/drone.sqlite"
 
-	// file is in new default("/var/lib/woodpecker/woodpecker.sqlite") / custom location
+	// file is at new default("/var/lib/woodpecker/woodpecker.sqlite") / custom location
 	_, err := os.Stat(path)
 	if err != nil && !os.IsNotExist(err) {
 		return "", err
@@ -91,13 +91,13 @@ func fallbackSqlite3File(path string) (string, error) {
 		return defaultPath, os.Rename(defaultDir, defaultPath)
 	}
 
-	// file is still old location
+	// file is still at old location
 	_, err = os.Stat(oldPath)
 	if err == nil {
 		return oldPath, nil
 	}
 
-	// file do not exist at all, use default / custom location
+	// file does not exist at all => use default / custom location
 	return path, nil
 }
 
--- a/cmd/server/setup.go
+++ b/cmd/server/setup.go
@@ -68,12 +68,30 @@ func setupStore(c *cli.Context) (store.Store, error) {
 
 // TODO: Remove this once we are sure users aren't attempting to migrate from Drone to Woodpecker (possibly never)
 func fallbackSqlite3File(path string) (string, error) {
-	const defaultPath = "/var/lib/woodpecker/woodpecker.sqlite"
-	const defaultDir = "/var/lib/woodpecker/drone.sqlite"
-	const oldPath = "/var/lib/drone/drone.sqlite"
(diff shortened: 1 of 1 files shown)


cmd/server/setup.go
```diff
@@ -46,41 +47,58 @@ import (
...
+	}
+	if err == nil {
+		// rename in same folder should be fine as it should be same docker volume
+		return defaultPath, os.Rename(defaultDir, defaultPath)
+	}
+
+	// file is still old location
+	_, err = os.Stat(oldPath)
+	if err == nil {
+		return oldPath, nil
```
anbraten, line 97: By this line `drone.sqlite` location wont be migrated, which was the whole idea of the original PRs. What about existing and showing a message to the user to move the file and update the volume if necessary.

cmd/server/setup.go
```diff
@@ -46,41 +47,58 @@ import (
...
 	}
 
-	return nil
+	// file do not exist at all, use default / custom location
```
anbraten, line 100:
```suggestion
	// file does not exist at all => use default / custom location
```

cmd/server/setup.go
```diff
@@ -46,41 +47,58 @@ import (
...
+	_, err = os.Stat(defaultDir)
+	if err != nil && !os.IsNotExist(err) {
+		return "", err
+	}
+	if err == nil {
+		// rename in same folder should be fine as it should be same docker volume
+		return defaultPath, os.Rename(defaultDir, defaultPath)
+	}
+
+	// file is still old location
```
anbraten, line 94:
```suggestion
	// file is still at old location
```

cmd/server/setup.go
```diff
@@ -46,41 +47,58 @@ import (
...
-		} else if !os.IsNotExist(err) {
-			return err
-		}
+// TODO: Remove this once we are sure users aren't attempting to migrate from Drone to Woodpecker (possibly never)
+func fallbackSqlite3File(path string) (string, error) {
+	const defaultPath = "/var/lib/woodpecker/woodpecker.sqlite"
+	const defaultDir = "/var/lib/woodpecker/drone.sqlite"
+	const oldPath = "/var/lib/drone/drone.sqlite"
+
+	// file is in new default("/var/lib/woodpecker/woodpecker.sqlite") / custom location
```
anbraten, line 75:
```suggestion
	// file is at new default("/var/lib/woodpecker/woodpecker.sqlite") / custom location
```

cmd/server/setup.go
```diff
@@ -46,41 +47,58 @@ import (
...
-	// default config for standalone installations
-	if c.String("datasource") == "woodpecker.sqlite" {
-		_, err := os.Stat("drone.sqlite")
-		if err == nil {
-			return os.Rename("drone.sqlite", "woodpecker.sqlite")
-		} else if err != nil && !os.IsNotExist(err) {
-			return err
-		}
+	// file is in new folder but not renamed
+	_, err = os.Stat(defaultDir)
```
anbraten, line 85: How should the database with is old name land at the new path (apart from moving it by hand to that location)?

cmd/server/setup.go
```diff
@@ -46,41 +47,58 @@ import (
...
-		if err == nil {
-			return os.Rename("/var/lib/drone/drone.sqlite", "/var/lib/woodpecker/woodpecker.sqlite")
-		} else if !os.IsNotExist(err) {
-			return err
-		}
+// TODO: Remove this once we are sure users aren't attempting to migrate from Drone to Woodpecker (possibly never)
+func fallbackSqlite3File(path string) (string, error) {
+	const defaultPath = "/var/lib/woodpecker/woodpecker.sqlite"
+	const defaultDir = "/var/lib/woodpecker/drone.sqlite"
+	const oldPath = "/var/lib/drone/drone.sqlite"
```
anbraten, line 73: I think this is currently missing the none-docker case where the location is `$(pwd)/drone.sqlite`.

cmd/server/setup.go
```diff
@@ -82,22 +100,25 @@ func fallbackSqlite3File(path string) (string, error) {
...
 	// file is in new folder but not renamed
-	_, err = os.Stat(defaultDir)
+	_, err = os.Stat(dockerDefaultDir)
 	if err != nil && !os.IsNotExist(err) {
 		return "", err
 	}
 	if err == nil {
 		// rename in same folder should be fine as it should be same docker volume
-		return defaultPath, os.Rename(defaultDir, defaultPath)
+		log.Warn().Msgf("found sqlite3 file at '%s', rename to '%s'", dockerDefaultDir, dockerDefaultPath)
```
anbraten, line 109:
```suggestion
		log.Warn().Msgf("found sqlite3 file at '%s' and moved to '%s'", dockerDefaultDir, dockerDefaultPath)
```

cmd/server/setup.go
```diff
@@ -68,12 +68,30 @@ func setupStore(c *cli.Context) (store.Store, error) {
...
+	}
+
+	// woodpecker run in standalone mode, file is in same folder but not renamed
+	_, err := os.Stat(standaloneOld)
+	if err != nil && !os.IsNotExist(err) {
+		return "", err
+	}
+	if err == nil {
+		// rename in same folder should be fine as it should be same docker volume
+		log.Warn().Msgf("found sqlite3 file at '%s', rename to '%s'", standaloneOld, standaloneDefault)
```
anbraten, line 89:
```suggestion
		log.Warn().Msgf("found sqlite3 file at '%s' and moved to '%s'", standaloneOld, standaloneDefault)
```

cmd/server/setup.go
```diff
@@ -82,22 +100,25 @@ func fallbackSqlite3File(path string) (string, error) {
...
+		log.Warn().Msgf("found sqlite3 file at '%s', rename to '%s'", dockerDefaultDir, dockerDefaultPath)
+		return dockerDefaultPath, os.Rename(dockerDefaultDir, dockerDefaultPath)
 	}
 
 	// file is still at old location
-	_, err = os.Stat(oldPath)
+	_, err = os.Stat(dockerOldPath)
 	if err == nil {
-		return oldPath, nil
+		log.Error().Msgf("found sqlite3 file at deprecated path '%s', please migrate to '%s'", dockerOldPath, dockerDefaultPath)
```
anbraten, line 116: Could we exit the server here?

anbraten:
```suggestion
		log.Error().Msgf("found sqlite3 file at deprecated path '%s', please move it to '%s' and update your volume path if necessary", dockerOldPath, dockerDefaultPath)
```

anbraten: > Could we exit the server here?

@6543 What do you think about existing / stopping the server if the user has a database at the old docker path?

6543: I would do that in the next version and remove fallback in v0.17

cmd/server/setup.go
```diff
@@ -46,41 +47,79 @@ import (
...
+	if err == nil {
+		// rename in same folder should be fine as it should be same docker volume
+		log.Warn().Msgf("found sqlite3 file at '%s' and moved to '%s'", dockerDefaultDir, dockerDefaultPath)
+		return dockerDefaultPath, os.Rename(dockerDefaultDir, dockerDefaultPath)
+	}
+
+	// file is still at old location
+	_, err = os.Stat(dockerOldPath)
+	if err == nil {
+		log.Error().Msgf("found sqlite3 file at deprecated path '%s', please move it to '%s' and update your volume path if necessary", dockerOldPath, dockerDefaultPath)
```
anbraten, line 117:
```suggestion
    // TODO: use log.Fatal()... in next version
		log.Error().Msgf("found sqlite3 file at deprecated path '%s', please move it to '%s' and update your volume path if necessary", dockerOldPath, dockerDefaultPath)
```

cmd/server/setup.go
6543: ajusted comment

anbraten (dismissed):

cmd/server/setup.go
```diff
@@ -46,41 +47,81 @@ import (
...
+		return standaloneDefault, os.Rename(standaloneOld, standaloneDefault)
+	}
+
+	// file is at new default("/var/lib/woodpecker/woodpecker.sqlite")
+	_, err = os.Stat(dockerDefaultPath)
+	if err != nil && !os.IsNotExist(err) {
+		return "", err
+	}
+	if err == nil {
+		return path, nil
```
6543, line 100:
```suggestion
		return dockerDefaultPath, nil
```

cmd/server/setup.go
```diff
@@ -46,41 +47,81 @@ import (
...
+	}
+
+	// file is at new default("/var/lib/woodpecker/woodpecker.sqlite")
+	_, err = os.Stat(dockerDefaultPath)
+	if err != nil && !os.IsNotExist(err) {
+		return "", err
+	}
+	if err == nil {
+		return dockerDefaultPath, nil
+	}
```
anbraten, line 101: This one could be in front of `standaloneOld` to skip one check after migration. Do we need a similar check to `dockerDefaultPath` for `standaloneDefault`? 🤔

6543: uh think so :/

6543: done

anbraten (approved):
crazy how complex such a simple looking thing can be 😅

6543:
well ... yes - this will be just 3 checks once we do fail hard (no fallback) in next version :)

Issue #435, opened 2021-10-12
Add golangci-lint


anbraten:
> That's a think golangci-lint suggests as well, right? Maybe we should add it?

I'm not sure if it checks filenames, but it does check package/struct names. 
Either way, I agree with adding it. It has some good linters (and some bad ones) which can be useful.

_Originally posted by @jolheiser in https://github.com/woodpecker-ci/woodpecker/issues/434#issuecomment-941292203_

PR #502, opened 2021-11-05
Add golangci-lint

lukashass:
Initial part of #435 
Remove Makefile tasks for `govet` and  `gofmt` since they are included in golangci-lint.
Additionally add `goimports` since it already passes.


Makefile
```diff
@@ -36,13 +33,10 @@ clean:
 	go clean -i ./...
 	rm -rf build
 
-.PHONY: vet
```
6543, line 39: I would still run `go vet`

lukashass: `go vet` is included in golangci-lint (see https://golangci-lint.run/usage/linters/). Is there any benefit to run it separately?

6543:
why this pull pass?

https://wp.laszlo.cloud/woodpecker-ci/woodpecker/1085/17 did fail and lots of that issues are valid & not fixed jet

6543 (approved):

PR #505, opened 2021-11-13
Only publish to latest on tags

belak:
I also mentioned this in discord, but it seemed like a small enough PR that I could just submit it and start a discussion here.

Woodpecker currently publishes to :latest on every push to master *and* new tags, however the general consensus around the Docker community seems to be that :latest should usually refer to the latest stable version. This PR removes publishing to latest on pushes to master (because :next still exists).

I personally prefer this behavior in this PR, but is there a reason to stick with the current behavior?


.woodpecker/main.yml
```diff
@@ -137,7 +137,7 @@ pipeline:
     dockerfile: docker/Dockerfile.server.alpine
     secrets: [ docker_username, docker_password ]
     # TODO: only release as next-alpine after 0.15.0 got released
```
6543, line 139: Rm reladed comment too please

6543:
Can be merged when v0.15.0 is in feature freeze (not jet here)

anbraten:
Suggested edit:
```diff
diff --git a/docs/docs/80-downloads.md b/docs/docs/80-downloads.md
--- docs/docs/80-downloads.md
+++ docs/docs/80-downloads.md
@@ -6,11 +6,11 @@
 
 ## Docker images
 
 Image variants:
-* The `next` images are based on the current master branch
-* The `latest` is same as next (will change after v0.15.0 release to be latest stable release)
+* The `latest` image is the latest stable release
 * The `x.x.x` version images are stable releases
+* The `next` images are based on the current master branch and should not be used for production environments
 
 ``` bash
 # server
 docker pull woodpeckerci/woodpecker-server:latest

```

6543:
@belak :point_up:

belak:
Good call - I've pushed a commit with the updated docs

docs/docs/80-downloads.md
```diff
@@ -7,9 +7,10 @@
 ## Docker images
 
 Image variants:
-* The `next` images are based on the current master branch
-* The `latest` is same as next (will change after v0.15.0 release to be latest stable release)
+* The `latest` image is the latest stable release
 * The `x.x.x` version images are stable releases
+* The `x.x` images are based on version specific release branche, they have the latest bugfixes backported
```
6543, line 12: @anbraten what do you thing about the wording of this?

6543 (approved):

docs/docs/80-downloads.md
anbraten: I am not really a fan of tagging `x.x`. We currently do not tag `x.x` in git anyways and we would have to re-tag for `x.x` each time we want to have a new version if we want to do it with the current way.

6543: https://hub.docker.com/layers/woodpeckerci/woodpecker-server/v0.14/images/sha256-4f728b67930fefa034781d34fb8dcded53e83449e0d3944308f16c8ffdf62fd6?context=explore

and it woks similar to `next` - so no we dont have to re-tag it just publish on merge/push

6543, line 12:
```suggestion
* The `vX.X.X` version images are stable releases
* The `vX.X` images are based on version specific release branche, they have the latest bugfixes backported
```

anbraten: But which job will create it 🤔 
`publish-server` => build next on default branch
`release-server` =>  `v{tag-name}` created on tags

6543: I changed the  `next` target on the release branch:
https://github.com/woodpecker-ci/woodpecker/blob/release/v0.14/.woodpecker/main.yml#L62-L70

we have two options:
 - change it for each release branch specific to it
 - add new step for this whit generic filters (branch: `release/*` ...)

anbraten: Okay, then we should do the second option, so we do not accidentally forget it 
> add new step for this whit generic filters (branch: release/* ...)

6543: OK, I'll add a task to  #684

PR #506, opened 2021-11-13
Update Agent and Pipeline syntax documentation

ticoombs:
- Creates 11-agent-config with examples of agents and configuration options
- Updates 20-pipeline-syntax about the different types of conditionals

Something that was unclear to me was the inherent differences between `when` on a step and `when` on a pipeline. So I fleshed out the original branches section and added an over arching Pipeline Conditionals section. As the previous branch conditional was related to top level pipelines.


docs/docs/30-administration/11-agent-config.md
```diff
@@ -1,3 +1,109 @@
...
+
+## All agent configuration options
+
+Here is the full list of configuration options and their default variables. 
+
+```yaml
+    - WOODPECKER_SERVER=localhost:9000
+    - WOODPECKER_AGENT_SECRET=""
+    - WOODPECKER_USERNAME=x-oauth-basic
+    - WOODPECKER_DEBUG=true
```
6543, line 96: deprecated use WOODPECKER_LOG_LEVEL

ticoombs: I'll remove this line.

I copied all the information from: https://github.com/woodpecker-ci/woodpecker/blob/master/cmd/agent/flags.go for these variables. I realise I also need to look at /agent/agent.go as well. Shall re-read and update this section so that the "defaults" are actually correct.

6543:
> Something that was unclear to me was the inherent differences between when on a step and when on a pipeline. 

when on a pipeline work same as on steps (it just do not support all types step-when does) but apply for all steps in that pipeline

6543:
to: dfc7769556213b36679ad9b9ad65fc45533666f1 -> It reverts a smal change I got wrong when reading and thought i have to fix it ...

6543:
ok did sort the config a bit:
Bildschirmfoto zu 2021-11-20 20-13-11

and also move `when` docu for steps into it's own page ... we might eventually add a sub-folder for that later ...

6543 (approved):

6543:
bit thank's for the initial work :)

PR #509, opened 2021-11-14
format do 'simplify' and check via CI

6543:
let linter check if code can be simplified


anbraten (approved):

6543:
ok I'll have to add --timeout ...

Issue #485, opened 2021-10-27
Fix repo permissions and only sync on demand


6543:
~Instead of pull all permission data of all repos for a user periodicaly, only check and update if needed.~

~(User request info for a repo ... and permission check is out of sync)~

ttl of permission data should be settable too

TODOs:
 - [x] revert #625
 - [ ] clarify permission model (currently 3 props `pull`, `push`, `admin`) and public repos currently get `pull` as fallback
   instead we should make clear a user has explicit pull request (member of the project) or is just able to see data but is not a member (this would fix some cases where users see repos not owned in their UI)
 - [ ] ...

6543:
will frontport https://github.com/woodpecker-ci/woodpecker/pull/491 ... since milestone changed :/

6543:
related https://codeberg.org/Codeberg-CI/feedback/issues/66

6543:
->https://github.com/woodpecker-ci/woodpecker/pull/5665

PR #510, opened 2021-11-14
Fix reggression, save not activated repos to database on sync again

### LINKED ISSUE #508: Reloading repositories from gitea does not work
### Component

server

### Describe the bug

Clicking "reload repositories" in the UI displays a success message ("Repository list reloaded") but the list remains empty. Checking the response from the server shows that it is always an empty array.

This can be reproduced with the docker compose below. Just register a user and add woodpecker as an authorized OAuth2 application (and adjust the compose file along the way).

### System Info

```shell
`{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-ca8e215c"}`
```
```yaml
version: "3"

networks:
  gitea:
    external: false

services:
  gitea:
    image: gitea/gitea:1.15.6
    container_name: gitea
    environment:
      - USER_UID=1000
      - USER_GID=1000
    restart: always
    networks:
      - gitea
    volumes:
      - ./gitea:/data
      - /etc/timezone:/etc/timezone:ro
      - /etc/localtime:/etc/localtime:ro
    ports:
      - "3000:3000"
      - "222:22"
  woodpecker-server:
    image: woodpeckerci/woodpecker-server:latest
    ports:
      - 8000:8000
    volumes:
      - ./woodpecker:/var/lib/woodpecker/
    environment:
      - WOODPECKER_OPEN=true
      - WOODPECKER_HOST=http://localhost:8000
      - WOODPECKER_GITEA=true
      - WOODPECKER_GITEA_PRIVATE_MODE=true
      - WOODPECKER_GITEA_URL=http://${IP_FOR_LOCALHOST}:3000
      - WOODPECKER_GITEA_CLIENT=${WOODPECKER_GITEA_CLIENT}
      - WOODPECKER_GITEA_SECRET=${WOODPECKER_GITEA_SECRET}
      - WOODPECKER_AGENT_SECRET=1234567890123456789012345678901234567890
```

### Additional context

Gitea logs show that `/api/v1/user/repos?limit=50&page=1` is called by woodpecker sucessfully. Also subsequent calls to each repository (`GET /api/v1/repos/{user/org}/{repository}`) are logged by gitea as processed successfully with a `200`.

Woodpeckers sqlite DB remains "empty" (contains only a record in the `user` table and the migrations run in the `migrations` table) after clicking "reload repositories".

### Validations

6543:
close  #508

caused by #474 

there a test was added to exclude repos with user id 0, but that's not jet posible ->  #485 ...

### DIFF
--- a/server/model/config.go
+++ b/server/model/config.go
@@ -28,7 +28,7 @@ type ConfigStore interface {
 type Config struct {
 	ID     int64  `json:"-"    xorm:"pk autoincr 'config_id'"`
 	RepoID int64  `json:"-"    xorm:"UNIQUE(s) 'config_repo_id'"`
-	Hash   string `json:"hash" xorm:"UNIQUE(s) INDEX 'config_hash'"`
+	Hash   string `json:"hash" xorm:"UNIQUE(s) 'config_hash'"`
 	Name   string `json:"name" xorm:"config_name"`
 	Data   []byte `json:"data" xorm:"config_data"`
 }
--- a/server/api/user.go
+++ b/server/api/user.go
@@ -93,7 +93,10 @@ func GetRepos(c *gin.Context) {
 	if flush || time.Unix(user.Synced, 0).Add(time.Hour*72).Before(time.Now()) {
 		log.Debug().Msgf("sync begin: %s", user.Login)
 		user.Synced = time.Now().Unix()
-		store_.UpdateUser(user)
+		if err := store_.UpdateUser(user); err != nil {
+			log.Err(err).Msgf("update user '%s'", user.Login)
+			return
+		}
 
 		config := ToConfig(c)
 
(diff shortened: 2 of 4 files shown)


server/model/config.go
```diff
@@ -28,7 +28,7 @@ type ConfigStore interface {
 type Config struct {
 	ID     int64  `json:"-"    xorm:"pk autoincr 'config_id'"`
 	RepoID int64  `json:"-"    xorm:"UNIQUE(s) 'config_repo_id'"`
-	Hash   string `json:"hash" xorm:"UNIQUE(s) INDEX 'config_hash'"`
+	Hash   string `json:"hash" xorm:"UNIQUE(s) 'config_hash'"`
```
6543, line 31: minor fix :)

anbraten (approved):

PR #512, opened 2021-11-14
Update only relevant cols on repo sync

### LINKED ISSUE #511: RepoSync disable enabled repos
#510 did fix a bug but create a new one ...

now let's do it properly!!!

6543:
close  #511

### DIFF
--- a/server/shared/userSyncer.go
+++ b/server/shared/userSyncer.go
@@ -79,11 +79,11 @@ func (s *Syncer) Sync(ctx context.Context, user *model.User) error {
 			perm := model.Perm{
 				UserID: user.ID,
 				Repo:   repo.FullName,
-				Pull:   true,
 				Synced: unix,
 			}
 			remotePerm, err := s.Remote.Perm(ctx, user, repo.Owner, repo.Name)
 			if err == nil && remotePerm != nil {
+				perm.Pull = remotePerm.Pull
 				perm.Push = remotePerm.Push
 				perm.Admin = remotePerm.Admin
 			}
--- a/server/shared/userSyncer.go
+++ b/server/shared/userSyncer.go
@@ -16,6 +16,7 @@ package shared
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/woodpecker-ci/woodpecker/server/model"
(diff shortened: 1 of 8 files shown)


6543:
now unit tests cover this bug: https://wp.laszlo.cloud/woodpecker-ci/woodpecker/1160/17

server/store/datastore/permission.go
```diff
@@ -29,19 +31,26 @@ func (s storage) PermFind(user *model.User, repo *model.Repo) (*model.Perm, erro
...
-	// TODO: check if needed
-	if len(perm.Repo) != 0 {
+	if err := s.permUpsert(sess, perm); err != nil {
+		return err
+	}
+
+	return sess.Commit()
+}
+
+func (s storage) permUpsert(sess *xorm.Session, perm *model.Perm) error {
```
anbraten, line 47: What was the reason for creating a new function here?

server/shared/userSyncer.go
```diff
@@ -70,24 +71,24 @@ func (s *Syncer) Sync(ctx context.Context, user *model.User) error {
...
 			}
 			remotePerm, err := s.Remote.Perm(ctx, user, repo.Owner, repo.Name)
-			if err == nil && remotePerm != nil {
-				perm.Push = remotePerm.Push
-				perm.Admin = remotePerm.Admin
+			if err != nil {
+				return fmt.Errorf("could not fetch permission of repo '%s': %v", repo.FullName, err)
 			}
-			perms = append(perms, &perm)
+			repo.Perm.Pull = remotePerm.Pull
```
anbraten, line 87: So by this change we wont save permissions on user-sync to the database? Wasn't the `s.Remote.Perm(ctx, user, repo.Owner, repo.Name)` call the expensive part?

server/store/datastore/repo.go
```diff
@@ -86,25 +88,42 @@ func (s storage) RepoBatch(repos []*model.Repo) error {
...
+				return err
+			}
 		} else {
 			// only Insert on single object ref set auto created ID back to object
 			if _, err := sess.Insert(repos[i]); err != nil {
 				return err
 			}
 		}
+
+		if repos[i].Perm != nil {
```
anbraten, line 120: This means: If we do get permissions for the current users for this repo back, we save those permissions, right?

anbraten (commented):
Sorry for my questions. I am still not 100% sure how permissions will be saved and updated. Maybe we should think about removing `repo.Perm` as those are not repo permissions, but the permissions a specific user has for this repo.

server/store/datastore/permission.go
6543: to have it within one sql transaction (**`sess *xorm.Session`**) - it's a private function so it's fine to do so :)

server/store/datastore/repo.go
6543: yes and that's what syncer now does

server/shared/userSyncer.go
6543: this is no fix to the other issue :/ - it make it easyer to fix sure but that's not the scope of that pull ... (it already got bit out of scope :/ )

server/store/datastore/repo.go
```diff
@@ -74,6 +74,8 @@ func (s storage) RepoList(user *model.User, owned bool) ([]*model.Repo, error) {
 		Find(&repos)
 }
 
+// RepoBatch Sync batch of repos (with permissions) to store (create if not exist else update)
```
anbraten, line 77:
```suggestion
// RepoBatch Sync batch of repos from SCM (with permissions) to store (create if not exist else update)
```

anbraten (approved):

PR #513, opened 2021-11-14
Update database in one transaction on syncing user repositorys

6543:
* userSyncer Sync commit changes in one transaction
* cover new cases with unit tests


server/store/datastore/repo.go
```diff
@@ -74,6 +74,8 @@ func (s storage) RepoList(user *model.User, owned bool) ([]*model.Repo, error) {
 		Find(&repos)
 }
 
+// RepoBatch Sync batch of repos (with permissions) to store (create if not exist else update)
```
6543, line 77:
```suggestion
// RepoBatch Sync batch of repos from SCM (with permissions) to store (create if not exist else update)
```

anbraten (approved):

PR #526, opened 2021-11-22
Show environ in every BuildProc

catdevnull:
Screenshot from 2021-11-22 21-16-14
Screenshot from 2021-11-22 21-16-02


anbraten:
@catdevnull I have updated it a bit for you. What do you think?

anbraten (approved):

catdevnull:
Yes, this is so much better. I didn't know what to go off by because there's no design, and I used `Object.entries` because I didn't know the Vue-specific syntax. Shouldn't the build error considering that I didn't add environ to the TypeScript definition? Either way +1

PR #529, opened 2021-11-22
fix back links

anbraten:


6543:
what exactly is this pull fixing?

anbraten:
> ~what exactly is this pull fixing?~ ah nice - that's a thing i also did notice - why is it not ready jet?

It is not really working as expected atm. If you open a new tab, the back button will now use the default, but if you come from another site as for example your Gitea instance, this page will be part of the history of the browser and it will return you to Gitea instead of going to the default. This problem is because the browser only allows us to go back and forth, but does not allow to list previous history entries caused by security concerns. I will look for a workaround.

6543:
on pipeline view: why not go back to repo view all time?

anbraten:
> on pipeline view: why not go back to repo view all time?

In most cases going to the parent route is just fine, but going back is normally a bit nicer as it allows the user to be for example on a build page of repo x, open the side-panel, click another build job from repo y, now the arrow-left button would go to the  overview of repo y instead of going back to the build job from repo x. 

But as fixing the history seems a bit more complicated I will for now change it to your suggestion.

PR #535, opened 2021-11-24
Add linter staticcheck

lukashass:


pipeline/backend/docker/docker.go
```diff
@@ -149,9 +149,9 @@ func (e *engine) Wait(ctx context.Context, proc *backend.Step) (*backend.State,
 	if err != nil {
 		return nil, err
 	}
-	if info.State.Running {
-		// todo
-	}
+	// if info.State.Running {
+	// todo
```
6543, line 153:
```suggestion
	// TODO
```

server/queue/fifo_test.go
```diff
@@ -496,7 +496,7 @@ func TestWaitingVsPending(t *testing.T) {
 	}
 
 	assert.NoError(t, q.Error(noContext, got.ID, fmt.Errorf("exitcode 1, there was an error")))
-	got, _ = q.Poll(noContext, func(*Task) bool { return true })
+	_, _ = q.Poll(noContext, func(*Task) bool { return true })
```
6543, line 500: instead of drop got and err - can we add an assert.True(t, got) ?

server/remote/gitea/gitea.go
```diff
@@ -118,7 +118,7 @@ func (c *Gitea) Login(ctx context.Context, w http.ResponseWriter, req *http.Requ
 		return nil, nil
 	}
 
-	token, err := config.Exchange(oauth2.NoContext, code)
+	token, err := config.Exchange(context.TODO(), code)
```
6543, line 121:
```suggestion
	token, err := config.Exchange(ctx, code)
```

6543 (approved):

PR #540, opened 2021-11-26
Server obtain remote from glob config not from context

6543:


cmd/server/server.go
```diff
@@ -253,6 +252,9 @@ func setupEvilGlobals(c *cli.Context, v store.Store, r remote.Remote) {
 	server.Config.Storage.Files = v
 	server.Config.Storage.Config = v
 
+	// remote
+	server.Config.Services.Remote = r
```
anbraten, line 256: @6543 Shouldn't we try to avoid using this `global evil` object? Apart from that code looks fine.

6543: It depends ... we do both currently we should either inject into context all things or go global.

I'm for last one, as it needs less cpu cycles, is easyer to understand and better to maintinence.

Con: there only can be one global config

About the "evil glob": if it is initialized once it's not evil. global vars could ptential lead to race conditions, if you are aware of that we dont have to worry

6543: & less code is needed

anbraten (approved):

cmd/server/server.go
anbraten: to be consistent we can stay with the global vars for now. if we decide to use context in the future we can do that with a signle PR for all vars

PR #543, opened 2021-11-26
Let normal repo admins change timeout to lower values

### LINKED ISSUE #179: Default Timeout value in build settings cannot be changed
In Woodpecker built from b52e404f93ccea05dc783aa929770c4a0fad2e74 one cannot change value of field `Timeout` in build task settings. After changing default value 60 to 30 error "Filed to update the respository settings" popups and setting is not changed.

No such problem with other settings on the same page i.e. `Pipeline Path`.

Regards,
Paweł

### LINKED ISSUE #534: Insufficient privileges when changing settings on my own project
### Component

server

### Describe the bug

Using Gitea and the latest Woodpecker, I added some projects and am able to run CI jobs on them. However, whenever I change settings, like the timeout, it says I have insufficient privileges. I only have the one user.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-4891f104"}
```

### Additional context

```shell
Thank you!
```

### Validations

6543:
close #534
close #179

### DIFF
--- a/server/api/repo.go
+++ b/server/api/repo.go
@@ -31,7 +31,7 @@ import (
 	"github.com/woodpecker-ci/woodpecker/shared/token"
 )
 
-// TODO: system wide default timeout and max timeout
+// TODO: make it set system wide via environment variables
 const defaultTimeout = 60 // 1 hour default build time
 const maxTimeout = defaultTimeout * 2
 
--- a/server/api/repo.go
+++ b/server/api/repo.go
@@ -115,7 +115,7 @@ func PatchRepo(c *gin.Context) {
 	}
 
 	if in.Timeout != nil && *in.Timeout > maxTimeout && !user.Admin {
-		c.Stringf(http.StatusForbidden, "Timeout is not allowed to be higher than max timeout (%dmin)", maxTimeout)
+		c.String(http.StatusForbidden, fmt.Sprintf("Timeout is not allowed to be higher than max timeout (%dmin)", maxTimeout))
 	}
 	if in.IsTrusted != nil && *in.IsTrusted != repo.IsTrusted && !user.Admin {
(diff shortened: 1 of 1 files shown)


server/api/repo.go
```diff
@@ -31,14 +31,18 @@ import (
 	"github.com/woodpecker-ci/woodpecker/shared/token"
 )
 
+// TODO: system wide default timeout and max timeout
```
6543, line 34:
```suggestion
// TODO: make it set system wide via environment variables
```

server/api/repo.go
```diff
@@ -108,8 +114,12 @@ func PatchRepo(c *gin.Context) {
 		return
 	}
 
-	if (in.IsTrusted != nil || in.Timeout != nil) && !user.Admin {
-		c.String(403, "Insufficient privileges")
+	if in.Timeout != nil && *in.Timeout > maxTimeout && !user.Admin {
+		c.String(http.StatusForbidden, "timeout is higher than max timeout on this instance")
+	}
+	if in.IsTrusted != nil && *in.IsTrusted != repo.IsTrusted && !user.Admin {
```
anbraten, line 120: Is this the fix for #534?

anbraten (approved):

server/api/repo.go
6543: `&& *in.IsTrusted != repo.IsTrusted` - yes it looks like the new ui send IsTrusted field even if not changed :)

so you could call it a "frontend"-issue but I like to harden the backend so this should just also do/fix it

anbraten: Ahh yip, I changed it to always send the complete from instead of the previous behavior were the form was directly submitted on click or key-input with only the specific part.

PR #552, opened 2021-11-27
Adding initial version of Kubernetes backend

### LINKED ISSUE #9: Add Kubernetes backend
Add Kubernetes as a backend to Woodpecker to allow running pipelines via the Kubernetes API in a Kubernetes Cluster.

anbraten:
close #9
close #784

# TODO
- [x] add docs
- [x] update helm charts
- [ ] make services work => will be done in a follow-up

# TEST it NOW
 - [`woodpeckerci/woodpecker-server:pull_552`](https://hub.docker.com/r/woodpeckerci/woodpecker-server/tags?name=pull_552)
 - [`woodpeckerci/woodpecker-agent:pull_552`](https://hub.docker.com/r/woodpeckerci/woodpecker-agent/tags?name=pull_552)
  - [`woodpeckerci/woodpecker-cli:pull_552`](https://hub.docker.com/r/woodpeckerci/woodpecker-cli/tags?name=pull_552)


dmolik:
is there a way to specify the storageClass and maybe even size, via the backend config?

now this is very pedantic, but it might make sense to abbreviate corev1 instead of v1

anbraten:
I think proper volumes support is one of the biggest challenges we somehow have to solve in this PR. We should allow to specify volume details. Do you know if we can create some kind of kubernetes storage provided by most kubernetes installations which allows us to mount the same volume to multiple pods?

616b2f:
@anbraten depends on, do you wanna ReadOnly mount or ReadWrite? There is a ReadOnlyMany (ROX) and ReadWriteMany (RWX) modes that can be set on a PersistentVolumeClaim: see https://stackoverflow.com/a/62545427

dmolik:
I think most deployments only support RWO ( read write once ), I'm thinking AWS EBS volumes and Rook/Ceph, I don't think we should expect or require anything more complicated.

pipelines can be mostly serial, so it should be enough.

Perhaps with parallel stages jobs can be schlepped into a "mega" pod

616b2f:
@dmolik it may be simpler to start with RWX first as they are some options out there:

- AWS with Amazon EFS CSI driver see https://aws.amazon.com/de/premiumsupport/knowledge-center/eks-persistent-storage/
- Google with Filestore CSI driver see https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/filestore-csi-driver
- Azure with Azure Files see https://docs.microsoft.com/en-us/azure/aks/azure-files-dynamic-pv
- For on Prem: Rook supports RWX via CephFS see https://github.com/ceph/ceph-csi and https://github.com/rook/rook/issues/543#issuecomment-388060949
- Or use NFS CSI driver see https://github.com/kubernetes-csi/csi-driver-nfs/blob/9811fe4c6fa00169b4f80832cd807c0203fa0059/deploy/example/pvc-nfs-csi-dynamic.yaml

How ever I am not sure how performant they are, but in my opinion it is simpler to start with them at the beginning then use some workarounds with RWO. The approach you described can still be explored after words if there are some issues with the RWX approach.

What do you think?

LamaAni:
You can, in general, use a network drive. https://kubernetes.io/docs/concepts/storage/volumes/#nfs

But these are sometime problematic (usually if they fail on the umount), and should be used with care.

kvaster:
Personally I don't think that sticking to network drives are good idea. We're using kubernetes with local storage mostly, cause it is much more faster.

616b2f:
> Personally I don't think that sticking to network drives are good idea. We're using kubernetes with local storage mostly, cause it is much more faster.

@kvaster This does not work for data that has to be shared across pods (because they can be located on different nodes) and even if you host them all on one node, this approach does not scale well and is not really failure tolerant, if your node dies all your workload and data does with it.
But maybe I misunderstood you so please correct me if I am wrong.

dmolik:
Existing Kubernetes deployments are just that, dind agents that have parallel jobs all on the same node, and don't have network RWX volumes, either.

kvaster:
Generally there can be two types of volumes. First one - create PVC for all pods created during build and this PVC can be RWX (and network) and second one - PVC per pod.

kvaster:
> @kvaster This does not work for data that has to be shared across pods (because they can be located on different nodes) and even if you host them all on one node, this approach does not scale well and is not really failure tolerant, if your node dies all your workload and data does with it. But maybe I misunderstood you so please correct me if I am wrong.

I think this is always question about cache, bandwidth, disk speed e.t.c. It is good to have flexibility. Sometimes it is much more better to use local disks while building with some kind of cache. Also this may be combined with network drive for sharing some part of the data while building artefacts in parallel.

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -36,37 +53,229 @@ func (e *kube) IsAvailable() bool {
...
+
+			switch pod.Status.Phase {
+			case v1.PodSucceeded, v1.PodFailed, v1.PodUnknown:
+				finished <- true
+			}
+		}
+	}
+
+	// TODO 5 seconds is against best practice, k3s didn't work otherwise
+	si := informers.NewSharedInformerFactory(e.client, 5*time.Second)
```
ericpp, line 148:
```suggestion
	si := informers.NewSharedInformerFactoryWithOptions(e.client, 5*time.Second, informers.WithNamespace(e.namespace))
```

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -36,37 +53,229 @@ func (e *kube) IsAvailable() bool {
...
+		if pod.Name == podName {
+			switch pod.Status.Phase {
+			case v1.PodRunning, v1.PodSucceeded, v1.PodFailed:
+				up <- true
+			}
+		}
+	}
+
+	// TODO 5 seconds is against best practice, k3s didn't work otherwise
+	si := informers.NewSharedInformerFactory(e.client, 5*time.Second)
```
ericpp, line 194:
```suggestion
	si := informers.NewSharedInformerFactoryWithOptions(e.client, 5*time.Second, informers.WithNamespace(e.namespace))
```

ericpp (commented):
The `NewSharedInformerFactory` calls seem to monitor all namespaces and error out if your service account is restricted to one namespace. Using the `NewSharedInformerFactoryWithOptions` version with `WithNamespace` restricts the monitoring to one namespace

xuecanlong:
can i inquire about the progress on this？

6543:
focus is on  #784 - after this one got in we can move forward to native by reusing code and refactoring more ...
the mentioned pull is not merged jet as int touches code outside of `pipeline/backend/*` - so this changes have to be proted and merged first seperatly

anbraten:
To be fair I am not 100% sure if we should take #784 over this one as I don't see the point in using the kubectl (not working for scratch images, rpm, deb packages etc out of the box). Suggestion would be to migrate all ideas / pieces from #784 into this one.

xuecanlong:
> 公平地说，我不是 100% 确定我们是否应该使用#784，因为我看不出使用 kubectl 的意义（不适用于开箱即用的临时图像、rpm、deb 包等）。建议将#784中的所有想法/片段迁移到这个中。

I agree with you, if we want to support k8s, use k8s-sdk is good. kubectl can use `local` type to use it

xuecanlong:
> To be fair I am not 100% sure if we should take #784 over this one as I don't see the point in using the kubectl (not working for scratch images, rpm, deb packages etc out of the box). Suggestion would be to migrate all ideas / pieces from #784 into this one.

I agree with you, if we want to support k8s, use k8s-sdk is good. kubectl can use `local` type to use it

6543:
> To be fair I am not 100% sure if we should take #784 over this one as I don't see the point in using the kubectl (not working for scratch images, rpm, deb packages etc out of the box). Suggestion would be to migrate all ideas / pieces from #784 into this one.

That would be even better - it just need a lot more work to be done ...

anbraten:
If you have a look at the code its not to different. We already have most of the general functions of the kubectl backend in this PR as well.

xuecanlong:
> If you have a look at the code its not to different. We already have most of the general functions of the kubectl backend in this PR as well.

i can't wait it be release

LamaAni:
Hi, the reason I used kubectl is simplicity in implementation (apply, delete). That said, by changing the functions on "client.go" you can use the kube go api. All the interactions with kubernets are only there. Sadly, I do not have the time to do that. 

We have been using the kubectl backend in development for about a month now with small issues (not detecting event based crashes is the worst one, it only detects BackOff for now). Once I have a fuller list I'll update these errors. 

@xuecanlong I can safely say that we are now in the beta mode of that, and creating an image that uses that is very simple.

xuecanlong:
agree, useing kubectl is faster to ops cluster, but not convenience for complex operations

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -1,28 +1,55 @@
...
-
 // New returns a new Kubernetes Engine.
-func New(namespace, endpoint, token string) types.Engine {
+func New() types.Engine {
+	namespace := os.Getenv("WOODPECKER_BACKEND_K8S_NAMESPACE")
+	if namespace == "" {
+		namespace = "woodpecker-ci"
+	}
+	storageClass := os.Getenv("WOODPECKER_BACKEND_K8S_NAMESPACE")
+	volumeSize := os.Getenv("WOODPECKER_BACKEND_K8S_NAMESPACE")
```
Rynoxx, line 43:
```suggestion
	storageClass := os.Getenv("WOODPECKER_BACKEND_K8S_STORAGE_CLASS")
	volumeSize := os.Getenv("WOODPECKER_BACKEND_K8S_VOLUME_SIZE")
```

6543:
Docs are missing

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -1,28 +1,54 @@
...
```
stintel, line 42: This causes a panic:

```
Defaulted container "woodpecker-agent" out of: woodpecker-agent, dind
{"level":"error","Repo":"stintel/vallumd","Build":"5","ID":"75","error":"pods \"0-7304064393753199929-clone\" not found","time":"2022-08-14T21:17:27Z","message":"could not destroy engine"}
panic: cannot parse 'woodpecker-agents': quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'

goroutine 15 [running]:
k8s.io/apimachinery/pkg/api/resource.MustParse({0xc000044021, 0x11})
        /src/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go:128 +0x1ce
github.com/woodpecker-ci/woodpecker/pipeline/backend/kubernetes.PersistentVolumeClaim({0xc000044021, 0x11}, {0xc00050c380, 0x1d}, {0xc000044021, 0x11}, {0xc000044021, 0x11})
        /src/pipeline/backend/kubernetes/volume.go:27 +0xf2
github.com/woodpecker-ci/woodpecker/pipeline/backend/kubernetes.(*kube).Setup(0xc0001199f0, {0x1ae9878, 0xc000216900}, 0xc000214600)
        /src/pipeline/backend/kubernetes/kubernetes.go:87 +0xf4
github.com/woodpecker-ci/woodpecker/pipeline.(*Runtime).Run(0xc000226690)
        /src/pipeline/pipeline.go:93 +0x2f5
github.com/woodpecker-ci/woodpecker/agent.(*Runner).Run(0xc0000b1b80, {0x1ae9878, 0xc0000b0f80})
        /src/agent/runner.go:319 +0xea5
main.loop.func3()
        /src/cmd/agent/agent.go:169 +0x432
created by main.loop
        /src/cmd/agent/agent.go:142 +0xe99
```

stintel: To elaborate, both storageClass and volumeSize use the same env var WOODPECKER_BACKEND_K8S_NAMESPACE

stintel: I worked around it by naming my namespace 10240m, but then I got stuck because I don't have a ReadWriteMany StorageClass in my poor man's K3S cluster.

anbraten: New version is currently building with support for `WOODPECKER_BACKEND_K8S_NAMESPACE`, `WOODPECKER_BACKEND_K8S_VOLUME_SIZE`, `WOODPECKER_BACKEND_K8S_STORAGE_CLASS`: https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/3160

stintel: Meanwhile I managed to get things to work using https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner. Still a poor man's K3S cluster, but it's working nicely. I'd say with those copy-paste errors fixed and some decent documentation, this is good to go. Thanks!

anbraten: New agent version got pushed.

cmd/agent/flags.go
```diff
@@ -106,4 +106,26 @@ var flags = []cli.Flag{
...
+		EnvVars: []string{"WOODPECKER_BACKEND_K8S_NAMESPACE"},
+		Name:    "backend-k8s-namespace",
+		Usage:   "backend k8s namespace",
+		Value:   "woodpecker-ci",
+	},
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_BACKEND_K8S_VOLUME_SIZE"},
+		Name:    "backend-k8s-volume-size",
+		Usage:   "backend k8s volume size (default 10GiB)",
+		Value:   "10GiB",
```
stintel, line 123:
```
panic: cannot parse '10GiB': quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
```

6543, line 123:
```suggestion
		Value:   "10G",
```

cmd/agent/flags.go
```diff
@@ -106,4 +106,26 @@ var flags = []cli.Flag{
...
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_BACKEND_K8S_NAMESPACE"},
+		Name:    "backend-k8s-namespace",
+		Usage:   "backend k8s namespace",
+		Value:   "woodpecker-ci",
+	},
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_BACKEND_K8S_VOLUME_SIZE"},
+		Name:    "backend-k8s-volume-size",
+		Usage:   "backend k8s volume size (default 10GiB)",
```
6543, line 122:
```suggestion
		Usage:   "backend k8s volume size (default 10G)",
```

anbraten:
This PR should be mainly done now. Some pipeline features are still missing, but I would add them from time to time. I will add the currently open points to https://github.com/woodpecker-ci/woodpecker/issues/9#issuecomment-483979755.

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -1,28 +1,44 @@
...
+
+	// To authenticate to GCP K8s clusters
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
 
 type kube struct {
-	namespace string
-	endpoint  string
-	token     string
+	logs         *bytes.Buffer // TODO remove
```
6543, line 28: ?

6543:
else for a starting point ... we can merge asap you call it "ready for review"

pipeline/backend/kubernetes/kubernetes.go
anbraten: I actually don't know how it should be resolved. I almost copied all of this code from laszlocphs PR #23.

charts/woodpecker-agent/values.yaml
```diff
@@ -7,15 +7,21 @@ image:
   # Overrides the image tag whose default is the chart appVersion.
   tag: ""
 
+env:
+  WOODPECKER_SERVER: "woodpecker-server.<namespace>.svc.cluster.local:9000"
+  WOODPCKER_BACKEND: kubernetes
```
anbraten, line 12:
```suggestion
  WOODPECKER_BACKEND: kubernetes
```

Rynoxx:
Not sure how to do this properly since this in and of itself is a PR, but I created a PR for this branch to implement a setting for using RWO access mode: https://github.com/anbraten/woodpecker/pull/3

pipeline/backend/kubernetes/volume_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	pvc := PersistentVolumeClaim("someNamespace", "someName", "local-storage", "1Gi", true)
+	j, err := json.Marshal(pvc)
+	assert.Nil(t, err)
+	assert.JSONEq(t, expected_rwx, string(j))
+
+	pvc = PersistentVolumeClaim("someNamespace", "someName", "local-storage", "1Gi", false)
+	j, err = json.Marshal(pvc)
+	assert.Nil(t, err)
+	assert.JSONEq(t, expected_rwo, string(j))
```
Rynoxx, line 61: Fix variable names to pass test/linting
```suggestion
	expectedRwx := `
	{
	  "metadata": {
	    "name": "someName",
	    "namespace": "someNamespace",
	    "creationTimestamp": null
	  },
	  "spec": {
	    "accessModes": [
	      "ReadWriteMany"
	    ],
	    "resources": {
	      "requests": {
	        "storage": "1Gi"
	      }
	    },
	    "storageClassName": "local-storage"
	  },
	  "status": {}
	}`

	expectedRwo := `
	{
	  "metadata": {
	    "name": "someName",
	    "namespace": "someNamespace",
	    "creationTimestamp": null
	  },
	  "spec": {
	    "accessModes": [
	      "ReadWriteOnce"
	    ],
	    "resources": {
	      "requests": {
	        "storage": "1Gi"
	      }
	    },
	    "storageClassName": "local-storage"
	  },
	  "status": {}
	}`

	pvc := PersistentVolumeClaim("someNamespace", "someName", "local-storage", "1Gi", true)
	j, err := json.Marshal(pvc)
	assert.Nil(t, err)
	assert.JSONEq(t, expectedRwx, string(j))

	pvc = PersistentVolumeClaim("someNamespace", "someName", "local-storage", "1Gi", false)
	j, err = json.Marshal(pvc)
	assert.Nil(t, err)
	assert.JSONEq(t, expectedRwo, string(j))
```

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -36,37 +54,237 @@ func (e *kube) IsAvailable() bool {
...
+					return err
+				}
+			}
+		}
+	}
+
+	for _, vol := range conf.Volumes {
+		pvc := PersistentVolumeClaim(e.namespace, vol.Name, e.storageClass, e.volumeSize, e.storageRwx)
+		err := e.client.CoreV1().PersistentVolumeClaims(e.namespace).Delete(ctx, pvc.Name, deleteOpts)
+		if err != nil {
```
Rynoxx, line 284: When canceling the job via the UI the ctx we receive is already canceled, and therefore the .Delete command errors and the containers don't get deleted.
And don't abort on NotFound errors.

Re-used the way around this from the docker backend, use a noContext instead

```suggestion
func (e *kube) Destroy(ctx context.Context, conf *types.Config) error {
	var gracePeriodSeconds int64 = 0 // immediately
	dpb := metav1.DeletePropagationBackground

	deleteOpts := metav1.DeleteOptions{
		GracePeriodSeconds: &gracePeriodSeconds,
		PropagationPolicy:  &dpb,
	}

	for _, stage := range conf.Stages {
		for _, step := range stage.Steps {
			e.logs.WriteString("Deleting pod\n")
			if err := e.client.CoreV1().Pods(e.namespace).Delete(noContext, podName(step), deleteOpts); err != nil && !errors.IsNotFound(err) {
				return err
			}
		}
	}

	for _, stage := range conf.Stages {
		if stage.Alias == "services" {
			for _, step := range stage.Steps {
				e.logs.WriteString("Deleting service\n")
				// TODO: support ports setting
				// svc, err := Service(e.namespace, step.Name, step.Alias, step.Ports)
				svc, err := Service(e.namespace, step.Name, step.Alias, []string{})
				if err != nil {
					return err
				}
				if err := e.client.CoreV1().Services(e.namespace).Delete(noContext, svc.Name, deleteOpts); err != nil && !errors.IsNotFound(err) {
					return err
				}
			}
		}
	}

	for _, vol := range conf.Volumes {
		pvc := PersistentVolumeClaim(e.namespace, vol.Name, e.storageClass, e.volumeSize, e.storageRwx)
		err := e.client.CoreV1().PersistentVolumeClaims(e.namespace).Delete(noContext, pvc.Name, deleteOpts)
		if err != nil && !errors.IsNotFound(err) {
```

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -1,28 +1,46 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+
+	// To authenticate to GCP K8s clusters
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
```
Rynoxx, line 26: Add noContext var for the previously suggested fix
```suggestion

var noContext = context.Background()

```

pipeline/backend/kubernetes/kubernetes.go
6543: Ok lets add a **descriptive** TODO

pipeline/backend/kubernetes/kubernetes.go
6543: Good catch can you open a pull for current docker & add some descriptive code comments ❤️

6543: If errNotFound we should trace log it ... so we can detect missbehaviour

Rynoxx: > Good catch can you open a pull for current docker & add some descriptive code comments
The docker-backend already has this stuff, where I got the idea to use the noContext: https://github.com/woodpecker-ci/woodpecker/blob/master/pipeline/backend/docker/docker.go#L203

As for the code comments and trace log, I've created a PR to @anbraten k8s branch again rather than editing this. These changes got very messy in to do in the change suggestions.

https://github.com/anbraten/woodpecker/pull/4

6543:
https://github.com/anbraten/woodpecker/pull/4 need a merge

pipeline/backend/backend.go
```diff
@@ -3,20 +3,22 @@ package backend
 import (
 	"fmt"
 
+	"github.com/urfave/cli/v2"
```
6543, line 6: would be nice if we could pass backend options somehow else into pipeline/** packages

pipeline/backend/kubernetes/kubernetes.go
anbraten: Okay I think I can just replace this "logs" buffer with zerolog trace log calls. It seems to be used for log messages like `Setting up Kubernetes primitives`, `Creating service`

6543: Thanks

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -36,37 +55,246 @@ func (e *kube) IsAvailable() bool {
...
+		pod := new.(*v1.Pod)
+		if pod.Name == podName {
+			switch pod.Status.Phase {
+			case v1.PodRunning, v1.PodSucceeded, v1.PodFailed:
+				up <- true
+			}
+		}
+	}
+
+	// TODO 5 seconds is against best practice, k3s didn't work otherwise
```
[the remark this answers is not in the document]
anbraten: Its k3s. I think k8s worked as expected in her, but k3s did not.

pipeline/backend/backend.go
6543: had an idea, will push

6543 (approved):

6543:
@anbraten hope you are fine with how it's done as ContextKey for now ...

I might refactor how backends do get added and so also add a new interface we can use for backend specific config, but that can be done later

stintel:
Been testing this for several weeks, also just built and deployed locally after the last push to this PR, things are still working, so, FWIW:
Tested-by: Stijn Tintel <stijn@linux-ipv6.be>

Dan6erbond:
What's the status on the docs and Helm chart? Getting this setup on my cluster would be a game-changer. Would love to help if there's anything I can do. Helm charts aren't the way I deploy my services, though, so I would mostly be open to contributing to docs.

616b2f:
@Dan6erbond you can generate plain yaml files from helm charts and deploy them, just in case you didn't knew.

PR #554, opened 2021-11-28
Add linter revive

lukashass:
Part of #435 

## Summary of fixes

- remove underscores from variable names (replace trailing with leading underscore to avoid shadowing)
- capitalize e.g. `URL`, `HTTP`, `ID` in variable names
- remove `else` after `return`
- comment for blank imports
- remove trailing punctuation in error messages
- make `error` the last return value

### DIFF
--- a/server/api/build.go
+++ b/server/api/build.go
@@ -54,7 +54,7 @@ func GetBuilds(c *gin.Context) {
 }
 
 func GetBuild(c *gin.Context) {
-	store_ := store.FromContext(c)
+	_store := store.FromContext(c)
 	if c.Param("number") == "latest" {
 		GetBuildLast(c)
 		return
@@ -67,13 +67,13 @@ func GetBuild(c *gin.Context) {
 		return
 	}
 
-	build, err := store_.GetBuildNumber(repo, num)
+	build, err := _store.GetBuildNumber(repo, num)
 	if err != nil {
 		_ = c.AbortWithError(http.StatusInternalServerError, err)
 		return
 	}
-	files, _ := store_.FileList(build)
-	procs, _ := store_.ProcList(build)
+	files, _ := _store.FileList(build)
+	procs, _ := _store.ProcList(build)
 	build.Procs = model.Tree(procs)
 	build.Files = files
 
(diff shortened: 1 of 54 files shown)


server/api/build.go
```diff
@@ -54,7 +54,7 @@ func GetBuilds(c *gin.Context) {
 }
 
 func GetBuild(c *gin.Context) {
-	store_ := store.FromContext(c)
+	_store := store.FromContext(c)
```
lukashass, line 57: We could even use `store` here and in a lot of other cases, as redefining the variable should not be a problem. Not sure what is best :thinking:

6543: well I'll change that anyway ... later :)

server/shared/configFetcher.go
```diff
@@ -13,27 +13,27 @@ import (
 	"github.com/woodpecker-ci/woodpecker/server/remote"
 )
 
-type configFetcher struct {
-	remote_ remote.Remote
-	user    *model.User
-	repo    *model.Repo
-	build   *model.Build
+type ConfigFetcher struct {
```
6543, line 16: why export this struct?

server/shared/procBuilder.go
```diff
@@ -175,7 +175,7 @@ func containsItemWithName(name string, items []*BuildItem) bool {
 	return false
 }
 
-func (b *ProcBuilder) envsubst_(y string, environ map[string]string) (string, error) {
+func (b *ProcBuilder) _envsubst(y string, environ map[string]string) (string, error) {
```
6543, line 178:
```suggestion
func (b *ProcBuilder) envsubst(y string, environ map[string]string) (string, error) {
```

shared/oauth2/oauth2.go
```diff
@@ -214,28 +214,28 @@ func (t *Transport) transport() http.RoundTripper {
 // AuthCodeURL returns a URL that the end-user should be redirected to,
 // so that they may obtain an authorization code.
 func (c *Config) AuthCodeURL(state string) string {
-	url_, err := url.Parse(c.AuthURL)
+	_url, err := url.Parse(c.AuthURL)
```
6543, line 217:
```suggestion
	url, err := url.Parse(c.AuthURL)
```

server/shared/configFetcher.go
lukashass: Ah forgot to mention this in the summary. Because of this rule: https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#unexported-return

6543 (approved):

PR #384, opened 2021-09-29
Remove ghodss/yaml

anbraten:
closes #352

Issue #517, opened 2021-11-17
Add support for YAML 1.2


seabass-labrax:
### Clear and concise description of the problem

It would be convenient to be able to use YAML 1.2 for `.woodpecker.yml` and other such workflow files. I believe many users are likely to assume that Woodpecker supports the latest version of YAML, but it currently supports only YAML 1.1 (which was published in 2005).

Adding an explicit version directive (`%YAML 1.2`) to `.woodpecker.yml` results in the CI not being run, and the CI status not being shown at all in the Gitea ([Codeberg](https://codeberg.org/)) user interface. Woodpecker's own user interface displays the following error message:

> Execution error
> yaml: line 3: found incompatible YAML document

This message is generated by the `go-yaml` library, in its [`yaml/parserc.go`](
https://github.com/go-yaml/yaml/blob/7649d4548cb53a614db133b2a8ac1f31859dda8c/parserc.go#L1032-L1035) file.

### Suggested solution

We could use a different YAML parsing library which does support YAML 1.2, or contribute support for YAML 1.2 upstream. This may be related to an existing issue and pull request relating to YAML support:

<https://github.com/woodpecker-ci/woodpecker/issues/352>
<https://github.com/woodpecker-ci/woodpecker/pull/384>

### Alternative

We could also update the documentation to make it clear that only YAML 1.1 is supported. Currently, it is potentially confusing that the ['Getting Started'](https://woodpecker-ci.org/docs/usage/intro) page on the Woodpecker documentation says that the format "is a superset of the widely used docker-compose file format", despite the fact that Docker Compose *does* support YAML 1.2.

### Additional context

This is the error message I referred to earlier:

![error-message](https://user-images.githubusercontent.com/79224200/142251320-666f63d8-2ce2-4d8b-9bd0-f62d508d2ffd.png)

...and this is a pop-up message that appears if one tries to restart the failed build:
![error-popup](https://user-images.githubusercontent.com/79224200/142251314-b59cb252-eb9f-4005-8e9b-e4ede08a87c1.png)

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

seabass-labrax:
[...]

PR #570, opened 2021-12-06
Write own yaml2json func

6543:
* fix regression of #384 
 * add more tests


6543:
that also should make #517 possible?

anbraten (approved):

PR #584, opened 2021-12-08
Add log-level option to cli

6543:


cli/build/build_approve.go
```diff
@@ -19,6 +19,7 @@ var buildApproveCmd = &cli.Command{
 }
 
 func buildApprove(c *cli.Context) (err error) {
+	common.SetupConsoleLogger(c)
```
anbraten, line 22: Isn't there something like a global hook / init fnc for `github.com/urfave/cli/v2` to call this logger setup?

anbraten (approved):

cli/build/build_approve.go
anbraten: from checking the docs there only seems to be an option for commands called `Before`:

```go
       Before: func(c *cli.Context) error {
          fmt.Fprintf(c.App.Writer, "brace for impact\n")
          return nil
        },
```

But I doesn't look like we could save code that way.

anbraten: Ohh, from the api it seems as we could add it to the app as global before hook as well.

6543: I'll have a look - that's why it's not merged jet :D

anbraten: Yeah no pressure. Just wanted to inform you about my investigations on it. 😉

6543: Before did the job ...

anbraten (approved):
nice. the new change looks very clean and DRY now

PR #586, opened 2021-12-09
add flag to set oauth redirect host in dev mode

anbraten:
This setting allows to set a special redirect host address (like `http://localhost:8000`) and still use `WOODPECKER_HOST` for the public address which is for example used for the webhooks.

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -287,6 +287,11 @@ func setupEvilGlobals(c *cli.Context, v store.Store, r remote.Remote) {
 	server.Config.Server.Key = c.String("server-key")
 	server.Config.Server.Pass = c.String("agent-secret")
 	server.Config.Server.Host = c.String("server-host")
+	if len(c.String("server-dev-oauth-host")) > 0 {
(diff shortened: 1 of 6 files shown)


anbraten:
Maybe it would also be a solution, to make it the other way around, so you can set the normal host (`WOODPECKER_HOST`) to `http://localhost:8000` and have an additional option `WOODPECKER_DEV_WEBHOOK_HOST` to set the host for webhooks differently.

cmd/server/server.go
```diff
@@ -287,6 +287,11 @@ func setupEvilGlobals(c *cli.Context, v store.Store, r remote.Remote) {
 	server.Config.Server.Key = c.String("server-key")
 	server.Config.Server.Pass = c.String("agent-secret")
 	server.Config.Server.Host = c.String("server-host")
+	if len(c.String("server-dev-oauth-host")) > 0 {
```
6543, line 290: c.IsSet()

6543 (approved):

PR #592, opened 2021-12-10
fix pr secret event names

anbraten:


6543:
can you add a check if Events are valid on secret creation
https://github.com/woodpecker-ci/woodpecker/blob/master/server/api/secret.go#L50

6543 (approved):

PR #604, opened 2021-12-13
Get secrets in settings

jolheiser:
This PR adds the ability to get a `setting` from a `secret`. 
I've opted to follow the format Drone does to ease migration for users, let me know if we want to go another route.

```yaml
settings:
  my_secret:
    from_secret: secret_name
```

-----

I'm not a huge fan of the hadoken logic, but with our current YAML unmarshaling I think it's the best we can do.

### DIFF
--- a/pipeline/frontend/yaml/compiler/params.go
+++ b/pipeline/frontend/yaml/compiler/params.go
@@ -13,7 +13,7 @@ import (
 
 // paramsToEnv uses reflection to convert a map[string]interface to a list
 // of environment variables.
-func paramsToEnv(from map[string]interface{}, to map[string]string) (err error) {
+func paramsToEnv(from map[string]interface{}, to map[string]string, secrets map[string]Secret) (err error) {
 	if to == nil {
 		return fmt.Errorf("no map to write to")
 	}
@@ -21,7 +21,7 @@ func paramsToEnv(from map[string]interface{}, to map[string]string) (err error) 
 		if v == nil || len(k) == 0 {
 			continue
 		}
-		to[sanitizeParamKey(k)], err = sanitizeParamValue(v)
+		to[sanitizeParamKey(k)], err = sanitizeParamValue(v, secrets)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


pipeline/frontend/yaml/compiler/params.go
```diff
@@ -66,6 +66,16 @@ func sanitizeParamValue(v interface{}) (string, error) {
 		return fmt.Sprintf("%v", vv.Float()), nil
 
 	case reflect.Map:
+		if fromSecret, ok := v.(map[string]interface{}); ok {
```
6543, line 69: if secrets != nil

pipeline/frontend/yaml/compiler/params_test.go
```diff
@@ -33,9 +35,44 @@ func TestParamsToEnv(t *testing.T) {
...
+bool: true
+slice: [1, 2, 3]
+my_secret:
+  from_secret: secret_token
+`)
+	var from map[string]interface{}
+	err := yaml.Unmarshal(fromYAML, &from)
+	t.Logf("%T", from["my_secret"])
+	assert.NoError(t, err)
+
```
6543, line 62: can you add an error case too?

pipeline/frontend/yaml/compiler/params.go
jolheiser: Since we are only reading we don't need a nil check. Nil maps behave like empty maps when reading,

pipeline/frontend/yaml/compiler/params_test.go
jolheiser: Added below separately

6543 (approved):

6543:
@jolheiser one thing i just remember - can you add some docu for this feature :laughing:

jolheiser:
> @jolheiser one thing i just remember (before hit the merge button ;) ) - can you add some docu for this feature laughing

Good catch! I've added some docs, will wait for review before merging.

6543 (approved):
not to short and not to long :)

PR #620, opened 2021-12-18
Use gitlab generic webhooks instead of drone-ci-service

anbraten:
Benefits:
- the webhook delivery history of the drone-ci-service is broken (no way to check if a webhook was successfully delivered by Gitlab)
- drone-ci-service has limited events support (for example no comment or branch deleted event)
- independent from drone integration in general

### DIFF
--- a/server/remote/gitlab/gitlab.go
+++ b/server/remote/gitlab/gitlab.go
@@ -388,6 +388,16 @@ func (g *Gitlab) Netrc(u *model.User, r *model.Repo) (*model.Netrc, error) {
 	}, nil
 }
 
+func (g *Gitlab) getTokenAndWebURL(link string) (token string, webURL string, err error) {
+	uri, err := url.Parse(link)
+	if err != nil {
+		return "", "", err
+	}
+	token = uri.Query().Get("access_token")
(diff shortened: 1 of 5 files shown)


6543:
you have to ajust the moc responces

server/remote/gitlab/gitlab.go
```diff
@@ -388,30 +388,48 @@ func (g *Gitlab) Netrc(u *model.User, r *model.Repo) (*model.Netrc, error) {
 	}, nil
 }
 
+func (g *Gitlab) getTokenAndWebURL(link string) (token string, webURL string, err error) {
+	uri, err := url.Parse(link)
+	if err != nil {
+		return "", "", err
+	}
+	token = uri.Query().Get("access_token")
+	webURL = fmt.Sprintf("%s://%s/api/hook", uri.Scheme, uri.Host)
```
6543, line 397: with this we do not support gitlab within subfolder - just wonder if we ever had?

6543: we didnt :D

6543 (approved):

PR #623, opened 2021-12-19
Add "woodpeckerci/plugin-docker-buildx" to privileged plugins

6543:


shared/constant/privileged_plugins.go
```diff
@@ -0,0 +1,8 @@
+package constant
+
+var PrivilegedPlugins = []string{
+	"plugins/docker",
+	"plugins/gcr",
+	"plugins/ecr",
+	"woodpeckerci/plugin-docker-buildx",
```
anbraten, line 7: @6543 Should we just add `woodpeckerci/plugin-docker` as preparation to the list? As noone could claim that image name it should have no risk doing so and we could simply start own docker plugin fork then.

anbraten (approved):

PR #624, opened 2021-12-19
pipeline env migrate CI_TAG & CI_PULL_REQUEST to new CI_COMMIT_TAG & CI_COMMIT_PULL_REQUEST

### LINKED ISSUE #498: [Helm]: problem while publishing charts on tags
noticed while using current master as ci test locally 

version replace do not work correct - need to test it again to get logs

6543:
close #498

### DIFF
--- a/docs/docs/91-migrations.md
+++ b/docs/docs/91-migrations.md
@@ -30,6 +30,7 @@ Some versions need some changes to the server configuration or the pipeline conf
     - CI_ARCH => use CI_SYSTEM_ARCH
     - CI_COMMIT => CI_COMMIT_SHA
     - CI_TAG => CI_COMMIT_TAG
+    - CI_PULL_REQUEST => CI_COMMIT_PULL_REQUEST
     - CI_REMOTE_URL => use CI_REPO_REMOTE
     - CI_REPO_BRANCH => use CI_REPO_DEFAULT_BRANCH
     - CI_PARENT_BUILD_NUMBER => use CI_BUILD_PARENT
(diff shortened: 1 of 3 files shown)


docs/docs/91-migrations.md
```diff
@@ -29,6 +29,7 @@ Some versions need some changes to the server configuration or the pipeline conf
   - Some variables got deprecated and will be removed in future versions. Please migrate to the new names. Same applies for `DRONE_` of them.
     - CI_ARCH => use CI_SYSTEM_ARCH
     - CI_COMMIT => CI_COMMIT_SHA
+    - CI_TAG => CI_COMMIT_TAG
```
anbraten, line 32: I thought about that change while doing the env renaming as well, but then I decided to stick with `CI_TAG` as it is probably used already by some projects at least the `DRONE_TAG` and the renaming wasn't so beneficial IMO. I just forgot to revert my changes for `CI_COMMIT_TAG` and for `CI_COMMIT_PULL_REQUEST` as well. So maybe we should just adjust the env docs to stick to the old naming.

6543:
will close #498

docs/docs/91-migrations.md
6543: we have a fallback for now - but this version do break things - so I like the new syntax! - and it's already addopted in a lot of repos ... so just fill CI_COMMIT_TAG witch content and drop CI_TAG in v0.16.0 ...

6543: I'll add changes for`CI_COMMIT_PULL_REQUEST` too ...

anbraten (approved):

anbraten (approved):

PR #626, opened 2021-12-19
Add cache and webhook filter

### LINKED ISSUE #395: Filter dublicated webhooks by uuid
webhooks at gitea/github/? do have a uuid in header witch we could use to prevent dubble exec. since it look like there are edgecases who submit uuid twice

6543:
TODOs:
 - [x] define cache interface
 - [x] implement default ram cache
 - [ ] implement redis cache
 - [ ] use it to close #395


server/cache/mcache/setup.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (m mem) Get(_ context.Context, key string) (interface{}, bool, error) {
+	item, exist := m.cache.Get(key)
+	return item, exist, nil
+}
+
+func (m mem) Del(_ context.Context, keys ...string) error {
+	for i := range keys {
+		if err := m.cache.Set(keys[i], nil, time.Nanosecond); err != nil {
```
6543, line 27: better way?

anbraten:
I haven't noticed any issues regarding duplicate / wrong webhook submissions in the past. Afaik each time there was an issue with it somehow it was a bug or a duplicate webhook.

Issue #598, opened 2021-12-12
Add Unit tests for storage migration


6543:
* create dump of old DBs
* add tests

6543:
we have migration tests now for long time ...

PR #630, opened 2021-12-20
Add tests framework for storage migration

6543:
close partially #598

* [x] Sqlite3
* [ ] PostgreSQL
* [ ] MySQL


6543:
sqlite3 migration test is ready, should fail until #629 got merged 🚀

anbraten (approved):
lgtm. I checked the db file and it seems that you replaced all secrets / private data with samples, right?

6543:
Yes took me quite a while ... filterd al logs manualy, delete most repos and related stuf, replaced all sensitive data (by using random data ... and even generate a new ssh ky just replace ...)

So the size went down from 114M to 27M (in git 4M)

PR #632, opened 2021-12-20
Use Woodpecker theme colors on Safari Tab Bar / Header Bar

ktprograms:
Add theme-color <meta> tag and dynamically change the theme-color when
switching into / out of dark mode.

I used the same colors as in the Navbar component, but I don't think it's possible to access the Windi CSS color definitions, so I had to hard core the hex values.

The theme-color <meta> tag needs to be present in the HTML, otherwise `querySelector` returns null.

### DIFF
--- a/web/src/compositions/useDarkMode.ts
+++ b/web/src/compositions/useDarkMode.ts
@@ -7,9 +7,11 @@ watch(isDarkModeActive, (isActive) => {
   if (isActive) {
     document.documentElement.classList.remove('light');
     document.documentElement.classList.add('dark');
+    document.querySelector('meta[name=theme-color]')?.setAttribute('content', '#2e323e'); // dark-gray-900 (see windi.config.ts)
   } else {
     document.documentElement.classList.remove('dark');
     document.documentElement.classList.add('light');
+    document.querySelector('meta[name=theme-color]')?.setAttribute('content', '#65a30d'); // lime-600
   }
 });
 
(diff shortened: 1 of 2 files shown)


web/src/compositions/useDarkMode.ts
```diff
@@ -7,9 +7,11 @@ watch(isDarkModeActive, (isActive) => {
   if (isActive) {
     document.documentElement.classList.remove('light');
     document.documentElement.classList.add('dark');
+    document.querySelector('meta[name=theme-color]')?.setAttribute('content', '#2e323e');
   } else {
     document.documentElement.classList.remove('dark');
     document.documentElement.classList.add('light');
+    document.querySelector('meta[name=theme-color]')?.setAttribute('content', '#65a30d');
```
anbraten, line 14:
```suggestion
    document.querySelector('meta[name=theme-color]')?.setAttribute('content', '#65a30d'); // lime-600
```

web/src/compositions/useDarkMode.ts
```diff
@@ -7,9 +7,11 @@ watch(isDarkModeActive, (isActive) => {
   if (isActive) {
     document.documentElement.classList.remove('light');
     document.documentElement.classList.add('dark');
+    document.querySelector('meta[name=theme-color]')?.setAttribute('content', '#2e323e');
```
anbraten, line 10: Which windicss color is this? Could you add a comment with the color name?

anbraten (commented):
nice change

6543:
ci failed related: https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/1536/28

anbraten (approved):

PR #649, opened 2022-01-01
Show yml config of pipeline in UI

### LINKED ISSUE #89: Show yaml file in the UI
https://github.com/laszlocph/woodpecker-ui/issues/6

anbraten:
Screenshot from 2022-01-09 01-49-26

closes #89


6543:
the new tab ui, do add to much "overhead", in my opinion

can we move the show config into a symbol like we do for repo config?

anbraten:
I added the tabs as I had in mind to add more tabs like changed files, artifacts later, ... later.

6543:
I still think its a to big header:

| before | after |
|----------|--------|
| Bildschirmfoto zu 2022-01-02 15-41-49 | Bildschirmfoto zu 2022-01-02 15-42-13 |

what about a side bar & icons with mouse-hover-over description text ?

idea:
Bildschirmfoto zu 2022-01-02 15-55-31

6543:
do you have a better idea?

anbraten:
> do you have a better idea?

I would like to keep the tabs idea as we already used such a way of navigating before, but I will have a look at the row containing the build details (pusher, commit hash, build time, ...) and the top row with the back arrow to minimize the header areas height.

anbraten:
Screenshot from 2022-01-03 17-41-51

@6543 What do you think about this arrangement?

6543:
@anbraten way better - I would call it "slim" :)

6543 (approved):

PR #652, opened 2022-01-03
hotfix gitlab sub repo listing

anbraten:
This fix allows us to list sub-repos from Gitlab, so users get a list of all gitlab repos and can activate "normal" repos.
Enabling / using sub-repos is not solved by this PR and will reult in an error for now.

close  #651

### DIFF
--- a/server/remote/gitlab/convert.go
+++ b/server/remote/gitlab/convert.go
@@ -28,6 +28,7 @@ import (
 
 func (g *Gitlab) convertGitlabRepo(_repo *gitlab.Project) (*model.Repo, error) {
 	parts := strings.Split(_repo.PathWithNamespace, "/")
+	// TODO(648) save repo id (support nested repos)
 	var owner = strings.Join(parts[:len(parts)-1], "/")
 	var name = parts[len(parts)-1]
 	repo := &model.Repo{
--- a/server/remote/gitlab/convert.go
+++ b/server/remote/gitlab/convert.go
@@ -28,9 +28,8 @@ import (
 
 func (g *Gitlab) convertGitlabRepo(_repo *gitlab.Project) (*model.Repo, error) {
 	parts := strings.Split(_repo.PathWithNamespace, "/")
-	// TODO: save repo id (support nested repos)
-	var owner = parts[0]
-	var name = parts[1]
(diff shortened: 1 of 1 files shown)


server/remote/gitlab/convert.go
```diff
@@ -28,9 +28,8 @@ import (
 
 func (g *Gitlab) convertGitlabRepo(_repo *gitlab.Project) (*model.Repo, error) {
 	parts := strings.Split(_repo.PathWithNamespace, "/")
-	// TODO: save repo id (support nested repos)
```
6543, line 31: wen we leave this todo and linke an issue to it?

6543: https://github.com/woodpecker-ci/woodpecker/issues/654

server/remote/gitlab/convert.go
```diff
@@ -28,9 +28,8 @@ import (
...
-	var owner = parts[0]
-	var name = parts[1]
+	var owner = strings.Join(parts[:len(parts)-1], "/")
```
6543, line 31:
```suggestion
	// TODO(653) save repo id (support nested repos)
	var owner = strings.Join(parts[:len(parts)-1], "/")
```

6543:
@jdoubleu can you check this works for you as hotfix?

jdoubleu:
Yes, at least the sync looks like it's been working 👍

6543:
can we have both?

->  #652 &  #656

jdoubleu:
> can we have both?
>
> -> #652 & #656

As far as I understand, #652 makes #656 obsolete, because we can now properly handle nested repositories and no longer need to filter them out?

6543 (approved):

6543:
@anbraten :point_up: your pull make it not fail - should we filter them until #648 Y/N

I'm for also filtering them ...

anbraten:
It kind of works. The listing is fine, but if you want to enable the repo, you get an 404 error as it tries to load `/api/repo/group/sub-group/my-repo/activate`, but the route is only listening under `/api/repo/owner/my-repo/activate`.

anbraten:
I would merge #652 (this PR) and #656 as well.

PR #653, opened 2022-01-03
pass repo to remote perm func

anbraten:


server/remote/mocks/remote.go
```diff
@@ -214,12 +214,12 @@ func (_m *Remote) Netrc(u *model.User, r *model.Repo) (*model.Netrc, error) {
 }
 
 // Perm provides a mock function with given fields: ctx, u, owner, repo
-func (_m *Remote) Perm(ctx context.Context, u *model.User, owner string, repo string) (*model.Perm, error) {
```
6543, line 217: did you re-generate the mocks file?

anbraten: Ups did that by hand as i just searched for the string.

6543: done for u

6543 (approved):

PR #656, opened 2022-01-03
skip nested GitLab repositories during sync

### LINKED ISSUE #648: support nested repositories (GitLab)

jdoubleu:
Implements https://github.com/woodpecker-ci/woodpecker/pull/651#issuecomment-1003794674, partly fixes #648

### DIFF
--- a/server/remote/gitlab/gitlab.go
+++ b/server/remote/gitlab/gitlab.go
@@ -251,6 +251,7 @@ func (g *Gitlab) Repos(ctx context.Context, user *model.User) ([]*model.Repo, er
 				return nil, err
 			}
 
+			// TODO(648) remove when woodpecker understands nested repos
 			if strings.Count(repo.FullName, "/") > 1 {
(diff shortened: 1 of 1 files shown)


jdoubleu:
!652 makes this obsolete.

server/remote/gitlab/gitlab.go
```diff
@@ -249,6 +250,12 @@ func (g *Gitlab) Repos(ctx context.Context, user *model.User) ([]*model.Repo, er
 			if err != nil {
 				return nil, err
 			}
+
+			if strings.Count(repo.FullName, "/") > 1 {
```
6543, line 254:
```suggestion
			// TODO(648) remove when woodpecker understands nested repos
			if strings.Count(repo.FullName, "/") > 1 {
```

6543 (approved):

6543:
@jdoubleu thanks :)

PR #666, opened 2022-01-07
Add documentation of all server configuration options

KN4CK3R:
While setting up my instance I searched for an overview of all available configuration options but could not find any.

Please have a look at the items with `TODO`. I could not find any information for these items. Feel free to extend these.


docs/docs/30-administration/10-server-config.md
```diff
@@ -71,3 +71,475 @@ services:
...
+> Default: `false`
+
+Enables the GitHub driver.
+
+### `WOODPECKER_GITHUB_URL`
+> Default: `https://github.com`
+
+Configures the GitHub server address.
+
+### `WOODPECKER_GITHUB_CONTEXT`
```
6543, line 287: will be deleted

docs/docs/30-administration/10-server-config.md
```diff
@@ -71,3 +71,475 @@ services:
...
+> Default: empty
+
+Configures the Gitea OAuth client id. This is used to authorize access.
+
+### `WOODPECKER_GITEA_SECRET`
+> Default: empty
+
+Configures the Gitea OAuth client secret. This is used to authorize access.
+
+### `WOODPECKER_GITEA_CONTEXT`
```
6543, line 382: will be deleted

6543 (changes requested):

6543:
update?

anbraten:
@6543 I update the options a bit to match the newly added and removed flags. Two things I noticed:
- Should we move the configs for the SCMs to their appropiate pages and just link to them at the end of this docs page, so we prevent duplicates?
- I would like to leave out documentation for some settings at the moment which are probably going to need a refactor in the next versions as documenting things will more likely direct users into using those settings which could lead to more migration problems in the case we deprecate those settings. Currently those are:

```yml
# all of those settings are quite close to the docker backend, but by adding new backend 
# like podman and kubernetes we probably have to think about more generic options for them or need to 
# rename them to something like WOODPECKER_DOCKER_...
WOODPECKER_VOLUME
WOODPECKER_DOCKER_CONFIG
WOODPECKER_ENVIRONMENT
WOODPECKER_NETWORK

# I am personally quite unsure how to use these endpoints and if they are really a approach we
# should keep or which should better be redesigned
WOODPECKER_SECRET_ENDPOINT
WOODPECKER_REGISTRY_ENDPOINT
WOODPECKER_GATEKEEPER_ENDPOINT
```

KN4CK3R:
Should we keep the documentation description with the one in the code? In this case we could generate parts of the documentation.

6543 (approved):

PR #669, opened 2022-01-08
Add migration tests for postgres

6543:


server/store/datastore/migration/testfiles/postgres.sql
```diff
@@ -0,0 +1,1310 @@
...
+
+--
+-- Data for Name: secrets; Type: TABLE DATA; Schema: public; Owner: root
+--
+
+COPY public.secrets (secret_id, secret_repo_id, secret_name, secret_value, secret_images, secret_events, secret_skip_verify, secret_conceal) FROM stdin;
+1	105	wow	\\x74657374	null\n	["push","tag","deployment","pull_request"]\n	f	f
+2	105	n	\\x6e	null\n	["deployment"]\n	f	f
+3	105	abc	\\x656466	null\n	["push"]\n	f	f
+4	105	quak	\\x66647361	null\n	["pull_request"]\n	f	f
```
6543, line 907:
```suggestion
4	105	quak	\\x66647361	null\n	["pull-request"]\n	f	f
```

pat-s:
@6543 Not sure what the state here is, but do you wanna continue here or should this be closed?

6543:
the problem is the cleanup between database tests ...

6543:
the `cleanDB` does not work ... I need to check if we can drop the whole database and recreate ...

PR #673, opened 2022-01-08
Build ref link point to commit not compare if only one commit was pushed

### LINKED ISSUE #401: Build ref link should point to commit not compare if only on commit has changed
example:
get: `https://codeberg.org/6543/woodpecker/compare/97554816f4b927328ed523a6506aa38a1b99f039...d086e8268fa7713d72c6f1d03bab52f762c99325`
want: `https://codeberg.org/6543/woodpecker/commit/d086e8268fa7713d72c6f1d03bab52f762c99325`

6543:
close #401

### DIFF
--- a/server/remote/gitea/helper.go
+++ b/server/remote/gitea/helper.go
@@ -112,11 +112,13 @@ func buildFromPush(hook *pushHook) *model.Build {
 func getChangedFilesFromPushHook(hook *pushHook) []string {
 	files := make([]string, 0)
 
-	for _, c := range hook.Commits {
-		files = append(files, c.Added...)
-		files = append(files, c.Removed...)
-		files = append(files, c.Modified...)
+	if len(hook.Commits) == 0 {
+		return files
 	}
+
+	files = append(files, hook.Commits[0].Added...)
+	files = append(files, hook.Commits[0].Removed...)
+	files = append(files, hook.Commits[0].Modified...)
 
 	return files
 }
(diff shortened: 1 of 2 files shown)


server/remote/gitea/helper.go
```diff
@@ -84,15 +84,19 @@ func buildFromPush(hook *pushHook) *model.Build {
 	}
 
 	message := ""
+	link := hook.Compare
 	if len(hook.Commits) > 0 {
 		message = hook.Commits[0].Message
+		if len(hook.Commits) == 1 {
+			link = hook.Commits[0].URL
+		}
 	}
```
anbraten, line 93:
```suggestion
	}

  if len(hook.Commits) == 1 {
	  link = hook.Commits[0].URL
  }
```

server/remote/gitea/helper.go
```diff
@@ -107,14 +111,12 @@ func buildFromPush(hook *pushHook) *model.Build {
 func getChangedFilesFromPushHook(hook *pushHook) []string {
 	files := make([]string, 0)
 
-	if len(hook.Commits) == 0 {
-		return files
+	for _, c := range hook.Commits {
+		files = append(files, c.Added...)
+		files = append(files, c.Removed...)
+		files = append(files, c.Modified...)
 	}
```
anbraten, line 118: Good catch. I would like to extract it and check with other SCMs as well.

6543: well I'm not sure if we can extract it because hook format is different for each forge (most similarity's are shared between github and gitea ...)

anbraten: Sorry, meant extract into a separate PR not code-wise.

6543: @anbraten done

anbraten (approved):

PR #561, opened 2021-12-01
Improve status updates

anbraten:
closes #410 
closes #297 
closes #459
closes #521 

# Changes
- [x] link to specific proc (only general build before)
- [x] set status for all procs (before: only for the whole build on some SCMs)
- [x] set status after restart
- [x] set status to pending after waiting for approval
- [x] make status of gitlab, gitea & github equal
- [x] dedupe status update code
- [x] dedupe `PostBuild` code

# Tests
- [x] Gitea: restart build
- [x] Gitea: push to branch
- [x] Gitea: push to PR 
- [x] Gitea: multi-file pipeline
- [x] Gitea: single file pipeline
- [ ] Gitea: matrix pipeline
- [x] Gitlab: push to PR
- [x] Gitlab: push to branch
- [x] Github: push to branch
- [x] Github: push to PR

PR #674, opened 2022-01-08
Allow to change status context again

6543:
regression of #561


anbraten:
Why should a user be able to change it? I don't see any real benefit, so I would just remove that option completely.

6543:
-> edge-case: two woodpecker instances on same forge

cmd/server/flags.go
```diff
@@ -209,6 +209,12 @@ var flags = []cli.Flag{
 	//
 	// remote parameters
 	//
+	&cli.StringFlag{
+		EnvVars: []string{"STATUS_CONTEXT", "WOODPECKER_GITHUB_CONTEXT", "WOODPECKER_GITEA_CONTEXT"},
```
anbraten, line 213:
```suggestion
		EnvVars: []string{"WOODPECKER_STATUS_CONTEXT", "WOODPECKER_GITHUB_CONTEXT", "WOODPECKER_GITEA_CONTEXT"},
```

anbraten (commented):
Please update docs to match this setup as well and add a deprecation notice to the migration guide so we can remove the `WOODPECKER_GITEA_CONTEXT` and `WOODPECKER_GITHUB_CONTEXT` in the next version.

anbraten (approved):

anbraten (approved):

Issue #555, opened 2021-11-29
Lists of changed files of pull-requests are empty


anbraten:
### Component

server

### Describe the bug

```yml
pipeline:
  build:
    image: node:14-alpine
    commands:
      - cd docs/
      - yarn install --frozen-lockfile
      - yarn build
    when:
      event: [push, pull_request]
      path: "docs/**"
```

Step is executed although no file in `docs/**` has been changed.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-7931bc78"}
```

### Additional context

```shell
https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/1317
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy).

anbraten:
It seems to work in general. One problem I saw is that `changed_files` for a commit of a PR only includes the changes of that single commit instead of all changes of the PR. This is felt somehow wrong to me and does not allow to use path changes as for example the woodpecker project itself is currently trying to.

PR #675, opened 2022-01-08
getChangedFilesFromPushHook: use all commits

6543:
similar to #555 for gitea


6543:
ps: for pull requests we need: https://github.com/go-gitea/gitea/issues/18216

anbraten (approved):

PR #681, opened 2022-01-09
Hide deprecated env "WOODPECKER_DEBUG"

6543:
as title


anbraten:
Should we just remove that option completely? It would be a breaking change, but not that relevant as it is just used for debugging.

6543:
ok then lets remove it ...

PR #697, opened 2022-01-16
Support ChangedFiles for Github & Gitlab PRs and Gitlab pushes

anbraten:
closes #555 

load changed files for Github and Gitlab PRs
Gitea is waiting for https://github.com/go-gitea/gitea/pull/18228

TODOs:
- [x] test gitlab in real
- [x] update gitlab webhook fixtures

PR #701, opened 2022-01-17
Use GitHub lib to parse webhooks

6543:
depend on #697


6543:
(got included into  #697)

PR #711, opened 2022-01-20
Add note about Gitlab & Gitea internal connections to docs

jdoubleu:
This PR adds a little note to the docs to enable internal connections when the Woodpecker CI server and GitLab are run on the same host. This caused the webhooks to fail.

> [..] similar thing applys to newest gitea too

@6543 also mentioned this might apply to other VCS (e.g. Gitea). Should we check for all of them before merging this PR?

## Previous monologue

Right now I cannot sync any repo from GitLab with WP.
I get an error and the request (on WP) logs:
```
POST https://gitlab.example.com/api/v4/projects/192/hooks: 422 {error: Invalid url given}
```
At least this route seems to exist: https://docs.gitlab.com/ee/api/projects.html#add-project-hook

The message "Invalid url given" probably refers to the message body and not the request (URL) itself.
HTTP 422 is for "Unprocessable".

Okay, this fixed it: https://gitlab.com/gitlab-org/gitlab/-/issues/25867#note_215577922
Probably because I was running the Woodpecker Server and the GitLab Server on the same host.

### DIFF
--- a/docs/docs/30-administration/20-vcs/30-gitea.md
+++ b/docs/docs/30-administration/20-vcs/30-gitea.md
@@ -24,6 +24,14 @@ services:
 
 Register your application with Gitea to create your client id and secret. You can find the OAuth applications settings of Gitea at `https://gitea.<host>/user/settings/`. It is very import the authorization callback URL matches your http(s) scheme and hostname exactly with `https://<host>/authorize` as the path.
 
+If you run the Woodpecker CI server on the same host as the Gitea instance, you might also need to allow local connections in Gitea, since version `v1.16`. Otherwise API requests will fail. Add the following lines to your Gitea configuration (usually at `/etc/gitea/conf/app.ini`).
+```ini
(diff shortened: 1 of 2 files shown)


6543:
to be specific https://github.com/go-gitea/gitea/pull/17482 - gitea >= v1.16.0 do need extra settings if you like to run it on the same host

6543:
@jdoubleu would be nice if you could add it to the docs too. thanks!

jdoubleu:
I'll look into it. However, I might not be able to test & verify this, just gonna search for the settings in Gitea and add a note about it.

6543:
Oh you dont have to verify just add it into docs, i can ensure you it is that way ... 

https://docs.gitea.io/en-us/config-cheat-sheet/#webhook-webhook

6543 (approved):

6543:
thanks :)

Issue #684, opened 2022-01-12
Issues from rc1 run


6543:
- [x] Changed file filter shuld be ignored on tag event
       see: https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/1707/7
       => #698
- [x] Gitea with multible commits have dublicated changed files
- [x] building binaries is missing / broken after #622 , #635 & #634 
- [ ] ~~link docker images to release~~ -> manual / maintainer docu / template ...
- [x] pipeline for latest release branch

6543:
release helm do not work: https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/1772/10

6543:
@anbraten can you describe `link docker images to release` ?

anbraten:
![Screenshot from 2022-01-29 06-43-47](https://user-images.githubusercontent.com/6918444/151649151-7bdfb19e-0a4f-4333-ac01-7fa30738954c.png)

I had something like this text in mind, so a user directly nows the name of the new image.

6543:
ah ... that's more a manual thing for now - we could create some sort of template for maintainers ...

Issue #86, opened 2019-11-11
Path conditions per pipeline


laszlocph:
Trigger the pipeline if the git changeset is part of predefined paths.

6543:
-included in #283

strk:
Is path condition now supported ? Can you give an example of that ?

6543:
https://github.com/woodpecker-ci/woodpecker/blob/master/.woodpecker/test.yml#L27-L29

PR #722, opened 2022-01-29
Run pipeline for binary files on pull if go code changed

6543:
will resolve a todo on #684


.woodpecker/docker.yml
```diff
@@ -1,3 +1,7 @@
+depends_on:
```
6543, line 1: formation nit

.woodpecker/binaries.yml
```diff
@@ -0,0 +1,112 @@
+depends_on:
+  - test
+  - web
+
+pipeline:
```
6543, line 5: is 1:1 the `main.yml` just added path condition to each step

anbraten (commented):
Instead of `when.path.include` we should be able to use `when.path` directly. #86 will be super helpful such configs at some point. 😂

.woodpecker/binaries.yml
```diff
@@ -18,6 +18,8 @@ pipeline:
           # go source code
           - "**/*.go"
           - "go.*"
+          # web source code
+          - "web/"
```
anbraten, line 22: Sorry, I think we have to adjust this to `web/**`. I tried it by adding a new test case in `pipeline/frontend/yaml/constraint/constraint_test.go` which was throwing an error:

```go
		{
			conf: "'docs/'",
			with: []string{"docs/README.md", "docs/sub/README.md", "docs/sub-sub/README.md"},
			want: true,
		},
```

6543: Its blob pattern ...

I'll update it ...

6543:
PS: remove `include:` result in jsonschema report an error - I think we have to fix that ... but it's another pull to do s

anbraten (approved):

PR #727, opened 2022-01-30
Add step to release latest images for release branches

### LINKED ISSUE #684: Issues from rc1 run

6543:
close #684

### DIFF
--- a/.woodpecker/docker.yml
+++ b/.woodpecker/docker.yml
@@ -68,7 +68,7 @@ pipeline:
       branch: ${CI_REPO_DEFAULT_BRANCH}
       event: push
 
-  publish-branch-server:
+  publish-release-branch-server:
     image: plugins/docker
     group: docker
     secrets: [docker_username, docker_password]
@@ -80,7 +80,7 @@ pipeline:
       branch: release/*
       event: push
 
-  publish-branch-server-alpine:
+  publish-release-branch-server-alpine:
     image: plugins/docker
     group: docker
     secrets: [ docker_username, docker_password ]
@@ -158,7 +158,7 @@ pipeline:
       branch: ${CI_REPO_DEFAULT_BRANCH}
       event: push
 
-  publish-branch-agent:
+  publish-release-branch-agent:
     group: docker
     image: woodpeckerci/plugin-docker-buildx
     secrets: [docker_username, docker_password]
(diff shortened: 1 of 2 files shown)


.woodpecker/docker.yml
```diff
@@ -68,13 +68,38 @@ pipeline:
       branch: ${CI_REPO_DEFAULT_BRANCH}
       event: push
 
+  publish-branch-server:
```
anbraten, line 71: What do you think about naming this `publish-stable-server` or `publish-release-branch-server` or something different? The meaning of `branch` wasn't directly clear to me. Maybe a comment would be an alternative, if you want to stick with `publish-branch-server`.

.woodpecker/docker.yml
```diff
@@ -68,13 +68,38 @@ pipeline:
...
+      event: push
+
   release-server:
     group: docker
     image: plugins/docker
     secrets: [docker_username, docker_password]
     settings:
       repo: woodpeckerci/woodpecker-server
       dockerfile: docker/Dockerfile.server
+      # remove 'latest' on older version branches to avoid accidental downgrade
```
anbraten, line 102: This would happen if  we would tag on a release branch, right?

6543: yes but if there are more release branches lets say 'release/v0.15' & 'release/v0.16' and we would tag for example 'v0.15.3' after 'v0.16.2' we would have a downgrade on latest.

this is because we can not determine a branch for a tag event else i could add a when.branch filter too it ...

so we have to remove 'latest' tag on pipeline config on all release branches who are not latest release branch ...

.woodpecker/docker.yml
6543: this are just pipeline step names ... if you think us maintainers wont understand them - just propose a new name ...

anbraten: Okay, lets use: `publish-release-branch-server`

6543: done

anbraten (approved):

PR #730, opened 2022-01-31
Fix pipeline timestamps

### LINKED ISSUE #705: If pipelin fail, finish time is not saved to database
source: https://codeberg.org/Codeberg-CI/feedback/issues/27

```
sqlite> select build_status,build_created,build_started,build_finished from builds where build_repo_id = 4053 AND build_number=57;
build_status  build_created  build_started  build_finished
------------  -------------  -------------  --------------
failure       1637414240     1637414240     0             
sqlite> select build_status,build_created,build_started,build_finished from builds where build_repo_id = 4053 AND build_number=59;
build_status  build_created  build_started  build_finished
------------  -------------  -------------  --------------
success       1637428693     1637428694     1637429120
```

6543:
close #705

### DIFF
--- a/server/model/build.go
+++ b/server/model/build.go
@@ -28,7 +28,7 @@ type Build struct {
 	Error        string       `json:"error"                   xorm:"build_error"`
 	Enqueued     int64        `json:"enqueued_at"             xorm:"build_enqueued"`
 	Created      int64        `json:"created_at"              xorm:"build_created"`
-	Updated      int64        `json:"updated_at"              xorm:"updated 'updated' NOT NULL"`
+	Updated      int64        `json:"updated_at"              xorm:"updated NOT NULL DEFAULT 0 'updated'"`
 	Started      int64        `json:"started_at"              xorm:"build_started"`
 	Finished     int64        `json:"finished_at"             xorm:"build_finished"`
 	Deploy       string       `json:"deploy_to"               xorm:"build_deploy"`
--- a/server/model/build.go
+++ b/server/model/build.go
@@ -28,6 +28,7 @@ type Build struct {
 	Error        string       `json:"error"                   xorm:"build_error"`
 	Enqueued     int64        `json:"enqueued_at"             xorm:"build_enqueued"`
 	Created      int64        `json:"created_at"              xorm:"build_created"`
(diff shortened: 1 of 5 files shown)


server/model/build.go
```diff
@@ -28,6 +28,7 @@ type Build struct {
 	Error        string       `json:"error"                   xorm:"build_error"`
 	Enqueued     int64        `json:"enqueued_at"             xorm:"build_enqueued"`
 	Created      int64        `json:"created_at"              xorm:"build_created"`
+	Updated      int64        `json:"updated_at"              xorm:"updated NOT NULL DEFAULT 0 'updated'"`
```
anbraten, line 31: Could you explain why we need an additional `updated_at` from my understanding of the issue we would need to set `finished_at` after a pipeline was marked as failed.

6543: It's a kind a hotfix ... update will get updated by xorm (build in feature) on each insert/update

make finished_at set in each case would require a whole more code ... and we are in freeze
-> the issue is, that the gRPC Done do set this but only if not multipipeline aka procs list only contain one parent proc
-> else it cant set finished as it has no info of other pipelines can set finish or not (since we dont have a dependency graph based on procs)

6543:
@anbraten to be specific the underlaying issue is: https://github.com/woodpecker-ci/woodpecker/pull/730/files#diff-0f4ca4733649eb6707a0dd7e0ca0083cdc587b5cdced5b3ac051fc32cc9353cbR350-R353

witch would have to build the whole config again and again on each step update ... -> we need to move some pipeline info int model.proc (specific: the depend_on) ... but that's a refactor not a bugfix

server/model/build.go
anbraten: Hmm, that's a bummer.

To be able to remove the `updated_at` field easier again if needed after we properly fixed `finished_at` could you split this PR from the refactoring of the moved helper functions.

6543: and if we like to get rid of it, just delete the line ... + some migration to drop it

6543: so reverting via squash-commit wont be a good idea so why split this pull?
touched code is in the scope of this topic

6543:
I'll add a todo with link to this - so it dont get loosed in the void ...!

anbraten (approved):

Issue #468, opened 2021-10-24
Feedback: UI :speech_balloon:


anbraten:
Summary of UI issues and smaller changes (based on new UI from #245) which can be posted as comments. For complex improvements please open a separate issue.

# UI improvements
- [x] improve responsive (mobile) usage
- [x] improve darkmode favicon #576 
- [x] favicon to show pipeline status #576 
- [x] use textarea for secret values (to support long texts like ssh keys) #574 
- [x] show PR name / id instead of target branch if `ref: refs/pull/xxx/merge`
- [x] Show matrix options in build view #526 
- [x] add page `/<user-or-group-name>` to view all enabled woodpecker repos of a user / group
- [x] use icons for build proc status (currently just colored points)

# UI bugs
- [x] #469 
- [x] fix overflowing commit messages #528
- [x] fix back button (direct page load) to go back in history or to some default if history is empty #529 
- [x] fix new build updates not being used / send (new builds are not added list)
- [x] fix flickering when adding new log-lines while step is running

Next Round:  #1314

6543:
screenshot in README.md need an update ;)

HerHde:
`// The new UI looks great, thanks!`
Display the version number or hash somewhere (like `next-f00baa` for examle before)

HerHde:
The sidebar with the jobs on the right becomes scrollable for long words in commit messages.

![image](https://user-images.githubusercontent.com/7630662/141981066-bdfd92c7-2db1-48ec-a26b-ada7546bf2c3.png)

HerHde:
Change secret values from input fields to textareas again (or change it dynamically) to allow multi-line values like SSH-Keys again.

HerHde:
From a repo's builds (and also settings), provide a link back to the repository itself. Preferably with the name of the repo itself. Maybe as a breadcrumb navigation? The arrow back button only goes back in history.

catdevnull:
https://github.com/woodpecker-ci/woodpecker/pull/526 fixes Show matrix options in build view

anbraten:
> `// The new UI looks great, thanks!` Display the version number or hash somewhere (like `next-f00baa` for examle before)

I still have that below the logo. Which version do you use? Could you attach a screenshot pls.

catdevnull:
> I still have that below the logo. Which version do you use? Could you attach a screenshot pls.

Not who you answered to, but here's my Woodpecker on the `:latest` container tag:

[...]

PR #741, opened 2022-02-02
Add page to view all projects of a user / group

valentinleistner:
- adds a page /\<user-or-group-name\> to view all enabled woodpecker repos of a user / group
- also includes a link from the repo page to the owner page
- related to #468

https://user-images.githubusercontent.com/77448526/152249235-95bee6d6-81f4-40f6-8608-77d0f0361476.mov

### DIFF
--- a/web/src/views/ReposOwner.vue
+++ b/web/src/views/ReposOwner.vue
@@ -5,7 +5,7 @@
       <TextField v-model="search" class="w-auto md:ml-auto md:mr-auto" placeholder="Search ..." />
     </div>
 
-    <div>
+    <div class="space-y-4">
       <ListItem
         v-for="repo in searchedRepos"
         :key="repo.id"
--- a/web/src/views/ReposOwner.vue
+++ b/web/src/views/ReposOwner.vue
@@ -48,6 +48,7 @@ export default defineComponent({
 
   setup(props) {
     const repoStore = RepoStore();
+    // TODO: filter server side
     const repos = computed(() => Object.values(repoStore.repos).filter((v) => v.owner === props.repoOwner));
     const search = ref('');
 
--- a/web/src/views/ReposOwner.vue
+++ b/web/src/views/ReposOwner.vue
@@ -1,0 +1,63 @@
+<template>
+  <FluidContainer class="flex flex-col">
+    <div class="flex flex-row flex-wrap border-b pb-4 mb-4 items-center dark:border-dark-200">
+      <h1 class="text-xl text-gray-500">{{ repoOwner }}</h1>
+      <TextField v-model="search" class="w-auto md:ml-auto" placeholder="Search ..." />
+      <Button class="md:ml-auto" :to="{ name: 'repo-add' }" start-icon="plus" text="Add repository" />
+    </div>
+
+    <div class="space-y-4">
+      <ListItem
+        v-for="repo in searchedRepos"
+        :key="repo.id"
+        clickable
+        @click="$router.push({ name: 'repo', params: { repoName: repo.name, repoOwner: repo.owner } })"
+      >
+        <span class="text-gray-500">{{ `${repo.name}` }}</span>
+      </ListItem>
+    </div>
+  </FluidContainer>
+</template>
+
+<script lang="ts">
+import { computed, defineComponent, onMounted, ref } from 'vue';
+
+import Button from '~/components/atomic/Button.vue';
+import ListItem from '~/components/atomic/ListItem.vue';
+import TextField from '~/components/form/TextField.vue';
+import FluidContainer from '~/components/layout/FluidContainer.vue';
+import { useRepoSearch } from '~/compositions/useRepoSearch';
+import RepoStore from '~/store/repos';
+
+export default defineComponent({
+  name: 'ReposOwner',
+
+  components: {
+    Button,
+    FluidContainer,
+    ListItem,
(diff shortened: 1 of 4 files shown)


6543:
CI issue is unrelated ...

web/src/views/ReposOwner.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+  <FluidContainer class="flex flex-col">
+    <div class="flex flex-row flex-wrap border-b pb-4 mb-4 items-center dark:border-dark-200">
+      <h1 class="text-xl text-gray-500">{{ repoOwner }}</h1>
+      <TextField v-model="search" class="w-auto md:ml-auto" placeholder="Search ..." />
+      <Button class="md:ml-auto" :to="{ name: 'repo-add' }" start-icon="plus" text="Add repository" />
```
anbraten, line 6: I think we can remove this button from this page. As it could be confusing to a user having a look at some public org / other user and seeing a "I can add something to your org" button.

```suggestion
```

web/src/views/ReposOwner.vue
```diff
@@ -0,0 +1,63 @@
...
+
+    <div class="space-y-4">
+      <ListItem
+        v-for="repo in searchedRepos"
+        :key="repo.id"
+        clickable
+        @click="$router.push({ name: 'repo', params: { repoName: repo.name, repoOwner: repo.owner } })"
+      >
+        <span class="text-gray-500">{{ `${repo.name}` }}</span>
+      </ListItem>
```
anbraten, line 17:
```suggestion
      </ListItem>
      <div v-if="searchedRepos.length === 0">
        <span class="text-gray-500 m-auto">This organization / user does not have any projects yet.</span>
      </div>
```

anbraten (commented):
Nice PR and great work. Just for interest: Do you used our [UI development docs](https://woodpecker-ci.org/docs/development/ui) or did you managed to setup the woodpecker UI development on your own? Seems like you did some frontend dev before 😉 

We (I) need to test if / how this currently works for other users. Will a user easily see all public repos of a different owner?

valentinleistner:
> Just for interest: Do you used our [UI development docs](https://woodpecker-ci.org/docs/development/ui) or did you managed to setup the woodpecker UI development on your own?

I first cloned the repo and tried to run the UI from VS Code which worked after I had found out that I needend to install `yarn` by taking a look at `launch.json`. To connect the UI with a server I used the [documentation](https://woodpecker-ci.org/docs/development/ui). Since I was already running Woodpecker in Docker on my server I mostly copied the config to my local machine and added `WOODPECKER_DEV_WWW_PROXY=http://host.docker.internal:8010` (localhost doesn't work because the container has its own IP address) as an environment variable.

> Seems like you did some frontend dev before 😉

I have done some plain HTML, CSS and PHP (using no frameworks and almost no JS (even for things like [tabs](https://css-tricks.com/the-checkbox-hack/))), a bit UIKit, a half finished iOS (& macOS) app using SwiftUI and an almost finished Flutter app. I also have a project which I stared using plain HTML, PHP and JS, added Vue.js, then switched to [Flutter](https://flutter.dev) because at the time I was also working on the other Flutter app and I was planning to eventually build a mobile app anyways. When the Fluter app was almost in a usable state I found that the glitches and huge bundle size were not ideal for a browser first app which recently caused me to completely rewrite the frontend using [React Native](https://reactnative.dev) (using [Expo](https://expo.dev) for crossplatform on web, iOS and Android).

> Will a user easily see all public repos of a different owner?

I am currently using the same `RepoStore` and `useRepoSearch(repos, search)` as for the [existing repos page](), where repos is filtered by the owner. It might be possible to optimize this by only fetching the owner's repos (and not all other repos, too) from the server.

6543:
I'm okay on how it works, but like to refactor the getRepos api ... to have optional filters based on name ... owner ... usw

but that's for another pull :)

6543:
@valentinleistner I only found one nit, before its good to go :)

the space for a user/org repo list between items do not exist:
Bildschirmfoto zu 2022-02-04 13-49-51

should look like:
Bildschirmfoto zu 2022-02-04 13-49-39

anbraten:
The problem is a different here tough. The current implementation of the getRepos endpoint only returns repos / projects you own somehow, but in this case we need to get all public projects someone else owns. We may need to think about adding a new property `owned: true` or similar to the project. This would also help with pipeline runs of public repos appearing in your sidebar as there is a difference in I own the project and I can access the project as it is public or internal.

6543:
well if that's already n issue, we can fix it within this pull ...

web/src/views/ReposOwner.vue
```diff
@@ -0,0 +1,63 @@
...
+  props: {
+    repoOwner: {
+      type: String,
+      required: true,
+    },
+  },
+
+  setup(props) {
+    const repoStore = RepoStore();
+    const repos = computed(() => Object.values(repoStore.repos).filter((v) => v.owner === props.repoOwner));
```
6543, line 51:
```suggestion
    // TODO: filter server side
    const repos = computed(() => Object.values(repoStore.repos).filter((v) => v.owner === props.repoOwner));
```

6543:
I did add a todo ... as this refactor wont make it into v0.15.0 ... :)

6543 (approved):

PR #750, opened 2022-02-06
Allow specifying dind container in values

fpiesche:
This allows:

- resource spec for the dind container different from the main agent
- environment variables for the dind container can also be specified in values, e.g. to change the default driver if one so wishes
- crucially: specifying a different dind image

I'm finding myself needing to use an unofficial image instead of the standard `docker` one as the official ones have been broken on `armv7` (e.g. Raspberry Pi 2, or any Pi3/Pi4 running a 32bit operating system) since some time in the 0.19 series.

### DIFF
--- a/charts/woodpecker-agent/templates/deployment.yaml
+++ b/charts/woodpecker-agent/templates/deployment.yaml
@@ -57,7 +57,7 @@ spec:
                 name: {{ . }}
             {{- end }}
         - name: dind
-          image: {{ Values.dind.image }}
+          image: {{ .Values.dind.image }}
           env:
             {{- range $key, $value := .Values.dind.env }}
             - name: {{ $key }}
--- a/charts/woodpecker-agent/templates/deployment.yaml
+++ b/charts/woodpecker-agent/templates/deployment.yaml
@@ -57,12 +57,14 @@ spec:
                 name: {{ . }}
             {{- end }}
         - name: dind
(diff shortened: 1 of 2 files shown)


6543:
`helm lint` fail :/ @fpiesche

charts/woodpecker-agent/templates/deployment.yaml
```diff
@@ -57,12 +57,14 @@ spec:
                 name: {{ . }}
             {{- end }}
         - name: dind
-          image: "docker:20.10.12-dind"
+          image: {{ Values.dind.image }}
```
6543, line 60:
```suggestion
          image: {{ .Values.dind.image }}
```

charts/woodpecker-agent/templates/deployment.yaml
```diff
@@ -57,12 +57,14 @@ spec:
                 name: {{ . }}
             {{- end }}
         - name: dind
-          image: "docker:20.10.12-dind"
+          image: {{ .Values.dind.image }}
           env:
-          - name: DOCKER_DRIVER
```
6543, line 62: do we have to iterate through `.Values.dind.env` and not just use `toYaml` ?

6543: look's like it would make it more complicated that how it should be

6543 (approved):

PR #756, opened 2022-02-07
UI: show date time on hover over time items

### LINKED ISSUE #707: UI: show date time on hover over time items

When hover over these items (and perhaps in more places):
image 

image

be able to see the actual date time.

### Suggested solution

Typically, this is done on:hover or on:mouseover.

### Alternative

_No response_

### Additional context

_No response_

### Validations

9p4:
Adds a tooltip that shows the date and time when hovering over "started at" times.

tooltip

Resolves #707


9p4:
Maybe ISO standard for the time isn't the way to do this but it's trivial to change.

9p4:
Do you all want the duration to be like this too?

6543:
@9p4 you have to look at https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/1955/27 format and linting :)

9p4:
I've fixed everything except for the restriction on the CSS import. The recommended method of importing the tooltip with `import 'floating-vue/dist/style.css';` has the keyword `dist` in it, and is marked as an error by eslint. I can't find a way to integrate the default styles easily, so would it be fine if I just had eslint ignore that error?

web/src/components/build-feed/BuildFeedItem.vue
```diff
@@ -7,7 +7,7 @@
       <div class="flex flex-col mt-2">
         <div class="flex space-x-2 items-center">
           <Icon name="since" />
-          <span>{{ since }}</span>
+          <span v-tooltip="'Started at ' + started">{{ since }}</span>
```
anbraten, line 10: Could you try to mount the directives locally just for this component. (Next to `setup()`).

web/src/main.ts
```diff
@@ -1,10 +1,12 @@
 import 'windi.css';
+import 'floating-vue/dist/style.css';
```
anbraten, line 2: You can disable the eslint - dist for this line. Maybe you can try to add this to the component as well.

web/src/components/build-feed/BuildFeedItem.vue
9p4: `v-tooltip` is in three files (`web/src/components/build-feed/BuildFeedItem.vue`, `web/src/components/repo/build/BuildItem.vue`, and `web/src/views/repo/build/BuildWrapper.vue`) so I'd prefer to not have it specialized to those files and instead use a global import. Let me know if I'm misinterpreting what you're saying or if you want it changed regardless.

anbraten: Okay, we could to try to shrink it with `app.directive('tooltip', FloatingVue.VTooltip)` instead of `app.use(FloatingVue)`, but keep it globally.

9p4: If CI passes, then it's all good!

9p4: CI passed on my Woodpecker instance. Resolving.

web/src/compositions/useBuild.ts
```diff
@@ -97,5 +97,15 @@ export default (build: Ref<Build | undefined>) => {
...
-  return { since, duration, message, prettyRef };
+  const started = computed(() => {
+    if (!build.value) {
+      return undefined;
+    }
+
+    const start = build.value.created_at || 0;
+
+    // sv-SE is in format YYYY-MM-DD HH:m:s
+    return new Date(start * 1000).toLocaleString('sv-SE');
```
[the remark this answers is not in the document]
9p4: From my understanding, the started at time in MS is UTC, and the conversions should automatically use the proper time zone. I can test this once I get back from school.

web/src/compositions/useBuild.ts
```diff
@@ -97,5 +97,15 @@ export default (build: Ref<Build | undefined>) => {
     return build.value?.ref;
   });
 
-  return { since, duration, message, prettyRef };
+  const started = computed(() => {
+    if (!build.value) {
+      return undefined;
+    }
+
+    const start = build.value.created_at || 0;
```
6543, line 105:
```suggestion
    const start = build.value.started_at || 0;
```

web/src/compositions/useBuild.ts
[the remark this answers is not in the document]
9p4: Timezone is properly selected!

web/src/compositions/useBuild.ts
6543: hmm we might should rename it ... as time in ui reflects created

web/src/compositions/useBuild.ts
```diff
@@ -97,5 +97,15 @@ export default (build: Ref<Build | undefined>) => {
     return build.value?.ref;
   });
 
-  return { since, duration, message, prettyRef };
+  const started = computed(() => {
```
6543, line 100: call it "created" ?

web/src/views/repo/build/BuildWrapper.vue
```diff
@@ -51,7 +51,7 @@
         <div class="flex justify-between gap-x-4 text-gray-500 flex-shrink-0 pb-2 md:p-0 mx-auto md:mr-0">
           <div class="flex space-x-1 items-center flex-shrink-0">
             <Icon name="since" />
-            <span>{{ since }}</span>
+            <span v-tooltip="'Started at ' + started">{{ since }}</span>
```
6543, line 54:
```suggestion
            <span v-tooltip="'Created at ' + started">{{ since }}</span>
```
this is about creation timestamp

9p4: If it's about the creation timestamp, then why this change https://github.com/woodpecker-ci/woodpecker/pull/756#discussion_r802090254

6543: I just did a code-lookup my-selfe was not sure if it's about start or create

6543 (approved):

6543:
thanks :) - i would say - let it merge :rocket:

web/src/main.ts
```diff
@@ -13,6 +15,12 @@ const app = createApp(App);
 
 app.use(router);
 app.use(notifications);
+
+app.directive('tooltip', VTooltip);
```
anbraten, line 19: I just tested adding these on-demand which works quite nicely.

web/src/compositions/useBuild.ts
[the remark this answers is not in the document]
anbraten: I would prefer to show it in the users local time format instead of using `sv-SE` for all users here. I will have a short test as it seems to not work out of the box on my browser for example.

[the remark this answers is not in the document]
6543: follow up pull ?!?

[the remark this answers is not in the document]
9p4: I prefer sv-SE as it is the least ambiguous date format (`YYYY-mm-dd HH:m:s`)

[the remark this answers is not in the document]
6543: well I can understand both ... but in the end I think we have to respect what the user do have chosen on there system ...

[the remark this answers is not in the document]
9p4: I'll see if I can put something together to add a user setting for it that defaults to the browser's setting in another PR.

PR #635, opened 2021-12-21
Multiarch build server docker images

6543:

PR #759, opened 2022-02-08
Add support for building in docker

LamaAni:
As a developer I would like to build woodpecker on any environment

Added:
1. Make command to build woodpecker on docker (recreate make build environment).
1. Dockerfile for make environment


Makefile
```diff
@@ -152,6 +168,10 @@ bundle-cli: bundle-prepare
 
 bundle: bundle-agent bundle-server bundle-cli
 
+
+
 .PHONY: version
 version:
 	@echo ${BUILD_VERSION}
+
+endif
```
6543, line 177: new line missing

Makefile
```diff
@@ -152,6 +168,10 @@ bundle-cli: bundle-prepare
 
 bundle: bundle-agent bundle-server bundle-cli
 
+
```
6543, line 171: ?

Makefile
```diff
@@ -1,3 +1,19 @@
...
```
6543, line 1:
```suggestion
```

6543 (changes requested):

anbraten:
Isn't this somehow redundant to the work done in #635? We at least use a similar approach for building the [cli](https://github.com/woodpecker-ci/woodpecker/blob/master/docker/Dockerfile.cli.multiarch) and [agent](https://github.com/woodpecker-ci/woodpecker/blob/master/docker/Dockerfile.agent.multiarch) already.

Makefile
LamaAni: Updated

Makefile
```diff
@@ -1,3 +1,24 @@
...
+  # use the rest as arguments for "in_docker"
+  MAKE_ARGS := $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
+  # Ignore the next args
+  $(eval $(MAKE_ARGS):;@:)
+
+  .ONESHELL:
+  in_docker:
+	docker build -f ./docker/Dockerfile.make -t woodpecker/make:local .
+	docker run -it \
+		-e VERSION="$(VERSION)" \
```
6543, line 12: nice catch ... you just have to move the block who sets this before the if block:

https://github.com/woodpecker-ci/woodpecker/pull/759/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R25-R40

6543:
also is it possible to preserver the permissions from user exec that command. at the moment build artefacts are owned by root if run within docker :/

Makefile
LamaAni: I want the defaults to come from inside the docker environment, in case someone updates the docker image.

Not sure how to update the user though, but I had no issues with it so far.

6543: let me play around a bit ... I'll see if i can see some nits/downsides/improvements

LamaAni: Any update on this?

6543: Yes still on my list 😅 , its awesome how much time you can dedicate, i'll try to catch up 🚀

6543:
```sh
make version
> next-22ae7fb8
VERSION=1.0 make version
> 1.0
make in_docker version # expect same as before get ""
>
```

6543 (approved):

PR #763, opened 2022-02-08
Use `DOCKER_HOST` env to search for docker before searching for a hardcode docker socket

### LINKED ISSUE #757: Ability to set custom path to Docker socket

Set an environment variable to have the Docker client connect to a specific socket. This would enable people to use Podman as their runner (using the Docker compatibility socket). Currently, I have `/var/run/docker.socket` symlinked to the user's Podman socket `/run/user/1000/podman.sock`, but this is an inelegant solution at best.

### Suggested solution

Add an environment variable (maybe `WOODPECKER_DOCKER_SOCKET`) that, if set, can set the `addr` value in the Docker client (https://github.com/moby/moby/blob/88bd93fa603870897400b7d1f0c836d554caf85e/client/client.go#L73)

### Alternative

A dedicated Podman runner, but that would be much more difficult to use than using the Podman compatibility layer.

### Additional context

https://github.com/woodpecker-ci/woodpecker/issues/85

### Validations

mscherer:
Fix #757

### DIFF
--- a/pipeline/backend/docker/docker.go
+++ b/pipeline/backend/docker/docker.go
@@ -36,6 +36,9 @@ func (e *docker) Name() string {
 }
 
 func (e *docker) IsAvailable() bool {
+	if os.Getenv("DOCKER_HOST") != "" {
+		return true
+	}
 	_, err := os.Stat("/var/run/docker.sock")
 	return err == nil
 }


wxiaoguang:
The DOCKER_HOST could be a URL:

https://docs.docker.com/engine/security/protect-access/

```
$ export DOCKER_HOST=ssh://docker-user@host1.example.com
$ docker info
```

I think as long as the environment exists, we do not to check the existence again.

wxiaoguang:
Now it looks good 😊

6543 (approved):

PR #766, opened 2022-02-08
Follow-up for date tooltip

anbraten:
I changed the code to:
- ​load the Tooltip component on demand
- to show date in format same to Github and Gitea use it for commit ago time tooltips


anbraten:
@6543 I saw you dicussed with 9p4 to show the created time instead of the started time. Should we just show both times in two lines as we can now format the tooltip content?

6543:
@anbraten looking at the code ... we only show **created time** - showing both is bit to bloated - we could do something like show created time & tooltip "create at ..." and change that to **started time** & "started at ..." once we have a newer started timestamp ...

anbraten:
I only mean showing two lines inside the tooltip, not in general.

6543:
Does this overload ui?
Screenschots?

6543 (approved):

PR #769, opened 2022-02-09
Add support for default clone image environment variable

LamaAni:
As a CI pipeline admin I would like to set the default clone image (if otherwise not specified). This allows for custom clone images for deployment in air-gap systems.

Changes:
1. Added WOODPECKER_DEFAULT_CLONE_IMAGE


LamaAni:
This has not merged.. test not running.

6543:
just wait ... the job scheduler is a bit dumb atm ... (it just picks up a proc ..., we need to sort them - but that's another issue)

6543:
@LamaAni some objections holding merge back?

Issue #686, opened 2022-01-12
Allow multiple when filters


anbraten:
Make `root.when` and `root.pipeline.[step].when` an array, so a user could  current options using the current settings to allow ORs

current:
```yml
when:
  branch: main
  event: push
```

new:
```yml
when:
  - branch: main
    event: push
  - event: pull_request
    path: "src/*"
```

PR #770, opened 2022-02-09
Add support for pipeline root.when conditions

### LINKED ISSUE #283: Use `when` keyword for pipeline conditions
Currently there is a single condition for whole pipeline files named `branches`. As woodpecker will probably support more conditions in the future like #86 and to match the way conditions for pipeline-steps are working it would be nice to move the branch condition below the `when` keyword.

```diff
pipeline:
  build:
    image: golang
    commands:
      - go build
      - go test
   when:
     branch: [master, develop]

-branches: [ master, develop ]
+when:
+  branch: [master, develop]
```

### LINKED ISSUE #688: global pipeline filter should behave on multi-pipeline same
at the moment, if only "some" pipelines in multi pipelines are skiped, there proc tree is still build but never exec (as intended)

example: 
`.build.yml`
```yml
# These checks already get performed on the feature branches
branches:
  exclude: [ main, dev ]

pipeline:
  ...
```
`.lint.yml`
```yml
branches:
  exclude: [ main, dev ]

pipeline:
  lint:
    ...
```
`.publish.yml`
```yml
branches: [main, dev]

pipeline:
  dev:
    image: plugins/docker
    ...
```
-> https://git.rustybever.be/Chewing_Bever/vieter/commit/b5224e8a63de54c122f178bf0052a4029953c4cc

# Expect:
`.lint` & `.build`  should not show up in woodpecker

LamaAni:
close #283
close  #688

As a ci user I would like to set a file/yaml global when condition, to disable pipeline runs. This allows easier to write control for the pipeline events, branches and metadata.

# Changes
- [x] add root when conditions:
  - [x] branch
  - [x] repo
  - [x] event
  - [x] tag
  - [x] platform
  - [x] instance
  - [x] path
- [x] add docs
- [x] update yaml schema
- [x] wait for #686 

# Fixes 
1. In hook.go, when one of the yaml(s) was filtered by branch the hook would not fire. Changed to if all are filtered hook is not fired.
1. changed function name 'branchFiltered' -> 'checkIfFiltered'


pipeline/frontend/yaml/config.go
```diff
@@ -10,19 +10,21 @@ import (
...
-		Clone     Containers
-		Pipeline  Containers
-		Services  Containers
-		Networks  Networks
-		Volumes   Volumes
-		Labels    types.SliceorMap
-		DependsOn []string `yaml:"depends_on,omitempty"`
-		RunsOn    []string `yaml:"runs_on,omitempty"`
-		SkipClone bool     `yaml:"skip_clone"`
+		// TODO: change to When (also in Pipeline)
```
anbraten, line 13: What would you like to change to When? Constraints => When?

server/shared/procBuilder.go
```diff
@@ -106,6 +106,12 @@ func (b *ProcBuilder) Build() ([]*BuildItem, error) {
 				return nil, err
 			}
 
+			// checking if filtered.
+			if !parsed.Constraints.Match((metadata)) {
+				proc.State = model.StatusSkipped
+			}
+
+			// legacy check.
```
anbraten, line 114:
```suggestion
			// TODO: deprecated branches filter => remove after some time
```

server/api/hook.go
```diff
@@ -269,7 +269,7 @@ func PostHook(c *gin.Context) {
 }
 
 // TODO: parse yaml once and not for each filter function
-func branchFiltered(build *model.Build, remoteYamlConfigs []*remote.FileMeta) (bool, error) {
+func checkIfFiltered(build *model.Build, remoteYamlConfigs []*remote.FileMeta) (bool, error) {
```
anbraten, line 272:
```suggestion
// Check if at least one pipeline config will be execute otherwise we will just ignore this webhook
func checkIfFiltered(build *model.Build, remoteYamlConfigs []*remote.FileMeta) (bool, error) {
```

server/api/hook.go
```diff
@@ -186,23 +186,23 @@ func PostHook(c *gin.Context) {
...
 		msg := "failure to parse yaml from hook"
-		log.Debug().Err(err).Str("repo", repo.FullName).Msg(msg)
+		log.Info().Err(err).Str("repo", repo.FullName).Msg(msg)
 		c.String(http.StatusBadRequest, msg)
 		return
 	}
 	if filtered {
-		msg := "ignoring hook: branch does not match restrictions defined in yaml"
-		log.Debug().Str("repo", repo.FullName).Msg(msg)
+		msg := "ignoring hook: branch/event does not match restrictions defined in (any) yaml"
```
anbraten, line 197:
```suggestion
		msg := "ignoring hook: none of the pipelines of the project would be triggered by this event"
```

pipeline/frontend/yaml/compiler/compiler.go
```diff
@@ -81,6 +81,14 @@ func New(opts ...Option) *Compiler {
 func (c *Compiler) Compile(conf *yaml.Config) *backend.Config {
 	config := new(backend.Config)
 
+	if !conf.Constraints.Match(c.metadata) {
+		// basically dont do nothing. No need to run in this case.
+		// This would also remove this item from the UI
+		// which seems to be as the case for each of the container
+		// contraints.
```
anbraten, line 88:
```suggestion
		// This pipeline does not match the configured filter so return an empty config and stop further compilation.
		// An empty pipeline will just be skipped and wont be shown in the UI as well. 
```

server/api/hook.go
```diff
@@ -284,11 +284,26 @@ func branchFiltered(build *model.Build, remoteYamlConfigs []*remote.FileMeta) (b
...
+
+		// if was filtered by the constraints (branch) continue
+		if !parsedPipelineConfig.Constraints.Branch.Match(build.Branch) {
+			continue
+		}
+
+		// if was filtered by the branch (legacy) continue
+		if !parsedPipelineConfig.Branches.Match(build.Branch) {
+			continue
 		}
```
anbraten, line 300: Could we use something like the complete constraint checking as well? Or is this the problem that we haven't compiled the yml yet? We should probably find a cleaner way of doing this as otherwise a pipeline would be skipped by the code from `pipeline/frontend/yaml/compiler/compiler.go`, but would pass this code block.

```suggestion
		if !parsedPipelineConfig.Constraints.Match(c.Metadata) {
			continue
		}
```

anbraten:
@6543 From the current changes I don't think this would be breaking.

pipeline/frontend/yaml/config.go
LamaAni: Yea. Since you write that in the yaml as when, it should be that in the config. Otherwise its confusing. Also, the yaml parsing dose not follow standard yaml. I made an issue about that here: https://github.com/woodpecker-ci/woodpecker/issues/771

server/api/hook.go
LamaAni: Should have a comment. This will also appear elsewhere but its nice.

LamaAni: Wait. C. metadata is not valid here. So we cannot do that.

6543:
I'll look at this code diff asap I'm ready to release v0.15.0 :)

LamaAni:
I'm thinking of adding support for multiple root when conditions. E.g.
```yaml
when:
- event:
    - pull_request
    - release
- branch: master
```

This would allow both things to be captured, e.g. Either on master, or on pr and release.
@anbraten @6543 What do you think?

anbraten:
@LamaAni We thought about such a change as well #686. Would be quite helpful I think. I would suggest we try to merge this change and then change both root-level when and step-level when to be an array or at least to allow to enter an array somehow.

LamaAni:
@anbraten We can use a whenArray for now :) And just add it to both without breaking. 

As I suggested in the issue I think a redesign of the yaml config is necessary, mostly since it dose not follow yaml guidelines and treats some dicts as lists (with name).  Then, in yaml 2.0 we can do a proper fix for this.

LamaAni:
@anbraten @6543 Added support for when array. To allow multiple whens. This would be a temp step before cleaning it?

6543:
I'll shortly have a look ...

to the "whenArray" ... we could do this similar to how it's done with the "path" config ... first try to pars it as map ... witch should fail if its an array, then parse it as array ... save this as array internally in all cases ...

LamaAni:
@6543 I'll give that a go. And see what we get.

6543:
well, hope you dont mid ... I'll have a look at this pull after i got some sleep and finished another review :D

6543:
ok ... pipeline/schema/schema.json needs an update & we need good documentation :)

LamaAni:
> 

> ok ... pipeline/schema/schema.json needs an update & we need good documentation :)

Would you mind adding that @6543 I am not sure what the configuration looks like. Also, we should add the validation as an array no?

6543:
ok I'll push it to this branch ...

pipeline/frontend/yaml/compiler/compiler.go
```diff
@@ -82,6 +82,12 @@ func New(opts ...Option) *Compiler {
 func (c *Compiler) Compile(conf *yaml.Config) *backend.Config {
 	config := new(backend.Config)
 
+	if !conf.MatchConstraints(c.metadata) {
+		// This pipeline does not match the configured filter so return an empty config and stop further compilation.
```
6543, line 86: this is a nice catch ... but has one downside: on multi pipeline builds, if one pipeline depend on this, it will fail 
-> we have to determine if a pipeline with no steps failed or is successful and somehow represent this

6543: (can we move this in a own pull ...? )

LamaAni: Sure thing. I'll grab that once back at my terminal. Its just efficiency though, so you can just remove it.

LamaAni: Hum... @6543 I don't understand. I think this would work as expected. I am returning an empty config in the case where the pipeline dose not match therefore multiple pipelines should execute. The pipeline config will be filtered on no steps later on.

I am relying on the check of empty number of steps so it would run.

E.g. these two pipelines work as expected (tested), only on_push runs.

1 on_release.yml
```yaml
when:
  - event:
      - pull_request
      - release

pipeline:
  envs:
    image: ubuntu:latest
    commands:
      - env
```

2 on_push.yml
```yaml
# Will only run afer first, and in the case of a pull request or release.
when:
  - event:
      - push

pipeline:
  always_run_on_push:
    image: ubuntu:latest
    commands:
      - echo "Runs on push"
```

6543:
I'm going to split out the two changes ...
A: root.when
B: multi-when

anbraten:
> I'm going to split out the two changes ... A: root.when B: multi-when

#1087

6543:
I think this should do it, but I like to have some more unit tests for pipeline.when filters ...
... so when I do refactor It wont break

server/pipeline/filter.go
```diff
@@ -49,11 +50,17 @@ func zeroSteps(build *model.Build, remoteYamlConfigs []*remote.FileMeta) bool {
 }
 
 // TODO: parse yaml once and not for each filter function
-func branchFiltered(build *model.Build, remoteYamlConfigs []*remote.FileMeta) (bool, error) {
+// Check if at least one pipeline config will be execute otherwise we will just ignore this webhook
```
anbraten, line 53:
```suggestion
// Check if at least one pipeline step will be execute otherwise we will just ignore this webhook
```

pipeline/frontend/yaml/compiler/compiler.go
```diff
@@ -85,6 +85,12 @@ func New(opts ...Option) *Compiler {
 func (c *Compiler) Compile(conf *yaml.Config) *backend.Config {
 	config := new(backend.Config)
 
+	if !conf.When.Match(c.metadata, true) {
+		// This pipeline does not match the configured filter so return an empty config and stop further compilation.
+		// An empty pipeline will just be skipped and wont be shown in the UI as well.
```
anbraten, line 90:
```suggestion
		// An empty pipeline will just be skipped completely.
```

It wont be saved to db, shown in UI, ... => that's what skips means ;)

pipeline/frontend/yaml/constraint/constraint.go
```diff
@@ -58,17 +58,17 @@ func (when *When) IsEmpty() bool {
 }
 
 // Returns true if at least one of the internal constraints is true.
-func (when *When) Match(metadata frontend.Metadata) bool {
+func (when *When) Match(metadata frontend.Metadata, golbal bool) bool {
 	for _, c := range when.Constraints {
-		if c.Match(metadata) {
+		if c.Match(metadata, golbal) {
```
anbraten, line 63:
```suggestion
		if c.Match(metadata, global) {
```

pipeline/frontend/yaml/constraint/constraint.go
```diff
@@ -58,17 +58,17 @@ func (when *When) IsEmpty() bool {
...
 			return true
 		}
 	}
 
 	if when.IsEmpty() {
 		// test against default Constraints
 		empty := &Constraint{}
-		return empty.Match(metadata)
+		return empty.Match(metadata, golbal)
```
anbraten, line 71:
```suggestion
		return empty.Match(metadata, global)
```

pipeline/frontend/yaml/constraint/constraint.go
```diff
@@ -126,24 +126,21 @@ func (when *When) UnmarshalYAML(value *yaml.Node) error {
...
-	if c.Event.IsEmpty() {
-		c.Event.Include = []string{
-			frontend.EventPush,
-			frontend.EventPull,
-			frontend.EventTag,
-			frontend.EventDeploy,
-		}
+func (c *Constraint) Match(metadata frontend.Metadata, golbal bool) bool {
+	match := true
+	if !golbal {
```
anbraten, line 131:
```suggestion
	if !global {
```

server/pipeline/filter.go
```diff
@@ -64,10 +71,20 @@ func branchFiltered(build *model.Build, remoteYamlConfigs []*remote.FileMeta) (b
 		}
 		log.Trace().Msgf("config '%s': %#v", remoteYamlConfig.Name, parsedPipelineConfig)
 
-		if parsedPipelineConfig.Branches.Match(build.Branch) {
-			return false, nil
+		// check filtered by match constraints.
```
anbraten, line 74:
```suggestion
		// ignore if the pipeline was filtered by matched constraints
```

server/pipeline/filter.go
```diff
@@ -64,10 +71,20 @@ func branchFiltered(build *model.Build, remoteYamlConfigs []*remote.FileMeta) (b
...
+		if !parsedPipelineConfig.When.Match(matchMetadata, true) {
+			continue
+		}
+
+		// if was filtered by the branch (legacy) continue
```
anbraten, line 79:
```suggestion
		// ignore if the pipeline was filtered by the branch (legacy)
```

anbraten (approved):

6543:
conflicts ...

6543 (approved):

PR #777, opened 2022-02-14
Increase margin for "No pipeline steps available!"

qwerty287:
Just a simple UI improvement, set margin to `1rem` on the "No pipeline steps available!" message.

Without this patch:

With this patch:


6543 (approved):
Tanks for the screenshot

PR #780, opened 2022-02-21
View better error if repo was deleted/renamed

### LINKED ISSUE #726: Enable non existing / renamed repo fail with useless error
get: error: `json: cannot unmarshal array into Go value of type gitea.Hook`

expect: error: ` ... repo not found on forge ...`

qwerty287:
If the repo was renamed, there's an issue with Gitea: it redirects the /api/v1/repos/\<owner>/\<repo>/hooks POST request to a GET request at the same URL (returns a 302 Found which always redirects to GET - it should return a 307 or woodpecker needs to request again with the correct repo).
This URL returns the list of all hooks, thus the Gitea SDK can't parse the response into a single gitea.Hook type. The PR includes a workaround for this case.

A better error is also visisble if the repo was deleted.

Closes #726

### DIFF
--- a/server/remote/gitea/gitea.go
+++ b/server/remote/gitea/gitea.go
@@ -381,7 +381,12 @@ func (c *Gitea) Activate(ctx context.Context, u *model.User, r *model.Repo, link
 	if err != nil {
 		return err
 	}
-	_, _, err = client.CreateRepoHook(r.Owner, r.Name, hook)
+	_, response, err := client.CreateRepoHook(r.Owner, r.Name, hook)
+	if (err != nil && response.StatusCode == 200) || response.StatusCode == 404 {
+		// if repo was renamed, Gitea redirects as a GET request which
+		// results in an error of the SDK which can not parse the response
+		return fmt.Errorf("Could not find repository")
+	}
 	return err
 }
 


server/remote/gitea/gitea.go
```diff
@@ -381,7 +381,12 @@ func (c *Gitea) Activate(ctx context.Context, u *model.User, r *model.Repo, link
 	if err != nil {
 		return err
 	}
-	_, _, err = client.CreateRepoHook(r.Owner, r.Name, hook)
+	_, response, err := client.CreateRepoHook(r.Owner, r.Name, hook)
+	if (err != nil && response.StatusCode == 200) || response.StatusCode == 404 {
+		// if repo was renamed, Gitea redirects as a GET request which
+		// results in an error of the SDK which can not parse the response
+		return fmt.Errorf("Could not find repository")
```
wxiaoguang, line 388: If `err != nil`, should the returned error `fmt.Errorf` include the `err`?

qwerty287: I'll look into it

6543: I would do:
```go
if err != nil {
  if response != nil && response.StatusCode == 404 {
    return fmt.Errorf("Could not find repository")
  }
  return err
}

6543: redirects should have a different status code ...
you can chech tat also ...
```go
if err != nil {
  if response != nil {
    if response.StatusCode == 404 {
    return fmt.Errorf("Could not find repository")
    }
    if response.StatusCode == xyz {
      ... // or use switch case ...
    }
  }
  return err
}

qwerty287: Please take a look at https://github.com/woodpecker-ci/woodpecker/pull/780/commits/5393195038a95d1913b71b23fbe5fe7617d4829f

6543 (approved):

PR #784, opened 2022-02-23
Kubernetes kubectl backend [open to pick patches that improve native from it]

LamaAni:
Support for a backend using the `kubectl` executable.
Requires:
1. The executable `kubectl` must on both the agent and the server.

Supports:
1. Environment variables for configuration.
1. Network isolation and internet access.
1. Configuration control through [environment variables](#envs).
1. Container features:
   ```yaml
   commands: [string],
   image: string,
   environment: {},
   pull: bool,
   detached: bool,
   privileged: bool,
   alias: string,
   dns: [string],
   dnssearch: [string],
   ```
Does not support:
1. Local volumes.
1. Multiple volumes.
1. Volumes on detached steps (or services), without custom PVC class.
1. Configuration control through CLI commands.

See readme in backend folder for more help.


anbraten:
@LamaAni Awesome work again thank you very much. Super interesting to see this approach of using kubectl. Why did you prefer using kubectl instead of the go-api or some otherway? Did you had a look on #552 and if so is there a difference to that approach?

It contains some pretty nice changes apart from the backend, but I would like to separate some of them as they can probably be merged independently from this PR. Could you do that, or should I try that and link you.

LamaAni:
@anbraten TBH, I have not looked hard at that PR. But I was not going to "overtake" someones draft. Seemed poor manners. 

As for the other changes. Yea. we can separate them. Feel free to do that or I'll do that once I solve some small issues I have at the moment.

I wrote this because I needed a kubernetes backend more quickly with full single pipeline network isolation and proper resource cleanup (as dose exist in docker, due to my clients security concerns) where multiple players can only access their "own run" and the internal company network or the internet. You will also see in my implementation that I prefer a more "server side" configuration approach, where the configuration is blocked on the Kubernetes side (see the README.md file for envs and how to configure the network pipeline isolation). This allows me as admin to clear the CI against the network and code security requirements and SLA.

As for the kubectl question :)

I general, I would prefer the go native client, but it has some issues I know from another o.s. project I have (python client). Mainly, the application of a yaml configuration, e.g. you cannot just configure a yaml and expect the native client to execute it as you would with,
```shell
kubectl apply -f [file]
```

This means the configuration would become very hard and would require the handling of multiple resource types (Not to mention inflate the code).

For future use I wrote the "client" library in such a way that a future PR may replace it with a native client, one that accepts yaml. You can find a python example of such a clientI wrote, that accepts yaml, [here](https://github.com/LamaAni/KubernetesJobOperator/tree/master/airflow_kubernetes_job_operator/kube_api).

If anyone else is willing to write client that uses the go native client as a backend I am game! You can replace the kubectl with something more native. In the next two-three pushes I am gonna remove all dependency on kubectl from the other files and achieve client.go isolation, which will allow this.

I hope that helps :) and you guys would consider this as a basis.

6543:
please resolve conflicts :)

LamaAni:
> please resolve conflicts :)

Nice!

LamaAni:
@6543 Updated. See another fix, for errors in new PR.

.vscode/launch.json
```diff
@@ -39,4 +56,4 @@
       ]
     }
   ]
-}
+}
```
6543, line 59:
```suggestion
}

```

cli/exec/exec.go
```diff
@@ -167,7 +167,8 @@ func execWithAxis(c *cli.Context, file, repoPath string, axis matrix.Axis) error
 		compiler.WithPrefix(
 			c.String("prefix"),
 		),
-		compiler.WithProxy(),
+		// TODO: should not be here, or use flag. Confusing when local testing.
```
6543, line 170: looks unrelated

pipeline/pipeline.go
```diff
@@ -81,18 +81,69 @@ func (r *Runtime) Run() error {
 	return r.err
 }
 
-//
-//
-//
+// Updates the current status of a step
```
6543, line 84: this is also unrelated :sweat: 

can we pick out this changes first? :)

LamaAni: I'll seperate it once I have a minute

6543:
https://github.com/woodpecker-ci/woodpecker/pull/552 is prevered ... but some snipeds from here should be moved over

LamaAni:
Hi, well. I would like to keep this since I am still using this controller. I can keep it in my branch or keep this as legacy. The current implementation works well.

6543:
in this patch here are some bit's worth porting to the native implementation!

so as long as I did not have looked through this completely I'll keep it open, after I did so I'll notify via a comment here ...

PR #785, opened 2022-02-23
[DONT-MERGE] Multiarch build server anbraten

anbraten:
copied from #635 => pls update original PR and merge that one


Makefile
```diff
@@ -102,13 +102,14 @@ check-xgo:
...
 	tree dist
+	tree /build
 	@$(foreach platform,$(subst ;, ,$(PLATFORMS)),TARGETOS=$(firstword $(subst |, ,$(platform))) TARGETARCH=$(word 2,$(subst |, ,$(platform))) make normalize-server-artifacts || exit 1;)
 
 cross-compile-server-build-loop:
 	$(foreach platform,$(subst ;, ,$(PLATFORMS)),TARGETOS=$(firstword $(subst |, ,$(platform))) TARGETARCH=$(word 2,$(subst |, ,$(platform))) make release-server-xgo || exit 1;)
 
 normalize-server-artifacts:
-	mv dist/server/$(TARGETOS)/$(TARGETARCH)/$(shell ls dist/server/$(TARGETOS)/$(TARGETARCH)/) dist/server/$(TARGETOS)/$(TARGETARCH)/woodpecker-server
+	mv /build/woodpecker-server-$(TARGETOS)-$(TARGETARCH) dist/server/$(TARGETOS)/$(TARGETARCH)/woodpecker-server
```
relnod, line 112: Hi, thanks fpr trying my suggestions.

I didn't try it with the `PLATFORMS` specified in `.woodpecker/docker.yml`, which is why I didn't run into the same problem. The problem is, that xgo produces binaries, where the `TARGETARCH` is not exactly the same as specified. Especially annoying is `arm64/v8` becomes `arm64` in the output binary, whereas `arm/v7` becomes `arm-7`.

The patch below should fix that, although it feels kind of hacky
```patch
-       mv /build/woodpecker-server-$(TARGETOS)-$(TARGETARCH) dist/server/$(TARGETOS)/$(TARGETARCH)/woodpecker-server
+       mv /build/woodpecker-server-$(TARGETOS)-$(subst arm64/v8,arm64,$(subst arm/v,arm-,$(TARGETARCH))) dist/server/$(TARGETOS)/$(TARGETARCH)/woodpecker-server
```

relnod:
@anbraten Not sure why ci failed. Maybe it needs to be restarted?

PR #796, opened 2022-02-25
Collected Nits

6543:


6543:
code remove is split out to #797 ...

6543:
hmm well that's the only commit thing left at this "parking log"

so we could just open issues for the to do and close it

PR #797, opened 2022-02-25
Remove unused code

6543:


anbraten:
@6543 What do you think about removing all empty tests in this PR as well?

anbraten (approved):

PR #799, opened 2022-02-25
Use shared func for registering Sigterm on a context

6543:
pick from  #796


anbraten:
@6543 Look quite similar to a change included in #784 in 
- cmd/agent/agent.go
- cli/exec/exec.go
- agent/runner.go
- shared/utils/context.go

Could you have a look if it's related and maybe take changes from there to this PR.

6543:
well this was in the wrong place and i moved it ... but we have similar func for server and agent too  - so yes we can make a shared lib for that ...

cli/exec/interrupt.go
```diff
@@ -0,0 +1,42 @@
...
+	"os/signal"
+)
+
+// registerInterruptSignal returns a copy of parent context that is canceled when an os interrupt signal is received.
+// it will print this information to stdout
+func registerInterruptSignal(ctx context.Context) context.Context {
+	ctx, cancel := context.WithCancel(ctx)
+	go func() {
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt)
```
LamaAni, line 30: SigTerm should be here as well

LamaAni:
@anbraten  I agree this should be in the shared lib. And should be something like ContextWithSignalTermination.

I would also include, sigint sigterm in there for kubernetes purpose.

6543:
done

shared/utils/context.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// Returns a copy of parent context that is canceled when
+// an os interrupt signal is received.
+func WithContextSigtermCallback(ctx context.Context, f func()) context.Context {
+	ctx, cancel := context.WithCancel(ctx)
+	go func() {
+		recivedSignal := make(chan os.Signal, 1)
+		signal.Notify(recivedSignal, syscall.SIGINT, syscall.SIGTERM)
+		defer signal.Stop(recivedSignal)
```
LamaAni, line 31: Nice!

LamaAni (approved):

anbraten (approved):

PR #800, opened 2022-02-26
Get Netrc machine from clone url

anbraten:
We previously got the machine hostname for Netrc from the url of the remote, but in cases where the clone-url does not match the api url this can lead to errors.


6543 (approved):

6543:
this is a thing we can backport ...

6543:
-> https://github.com/woodpecker-ci/woodpecker/pull/803

PR #802, opened 2022-02-26
refactor pipeline compiling

anbraten:
- clean up types for internal representation of pipelines
  - one for the backend to be executable as step array
  - one to do further checks on
- migrate procBuilder into pipeline compiler & move all compiling related code into own package
- make pipeline compiler / procBuilder generic by using an interface to allow alternative implementations (new versions, different config languages, ...)
  - it should "simply" get some string and output an internal data representation 
- add some general utils for a compiled pipeline like zerosteps etc

### DIFF
--- a/server/pipeline/compiler/compiler.go
+++ b/server/pipeline/compiler/compiler.go
@@ -1,0 +1,6 @@
+package compiler
+
+type Compiler interface {
+	area() float64
+	perim() float64
+}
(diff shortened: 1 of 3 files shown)


6543:
@anbraten the image should go into docs -> development -> architecture

anbraten:
@6543 It currently only showing the path which is implemented atm. I mainly did this image as I was super confused how the yml files are converted. Its a hell of back and forth of data passing. My plan is to change it so you pass the configs as string and some metadata like repo, user, ... to a pipeline-builder and get back a model.Build for the database / UI and the finished config to execute (backend.Config).

anbraten:
closing for now as it is a quite complex topic

6543:
let's do it pice by pice :)

Issue #811, opened 2022-02-26
Yaml parse issues are only reported back to hook and not shown in the UI


s00500:
### Component

server

### Describe the bug

When a error exists in the commited yaml file this information is set back to the source control webhook... The build then never shows up and one has to check the source control webhook log to debug the issue...

One could argue that this is ok, but we could also just create a build that shows as failed and displays the error ? this to me seems like the expected behavior actually....

What do you think ? 

### System Info

```shell
Currently running master on my mac, agent also running right next to it

SCM: Gitea @ git.lbsfilm.at
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy).

PR #812, opened 2022-02-27
Show builds failed because of invalid yaml in woodpeckers webui

s00500:
I am not yet 100% sure but this feels like a nicer behavior, related to #811 

I think I need to play with it a little more still... What do you think ?


6543:
-> #1066

PR #815, opened 2022-02-28
Allow loading sensitive flags from files

s00500:
With systems like docker swarm or docker compose it is usually a little awkward to manage secrets. 
There is no way to directly inject them into the environment config. So you often have to write your secrets directly into the compose file

There are hacky workarounds such as overriding the entry-point of the container and loading a script which then fetches secrets from /run/secrets and replaces the environment variables, but this becomes very difficult once we are using docker images built from "scratch" (which is a really great practice otherwise) as there is no shell or standard tooling available

This is why containers like Postgres usually allow for a *_FILE variant of their Environment config values to work around this issue.

This PR basically does exactly the same

Let me know what you think of this, or if I have missed a more obvious solution to the problem

Greetings


s00500:
(Still need to add docs if you think this change is a good idea to add)

6543:
yes some docu would be nice

anbraten:
> yes some docu would be nice

I would suggest to add all new environment variables to `docs/30-administration/10-server-config.md` and a general information above explaining that `_FILE` is used for secrets to load those secrets from an file instead of using the env-varaible.

s00500:
Added some docs and cleaned up some variables

PR #817, opened 2022-02-28
Added support for step errors when executing backend

LamaAni:
When executing a backend step, in case of failure of the specific step, the run is marked as errored but the step error is missing.

Added:
1. Log for the backend error (without trace)
2. Mark the step as errored with exit code 126 (Could not execute).


agent/runner.go
```diff
@@ -98,6 +99,13 @@ func (r *Runner) Run(ctx context.Context) error {
 	ctx, cancel := context.WithTimeout(ctxmeta, timeout)
 	defer cancel()
 
+	// Add sigterm support for internal context.
+	// Required when the pipeline is terminated by external signals
```
6543, line 103: it's already registerd for the agent -> cmd/agent/agent.go

LamaAni: Dose not work here. Its registered from a background context. See: 
```golang
ctxmeta := metadata.NewOutgoingContext(context.Background(), meta)
```

anbraten: Does one of you know what line 66 does / is needed?

pipeline/pipeline.go
```diff
@@ -81,18 +81,69 @@ func (r *Runtime) Run() error {
...
+
+// Executes a set of parallel steps
+func (r *Runtime) execAll(steps []*backend.Step) <-chan error {
 	var g errgroup.Group
 	done := make(chan error)
 
-	for _, proc := range procs {
-		proc := proc
+	for _, step := range steps {
+		step := step
```
anbraten, line 119: `step := step` 🤔

anbraten (commented):
Does your change already solve "the problem" that an error is set to the parent proc / step and not to the step itself if for example the docker image wasn't found? Currently it does not set it on the step, but on the parent which makes it difficult to properly show it in the UI.

LamaAni:
> Does your change already solve "the problem" that an error is set to the parent proc / step and not to the step itself if for example the docker image wasn't found? Currently it does not set it on the step, but on the parent which makes it difficult to properly show it in the UI.
@anbraten 

It dose, since I am writing the error to the log of the step. You can try it :)

pipeline/pipeline.go
LamaAni: Yea :)

LamaAni:
@anbraten Fixed

pipeline/pipeline.go
LamaAni: WE do need it ... it seems go replaces the value of step from the loop.

anbraten:
@LamaAni I am still getting no error property for the step itself, but its parent. I think the problem is that we set the error by calling `rpc.Done()` for the parent step, but not when calling `rpc.Update()` for the step itself. Although your current change would allow us to see an error in the UI, I am not sure if it is the best idea to stream the error as log data. I would still prefer to have it assigned as error property to the step data:
Screenshot from 2022-03-01 23-01-55

LamaAni:
> @LamaAni I am still getting no error property for the step itself, but its parent. I think the problem is that we set the error by calling `rpc.Done()` for the parent step, but not when calling `rpc.Update()` for the step itself. Although your current change would allow us to see an error in the UI, I am not sure if it is the best idea to stream the error as log data. I would still prefer to have it assigned as error property to the step data: Screenshot from 2022-03-01 23-01-55

@anbraten In this case you would have to change the state.Error to match. That is, when an error occurs you would have to set that error, to the state instead of the rpc error. Line 104 in the file,
```golang
state.Pipeline.Error = r.err 
```
Change it to,
```golang
state.Pipeline.Error = r.err 
if err!=nil {
state.Pipeline.Error = err
}
```

I am not sure what are the implications of this. For me, I need the error to be shown in the UI and the execution and not in the proc iteslef (as long as its marked as errored). This is more important since it give the user a notice of what is wrong with the pipeline. IMO The is no reason not to print the error to the log, it gives more information and it is part of the pipeline.

Feel free to test the implication of setting this. If you like I'll add that to the PR.

pipeline/pipeline.go
6543: yes because the step var is used in a go call afterwards ...

anbraten:
Note: Should be working as originally planned and only waiting for https://github.com/LamaAni/woodpecker/pull/1 to be tested and merged

LamaAni:
@anbraten Sorry for the late responses. I am swamped at the moment. I'll try and do a woodpecker session on monday for cleanup for everything but the kubectl backend.

anbraten:
@6543 I added my code from https://github.com/LamaAni/woodpecker/pull/1 and merged master. It seems to work fine for me. I tested it with a woodpecker config using a none-existing container image.

PR #818, opened 2022-03-01
Improve agent backend loading and suppress expectable errors

anbraten:


pipeline/backend/docker/docker.go
```diff
@@ -232,3 +233,10 @@ var (
 		Timestamps: false,
 	}
 )
+
+func isErrContainerNotFoundOrNotRunning(err error) bool {
```
6543, line 237: it's ok to filter them out now ... but that's an open TODO ... since we should finde the root issue

anbraten: This is somehow expected. For example if we start a container with the command `exit 1` Kill will always run into a `container is not running issue`.

6543 (approved):

PR #824, opened 2022-03-01
fix redirect after login

anbraten:
fix a comment https://github.com/woodpecker-ci/woodpecker/issues/468#issuecomment-1027716962


6543:
@anbraten does it fix https://github.com/woodpecker-ci/woodpecker/issues/468#issuecomment-1027716962 ?

web/src/views/repo/RepoWrapper.vue
```diff
@@ -83,6 +85,12 @@ export default defineComponent({
     provide('builds', builds);
 
     async function loadRepo() {
+      // TODO: remove after we fixed #485 as we should then be able to have a look on public repos as well
+      if (!isAuthenticated) {
+        await router.replace({ name: 'login', query: { url: route.fullPath } });
+        return;
+      }
```
anbraten, line 92:
```suggestion
      // TODO: remove after we fixed #485 as we should then be able to have a look on public repos as well
      /* if (!isAuthenticated) {
        await router.replace({ name: 'login', query: { url: route.fullPath } });
        return;
      } */
```

6543:
oauth broke :/

6543:
oauth broken :/
Screenshot at 2022-03-24 02-58-17

anbraten:
@6543 Are you sure your error is related to this PR? I can't reproduce it.

6543:
yes same test instance, if i use master branch it works if i jump to this branch it fails.

I'll try to collect more info for  you since you can not reproduce it

anbraten:
@6543 I tested again using my local gitea setup and it works just perfectly. I don't see how the changes could even lead to your error as they do not really touch the oauth setup at all 🤔 

https://user-images.githubusercontent.com/6918444/166634649-edfdc14a-eb64-46fe-b57a-47167ea00852.mp4

anbraten:
> @anbraten https://cloud.obermui.de/s/T3jEYpPeo9b9yc6

Url looks fine. The error you have in your video at the beginning was expected. Should we remove it? Apart from that could you do the login and are you send to the page initially requested or are you still getting this  `OAuth Error` you posted before?

6543:
@anbraten it s a public repo so i expected no error

anbraten:
@6543 Do you have public in the project settings of the woodpecker project? For me this works fine.

6543:
is_private=false

anbraten:
> is_private=false

Which prop do you refer to? Repos in WP have `repo_visibility` which handles the visibility for WP (needs  to be `public` to see repo project logs / data without login) and we have  `repo_private` (`IsSCMPrivate`) which refers to the visibility of the repo in your forge which isn't used for the WP authentication only for clone at least if I am not mixing sth up here.

6543:
```
MariaDB [oo]> select repo_private,repo_visibility from repos where repo_full_name like 'hahaO/testCIservices';
+--------------+-----------------+
| repo_private | repo_visibility |
+--------------+-----------------+
|            0 |                 |
+--------------+-----------------+
1 row in set (0.000 sec)
```

6543:
I'll try to investigate on my own ...

6543 (approved):

PR #831, opened 2022-03-03
Add support for superseding runs

### LINKED ISSUE #11: Superseded builds - Skip outdated builds
If a new commit arrives to a branch, cancel the older build.

Today with quickly repeated commits and pushes, the CI builds for all commits. The goal would be to cancel the build running on the older commit, since it's not relevant anymore, just holds up resources.

LamaAni:
closes #11 

As an admin I would like not to allow multiple runs to execute, for specific events (ref matches)

Added support:
1. Environment variable `WOODPECKER_DELETE_MULTIPLE_RUNS_ON_EVENTS` (Default pull_request, push)
2. Builds will be marked as killed when they "override" another build

Screenshot from 2022-05-07 15-18-53


anbraten:
@LamaAni  Is this solving #11 ?

LamaAni:
Yea it does

anbraten:
@LamaAni Nice. What do you think about making it a repo setting instead of a global setting as in most cases I expect that it needs to be set differently for specific repos. Would you really need to set it as a server admin or is setting it as repo admin enough? If you want to be able to set if for a large amount of repos we could make that possible via the cli. 

Some context why i would like to make most similar settings a repo setting: Quite some Woodpecker installations are used by multiple parties (not just a single org / company / user) and for such installations enforcing such settings sometimes really limits the possibilities of those different parties to use WP for different use-cases.

LamaAni:
> @LamaAni Nice. What do you think about making it a repo setting instead of a global setting as in most cases I expect that it needs to be set differently for specific repos. Would you really need to set it as a server admin or is setting it as repo admin enough? If you want to be able to set if for a large amount of repos we could make that possible via the cli.
> 
> Some context why I would like to make most similar settings a repo setting: Quite some Woodpecker installations are used by multiple parties (not just a single org / company / user) and for such installations enforcing such settings sometimes really limits the possibilities of those different parties to use WP for different use-cases. So the general rule of thumb is adding settings on a per repo level and if needed adding an additional global setting via environment variable to enforce / overwrite it.

Yea. A repo setting would be much better but I am not familiar with your web design. I did need to work so not to overload my cluster, so I added it as a global value for now. We can add a setting later on and use the global configuration as a default. This way it would allow both. Feel free to add on to it or make a new PR, I would like that. If you do, please add a setting to disable the per repo option.

anbraten:
Awesome. I will then add the UI part to your PR.

cmd/server/server.go
```diff
@@ -287,6 +288,11 @@ func setupEvilGlobals(c *cli.Context, v store.Store, r remote.Remote) {
 	// Cloning
 	server.Config.Pipeline.DefaultCloneImage = c.String("default-clone-image")
 
+	// Execution
+	server.Config.Pipeline.DeleteMultipleRunsOnEvents = regexp.
```
6543, line 292: c.StringSlice ?

cmd/server/flags.go
```diff
@@ -270,6 +276,12 @@ var flags = []cli.Flag{
 		Name:    "github-skip-verify",
 		Usage:   "github skip ssl verification",
 	},
+	&cli.StringFlag{
```
6543, line 279: unrelated?

cmd/server/flags.go
```diff
@@ -97,6 +97,12 @@ var flags = []cli.Flag{
 		Name:    "authenticate-public-repos",
 		Usage:   "Always use authentication to clone repositories even if they are public. Needed if the SCM requires to always authenticate as used by many companies.",
 	},
+	&cli.StringFlag{
```
6543, line 100: cli.StringSliceFlag ?

server/store/datastore/build.go
```diff
@@ -80,6 +80,24 @@ func (s storage) GetBuildList(repo *model.Repo, page int) ([]*model.Build, error
...
+	if page > 0 {
+		query = query.Limit(perPage, perPage*(page-1))
+	}
+	return builds, query.Find(&builds)
+	// return builds, s.engine.
+	// 	Where("build_repo_id = ?", repo.ID).
+	// 	Where("build_status = ? or build_status = ?", "pending", "running").
+	// 	Desc("build_number").
+	// 	Limit(perPage, perPage*(page-1)).
+	// 	Find(&builds)
```
6543, line 98:
```suggestion
```
?

cmd/server/flags.go
LamaAni: Should be a string. Why use string slice?

6543:
@LamaAni just as a sidenote, if you like to know more about xorm: https://xorm.io/docs/chapter-13/readme/

6543 (approved):
code looks good - didn't hat time to test but I guess it already had 4 :eyes: ... so

cmd/server/server.go
anbraten: `StringSlice` is a way to get multiple options / strings from urfave-cli. I adjusted the code a bit to use it.

PR #842, opened 2022-03-19
Fix uppercase from_secrets

anbraten:
Secret names where matched based on their lowercase value already just the conversion to lowercase for `from_secrets` was missing.


6543 (approved):

6543:
please backport ;)

anbraten:
I will try to fix the image filtering in from_secret first and then backport both to 0.15

PR #843, opened 2022-03-19
Add test for e2e pipeline compilation

anbraten:
# Changes
- add test for end to end pipeline compilation
- return errors when compiling instead of only logging them


anbraten:
moved into #1572

PR #857, opened 2022-03-31
support docker credential and add docker-credential-ecr-login into alpine image

xuecanlong:
## Changes
- support docker credential and add docker-credential-ecr-login into alpine image

### DIFF
--- a/docker/Dockerfile.server.alpine.multiarch
+++ b/docker/Dockerfile.server.alpine.multiarch
@@ -1,6 +1,6 @@
 FROM alpine:3.14
 ARG TARGETOS TARGETARCH
-RUN apk add -U --no-cache ca-certificates
+RUN apk add -U --no-cache ca-certificates docker-credential-ecr-login
 ENV GODEBUG=netdns=go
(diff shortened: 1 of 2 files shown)


6543:
do you have a link to what this ecr-login is?

xuecanlong:
> * docker-credential-ecr-login

https://github.com/awslabs/amazon-ecr-credential-helper

xuecanlong:
> do you have a link to what this ecr-login is?

https://github.com/awslabs/amazon-ecr-credential-helper

anbraten:
I don't know if this should be part of the main container image as to be fair we would need to support Google, Azure and all the other hosters smaller ones as well. Could we add it as some kind of plugin?

xuecanlong:
> I don't know if this should be part of the main container image as to be fair we would need to support Google, Azure and all the other hosters smaller ones as well. Could we add it as some kind of plugin?

just need add the `docker-credential-*` in system, `CredentialHelpers` is a standard. https://github.com/docker/docker-credential-helpers

6543:
#1075 got merged

PR #851, opened 2022-03-26
Update docs

anbraten:
closes #849 

- migrate step conditions back into pipeline syntax, but show 2-4 level in toc to be able to see `when` keywords
- create new backend section in admin docs
- update docusaurus

PR #709, opened 2022-01-19
Add support to run pipelines using a local backend

Ta180m:
This adds support for #559. I tested using [this .woodpecker.yml](https://git.exozy.me/Ta180m/Hello-world/src/branch/main/.woodpecker.yml) on my self-hosted [Woodpecker instance](https://ci.exozy.me/Ta180m/Hello-world). I was also able to get this to build [Hugo websites](https://ci.exozy.me/Ta180m/howtuwu/build/1). It's currently very simplistic but works!

close #559

Issue #278, opened 2021-08-26
Allow to change entrypoint / shell for ci step


anbraten:
# Problem

Currently `/bin/sh` is used as entrypoint for every ci step. If an image does not contain `/bin/sh` it can not be used with woodpecker ATM. For example `gcr.io/kaniko-project/executor:debug` only ships with `/bin/busybox`. 

```go
// File: /pipeline/frontend/yaml/compiler/convert.go

entrypoint = []string{"/bin/sh", "-c"}
command = []string{"echo $CI_SCRIPT | base64 -d | /bin/sh -e"}
```

# Solution

Allow to exec multiple commands with the agent instead of a single combined one. This could allow to use the default entrypoint or even a specific one.  Commands could be executed with something like `docker exec` instead.

# Workaround

Create a plugin as plugins can define their entrypoint.

zc-devs:
As a temporary workaround we can try to create symbolic link:
```
pipeline:
  build-package:
    image: python:3.11.4-slim-bullseye
    commands:
      - ln -sf /bin/bash /bin/sh
```

6543:
Well from securety standpoint: as long as we handle it like normal steps there should not be a downside ... we just make sure that asap it is set ... it's not a plugin anymore

zc-devs:
[For linux-like OS entrypoint and command](https://github.com/woodpecker-ci/woodpecker/blob/cef135eba5f21a3e4ad429f9af51f535ce907372/pipeline/backend/common/script.go#L30) look like mentioned above:
```
env["CI_SCRIPT"] = base64.StdEncoding.EncodeToString([]byte(generateScriptPosix(commands)))
env["HOME"] = "/root"
env["SHELL"] = "/bin/sh"
entry = []string{"/bin/sh", "-c"}
cmd = []string{"echo $CI_SCRIPT | base64 -d | /bin/sh -e"}
```

So, we depend on some `shell` _and_ `base64`. Seems we can change shell only and hope image would contain `base64` command. Am I right? Things are getting worse with Windows.

If we want to execute the `command images` like a general `kaniko`, then we can put entrypoint used by image instead of shell like below:
```
  build-image:
    image: gcr.io/kaniko-project/executor
    entrypoint: /kaniko/executor
```
```
env["SHELL"] = ""
var stepEntrypoint string
entry = []string{stepEntrypoint}
cmd = []string{"echo $CI_SCRIPT"}
```
[...]

PR #861, opened 2022-04-02
Add SSH backend

### LINKED ISSUE #848: Add SSH backend

People miss this feature Drone CI has: https://docs.drone.io/runner/ssh/overview/

It allows you to connect to a system by ssh and execute a build on the default shell, without isolation.

### Suggested solution

A user should be able to add and choose agents for a repo.

### Alternative

People keep using Drone CI.

### Additional context

_No response_

### Validations

qwerty287:
Add SSH backend that runs commands via SSH. Uses https://github.com/melbahja/goph. Unfortunately, I don't know how to set the working directory for the commands, any ideas?

Closes #848


pipeline/backend/ssh/ssh.go
```diff
@@ -0,0 +1,141 @@
...
+func New() types.Engine {
+	return &ssh{}
+}
+
+func (e *ssh) Name() string {
+	return "ssh"
+}
+
+func (e *ssh) IsAvailable() bool {
+	return true
```
anbraten, line 43:
```suggestion
	return os.Getenv("WOODPECKER_SSH_KEY_PASSWORD") != "" || os.Getenv("WOODPECKER_SSH_PASSWORD") != ""
```

Something like this could be used.

docs/docs/30-administration/15-agent-config.md
```diff
@@ -153,4 +153,29 @@ Configures if the gRPC server certificate should be verified, only valid when `W
 ### `WOODPECKER_BACKEND`
 > Default: `auto-detect`
 
-Configures the backend engine to run pipelines on. Possible values are `auto-detect`, `docker`, or `local`.
+Configures the backend engine to run pipelines on. Possible values are `auto-detect`, `docker`, `local` or `ssh`.
```
anbraten, line 156: Maybe we should merge #851 first to be able to create a doc entry for this new backend as well. 

CC @6543

pipeline/backend/ssh/ssh.go
```diff
@@ -0,0 +1,141 @@
...
+	if file, has := os.LookupEnv("WOODPECKER_SSH_KEY"); has {
+		var err error
+		auth, err = goph.Key(file, os.Getenv("WOODPECKER_SSH_KEY_PASSWORD"))
+		if err != nil {
+			return err
+		}
+	} else {
+		auth = goph.Password(os.Getenv("WOODPECKER_SSH_PASSWORD"))
+	}
+	client, err := goph.New(os.Getenv("WOODPECKER_SSH_USER"), os.Getenv("WOODPECKER_SSH_HOST"), auth)
```
anbraten, line 57: We should add check for empty vars with proper error messages. Like you have to set either `WOODPECKER_SSH_KEY` or `WOODPECKER_SSH_PASSWORD` and always user and host.

pipeline/backend/ssh/ssh.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Get default clone image
+	defaultCloneImage := "docker.io/woodpeckerci/plugin-git:latest"
+	if len(server.Config.Pipeline.DefaultCloneImage) > 0 {
+		defaultCloneImage = server.Config.Pipeline.DefaultCloneImage
+	}
+
+	if proc.Image == defaultCloneImage {
+		// Default clone step
+		Command = append(Command, "CI_WORKSPACE=/tmp/woodpecker/"+proc.Environment["CI_REPO"])
```
mscherer, line 88: Using /tmp with a predictable name is a bad idea. If someone has a access to the server, that person could simply inject content in /tmp in various fashion, which in turn mean code execution under the user that is running the CI (and or inject content if the CI is used to make artifact).

It would also fail if 2 jobs are running at the same time, and trust me, that's always extra fun to debug.

docs/docs/30-administration/15-agent-config.md
```diff
@@ -153,4 +153,29 @@ Configures if the gRPC server certificate should be verified, only valid when `W
...
+
+### `WOODPECKER_SSH_HOST`
```
mscherer, line 158: I think using HOST is misleading. 

Since it also seems to support "ssh.example.org:2234", it would be more than the host, as it include the port. So either the post should be a separate variable, or it should reuse the same nomenclature as goph, eg, use address (or addr).

pipeline/backend/ssh/ssh.go
```diff
@@ -0,0 +1,141 @@
...
+func (e *ssh) Load() error {
+	var auth goph.Auth
+	if file, has := os.LookupEnv("WOODPECKER_SSH_KEY"); has {
+		var err error
+		auth, err = goph.Key(file, os.Getenv("WOODPECKER_SSH_KEY_PASSWORD"))
+		if err != nil {
+			return err
+		}
+	} else {
+		auth = goph.Password(os.Getenv("WOODPECKER_SSH_PASSWORD"))
```
mscherer, line 55: What about people using key + password as a form a poor-man 2 factor authentication ?
(that's supported since years: https://security.stackexchange.com/questions/17931/possible-to-use-both-private-key-and-password-authentication-for-ssh-login )

mscherer:
I also wonder how is the system dealing with host keys. Jenkins has several configuration options (trust on first us, manually approve, always approve, etc, etc). Here, the documentation do not really say what to do to approve a host, nor what to do if the ssh host key change.

pipeline/backend/ssh/ssh.go
```diff
@@ -0,0 +1,141 @@
...
+	return nil
+}
+
+// Exec the pipeline step.
+func (e *ssh) Exec(ctx context.Context, proc *types.Step) error {
+	// Get environment variables
+	Command := []string{}
+	for a, b := range proc.Environment {
+		if a != "HOME" && a != "SHELL" { // Don't override $HOME and $SHELL
+			Command = append(Command, a+"="+b)
```
mscherer, line 76: There should be quotes around b, and b should be properly escaped. Otherwise, a environment variable with space would result in weird failure, and potential code execution (albeit, if you control the environment, there is much easier way):
if PATH has /bin:/usr/local/directory with space/bin, it will be PATH=/bin:/usr/local/directory with space/bin command, eg, executing "with" as a command, and not command. 

I give PATH as a example, but on my current system, I have 3 env variable with spaces:

PERL_MB_OPT=--install_base "/home/user/perl5"
LESSOPEN=||/usr/bin/lesspipe.sh %s
which_declare=declare -f


LESSOPEN is set in /etc/profile.d/less.sh, from less-590-2.fc35.x86_64. 

I never really understood the semantics of profile vs bashrc and I run the agent with systemd, so the environment would be clean, but some people might not, and so might face weird issue due to that.

pipeline/backend/ssh/ssh.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+	}
+
+	// Get default clone image
+	defaultCloneImage := "docker.io/woodpeckerci/plugin-git:latest"
```
mscherer, line 81: The value shouldn't be hardcoded here, in case it need to be changed in the future, the default server value should be the default one.

pipeline/backend/ssh/ssh.go
mscherer: In fact, I wonder why this code is here. For example, this override HOSTNAME with the agent one, which would be surprising. It also override PATH or LANG, both variable that would have side effect on execution (for example, the side effect of translating error messages and/or changing sort order, as my team found out one day).

I am surely missing something, but could the reason for passing variable be explained a bit ?

mscherer:
This backend is also making platform match (eg, "when: platform: linux/amd64") less useful, since the platform is the one of the agent, not the one where the code is executed (which could be anything else).

While that's not a big deal, I suspect users might expect it to match the remote system and not the agent.

qwerty287:
Most parts of your review is already in `master` as backend `local`, which is the base for this.

mscherer:
Indeed. I think the code should be fixed on that backend too, especially the /tmp part. I didn't look at the code in details because I stopped at the UX (eg, the whole reuse of the image), and it was suddenly merged, but you are right, we need to fix on both side (and maybe refactor if that's doable).

mscherer:
The PR ( for reference: https://github.com/woodpecker-ci/woodpecker/pull/709 )

pipeline/backend/ssh/ssh.go
mscherer: So that part is changed in local.go with #868

pipeline/backend/ssh/ssh.go
qwerty287: I don't see any way to do this in goph or I didn't see it.

mscherer: I guess that's a rather niche use case. Maybe a upstream bug could be opened on it, and/or documented ?

qwerty287: https://github.com/melbahja/goph/issues/36

pipeline/backend/ssh/ssh.go
mscherer: It also seems to leak WOODPECKER_SSH_PASSWORD and WOODPECKER_SSH_KEY_PASSWORD to the remote system. It doesn't seems like a great idea, especially if you are running untrusted pipeline or something.

pipeline/backend/ssh/ssh.go
mscherer: That part is explained on #872

mscherer:
For the working directory question in the 1st comment, I see there is a option -C to env on my Linux system:
```  -C, --chdir=DIR      change working directory to DIR```

But that's not part of POSIX (so not on FreeBSD), and while the patch was added 5 years ago (https://github.com/coreutils/coreutils/commit/57dea5ed07471b2192cc5edf08993e663a3f6802 ), it is not present on RHEL 6 nor RHEL 7 systems. 

However, it is on Debian stable and Centos Stream 8.

**Comment by qwerty287** [comment]
I also thought about just adding a `cd ... && ` in front of the actual command (so the command will look like this with `bash`: `bash -c "cd ... && ...`. I think that worked when I tried it, but I don't know if this is POSIX-compatible.

**Comment by mscherer** [comment]
I think it would be POSIX compatible, but that requires a Unix shell. You are not going to run commands on something exotic like a Netapp filer or a windows machine with Powershell. That agent remind me a lot of Ansible who faced the same problems to run python on remote systems.

**Comment by qwerty287** [comment]
Sure but here I just copied the `local` backend again which also requires a Unix shell (i.e. this backend requires a Unix shell right now as well).

File: docs/docs/30-administration/15-agent-config.md (hunk 2 above)
Comment (6543) [reply docs/docs/30-administration/15-agent-config.md:156]: @qwerty287 @mscherer #851 got merged :)

would be nice if you could add some hints to this specific backend and describe where it should be used and where not

File: pipeline/backend/ssh/ssh.go (hunk 7 above)
Comment (mscherer) [reply pipeline/backend/ssh/ssh.go:85]: Reading the code a bit more, it seems I was wrong proc.Environment is not the same as os.Environment, so there is no variable leak. I however still think to verify the space issue is valid, since it get passed to a remote shell by ssh (and I think local backend should be ok, but I need to verify)

File: docs/docs/30-administration/15-agent-config.md (hunk 2 above)
Comment (qwerty287) [reply docs/docs/30-administration/15-agent-config.md:156]: https://github.com/woodpecker-ci/woodpecker/pull/861/commits/8f2941ff6480456fd520d905dd3132dab94502e6

File: pipeline/backend/ssh/ssh.go [hunk 9]
```diff
@@ -0,0 +1,150 @@
...
+	"encoding/base64"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/melbahja/goph"
+
+	"github.com/woodpecker-ci/woodpecker/pipeline/backend/types"
+	"github.com/woodpecker-ci/woodpecker/server"
```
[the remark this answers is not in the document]
qwerty287: https://github.com/woodpecker-ci/woodpecker/pull/861/commits/cdf99718e4a2ff78eba9191e982d51ec7a9a577e

pipeline/backend/ssh/ssh.go
qwerty287: I ported the changes to the `ssh` backend.

docs/docs/30-administration/22-backends/20-ssh.md
```diff
@@ -0,0 +1,11 @@
+# SSH backend
+
+The local backend will execute the pipelines using SSH on a remote system without any isolation of any kind.
```
mscherer, line 3: I think the backend name should be changed here.

docs/docs/30-administration/22-backends/20-ssh.md
```diff
@@ -0,0 +1,11 @@
...
+
+Since the code run directly in the same context as the agent (same user, same filesystem), a malicious pipeline could
```
mscherer, line 5: However, it doesn't run in the same context.

docs/docs/30-administration/22-backends/20-ssh.md
```diff
@@ -0,0 +1,11 @@
...
+
+The local backend will execute the pipelines using SSH on a remote system without any isolation of any kind.
+
+Since the code run directly in the same context as the agent (same user, same filesystem), a malicious pipeline could 
+be used to access the agent configuration especially the `WOODPECKER_AGENT_SECRET` variable.
+
+It is recommended to use this backend only for private setup where the code and pipeline can be trusted. You shouldn't
+use it for a public facing CI where anyone can submit code or add new repositories.
+
+The backend will use a random directory in $TMPDIR to store the clone code and execute commands.
```
mscherer, line 11: That's not the case at the moment.

qwerty287: Wasn't that done here: https://github.com/woodpecker-ci/woodpecker/pull/861/commits/2abd0743fed0e955f57cc360b065bc428dcd45eb?

docs/docs/30-administration/22-backends/20-ssh.md
qwerty287: I should have looked at the docs after I copied them before commiting 😆

docs/docs/30-administration/22-backends/20-ssh.md
mscherer: Strange, I didn't see that commit, guess there was a mixup in github notications or something, sorry. However, this create the directory on the agent, not on the system where ssh is run.

mscherer:
Also, while the issue is not just in this backend (but also on local), I wonder how  [services](https://woodpecker-ci.org/docs/usage/services) would work with this backend.  The services are implemented by using the image as the image of a container to use, and local (and this backend) use image as the path of the shell.

I am not sure what the best way to handle that. Either the backend should detect that and fail, or detect that and start a container. Or the code should stop using "image" as something that is not a image, and use "shell" (where it would have meaning for all backend).

pipeline/backend/ssh/ssh.go
qwerty287: If somebody would like to fix this (would recommend a separate PR since this is not limited to this backend), you can take a look at https://github.com/alessio/shellescape

qwerty287: If somebody would like to fix this (would recommend a separate PR since this is not limited to this backend), you can take a look at https://github.com/alessio/shellescape

qwerty287:
I tried to fix the workdir issue in https://github.com/woodpecker-ci/woodpecker/pull/861/commits/f2143b620042659c86481afb831e17e3c51b8bb8. It is using `cd ... && ...` scheme to change it.

pipeline/backend/ssh/ssh.go
```diff
@@ -47,11 +46,17 @@ func (e *ssh) IsAvailable() bool {
 }
 
 func (e *ssh) Load() error {
-	dir, err := ioutil.TempDir("", "woodpecker-local-*")
+	cmd, err := e.client.Command("/bin/env", "mktemp", "-d")
```
mscherer, line 49: I tend to prefer having a prefix for directory created by mktemp, ( just use "woodpecker-ssh-XXXXXXX" as a argument for mktemp, per the man pages), as it help to later diagnose if there is a problem (like Cleanup not cleaning the directories)

pipeline/backend/ssh/ssh.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Tail the pipeline step logs.
+func (e *ssh) Tail(context.Context, *types.Step) (io.ReadCloser, error) {
+	return e.output, nil
+}
+
+// Destroy the pipeline environment.
+func (e *ssh) Destroy(context.Context, *types.Config) error {
+	e.client.Close()
+	//_ = os.RemoveAll(e.cmd.Dir)
```
mscherer, line 147: The directory created for the test should be removed, as it might otherwise exhaust inodes on the server (in theory, cause I think in practice, that should be fine), or just use too much space in /tmp (as the source code is never erased)

6543 (approved):

pipeline/backend/ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+func (e *ssh) Name() string {
+	return "ssh"
+}
+
+func (e *ssh) IsAvailable() bool {
+	return os.Getenv("WOODPECKER_SSH_KEY") != "" || os.Getenv("WOODPECKER_SSH_PASSWORD") != ""
+}
+
+func (e *ssh) Load() error {
+	cmd, err := e.client.Command("/bin/env", "mktemp", "-d", "-p", "/tmp", "woodpecker-ssh-XXXXXXXXXX")
```
anbraten, line 49: @qwerty287 Am I right that the code would currently create a `/tmp/woodpecker-ssh-XXXXXXXXXX` folder once and place one folder per pipeline with the name of the pipeline inside of it?

pipeline/backend/ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+		Command = append(Command, "CI_WORKSPACE="+e.workingdir+"/"+proc.Environment["CI_REPO"])
+		Command = append(Command, "plugin-git")
+	} else {
+		// Use "image name" as run command
+		Command = append(Command, proc.Image[18:len(proc.Image)-7])
+		Command = append(Command, "-c")
+
+		// Decode script and delete initial lines
+		// Deleting the initial lines removes netrc support but adds compatibility for more shells like fish
+		Script, _ := base64.RawStdEncoding.DecodeString(proc.Environment["CI_SCRIPT"])
```
anbraten, line 112: @6543 We should really try to get rid of this single `Exec` call using this base64 command passed to sh and instead execute each command of a step separately 🙈 (see #278)

anbraten (commented):
Thanks for all the work and discussion you put into this. Apart from those tiny adjustments I am fine with merging it and trying it out as some kind of beta testing.

anbraten (approved):

PR #864, opened 2022-04-02
Cleanup models

anbraten:
I checked the database from ci.woodpecker-ci.org and noticed that quite some properties of the models are unused and other tables are not used at all.

# Changes
- [x] remove sender model (used for gated service to allow specific users to run pipelines)
- [x] remove file model (used / implemented for log file upload)
- [ ] remove unused properties from build model
- [ ]


server/model/file.go
```diff
@@ -1,47 +0,0 @@
...
-	Mime    string `json:"mime"    xorm:"file_mime"`
-	Time    int64  `json:"time"    xorm:"file_time"`
-	Passed  int    `json:"passed"  xorm:"file_meta_passed"`
-	Failed  int    `json:"failed"  xorm:"file_meta_failed"`
-	Skipped int    `json:"skipped" xorm:"file_meta_skipped"`
-	Data    []byte `json:"-"       xorm:"file_data"` // TODO: dont store in db but object storage?
-}
-
-// TableName return database table name for xorm
-func (File) TableName() string {
```
6543, line 45: drop table from db?

PR #868, opened 2022-04-04
Do not hardcode default in several place

mscherer:
Since DefaultCloneImage is always defined by the code, there is no
need to have a 2nd place where the default is defined.

Mentionned on #861


6543:
Uh there was a reason for this if statement ... let me check first

6543:
see #873 for the whole explanation :) - and thanks for finding this issue!!!

mscherer:
So to be clear, i should close that PR ?

6543:
You can otherwhise it will be done automaticaly once refered pr got merged

PR #872, opened 2022-04-05
Fix insecure /tmp usage in local backend

mscherer:
Since /tmp is writable by everybody, a user could precreate
/tmp/woodpecker with 777 permissions, allowing them to modify the
pipeline while it is being run, or preventing the pipeline from running.

And since os.MkdirAll error code wasn't checked, the same attacker
could have precreated the directory where the pipeline is executed to
mess with the run, allowing code execution under the UID of the
agent (who has access to the toke, to communicate with the server, which
mean a attacker could inject a fake agent, steal credentials, etc)

### DIFF
--- a/pipeline/backend/local/local.go
+++ b/pipeline/backend/local/local.go
@@ -3,18 +3,19 @@ package local
 import (
 	"context"
 	"encoding/base64"
+	"github.com/woodpecker-ci/woodpecker/pipeline/backend/types"
+	"github.com/woodpecker-ci/woodpecker/server"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
-
-	"github.com/woodpecker-ci/woodpecker/pipeline/backend/types"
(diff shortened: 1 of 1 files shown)


6543:
"local backend" will not be as secure as the container - it's more for local deployments and to tinker around

but if it can be fixed we should do so

pipeline/backend/local/local.go
```diff
@@ -3,18 +3,19 @@ package local
 import (
 	"context"
 	"encoding/base64"
+	"github.com/woodpecker-ci/woodpecker/pipeline/backend/types"
```
6543, line 6: this will likely hit the linter

just use `gofumpt -w --extra .` to make sure ;)
-> in this just commit changes of the result and push - we will squashmerge pulls

mscherer: I used gofmt -w, but indeed the change was suspect.

6543 (approved):

PR #874, opened 2022-04-06
WIP: add dependency checks to linter

6543:
try to avoid things like: #873


xoxys:
Sure I can. But maybe we should even split the other PR into smaller pieces to get it reviewed/merded faster?

Issue #778, opened 2022-02-14
Do not run clone step if no pipeline step will run


qwerty287:
### Clear and concise description of the problem

Woodpecker clones the repository for every pipeline, even if no step will run. Woodpecker's own CI is doing this every time on PRs, e.g. https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/2025/22 clones the repo, but no steps will run.

### Suggested solution

To save time and resources (#772), it would be good to just skip the step and hide the pipeline in the UI or even skip the pipeline completely.

### Alternative

You could use a custom clone step that only runs on some conditions.

### Additional context

For an example, see https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/2025/22

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

6543:
we have to handle the case that the pipeline should still be reported as "successfull" even if nothing will run ... as it would do now

xuecanlong:
> we have to handle the case that the pipeline should still be reported as "successfull" even if nothing will run ... as it would do now

but if we skip clone, and no pipeline step, what will report? i support if no pipeline step then skip clone

qwerty287:
A proper implementation would report a "success" state back (it is successful, so this should be reported back to the forges).

xuecanlong:
> A proper implementation would report a "success" state back (it is successful, so this should be reported back to the forges).

but why we don't skip clone, just report success. for this case, clone isn't necessary

qwerty287:
That's what this issue is for.

6543:
reopend by e2e094cfda87becedf7a37d6077fabea2c97ebb5

Powersource:
Still getting this

runiq:
Still getting this here too.

6543:
## workaround

define the clone step as default would do and the add an when filter to it ... 
... and a second clone step that is a dummy that just prints out success ;)

## state

this is a long standing issue that is because this is wanted:
 -> workflow when filter filters out workflow -> no status report
 -> all steps are filtered out but workflow is not -> should still report status as success (usefull for branch protection rules to require it etc...)

[...]

PR #877, opened 2022-04-09
Do not run clone step if no pipeline step will run

qwerty287:
Skip the clone step and ignore hook/pipeline if no pipeline step except clone would run. The status reported back to the forge is `success`.

Closes https://github.com/woodpecker-ci/woodpecker/issues/778


anbraten:
> With one pipeline, an issue occurs: the time says not started yet because the duration is 0 which results in displaying this message. I don't know a simple way to fix it. Also, the large white "box" next to the pipeline list looks a bit odd.

I improved the "white box" a bit in #846. Maybe that adjustment would be helpful for you in this PR as well.

What do you think about hiding empty pipelines and maybe the complete "build" in case they are empty as we are doing a similar thing for pipelines steps which are not executed => they are not shown to the user as well?

qwerty287:
Yeah I thought about this too but wasn't able to implement it after some experiments (I don't know any web frontend dev, which I'd need to add such a weird `v-if` statement to the vue files, sorry). Hiding the complete build would be possible, then an easy solution would also to just return if coming from the hook (i.e. don't initiate a build). This wouldn't report a status back to the forge, which can be problematic if using pull requests that require the checks. Hinding it only from the list is a good idea, but would require more effort which I'm not able to do because I don't understand woodpecker's structure enough (and don't vue in general).

6543:
I would: if there are no steps, dont add clone but report success back

and only if global pipeline filter #when (#283 ) do filter out make no response

agent/runner.go
```diff
@@ -75,6 +75,16 @@ func (r *Runner) Run(ctx context.Context) error {
 		return nil
 	}
 
+	if len(work.Config.Stages) < 1 {
```
6543, line 78: I dont think this should affect the agent at all - it's a pipeline compiler thing

anbraten: You are right it should work without.

6543:
> After the commits I pushed now it just skips and ignores the hook. Should it send a "success" status back and should this always happen for empty builds?

If no steps but pipeline is not filtered golbaly -> success
if pipline is filtered globaly -> none

server/api/hook.go
```diff
@@ -200,10 +200,18 @@ func PostHook(c *gin.Context) {
 		return
 	}
 
-	if zeroSteps(build, remoteYamlConfigs) {
+	if z, steps := zeroSteps(build, remoteYamlConfigs); z {
```
6543, line 203:
```suggestion
	// TODO: move global pipeline filters into own check functions ...
	if z, steps := zeroSteps(build, remoteYamlConfigs); z {
```

server/api/hook.go
```diff
@@ -200,10 +200,19 @@ func PostHook(c *gin.Context) {
...
 		msg := "ignoring hook: step conditions yield zero runnable steps"
 		log.Debug().Str("repo", repo.FullName).Msg(msg)
 		c.String(http.StatusOK, msg)
+		build.Status = model.StatusSuccess
+		for _, step := range steps {
+			step.Proc.State = model.StatusSuccess
+			build.Procs = append(build.Procs, step.Proc)
+		}
+		if err := updateBuildStatus(c, build, repo, repoUser); err != nil {
```
6543, line 213:
```suggestion
		// TODO: this wont create any builds the status can link to ...
		if err := updateBuildStatus(c, build, repo, repoUser); err != nil {
```

6543 (approved):
some todos ... but that should be covered in another pull (refactoring etc ...)

PR #886, opened 2022-04-19
Initial version of CLI docs

jolheiser:
I think in the long run this may be a bit verbose, but for the time being it's ostensibly better than TODO :wink:

### DIFF
--- a/cmd/cli/docs.go
+++ b/cmd/cli/docs.go
@@ -1,0 +1,64 @@
+// Copyright 2021 Woodpecker Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


cmd/cli/docs.go
```diff
@@ -0,0 +1,64 @@
...
+	app := newApp()
+	for _, cmd := range app.Commands {
+		fixHiddenFlags(cmd)
+	}
+	md, err := app.ToMarkdown()
+	if err != nil {
+		panic(err)
+	}
+	// Still a bug in our version of urfave/cli/v2
+	// https://github.com/urfave/cli/pull/1311
```
anbraten, line 38: Should we update the cli lib?

anbraten (approved):

cmd/cli/docs.go
jolheiser: May as well wait for the other PR to be merged, but yeah.
I also wanted it to be a separate PR just to keep this one contained.

PR #889, opened 2022-04-23
fix key/val format for dind env vars

### LINKED ISSUE #866: Erroneous env var definition in the woodpecker agent Helm chart
### Component

agent

### Describe the bug

In `charts/woodpecker-agent/values.yaml`, `dind.env` variables [are defined](https://github.com/woodpecker-ci/woodpecker/blob/c310363c99d5b1422a9abfeba1a74d8b520a50a0/charts/woodpecker-agent/values.yaml#L12-L14) using a list of `{name,value}` objects. However, the template defined in `woodpecker-agent/templates/deployment.yaml` expects `dind.env` to be a key-value map instead.

This results in the following error when trying to install the chart:

```
Error: INSTALLATION FAILED: Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.SecurityContext: Resources: Name: Image: Env: []v1.EnvVar: v1.EnvVar.Value: Name: ReadString: expects " or n, but found 0, error found in #10 byte of ...|[{"name":0,"value":"|..., bigger context ...|:"/var/run","name":"sock-dir"}]},{"env":[{"name":0,"value":"map[name:DOCKER_DRIVER value:overlay2]"}|...
```

This may be fixed by specifying the default values as a key-value map ([the way the `env` value is defined](https://github.com/woodpecker-ci/woodpecker/blob/c310363c99d5b1422a9abfeba1a74d8b520a50a0/charts/woodpecker-agent/values.yaml#L17-L18)) instead:

```yaml
dind:
  env:
    DOCKER_DRIVER: overlay2
```

I can provide the patch :)

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"0.15.0"}
```

### Additional context

_No response_

### Validations

IDerr:
Fixes the key/value format used in the agent values for Helm
Should fix the issue (https://github.com/woodpecker-ci/woodpecker/issues/866)

Closes #866

### DIFF
--- a/charts/woodpecker-agent/values.yaml
+++ b/charts/woodpecker-agent/values.yaml
@@ -10,8 +10,7 @@ image:
 dind:
   image: "docker:20.10.12-dind"
   env:
-  - name: DOCKER_DRIVER
-    value: overlay2
+    DOCKER_DRIVER: "overlay2"
   resources: {}
 
 env:


6543:
->  #890

PR #893, opened 2022-04-27
Add support to define a custom docker network and enable docker ipv6

### LINKED ISSUE #813: Add option for pipeline to use custom network

Currently, I'm trying to make Woodpecker work with Gitea, both locally, using Docker Compose.

I've set both server and agent, along with Gitea, to use a custom Docker network under the 10.0.0.0/8 subnet.

The issue lies when running a pipeline: containers from pipelines can't reach that custom network, so they fail when trying to clone the repository from Gitea.

### Suggested solution

In Drone CI, it is possible to set `DRONE_RUNNER_NETWORKS`, so pipelines are able to use custom networks. Woodpecker CI could add an environment variable for that as well.

### Alternative

N/A

### Additional context

Docs for `DRONE_RUNNER_NETWORKS`: https://docs.drone.io/runner/docker/configuration/reference/drone-runner-networks/
The `docker-compose.yml` file for running Gitea and Woodpecker locally: https://codeberg.org/gbrlsnchs/void-pkgs/src/branch/trunk/docker-compose.yml

### Validations

### LINKED ISSUE #845: IPv6 connectivity inside docker pipeline containers

Currently it seems impossible to get IPv6 connectivity inside containers spawned by woodpecker, as it creates docker networks without IPv6 capability. Even if IPv6 support is configured for docker itself (see https://docs.docker.com/config/daemon/ipv6/), no IPv6 connectivity is available inside containers, because the mentioned settings only affect the dockers default network.

The container-specific network is created here and neither sets `EnableIPv6` to `true`, nor does it specify any IPv6 subnets to configure:
https://github.com/woodpecker-ci/woodpecker/blob/905350fa1573b39a41ba289b4d4e76b7f652df3b/pipeline/frontend/yaml/compiler/compiler.go#L91-L102

There seems to be an undocumented `network_mode` option, which should allows new networks to use the "host"-mode, but this option is understandably only available to trusted repositories (admin users). Also, I have not been able to make this option work; the default network is still created using the bridge driver (see above, probably related to the recent introduction of multiple backends).

### Suggested solution

As docker - for good or for worse - does not support IPv6 NAT (and IPv6 therefore cannot be configured in a similar way to IPv4), I can think of multiple solutions to this problem:
- Add the ability to use already existing docker networks for new containers (via server/agent configuration)
- Add the ability to configure the use of the docker "host" network driver
- Add the ability to configure a pool of IPv6 subnets for the newly spawned docker networks

### Alternative

_No response_

### Additional context

_No response_

### Validations

anbraten:
closes #845 
closes #813


pipeline/backend/docker/docker.go
```diff
@@ -52,6 +55,14 @@ func (e *docker) Load() error {
 	}
 	e.client = cli
 
+	if enableIPV6, err := strconv.ParseBool(os.Getenv("WOODPECKER_BACKEND_DOCKER_ENABLE_IPV6")); err == nil {
+		e.enableIPv6 = enableIPV6
+	} else {
+		e.enableIPv6 = false
+	}
```
6543, line 62:
```suggestion
	e.enableIPv6, _ = strconv.ParseBool(os.Getenv("WOODPECKER_BACKEND_DOCKER_ENABLE_IPV6"))
```

if parse return err, value is already false

6543 (approved):

Issue #855, opened 2022-03-28
Context of commit statuses is not really "nice"


qwerty287:
### Clear and concise description of the problem

That's not a big problem, but currently the context is reported as `ci/woodpecker/<event>/<pipeline>`. This doesn't look really nice in UIs.

### Suggested solution

1. just use `Woodpecker CI` as context and put more information into the description
2. Only report the pipeline name and event as context (this is similar to GitHub Actions) and add more information in the description

### Additional context

GitHub Actions example:
![Screenshot_20220328-193038](https://user-images.githubusercontent.com/80460567/160454474-ce133516-94d2-4bb1-bb34-2f4d9fcb8066.png)
Woodpecker example:
![Screenshot_20220328-193115](https://user-images.githubusercontent.com/80460567/160454513-60524a03-55ef-4438-a811-4da3139081a1.png)

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

6543:
we  can experiment around a bit - but you have to take in mid, that branch protection roules depend on context be be different
so we realy have to look into each forge if there is a potion to split UI view and actual context

6543:
if we have to change the context - it will be breaking - just to mention

PR #900, opened 2022-05-04
Allow to change forge status messages

qwerty287:
Allow to change the status message via template option

Closes https://github.com/woodpecker-ci/woodpecker/issues/855


confusedsushi:
I like the current context text. From a programmers point of view it is short, informative and probably easy to parse.

I see that someone maybe wants a more human friendly text. What about making the format string which builds the text configurable?

server/remote/common/status.go
```diff
@@ -23,10 +23,10 @@ func GetBuildStatusContext(repo *model.Repo, build *model.Build, proc *model.Pro
 		name += "/" + proc.Name
 	}
 
-	return name
+	return fmt.Sprintf("%s - %s (%s)", server.Config.Server.StatusContext, proc.Name, build.Event)
```
6543, line 26: will make the other logic of GetBuildStatusContext useless & is breaking

6543:
well just let it ope until we could run down the 2 points :)

cmd/server/flags.go
```diff
@@ -206,6 +206,12 @@ var flags = []cli.Flag{
 		Usage:   "status context prefix",
 		Value:   "ci/woodpecker",
 	},
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_STATUS_CONTEXT_FORMAT", "WOODPECKER_GITHUB_CONTEXT_FORMAT", "WOODPECKER_GITEA_CONTEXT_FORMAT"},
```
6543, line 210:
```suggestion
		EnvVars: []string{"WOODPECKER_STATUS_CONTEXT_FORMAT"},
```
the gitea/github env var above where just keeped, to not break too much

6543:
adding tests is always a good idea ;)

6543 (approved):

docs/docs/30-administration/10-server-config.md
```diff
@@ -305,6 +305,17 @@ Read the value for `WOODPECKER_PROMETHEUS_AUTH_TOKEN` from the specified filepat
 
 Context prefix Woodpecker will use to publish status messages to SCM. You probably will only need to change it if you run multiple Woodpecker instances for a single repository.
 
+### `WOODPECKER_STATUS_CONTEXT_FORMAT`
+> Default: `%context/%event/%pipeline`
+
+Template for the status messages published to SCM.
```
anbraten, line 311:
```suggestion
Template for the status messages published to forges.
```

server/remote/common/status.go
```diff
@@ -2,31 +2,32 @@ package common
...
+	"strings"
 
 	"github.com/woodpecker-ci/woodpecker/server"
 	"github.com/woodpecker-ci/woodpecker/server/model"
 )
 
 func GetBuildStatusContext(repo *model.Repo, build *model.Build, proc *model.Proc) string {
-	name := server.Config.Server.StatusContext
-
+	ctx := server.Config.Server.StatusContextFormat
```
anbraten, line 12: What do you think about using GOs template function instead?

anbraten: https://pkg.go.dev/text/template

jolheiser: Either text template or https://pkg.go.dev/os#Expand style imo

6543: @qwerty287 u may want to look into this ;) ?

6543:
please update the unittests too https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/2472/31

6543 (approved):

PR #902, opened 2022-05-07
Add agent tagging / filtering for pipelines 

### LINKED ISSUE #304: `filters` / `Labels` are not very well documented.
In the agent's flags, there seems to be the ability to specify some filters:
https://github.com/woodpecker-ci/woodpecker/blob/a07f120d1245bab2c6eb4f896cbc3348f5e04d72/cmd/drone-agent/flags.go#L67-L71

It says we can restrict builds by label — perfect, but it's not clear what those labels are or what the filters look like.

I've been digging through the code a little bit, and it turns out that:

1: this has examples of the filter syntax: https://github.com/woodpecker-ci/expr/blob/164b8b3d0915b31a990c04f93e20928ed67a45fb/selector_test.go#L11

2: the below defines the 'filter function' — one thing I notice is that it seems to be that if you specify a filter, you MUST manually specify `platform == ...` in that filter otherwise you won't be filtering by platform anymore (The 'platform' is special-cased into the `filter.Labels` map at some point leading up to here, but I've lost track)
https://github.com/woodpecker-ci/woodpecker/blob/a07f120d1245bab2c6eb4f896cbc3348f5e04d72/server/rpc.go#L494-L518

3: this below is where the labels are attached to the Task
https://github.com/woodpecker-ci/woodpecker/blob/master/server/hook.go#L390-L395

4: it seems like this might be specifiable as a `labels` map in the pipeline YAML, but that part is not entirely clear to me — seems to be `libcompose.SliceorMap`.

Point (2) seems like a bug to me — surely you shouldn't be able to bypass the platform filter.
But in any case, none of this seems documented (please point it out if it is :)) and it'd be nice to change that (I'm happy to do so, just wanted to know if I'm missing anything / not understanding properly).

### LINKED ISSUE #860: all jobs stall as "pending" in queue since v0.15.0_RC2
### Component

server

### Describe the bug

I'm having job running issues with woodpecker v0.15.0. Builds stay stuck into queue ; they display as "not yet started" in the web interface. According to logs, the server is up and the agent seems able to connect to it. The only log I was able to get is when I try and jump to a pending job in the right panel in the web interface. Then the server complains : 
``` json
{"level":"error","error":"Error #01: sql: no rows in result set\n","ip":"192.168.1.1","latency":15.180251,"method":"GET","path":"/api/repos/lafleur/wanderings/logs/19/1","status":404,"time":"2022-04-01T13:34:57Z","user-agent":"Mozilla/5.0 (X11; Linux x86_64; rv:98.0) Gecko/20100101 Firefox/98.0","time":"2022-04-01T13:34:57Z"}
{"level":"error","error":"Error #01: sql: no rows in result set\n","ip":"::1","latency":14.972052,"method":"GET","path":"/api/repos/lafleur/wanderings/logs/19/1","status":404,"time":"2022-04-01T13:34:57Z","user-agent":"Mozilla/5.0 (X11; Linux x86_64; rv:98.0) Gecko/20100101 Firefox/98.0","time":"2022-04-01T13:34:57Z"}
```

I'm running on a barebone ARMv7 Arch system, installing through its [PKGBUILD](https://aur.archlinux.org/packages/woodpecker).

I tried rolling back to v0.15.0-rc2. The result is the same, although this time the agent seems unable to connect to the server. Logs from the agent :
``` json
{"level":"error","error":"rpc error: code = Unavailable desc = closing transport due to: connection error: desc = \"error reading from server: EOF\", received prior goaway: code: ENHANCE_YOUR_CALM, debug data: \"too_many_pings\"","time":"2022-04-01T13:47:45Z","message":"grpc error: done(): code: Unavailable: rpc error: code = Unavailable desc = closing transport due to: connection error: desc = \"error reading from server: EOF\", received prior goaway: code: ENHANCE_YOUR_CALM, debug data: \"too_many_pings\""}
```
My RC2 version is `next-c79aa2b1`.

Rolling back to [v0.15.0-rc1](https://github.com/woodpecker-ci/woodpecker/releases/tag/v0.15.0-rc1) (version `next-150c5e1c`) does workaround the issue.

I compared my issue with #841 , and the logs do not match, hence this new issue.

Any clue appreciated !

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-28b15120"}
```

### Additional context

_No response_

### Validations

anbraten:
closes #304 

closes #860 

Officially support labels for pipelines and agents to improve pipeline picking.


6543:
this is breaking but I'll support this simplification

6543:
one downside: we can't backport this ...

anbraten:
@6543 I adjusted the filtering a bit to fix some problems. But there are still some things left I am not sure about:

A whole pipeline file is always executed by a single agent (right?), so how should the `platform` filter of a single step work? 

```yml
plaform: * # not set or wildcard whatever => picks an random agent ignoring the platform
pipeline:
  slack:
    image: alpine
   when:
     platform: linux/amd64 # only executes the step if the random picked agent is `linux/amd64` - doesn't really make sense IMO
```

anbraten:
> one downside: we can't backport this ...

I think we can kindly ask the arm users to use latest for now as arm support is pretty new

pipeline/rpc/peer.go
```diff
@@ -10,7 +10,6 @@ type (
 	// Filter defines filters for fetching items from the queue.
 	Filter struct {
 		Labels map[string]string `json:"labels"`
-		Expr   string            `json:"expr"`
```
6543, line 13: you have to update the woodpecker.proto too

6543:
please resolve conflicts  ;)

pipeline/rpc/peer.go
anbraten: Did that and updated the notes how to install use protoc as it was my first time using it and the old guide wasn't complete.

6543: well each time it gets better, before i touched it there was none ;) - thanks

6543 (approved):

mcasviper:
I am a arm64 user, too but even with starting "platform: linux/arm*" I didn't get it run. After adding an amd64 agent it directly starts the tasks. 
Maybe you could post a dead simple example for me or can I help with some debugging? I am running v.0.15.3

6543:
> I am running v.0.15.3

that's the issue, you have to run `next`

mcasviper:
Ok I am using now next-alpine for my server and the arm agent with this example
`

  | # .woodpecker.yml
-- | --
  | platform: linux/arm*
  | pipeline:
  | build:
  | image: koalaman/shellcheck-alpine:latest
  | commands:
  | - find . -type f -name '*.sh' -print0  \| xargs -0r shellcheck
  | when:
  | platfrom: linux/arm*

`
Same issue as before and even without when: it is not working

mcasviper:
> > I am running v.0.15.3
> 
> that's the issue, you have to run `next`

Thanks. Now it works.

PR #905, opened 2022-05-10
auto cancel blocked pipelines

### LINKED ISSUE #904: Cancel blocked pipelines if auto-cancel is enabled
Followup of #831 

When a pipeline is blocked (awaiting approval) it should be auto-canceled as well.

anbraten:
closes #904 

follow-up #831

# TODO
- [x] send cancel event when auto canceling
- [ ] cancel before accepting latest?

### DIFF
--- a/server/api/build.go
+++ b/server/api/build.go
@@ -295,19 +295,15 @@ func cancelBuild(
 		return http.StatusInternalServerError, err
 	}
 
-	// For pending builds, we stream the UI the latest state.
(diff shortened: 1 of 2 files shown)


6543 (approved):

6543:
> cancel before accepting latest
I would cancel ON accept

anbraten:
I think so as well, as otherwise someone could kill approved pipelines by triggering an unapproved one.

PR #914, opened 2022-05-14
Allow gitea dev version

6543:
As before try to autodetect gitea version, if this does not work, assume it's latest version (v1.17.0 atm)


server/remote/gitea/gitea.go
```diff
@@ -458,7 +461,13 @@ func (c *Gitea) newClientToken(ctx context.Context, token string) (*gitea.Client
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 		}
 	}
-	return gitea.NewClient(c.URL, gitea.SetToken(token), gitea.SetHTTPClient(httpClient), gitea.SetContext(ctx))
+	client, err := gitea.NewClient(c.URL, gitea.SetToken(token), gitea.SetHTTPClient(httpClient), gitea.SetContext(ctx))
```
anbraten, line 464: So it will try to auto detect the version and use `v1.17.0` if it fails?

6543: right

anbraten (approved):

server/remote/gitea/gitea.go
6543: I wont like to make it a config option, as you only should have this problem if you develop gitea ...

PR #916, opened 2022-05-14
Use asym key to sign webhooks

### LINKED ISSUE #865: Use rsa for http signing
Using rsa for http would allow us to keep the secret secret :wink:
This will be especially helpful when allowing more extensions in addition to the current configuration service to be settable per repo.
The current lib used for signing doesn't support that but we could try https://github.com/go-fed/httpsig

anbraten:
closes #865 

- [x] use public private key for http signing
- [x] generate global ed25519 key pair
- [x] save generated key pair to db and load on next restart
- [x] add endpoint for getting public key
- [x] add migration notice
- [x] update docs
- [ ] update sample config service => will be done after merging this to not confuse users


server/shared/configFetcher_test.go
```diff
@@ -341,17 +344,28 @@ func TestFetchFromConfigService(t *testing.T) {
 		},
 	}
 
-	httpSigSecret := "wykf9frJbGXwSHcJ7AQF4tlfXUo0Tkixh57WPEXMyWVgkxIsAarYa2Hb8UTwPpbqO0N3NueKwjv4DVhPgvQjGur3LuCbiGHbBoaL1X5gZ9oyxD2lBHndoNxifDyNH7tNPw3Lh5lX2MSrWP1yuqHp8Sgm7fX8pLTjaKKFgFIKlODd"
+	pubEd25519Key, privEd25519Key, err := ed25519.GenerateKey(rand.Reader)
+	if err != nil {
+		panic(err)
```
6543, line 349: dont panic!
t.FailNow()

server/store/datastore/server_config.go
```diff
@@ -0,0 +1,36 @@
...
+	config := &model.ServerConfig{
+		Key: key,
+	}
+
+	err := wrapGet(s.engine.Get(config))
+	if err != nil {
+		return "", err
+	}
+
+	return config.Value, nil
```
6543, line 15: a simple get set test unit test would be nice

anbraten: done

6543 (approved):

PR #918, opened 2022-05-17
Clarify the `when.tag` example

Windvis:
The `tag` filter is only used if it's also a `tag` event. This makes the example clearer for new users.


Windvis:
Is it "valid" to define a tag filter without also the accompanying tag event? If not, it might be nice to add a check for this to the `woodpecker lint` command?

6543 (approved):

PR #921, opened 2022-05-17
Draft: CI add lint-pipeline

6543:


anbraten:
Nice idea 👍🏾

6543:
I think woodpecker should ignore errors etc while `cli lint` should warn you if your config do have strange things ...
- conditions that will never be true
- filters that will never apply (event: tag & branch: main -> branch is ignored in all cases)
- ...

anbraten (approved):

PR #923, opened 2022-05-17
remove senders

anbraten:
The senders service datastore, model and plugin doesn't seem to be used atm. To keep the code clean I would like to remove it (at least for now we could re-add it later if needed).


anbraten:
Should we leave the senders table (no migration removing it) as it is empty for most users and if someone had data we could maybe use it later again if we need it at some point?

6543:
I would remove tables, as we then later dont have to worry to migrate anything meaningfull out ot it

Issue #931, opened 2022-05-19
Pipelines are skipped where they should not


6543:
#877 did introduce an regression (I did also not notice while reviewing :/ ) 

WP only exec 1 pipeline but should do 2

## Before
https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/2558
## Now :/
https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/2563

anbraten:
Stange its not working for the push event? Skip pipeline shouldn't be touching push at all as it is only enabled for PR in our setup ...

6543:
closed by e2e094cfda87becedf7a37d6077fabea2c97ebb5 ...  - lets move on at #932 to make it work

PR #932, opened 2022-05-19
Do not run clone step if no pipeline step will run

qwerty287:
1. Fix empty build link
2. Fix issue if not all builds are skipped

TODO:
 - [x] fix the issue https://github.com/woodpecker-ci/woodpecker/issues/931
 
 Closes https://github.com/woodpecker-ci/woodpecker/issues/778


server/shared/procBuilder.go
```diff
@@ -115,7 +115,7 @@ func (b *ProcBuilder) Build() ([]*BuildItem, error) {
 
 			ir := b.toInternalRepresentation(parsed, environ, metadata, proc.ID)
```
6543, line 117: we should explane this if statement here ...

6543:
I did revert https://github.com/woodpecker-ci/woodpecker/pull/877 via https://github.com/woodpecker-ci/woodpecker/commit/e2e094cfda87becedf7a37d6077fabea2c97ebb5

you can add it back in this pull - it need just more work ...

PS: this patch do not solve the underlaying issue

6543:
@qwerty287 #949 should make the whole pipeline functions way more readable :)

6543:
to reproduce:
1. clone this repo into your local test env
2. enable ci for it
3. change a line in Makefile

expect:
- multible piplenes still run like "docker"
get:
- only "test" runn and report build successfull

6543:
at some point I'll also have a look at this ... as I dislike how global pipeline filter do work atm

anbraten:
I had a look and debugged it as well a few days / maybe even weeks ago (sorry for that). I also experienced some issues with it while testing. In general this is pretty related to #770 and I would like to have a more generic way of checking dependencies and if a required pipeline (depends_on) failed / is existing at all. Having properties like `isEmpty` / `skippedClone` sounds a bit to specific to me right now 🙈

anbraten:
Okay lets write down the cases again as I am not sure how you planned on handling each specific:
- pipeline is required by `depends_on`, but hasn't been defined at all (typo, not added yet) => the config is marked as wrong by the linter?!
- pipeline is required by `depends_on`, but was skipped as it was empty / only clone steps => treat as successful?
- pipeline is required by `depends_on`, but was skipped as it hasn't matched root.when #770) => skip dependent pipelines or treat as successful?
- pipeline is required by `depends_on`, but was failing => stop pipeline execution of dependent pipelines
- pipeline is required by `depends_on` and was successful => go on

anbraten:
Nice. I have updated my comment.

> The corresponding depends_on is ignored completely, not sure if the linter says something.

I just checked it. The linter is not checking that a `depends_on` target pipeline exists, but I think that would be a helpful additional feature for another PR.

PR #943, opened 2022-05-30
Improve UI colors to have more contrast 

egor-tensin:
Hello! I'd love to get rid of my personal eye sore (maybe something wrong with my eyes). First, in the build view, I can barely read the username and the branch on the sidebar; I'd like to make it more legible from my point of view, if you don't mind.

BuildProcList_before
BuildProcList_after

Second, in the repository settings, isn't the settings panel too gray for the text? I switched it to white, and it does seem to look a bit clearer, wouldn't you agree?

Panel_before
Panel_after

To be honest, even here I would prefer the muted gray text to be slightly darker, but that I can definitely live with.

### DIFF
--- a/web/src/components/layout/Panel.vue
+++ b/web/src/components/layout/Panel.vue
@@ -1,7 +1,7 @@
 <template>
-  <div class="rounded-md w-full shadow overflow-hidden text-gray-500 bg-gray-300 dark:bg-dark-gray-700">
-    <div v-if="title" class="font-bold bg-gray-400 dark:bg-dark-gray-800 p-2">{{ title }}</div>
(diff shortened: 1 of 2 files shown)


anbraten:
@egor-tensin Thanks for submitting this PR. There are quite some places which need some improvement regarding the contrast of the colors currently used. We are totally open for changes improving it.

6543:
thanks for the pull, as mentioned I'd love to see more suggestions :)

6543:
PS: code is unformated ... (https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/2587/27) - if you need help just tell us

6543:
well sure I just like to check against dark theme locally ...

PR #956, opened 2022-06-03
Change language for build badge to a more generic term that describes…

### LINKED ISSUE #955: Rename build to pipeline in bagde

This one falls into a very gray area, but I think the `build | status` badge could be more generic and be presented as `pipeline | status`, because `build` is a targeted process within the entire pipeline — if present.

Some pipelines don't have a "build" process, but like I said, it's a grays area, because the word can be used to describe the entire process as well.

### Suggested solution

Badge should be `pipeline | status`.

### Alternative

N/A

### Additional context

N/A

### Validations

x80486:
Closes #955


anbraten (approved):

x80486:
That should do it. I went ahead and changed the variable, but the actual type requires quite some changes, so I think if you change `Build` at some point, it might be as part of the RFC issue mentioned.

x80486:
That will be because the size in the SVG doesn't fit with the size of the text, right?

Issue #362, opened 2021-09-25
Global Environment Variable UI editor


hairlesshobo:
### Clear and concise description of the problem

At the moment, the only way to set a global environment variable is to change the configuration of the service itself and restart the Woodpecker application. [See here](https://woodpecker-ci.github.io/docs/usage/environment#global-environment-variables)

An example of where I would find this incredibly useful is for image names. Let say that I use one specific docker image throughout a number of projects and I know ahead of time that, periodically, I may need to change that image name. It would be great to easily do so without having to restart the entire Woodpecker platform.

### Suggested solution

I propose, in the spirit of #7 and #345 that there be Global and/or Project level environment variables and the corresponding UI editor. These environment variables could be stored in the database and would then be easily made available to all projects.

### Alternative

 At the moment, the closest solution to this would be to create a secret per-project and then map that to the environment variable, but this comes with limitations. With the fact that secrets are, by design, secret.. there would be no way to easily see what the current value of that secret is. For anything that isn't actually a secret, this would be counter-productive. 

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.github.io/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

anbraten:
We added global secrets with #1027. By doing #1007 this should be solved as well.

6543:
as per comment, it should be 'addrest', the rest is polishing

PR #968, opened 2022-06-15
Provide global environment variables for pipeline substitution

fracai:
This makes global environment variables (as specified by `WOODPECKER_ENVIRONMENT`) available for pipeline substitution.

Notably, it prepends `GLOBAL_` to each to ensure existing variables are not overwritten.

I should probably add a commit that documents this behavior (I'll do that soon).

The variables would still be available without a prefix within the pipeline task.

The use case I created this for is for specifying things like images, common to many steps, that should be updated together, or specifying a common registry that many pipelines will push to. Things that aren't really "secrets", but don't belong statically in the pipeline yaml.

As in:
~~~~
pipeline:
  build-and-publish:
    group: docker
    settings:
      dockerfile: Dockerfile
      repo: ${GLOBAL_REGISTRY}/${GLOBAL_USER}/woodpecker-plugin
~~~~

## TODO:
* [x] support `cli exec --env`
* [x] tests
* [x] documentation


6543:
yes some tests would be nice

server/shared/procBuilder.go
```diff
@@ -87,6 +87,16 @@ func (b *ProcBuilder) Build() ([]*BuildItem, error) {
 			metadata := metadataFromStruct(b.Repo, b.Curr, b.Last, proc, b.Link)
 			environ := b.environmentVariables(metadata, axis)
 
+			// add GLOBAL_ environment variables for substituting
+			for k, v := range b.Envs {
+				new_k := "GLOBAL_" + k
```
anbraten, line 92: Couldn't we swap the global env vars with `b.environmentVariables(metadata, axis)` in line 88. That way the global value can be overwritten, but the user wouldn't have to remember to prefix his vars with `GLOBAL_`.

anbraten:
> The use case I created this for is for specifying things like images, common to many steps, that should be updated together, or specifying a common registry that many pipelines will push to. Things that aren't really "secrets", but don't belong statically in the pipeline yaml.

I was thinking about some similar use-case as well and had the idea to at some point rename secrets to variables and allow the user to mask them (in case of a secret).

6543:
I think this should also apply to `cli exec`

fracai:
> > The use case I created this for is for specifying things like images, common to many steps, that should be updated together, or specifying a common registry that many pipelines will push to. Things that aren't really "secrets", but don't belong statically in the pipeline yaml.
> 
> I was thinking about some similar use-case as well and had the idea to at some point rename secrets to variables and allow the user to mask them (in case of a secret).

That'd be useful, as long as there's a capability for viewing and changing existing values and there's an option for them to be available across projects. I think I recall another ticket for that… #362

fracai:
> I think this should also apply to `cli exec`

Gotcha, so anything passed to the `--env` flag? I'll take a look.

server/shared/procBuilder.go
fracai: I'm not sure I follow. `b.eV(m,a)` provides the `CI_` variables (which probably shouldn't be overwritten). I can easily remove the prefix and either just skip anything that starts with `CI_` or not check and allow them to be overwritten. Despite the annoyance of having to remember the prefix, it seemed like a good safety measure to avoid overwriting the builtin variables.
Am I misunderstanding your suggestion?

fracai:
I added `cli exec` support and a procBuilder test.
I still need a `cli` test, documentation, and consensus on the naming behavior regarding the `GLOBAL_` prefix. And I suppose a consensus on whether variables should be allowed to overwrite existing variables. Allowing this would include variables like `CI_COMMIT_SHA`, so I'm weary of that idea.
I think I lean towards dropping the prefix and not allowing overwriting.

server/shared/procBuilder_test.go
```diff
@@ -22,6 +22,74 @@ import (
...
+	t.Parallel()
+
+	b := ProcBuilder{
+		Envs: map[string]string{
+			"KEY_K":    "VALUE_V",
+			"NO_IMAGE": "scratch",
+		},
+		Repo: &model.Repo{},
+		Curr: &model.Build{
+			Message: "aaa",
```
6543, line 69: a clasic message - i use it all the time :+1: :D

server/shared/procBuilder.go
6543: I'm fine with this as is

6543:
> > The use case I created this for is for specifying things like images, common to many steps, that should be updated together, or specifying a common registry that many pipelines will push to. Things that aren't really "secrets", but don't belong statically in the pipeline yaml.
> 
> I was thinking about some similar use-case as well and had the idea to at some point rename secrets to variables and allow the user to mask them (in case of a secret).

well I guess we can introduce `GLOBAL_` and set vars should then have no prefix and **do** overwrite stuff ...
for the ui part I'm fine with re-using the secrets stuff, but in backend we should difference strongly between them, for security reasons

6543 (dismissed):
oh well documentation missing :D

fracai:
OK, so this drops the `GLOBAL_` prefix, but still avoids overwriting existing variables. Another reason to avoid the prefix is that running from `cli exec` meant the variable was prefixed for the yaml substitution, but not prefixed inside the container. Leaving off the prefix, but not overwriting, seems like a safe alternative without having to understand multiple variable names in different contexts.

I added documentation and examples (let me know if anything is missing, unclear, or needs to be expanded).

I have a procBuilder test, but I'm struggling to put together something for `cli/exec`. Specifically, I think I'd like the test entry point to be `run(…)` and I'd ideally mock the input file, but there are multiple places that read from that input file. Even entering at the method that I modified (`execWithAxis(…)`) would require reading from the file. I did find some tips on mocking a file system, but that would require making additional changes to how the file is read in the `exec` methods. Any thoughts on a better way to create a test?

fracai:
Just wanted to check in and see if there's anything else requested for this? My only outstanding thought is the cli test, which I think would require changes to the operational parts of the cli code. I'd actually be more inclined to put that in a separate PR if it's desired.
Anyway, let me know if anything is missing or needs work.

6543:
I dont understand why you droped `GLOBAL_`

fracai:
> I dont understand why you droped `GLOBAL_`

As I thought more about the prefix, it seemed to just be confusing.

You add a variable name for the environment, but then to use it in the pipeline configuration you have to add a prefix to it. You can also use the variable inside the pipeline container, but there it doesn't have a prefix.

The only thing positive about the prefix is that it identifies the source of that variable, but only in the pipeline.

I suppose the alternative would be to add the prefix when the variables are injected to the container environment as well, but that'd be a breaking change to existing behavior. I figured it'd be safer to modify the behavior I was introducing, rather than the current state.

If you disagree I can modify the PR, including updating the documentation.

Options:
- Leave as is with no prefix
- Restore prefix when processing the pipeline. Leave the unprefixed version within the container.
- Restore the prefix and update the container environment to also include a prefix.
- Restore the prefix and update the container to include both a prefixed and unprefixed version. And mark the unprefixed version as deprecated.

I personally prefer the first option for simplicity, but I understand there might be other considerations or future interactions to consider.

anbraten:
+1 for the current option.

fracai:
Just checking in regarding if there's any further requests on this PR, or if I can "ignore it" until you're ready to merge. Thanks.

docs/docs/20-usage/50-environment.md
```diff
@@ -136,6 +136,28 @@ services:
...
+
+```diff
+services:
+  woodpecker-server:
+    [...]
+    environment:
+      - [...]
+-     - WOODPECKER_ENVIRONMENT=GOLANG_VERSION:1.17
++     - WOODPECKER_ENVIRONMENT=GOLANG_VERSION:1.18
+```
```
anbraten, line 159:
```suggestion
```diff
pipeline:
  build:
-    image: golang:1.18
+   image: golang:${GOLANG_VERSION}
    commands:
      - export PATH=$${PATH}:/go
      - go build
      - go test
    environment:
      - [...]
+     - WOODPECKER_ENVIRONMENT=GOLANG_VERSION:1.18
```
```

anbraten (approved):

PR #975, opened 2022-06-16
Return to page after clicking login at navbar

anbraten:
By this change the user will be returned to the currently opened page (for example a public repo), after pressing the login button att the nav-bar.


6543:
I think we can&shuld backport that :)

6543 (approved):

6543:
->  #976

PR #977, opened 2022-06-16
More go linter

6543:


6543:
missed it

6543:
well it is just to outdated now :/

PR #987, opened 2022-06-17
Show forge icons in UI

qwerty287:
Show icons of the forge in the web UI, depending on the configured forge. Currently only supports GitHub and Gitea since the other ones don't have icons (how can I add them?).


anbraten:
You should be able to add every icon from https://icones.js.org

server/remote/remote.go
```diff
@@ -27,6 +27,9 @@ import (
 // TODO: add Driver() who return source forge back
 
 type Remote interface {
+	// Name returns the string name of this driver
+	Name() string
```
6543, line 31: with this you have to codegenerate the mock

6543 (approved):

PR #994, opened 2022-06-19
add graphql

anbraten:
# todo
- [x] map graphql types & fields correctly to models
- [x] add some basic queries
- [x] add middlewares / directives
- [ ] add dataloaders
- [ ] add subscription
- [ ] add some docs


gapodo:
I almost feel like necro posting, but I'm currently looking over open PRs / Issues looking for interesting things...

Have you considered some complexity calcs before running queries, as graphql with resolvers is a "nice" way to (intentionally or just by accident) DOS a server with a single request (don't ask me how I know...) or is this baked into the graphql-lib you are using?

anbraten:
Although its super anoying that event-streams and websockets have nothing like openapi to be schema defined which could be done by graphql and using graphs for nested structures like pipelines, workflows and steps would be perfect this seems to be a lot of work atm. Maybe we will have a look at it at some point again.

PR #1000, opened 2022-06-25
Fix terminal DOM existing multiple times

thestr4ng3r:
When switching between the tasks/config/changed files tabs in the build
view, the DOM for the log xterm would be inserted multiple times,
causing the current terminal to be shifted down weirdly:

https://user-images.githubusercontent.com/1460997/175767842-ad72bbec-c68f-42a7-b622-68d55a5f0615.mov

This fixes this behavior by cleaning any custom DOM before the log is
unmounted.

Suggestions welcome if there are better ways clean up such custom elements. It has been a while since I last dived deeper into vue.

### DIFF
--- a/web/src/components/repo/build/BuildLog.vue
+++ b/web/src/components/repo/build/BuildLog.vue
@@ -167,13 +167,13 @@ export default defineComponent({
       fitAddon.value.fit();
     }
 
-    let unmounted = false;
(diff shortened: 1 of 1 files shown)


anbraten (approved):
I still don't get why the onmounted hook is called several times and the dom is not updated in between, but thank you so much for the fix.

PR #1018, opened 2022-07-08
Fix incorrect port used in docs and docker-compose example

rluetzner:
The docker-compose examples in the docs and the example file incorrectly use port 8000 instead of 9000.

### DIFF
--- a/docker-compose.example.yml
+++ b/docker-compose.example.yml
@@ -7,7 +7,7 @@ services:
       dockerfile: Dockerfile
     image: woodpeckerci/woodpecker-server:local
     ports:
-      - 8000:8000
+      - 9000:9000
     volumes:
       - /var/lib/woodpecker:/var/lib/woodpecker/
     environment:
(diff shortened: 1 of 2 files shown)


jackbondpreston:
I think these were correct before: 8000 is the external facing web server, 9000 is the rpc endpoint for the agents.

I may be wrong though

rluetzner:
Okay, I checked `server.go` and `flags.go` which confirmed my suspicion.
When run without Let's Encrypt integration the default port 8000 is used for the web UI. Otherwise ports 80 and 443 are used instead.
Sorry. I hope I didn't waste too much of your time. 🙂

jackbondpreston:
> Interesting. I see that port 8000 is exposed in the Dockerfiles, but I don't quite understand what it does.
> I just now realized that due to the internal network the agent won't have any issues connecting to the server. We only have to expose 9000 when we want to connect other (external) agents.
> Is port 8000 maybe the default web port when not using Let's Encrypt?

Yeah, I think it is

PR #1027, opened 2022-07-17
Global and organization secrets

lafriks:
Adds way to add server wide and organization wide secrets

Organization page

attels

Organization settings page

attels

Global settings page

attels

* [x] Database and store changes
* [x] #1037
* [x] Server API changes
* [x] Organization settings UI
* [x] Server settings UI
* [x] Translation fixes


web/src/router.ts
```diff
@@ -28,6 +28,26 @@ const routes: RouteRecordRaw[] = [
...
+  {
+    path: '/org/:repoOwner',
+    name: 'org-wrapper',
+    component: (): Component => import('~/views/org/OrgWrapper.vue'),
+    props: true,
+    children: [
+      {
+        path: '',
+        name: 'org',
+        redirect: (route) => ({ name: 'repos-owner', params: route.params }),
```
anbraten, line 40: Or should we do it the other way around and redirect `/:repoOwner` to `/org/:repoOwner` 🤔

web/src/router.ts
```diff
@@ -28,6 +28,26 @@ const routes: RouteRecordRaw[] = [
     component: (): Component => import('~/views/ReposOwner.vue'),
```
anbraten, line 28: Could you rename `ReposOwner` and `repos-owner` to something like `Orgs` and `org`?

web/src/assets/locales/en.json
```diff
@@ -205,6 +205,66 @@
...
+        },
+        "events": {
+          "events": "Available at following events",
+          "pr_warning": "Please be careful with this option as a bad actor can submit a malicious pull request that exposes your secrets."
+        }
+      }
+    }
+  },
+
+  "admin": {
```
anbraten, line 238: I am not sure if it is a good idea to duplicate all of those values, although I like how you simply pass `i18n-prefix` to get the correct translations.

server/store/datastore/migration/common.go
```diff
@@ -212,6 +212,42 @@ func dropTableColumns(sess *xorm.Session, tableName string, columnNames ...strin
...
+	dialect := sess.Engine().Dialect().URI().DBType
+	switch dialect {
+	case schemas.MYSQL:
+		_, err := sess.Exec(fmt.Sprintf("ALTER TABLE `%s` COLUMN `%s` SET DEFAULT %s;", table, column, defValue))
+		return err
+	case schemas.POSTGRES:
+		_, err := sess.Exec(fmt.Sprintf("ALTER TABLE `%s` ALTER COLUMN `%s` SET DEFAULT %s;", table, column, defValue))
+		return err
+	case schemas.SQLITE:
+		return nil
```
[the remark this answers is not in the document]
lafriks: SQLite has no way to alter column default value/not null

web/src/router.ts
anbraten: Ahh I see good point.

server/store/datastore/migration/common.go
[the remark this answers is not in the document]
anbraten: So we would need to do this temp table copy work-around again?

server/model/secret.go
```diff
@@ -65,6 +83,16 @@ func (Secret) TableName() string {
...
 
+// Global secret.
+func (s Secret) Global() bool {
+	return s.RepoID == 0 && s.Owner == ""
+}
+
+// Organization secret.
+func (s Secret) Organization() bool {
+	return s.RepoID == 0 && s.Owner != ""
+}
```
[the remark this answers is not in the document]
lafriks: you mean adding new column? I don't see much point in that

server/store/datastore/migration/common.go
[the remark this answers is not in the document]
lafriks: That would be only option but do we need that, I added that just to have more database checks and sqlite3 is used probably mostly for small instances, so imho it's not worth the trouble :)

web/src/components/secrets/SecretList.vue
```diff
@@ -0,0 +1,82 @@
...
+        >
+          {{ event }}
+        </span>
+      </div>
+      <IconButton icon="edit" class="ml-2 w-8 h-8" @click="editSecret(secret)" />
+      <IconButton
+        icon="trash"
+        class="ml-2 w-8 h-8 hover:text-red-400 hover:dark:text-red-500"
+        :is-loading="isDeleting"
+        @click="deleteSecret(secret)"
```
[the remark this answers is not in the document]
lafriks: I would rather leave them as I it's better to leave business logic out of view. For example for delete I would like to add confirmation dialog in the future

web/src/assets/locales/en.json
lafriks: Not all are duplicating and I did not like to having repo prefix in views that are not repo specific. I could probably move some out to some specific section maybe?

web/src/views/org/OrgWrapper.vue
```diff
@@ -0,0 +1,61 @@
...
+    const repoOwner = toRef(props, 'repoOwner');
+    const apiClient = useApiClient();
+    const org = computed<Org>(() => ({ name: repoOwner.value }));
+
+    const orgPermissions = ref<OrgPermissions>();
+    provide('org', org);
+    provide('org-permissions', orgPermissions);
+
+    async function load() {
+      orgPermissions.value = await apiClient.getOrgPermissions(repoOwner.value);
```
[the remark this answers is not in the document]
lafriks: I have made that it should not error (it will return values even for non-existing organizations). Fixed to return false values also for unauthorized users

web/src/components/admin/settings/AdminSecretsTab.vue
```diff
@@ -0,0 +1,140 @@
+<template>
+  <Panel>
+    <div class="flex flex-row border-b mb-4 pb-4 items-center dark:border-gray-600">
+      <div class="ml-2">
+        <h1 class="text-xl text-color">{{ $t('admin.settings.secrets.secrets') }}</h1>
+        <p class="text-sm text-color-alt">
```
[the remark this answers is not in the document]
lafriks: attels

lafriks:
@anbraten added suggested changes

web/src/components/atomic/Warning.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <div class="text-sm font-bold text-light-600 rounded-md border border-solid p-2 border-red-900 bg-red-600">
```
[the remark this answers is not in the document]
lafriks: done

attels

attels

server/store/datastore/migration/common.go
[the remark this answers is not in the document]
anbraten: https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker is using sqlite 🙈. Would it still work when having the old column type? Otherwise it would be a breaking change IMO so we would need to mention or migrate it.

web/src/assets/locales/en.json
anbraten: Linking them sounds great. I had a look at the docs and its actually possible: https://kazupon.github.io/vue-i18n/guide/messages.html#linked-locale-messages

server/store/datastore/migration/common.go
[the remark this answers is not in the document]
lafriks: It would still work even without migration, constraints are added to be more on safe side

web/src/assets/locales/en.json
anbraten: I tried to link the translations. Which worked in general, but as just a few were duplicated it was more work than benefit in the end somehow.

server/api/global_secret.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"net/http"
+
+	"github.com/woodpecker-ci/woodpecker/server"
+	"github.com/woodpecker-ci/woodpecker/server/model"
+
+	"github.com/gin-gonic/gin"
+)
+
+// GetGlobalSecretList get the global secret list from
```
anbraten, line 26:
```suggestion
// GetGlobalSecretList gets the global secret list from
```

server/api/global_secret.go
```diff
@@ -0,0 +1,123 @@
...
+	name := c.Param("secret")
+	secret, err := server.Config.Services.Secrets.GlobalSecretFind(name)
+	if err != nil {
+		c.String(404, "Error getting global secret %q. %s", name, err)
+		return
+	}
+	c.JSON(200, secret.Copy())
+}
+
+// PostGlobalSecret persists the global secret to the database.
```
anbraten, line 54:
```suggestion
// PostGlobalSecret persists a global secret to the database.
```

server/api/global_secret.go
```diff
@@ -0,0 +1,123 @@
...
+		return
+	}
+	if err := server.Config.Services.Secrets.GlobalSecretCreate(secret); err != nil {
+		c.String(500, "Error inserting global secret %q. %s", in.Name, err)
+		return
+	}
+	c.JSON(200, secret.Copy())
+}
+
+// PatchGlobalSecret updates the global secret in the database.
```
anbraten, line 78:
```suggestion
// PatchGlobalSecret updates a global secret in the database.
```

server/api/org_secret.go
```diff
@@ -0,0 +1,136 @@
...
+	)
+	secret, err := server.Config.Services.Secrets.OrgSecretFind(owner, name)
+	if err != nil {
+		c.String(404, "Error getting org %q secret %q. %s", owner, name, err)
+		return
+	}
+	c.JSON(200, secret.Copy())
+}
+
+// GetOrgSecretList get the organization secret list from
```
anbraten, line 41:
```suggestion
// GetOrgSecretList gest the organization secret list from
```

server/api/org_secret.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	// copy the secret detail to remove the sensitive
+	// password and token fields.
+	for i, secret := range list {
+		list[i] = secret.Copy()
+	}
+	c.JSON(http.StatusOK, list)
+}
+
+// PostOrgSecret persists the organization secret to the database.
```
anbraten, line 58:
```suggestion
// PostOrgSecret persists an organization secret to the database.
```

server/api/org_secret.go
```diff
@@ -0,0 +1,136 @@
...
+		return
+	}
+	if err := server.Config.Services.Secrets.OrgSecretCreate(owner, secret); err != nil {
+		c.String(500, "Error inserting org %q secret %q. %s", owner, in.Name, err)
+		return
+	}
+	c.JSON(200, secret.Copy())
+}
+
+// PatchOrgSecret updates the organization secret in the database.
```
anbraten, line 85:
```suggestion
// PatchOrgSecret updates an organization secret in the database.
```

anbraten (approved):

anbraten:
@lafriks https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/3121/41

lafriks:
@anbraten fixed

PR #1028, opened 2022-07-17
use external lib to convert yaml to json

6543:
this move shared/yml/* into an independent lib


anbraten (approved):
Looks fine codewise. Whats the purpose of moving it out. Do you plan to use it somewhere else?

6543:
yes mostly for smal side projects ...
... but this lib is usefull for all go projects that do like to convert yaml 2 json without depending on old libs

6543:
@anbraten added you as contributor to that lib repo in case there is something and I'm afk ...

PR #1032, opened 2022-07-18
Move constrain to only have a single command in backend to run to dedicated backends

6543:
at the moment we compile a script that we can pipe in as single command
this is because of the constrains the docker backend gives us.

so we move it into the docker backend and eventually get rid of it altogether


pipeline/backend/docker/convert.go
```diff
@@ -20,11 +21,26 @@ func toConfig(proc *types.Step) *container.Config {
...
+			proc.Environment["HOME"] = "c:\\root"
+			proc.Environment["SHELL"] = "powershell.exe"
+			proc.Entrypoint = []string{"powershell", "-noprofile", "-noninteractive", "-command"}
+			proc.Commands = []string{"[System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($Env:CI_SCRIPT)) | iex"}
+		} else {
+			proc.Environment["CI_SCRIPT"] = generateScriptPosix(proc.Commands)
+			proc.Environment["HOME"] = "/root"
+			proc.Environment["SHELL"] = "/bin/sh"
+			proc.Entrypoint = []string{"/bin/sh", "-c"}
+			proc.Commands = []string{"echo $CI_SCRIPT | base64 -d | /bin/sh -e"}
```
6543, line 37:
```suggestion
			config.Cmd = []string{"echo $CI_SCRIPT | base64 -d | /bin/sh -e"}
```

anbraten:
Sounds nice. My idea for the docker backend was to run each command step with `docker exec` instead of using the script and executing it via `docker run` once. This would allow us to change the entrypoint as well: #278

pipeline/backend/docker/convert.go
```diff
@@ -20,12 +21,26 @@ func toConfig(proc *types.Step) *container.Config {
 		AttachStdout: true,
 		AttachStderr: true,
 	}
+
+	if len(proc.Commands) != 0 {
+		if runtime.GOOS == "windows" {
```
anbraten, line 26: @6543 Isn't this executed on the server? So it would use the servers OS instead of the OS selected in the pipeline config?

pipeline/frontend/yaml/container.go
```diff
@@ -37,7 +36,6 @@ type (
 		Tmpfs         []string               `yaml:"tmpfs,omitempty"`
 		DNS           types.Stringorslice    `yaml:"dns,omitempty"`
 		DNSSearch     types.Stringorslice    `yaml:"dns_search,omitempty"`
-		Entrypoint    types.Command          `yaml:"entrypoint,omitempty"`
```
anbraten, line 40: @6543 Why do you plan to remove the entrypoint completely?

6543: there was even a liter that permit the usage for normal steps ...

pipeline/backend/kubernetes/pod.go
```diff
@@ -97,10 +88,10 @@ func Pod(namespace string, step *types.Step) *v1.Pod {
 				Name:            podName(step),
 				Image:           step.Image,
 				ImagePullPolicy: pullPolicy,
-				Command:         command,
-				Args:            args,
+				Command:         step.Entrypoint,
+				Args:            step.Commands,
```
6543, line 92: can somebody test this?

-> so that it's possible to forward commands directly into kubernetes

gapodo:
After the changes put in via 30f113f it's working on Kubernetes with the Kubernetes backend active
image
https://test-ci.kle.li/sandbox/sq-test/pipeline/10/3
thanks again for the patience while I was stumbling through the setup

6543:
ok now it's just a refactor and do not change any behavior at all

pipeline/backend/kubernetes/pod.go
```diff
@@ -97,10 +103,10 @@ func Pod(namespace string, step *types.Step) *v1.Pod {
 				Name:            podName(step),
 				Image:           step.Image,
 				ImagePullPolicy: pullPolicy,
-				Command:         command,
+				Command:         entrypoint,
```
anbraten, line 106: Why not call it command?

pipeline/backend/common/script.go
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import "runtime"
+
+func GenerateDockerConf(commands []string) (env map[string]string, entry, cmd []string) {
```
anbraten, line 19:
```suggestion
func GenerateContainerConf(commands []string) (env map[string]string, entry, cmd []string) {
```

pipeline/backend/kubernetes/pod.go
6543: Because it's the entrypoint, if you read the description of the struct it also tels you tjat

6543:
```
	// Entrypoint array. Not executed within a shell.
	// The container image's ENTRYPOINT is used if this is not provided.
	// Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
	// cannot be resolved, the reference in the input string will be unchanged. Double $$ are reduced
	// to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e. "$$(VAR_NAME)" will
	// produce the string literal "$(VAR_NAME)". Escaped references will never be expanded, regardless
	// of whether the variable exists or not. Cannot be updated.
	// More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
	// +optional
```

anbraten (approved):

PR #1037, opened 2022-07-19
Add method to check organization membership

lafriks:
Needed for right checks for organization wide methods


server/remote/bitbucket/bitbucket.go
```diff
@@ -301,6 +301,16 @@ func (c *config) Hook(ctx context.Context, req *http.Request) (*model.Repo, *mod
 	return parseHook(req)
 }
 
+// OrgMembership returns if user is member of organization and if user
+// is admin/owner in this organization.
+func (c *config) OrgMembership(ctx context.Context, u *model.User, owner string) (bool, bool, error) {
```
6543, line 306: do named returns or document values in desc

server/remote/remote.go
```diff
@@ -79,6 +79,10 @@ type Remote interface {
 	// Hook parses the post-commit hook from the Request body and returns the
 	// required data in a standard format.
 	Hook(ctx context.Context, r *http.Request) (*model.Repo, *model.Build, error)
+
+	// OrgMembership returns if user is member of organization and if user
+	// is admin/owner in that organization.
+	OrgMembership(ctx context.Context, u *model.User, owner string) (bool, bool, error)
```
6543, line 85: do named returns or document values in desc

anbraten:
For caching I have created a separate issue as this could be interesting for the branches and permission calls.

server/remote/remote.go
lafriks: added named return values in interface

server/remote/bitbucket/bitbucket.go
lafriks: having that in interface should be enaugh

lafriks:
I don't think PR failure is related to this PR as there is no changes related to database or datasource module

server/remote/remote.go
```diff
@@ -79,6 +79,10 @@ type Remote interface {
 	// Hook parses the post-commit hook from the Request body and returns the
 	// required data in a standard format.
 	Hook(ctx context.Context, r *http.Request) (*model.Repo, *model.Build, error)
+
+	// OrgMembership returns if user is member of organization and if user
+	// is admin/owner in that organization.
+	OrgMembership(ctx context.Context, u *model.User, owner string) (member, admin bool, err error)
```
anbraten, line 85: What do you think about using your `membership` struct instead of the two booleans.  `owner string` should probably rather be `name string`.

```go
type membership struct {
	Member bool
	Admin  bool
}
```

server/cache/membership.go
```diff
@@ -0,0 +1,88 @@
...
+func NewMembershipService(r remote.Remote) MembershipService {
+	return &membershipCache{
+		TTL:    10 * time.Minute,
+		Remote: r,
+		Cache:  ttlcache.New(ttlcache.WithDisableTouchOnHit[string, membership]()),
+	}
+}
+
+func (c *membershipCache) get(ctx context.Context, u *model.User, owner string) (bool, bool, error) {
+	key := u.Login + "/" + owner
```
anbraten, line 57:
```suggestion
func (c *membershipCache) get(ctx context.Context, u *model.User, name string) (bool, bool, error) {
	key := u.Login + "/" + name
```

server/cache/membership.go
```diff
@@ -0,0 +1,88 @@
...
+	member, _, err := c.get(ctx, u, owner)
+	if err != nil {
+		return false, err
+	}
+	return member, nil
+}
+
+// IsAdmin returns true if the user is an admin of the organization.
+func (c *membershipCache) IsAdmin(ctx context.Context, u *model.User, owner string) (bool, error) {
+	_, admin, err := c.get(ctx, u, owner)
```
anbraten, line 83:
```suggestion
func (c *membershipCache) IsAdmin(ctx context.Context, u *model.User, name string) (bool, error) {
	_, admin, err := c.get(ctx, u, name)
```

server/cache/membership.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return false, false, err
+	}
+	c.Cache.Set(key, membership{Member: member, Admin: admin}, c.TTL)
+	return member, admin, nil
+}
+
+// IsMember returns true if the user is a member of the organization.
+func (c *membershipCache) IsMember(ctx context.Context, u *model.User, owner string) (bool, error) {
+	member, _, err := c.get(ctx, u, owner)
```
anbraten, line 74:
```suggestion
func (c *membershipCache) IsMember(ctx context.Context, u *model.User, name string) (bool, error) {
	member, _, err := c.get(ctx, u, name)
```

server/cache/membership.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/lafriks/ttlcache/v3"
+)
+
+// MembershipService is a service to check for user membership.
+type MembershipService interface {
+	// IsMember returns true if the user is a member of the organization.
+	IsMember(ctx context.Context, u *model.User, owner string) (bool, error)
+
+	// IsAdmin returns true if the user is an admin of the organization.
+	IsAdmin(ctx context.Context, u *model.User, owner string) (bool, error)
```
anbraten, line 33:
```suggestion
	// IsMember returns true if the user is a member of the organization.
	IsMember(ctx context.Context, u *model.User, name string) (bool, error)

	// IsAdmin returns true if the user is an admin of the organization.
	IsAdmin(ctx context.Context, u *model.User, name string) (bool, error)
```

server/remote/remote.go
lafriks: Refactored to have single method in membership service

lafriks:
@anbraten refactored remote method to return struct and simplified membership service

anbraten (approved):

PR #1044, opened 2022-07-22
make ListItem component more accessible by using a button tag when clickable

### LINKED ISSUE #1043: ListItem (as used on /repos) is not keyboard tab-able
### Component

web-ui

### Describe the bug

[`ListItem`](https://github.com/woodpecker-ci/woodpecker/blob/master/web/src/components/atomic/ListItem.vue) (as used on `/repos`) is not keyboard tab-able, preventing navigation by keyboard, screen readers and other non-mouse users.

by using a `button` tag instead of a `div` for when the  `ListItem`component is `clickable`, we can improve the accessibility of this component.

### System Info

```shell
0.15.3
```

### Additional context

_No response_

### Validations

eleith:
resolves #1043


anbraten:
LGTM. What happens with buttons inside the ListItem. Is it a problem to have nested buttons?

eleith:
a [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) is an interactive element and should not contain other interactive elements inside of it (similar to `a` tags or for that matter, a clickable div).

on the add repo page, this is handled correctly by making the ListItem not clickable when a button (to add the repo) is rendered within ListItem.

to your point, if you wanted to use a ListItem that was clickable to expand, and inside the expanded area could be buttons. if that is an allowable use case of ListItem, then it would need to handle remaining a div, even if clickable.

6543 (approved):
works

6543:
I think we can and should backport it ... @eleith would you like to do so?

eleith:
yup. i'll give it a try. should i start from https://github.com/woodpecker-ci/woodpecker/tree/release/v0.15 ?

6543:
yes I usually cherry-pick the merged sha ...

PR #1050, opened 2022-07-28
Update type of 'log_data' from blob to longblob

simmstein:
fix #1022

PR #1052, opened 2022-07-29
Update type of 'log_data' from blob to longblob (#1050)

simmstein:
Backport #1050

* Update type of 'log_data' from blob to longblob
* Add migration task to update log_data type

### DIFF
--- a/server/store/datastore/migration/migration.go
+++ b/server/store/datastore/migration/migration.go
@@ -33,7 +33,7 @@ var migrationTasks = []*task{
 	&fixPRSecretEventName,
 	&alterTableReposDropCounter,
 	&dropSenders,
-	&alterTableLogUpdateColumnLogDataType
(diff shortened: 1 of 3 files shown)


server/store/datastore/migration/migration.go
```diff
@@ -32,6 +32,8 @@ var migrationTasks = []*task{
 	&alterTableReposDropAllowDeploysAllowTags,
 	&fixPRSecretEventName,
 	&alterTableReposDropCounter,
+	&dropSenders,
```
6543, line 35:
```suggestion
```

6543 (approved):

server/store/datastore/migration/migration.go
6543, line 35:
```suggestion
```

6543 (approved):

PR #1060, opened 2022-07-31
Make build proc list to be returned always in correct order

lafriks:
Currently processes can be returned in incorrect order depending on how database returns them that could be random and even changing order


6543:
please backport :)

lafriks:
@6543 done

PR #1066, opened 2022-08-02
Fix to show build pipeline parse error

### LINKED ISSUE #811: Yaml parse issues are only reported back to hook and not shown in the UI

### LINKED ISSUE #1045: Improve Visibility of Errors on Invalid CI Config Yaml

I am using Woodpecker with Gitea using webhooks to trigger CI runs on every push/commit.

Currently, if a commit contains invalid CI yaml, it silently fails to run the CI job, unless you look at the webhook configuration (which may not be available to lower-privileged users) and is not obvious unless you know what you're looking for.

The following screenshot was taken from `https://gitea.example.com/[project]/[repo]/settings/hooks/[hook-id]`:

image

However, if you're not sure what you're looking for, it looks like Ci just silently failed.

### Suggested solution

I would propose that a far more self-evident solution would be creating a job entry in the CI jobs list, but marking it as failed, and also providing failed feedback to the git repo to turn the build red, etc. This fits far more closely in the mental model of how a developer would expect a CI system to interact: fail the build if it couldn't complete successfully due to bad files in the repo.

This is similar to how gitlab CI approaches the problem. Below is an example of one of my projects - the bottom entry is a normal successful run, and then I intentionally committed invalid yaml, which generated the top entry in CI run entries.

image

Similarly, it's treated like any other failed job, and marks the commit as having had CI not pass (instead of gitea's solution of not creating any CI runs in the interface).

image

The run details has more specific information as well.

image

### Alternative

_No response_

### Additional context

_No response_

### Validations

lafriks:
Also handle if no pipeline config is found in repository

Replaces #812

Fixes #811

Fixes #1045

Screenshots:
attels

attels

### DIFF
--- a/pipeline/frontend/yaml/error.go
+++ b/pipeline/frontend/yaml/error.go
@@ -22,3 +22,9 @@ type PipelineParseError struct {
 func (e PipelineParseError) Error() string {
 	return e.Err.Error()
 }
+
+func (e PipelineParseError) Is(target error) bool {
+	_, ok1 := target.(PipelineParseError)
+	_, ok2 := target.(*PipelineParseError)
+	return ok1 || ok2
+}
--- a/pipeline/frontend/yaml/error.go
+++ b/pipeline/frontend/yaml/error.go
@@ -1,0 +1,24 @@
+// Copyright 2022 Woodpecker Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package yaml
+
+// PipelineParseError is an error that occurs when the pipeline parsing fails.
+type PipelineParseError struct {
+	Err error
+}
+
+func (e PipelineParseError) Error() string {
+	return e.Err.Error()
+}
--- a/server/pipeline/restart.go
+++ b/server/pipeline/restart.go
@@ -95,8 +95,7 @@ func Restart(ctx context.Context, store store.Store, lastBuild *model.Build, use
 
 	newBuild, buildItems, err := createBuildItems(ctx, store, newBuild, user, repo, pipelineFiles, envs)
 	if err != nil {
-		var parseErr *yaml.PipelineParseError
-		if errors.As(err, &parseErr) {
+		if errors.Is(err, &yaml.PipelineParseError{}) {
 			return newBuild, nil
 		}
 		msg := fmt.Sprintf("failure to createBuildItems for %s", repo.FullName)
--- a/server/pipeline/restart.go
+++ b/server/pipeline/restart.go
@@ -16,11 +16,13 @@ package pipeline
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/rs/zerolog/log"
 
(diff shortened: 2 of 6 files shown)


lafriks:
Also fixes to handle restart of such jobs

server/pipeline/restart.go
```diff
@@ -85,6 +87,10 @@ func Restart(ctx context.Context, store store.Store, lastBuild *model.Build, use
 
 	newBuild, buildItems, err := createBuildItems(ctx, store, newBuild, user, repo, pipelineFiles, envs)
 	if err != nil {
+		var parseErr *yaml.PipelineParseError
+		if errors.As(err, &parseErr) {
```
6543, line 91: why not .Is ?

lafriks: I don't think you can check typed errors with fields with `Is` 🤔

6543: did not looked up this ... I'll check

lafriks:
Added to display error in UI if no pipeline config file is found

6543:
please merge master into

server/pipeline/restart.go
6543: "... sets target to that error value and returns true" ...

we should not use As at all ...

6543: https://go.dev/play/p/zcJL_wTcC0q

it will always be true ...

pipeline/frontend/yaml/error.go
```diff
@@ -0,0 +1,24 @@
...
+package yaml
+
+// PipelineParseError is an error that occurs when the pipeline parsing fails.
+type PipelineParseError struct {
+	Err error
+}
+
+func (e PipelineParseError) Error() string {
+	return e.Err.Error()
+}
```
6543, line 24:
```suggestion
}

// IsPipelineParseError checks if an error is a PipelineParseError
func IsPipelineParseError(err error) bool {
	_, ok1 := err.(PipelineParseError)
	_, ok2 := err.(*PipelineParseError)
	return ok1 || ok2
}
```

server/pipeline/restart.go
```diff
@@ -85,6 +95,10 @@ func Restart(ctx context.Context, store store.Store, lastBuild *model.Build, use
...
```
6543, line 99:
```suggestion
		if yaml.IsPipelineParseError(err) {
```

pipeline/frontend/yaml/error.go
6543:
```go
// IsPipelineParseError checks if an error is a PipelineParseError
func IsPipelineParseError(err error) bool {
	_, ok := err.(*PipelineParseError)
	return ok
}
```
would do it too ... if we are sure we only use &PipelineParseError{} version ...

lafriks: Why not just use errors.As? As that would work also for wrapped errors

6543: https://pkg.go.dev/errors#As

6543: https://go.dev/play/p/pT0g50DWTku vs https://go.dev/play/p/2X1yEErc5XO I would say ...

in the end if this is the only blocking thing I'm fine

lafriks: changed to `errors.Is`: https://go.dev/play/p/hpv3yhmlvzk

6543 (approved):

PR #1069, opened 2022-08-03
Do not show pipeline name if it's a single file

qwerty287:
Only set the pipeline name if there's a directory, not a single file.This should prevent pipeline names like "woodpecker" and "drone".


lafriks (approved):

anbraten:
@qwerty287 Why don't you want to set that name? Is there a functional reason or do you just want to hide it from UI? (In that case I would do the change in the UI)

lafriks:
maybe hide name in UI only if there is single pipeline in job?

6543:
status report will also be affected ...

Issue #1071, opened 2022-08-03
[Important] Limit secret access to plugins only


6543:
similar to the images filter for secrets, add a checkbox to only inject secret if step is exec as plugin.

this ensure only the intended entry-point do get the secrets to handle.

this does help if a plugin is not based from scratch image but do contain an shell that could be used.

---
bounty: 50$

anbraten:
I don't see how this could protect secrets? You could just create a plugin (normal docker image) that leaks the secrets, couldn't you?

lafriks:
this could be helpful in combination with trusted images ex. currently setting trusted image `woodpeckerci/plugin-git` you can still use:
```
image woodpeckerci/plugin-git
commands:
  - echo ${SECRET}
```

adding such option would prevent that

6543:
```
image: woodpeckerci/plugin-git
commands:
  - echo ${SECRET} | base58
  ```
-> it's a trusted image
-> secret is leaked

6543:
bounty: 50$

PR #1072, opened 2022-08-03
Publish preview docker images of pulls

6543:
publish docker images of pulls as preview under `pull_${CI_COMMIT_PULL_REQUEST}` tags

secrets are also restricted by image so secret exfiltration should not be possible

- [server images](https://hub.docker.com/r/woodpeckerci/woodpecker-server/tags?page=1&name=pull_)
- [cli images](https://hub.docker.com/r/woodpeckerci/woodpecker-cli/tags?page=1&name=pull_)
- [agent images](https://hub.docker.com/r/woodpeckerci/woodpecker-agent/tags?page=1&name=pull_)


lafriks:
would be cool to also delete them if PR is closed/merged

6543:
I think we can remove block of  #1071 as we have all pipelines gated too

anbraten (approved):

PR #1078, opened 2022-08-05
Fetch repositories with remote ID if possible

### LINKED ISSUE #648: support nested repositories (GitLab)

### LINKED ISSUE #854: Rename of repo in forge break woodpecker repos
to reproduce:
1. create repo `test`
2. enable repo in woodpecker
3. rename repo to `Test`
4. broken

to fix it atm:
1. Delete Repo in woodpecker & hooks if any exist afterwards in forge
2. sync & activate

expect:
1. url path should be case insensitive
2. if api return permanent redirect update record based on this

qwerty287:
Use IDs of the forge to fetch repositories instead of their names and owner names. This improves handling of renamed and transferred repos.

TODO

- [ ] try to support as many forges as possible
    - [x] Gogs (no API)
    - [ ] Bitbucket Server
    - [x] Coding (no API?)
- [x] update repo every time it is fetched or received from the forge
- [x] if repo remote IDs are not available, use owner / name to get it
- [x] handle redirections (redirect a renamed repo to its new path)
- [x] ~~pull all repos once during migration to update ID (?)~~ issue fixed by on-demand loading of remote IDs
- [x] handle redirections in web UI
- [ ] improve handling of hooks after a repo was renamed (currently it checks for a redirection to the repo)
- [x] tests
- [x] `UNIQUE` constraint for remote IDs after migration shouldn't work (all repos have an empty string as remote ID)

close #854
close #648 partial
close https://codeberg.org/Codeberg-CI/feedback/issues/46

Possible follow-up PRs
- apply the same scheme on everything fetched from the remote (currently only users)


server/model/repo.go
```diff
@@ -26,6 +26,7 @@ import (
 type Repo struct {
 	ID                           int64          `json:"id,omitempty"                    xorm:"pk autoincr 'repo_id'"`
 	UserID                       int64          `json:"-"                               xorm:"repo_user_id"`
+	RemoteID                     string         `json:"-"                               xorm:"UNIQUE 'remote_id'"`
```
lafriks, line 29: It would be preferred to also add migration that adds this column using Sync2

lafriks (changes requested):

server/model/repo.go
lafriks: I know but from experience with Gitea if this is not done as migration and this field will be used in future migrations it can be problem that migration will fail as there will be no such column in table and thus will create problems if upgrades are not made regularly

Makefile
```diff
@@ -57,8 +57,8 @@ vendor:
 	go mod tidy
 	go mod vendor
 
-format:
-	@gofmt -s -w ${GOFILES_NOVENDOR}
+format: install-tools
```
6543, line 60: ack :+1:

pipeline/rpc/proto/woodpecker.pb.go
```diff
@@ -7,10 +7,11 @@
 package proto
 
 import (
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
```
6543, line 10: generated files changed - whitch protoc version do you use?

server/remote/bitbucketserver/convert.go
```diff
@@ -51,6 +52,7 @@ func convertStatus(status model.StatusValue) string {
 // structure to the common Woodpecker repository structure.
 func convertRepo(from *internal.Repo) *model.Repo {
 	repo := model.Repo{
+		RemoteID:     strconv.FormatInt(int64(from.ID), 10),
```
6543, line 55:
```suggestion
		RemoteID:     fmt.Sprint(from.ID),
```
isn't Sprint the new suggested way?

server/remote/coding/coding.go
```diff
@@ -172,6 +172,7 @@ func (c *Coding) Repo(ctx context.Context, u *model.User, owner, name string) (*
 		return nil, err
 	}
 	return &model.Repo{
+		// TODO RemoteID:     project.ID,
```
6543, line 175:
```suggestion
		// TODO(1138) RemoteID:     project.ID,
```

and add new headline at #1138 and describe the issue ...

server/remote/coding/coding.go
```diff
@@ -199,6 +200,7 @@ func (c *Coding) Repos(ctx context.Context, u *model.User) ([]*model.Repo, error
 			return nil, err
 		}
 		repo := &model.Repo{
+			// TODO RemoteID:     project.ID,
```
6543, line 203: #1138

server/store/datastore/migration/008_repo_add_remote_id.go
```diff
@@ -0,0 +1,39 @@
...
+
+package migration
+
+import (
+	"github.com/woodpecker-ci/woodpecker/server/model"
+	"xorm.io/xorm"
+)
+
+type RepoV008 struct {
+	RemoteID string `json:"-" xorm:"'remote_id'"`
```
6543, line 23: migration not needed, will be executed anyway ...!

https://woodpecker-ci.org/docs/development/guides#add-a-new-migration

server/store/datastore/redirection.go
```diff
@@ -0,0 +1,33 @@
+package datastore
+
+import (
+	"github.com/woodpecker-ci/woodpecker/server/model"
+	"xorm.io/xorm"
+)
+
+func (s storage) GetRedirection(fullName string) (*model.Redirection, error) {
```
6543, line 8: redicrection need unit tests

server/model/perm.go
```diff
@@ -25,15 +25,15 @@ type PermStore interface {
...
-	Repo    string `json:"-"       xorm:"-"` // TODO: better caching (use type *Repo)
-	Pull    bool   `json:"pull"    xorm:"perm_pull"`
-	Push    bool   `json:"push"    xorm:"perm_push"`
-	Admin   bool   `json:"admin"   xorm:"perm_admin"`
-	Synced  int64  `json:"synced"  xorm:"perm_synced"`
-	Created int64  `json:"created" xorm:"created"`
-	Updated int64  `json:"updated" xorm:"updated"`
+	UserID  int64 `json:"-"       xorm:"UNIQUE(s) INDEX NOT NULL 'perm_user_id'"`
+	RepoID  int64 `json:"-"       xorm:"UNIQUE(s) INDEX NOT NULL 'perm_repo_id'"`
+	Repo    *Repo `json:"-"       xorm:"-"`
```
6543, line 30: thanks for solving this todo!

server/store/datastore/migration/008_repo_add_remote_id.go
lafriks: But it will fail if any future migration uses this field and upgrade is happening from version where this field is not added yet

6543: no it wont have a problem either ... if field is realy needed, because of a complex transformation then a sync2 with onld struct should be added there ... migrations are handled theme based not version based (that makes them backportable)

6543: woodpecker migrations != gitea migrations

6543:
I think this is also a way towards #648

server/api/hook.go
```diff
@@ -26,7 +26,6 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/rs/zerolog/log"
```
6543, line 28:
```suggestion
	"github.com/rs/zerolog/log"

```

server/store/datastore/redirection_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+func TestHasRedirectionForRepo(t *testing.T) {
+	store, closer := newTestStore(t, new(model.Redirection))
+	defer closer()
+
+	redirection := &model.Redirection{
+		RepoID:   1,
+		FullName: "foo/bar",
+	}
+	if err := store.CreateRedirection(redirection); err != nil {
```
6543, line 64: try to create a redirect that should not work and check if it errors too

good tests not only test against success but also against expected errors

server/store/datastore/repo_test.go
```diff
@@ -191,32 +194,36 @@ func TestOwnedRepoList(t *testing.T) {
 		Owner:    "bradrydzewski",
```
6543, line 194: can you add test cases for redirects and test GetRepoNameFallback ?

server/store/datastore/repo.go
```diff
@@ -128,23 +165,29 @@ func (s storage) RepoBatch(repos []*model.Repo) error {
...
+				// create redirection
+				err := s.createRedirection(sess, &model.Redirection{RepoID: repo.ID, FullName: repo.FullName})
+				if err != nil {
+					return err
+				}
+			}
 			if _, err := sess.
-				Where("repo_owner = ? AND repo_name = ?", repos[i].Owner, repos[i].Name).
-				Cols("repo_scm", "repo_avatar", "repo_link", "repo_private", "repo_clone", "repo_branch").
+				Where("remote_id = ?", repos[i].RemoteID).
```
6543, line 183: if repo_id == 0 ?

server/api/repo.go
```diff
@@ -302,7 +311,7 @@ func MoveRepo(c *gin.Context) {
 		return
 	}
 
-	from, err := remote.Repo(c, user, owner, name)
+	from, err := remote.Repo(c, user, "0", owner, name)
```
6543, line 314:
```suggestion
	from, err := remote.Repo(c, user, "", owner, name)
```

6543:
@qwerty287 what does represent an empty remote_id in backend. "0" or "" or both ? -> we need to handle these cases or make sure they are validated before inserted

6543:
no we just need only one place where a func can determine if a repo has an id or not.
-> single source of truth

server/remote/bitbucket/bitbucket.go
```diff
@@ -142,8 +142,11 @@ func (c *config) Teams(ctx context.Context, u *model.User) ([]*model.Team, error
 }
 
 // Repo returns the named Bitbucket repository.
-func (c *config) Repo(ctx context.Context, u *model.User, owner, name string) (*model.Repo, error) {
-	repo, err := c.newClient(ctx, u).FindRepo(owner, name)
+func (c *config) Repo(ctx context.Context, u *model.User, id, owner, name string) (*model.Repo, error) {
+	if id == "" {
```
6543, line 146: like this would do with import "0" a lookup that fail -> unexpected behaviour

server/model/repo.go
```diff
@@ -99,3 +104,9 @@ type RepoPatch struct {
 	AllowPull                    *bool           `json:"allow_pr,omitempty"`
 	CancelPreviousPipelineEvents *[]WebhookEvent `json:"cancel_previous_pipeline_events"`
 }
+
+type RemoteID string
+
+func (r RemoteID) IsSet() bool {
```
6543, line 110: if you have a better name go on ...

server/model/repo.go
```diff
@@ -99,3 +104,9 @@ type RepoPatch struct {
 	AllowPull                    *bool           `json:"allow_pr,omitempty"`
 	CancelPreviousPipelineEvents *[]WebhookEvent `json:"cancel_previous_pipeline_events"`
 }
+
+type RemoteID string
```
6543, line 108: with this we could in future make it an interface and do more granular validation per remote if we ever need it, else it's just make sure the logic is only implemented once -> IsSet()

6543: PS: the whole conversion from and to RemoteID from int64... we could also write a helper function if desired ...

6543 (approved):

server/model/repo.go
6543: @qwerty287 can you rename I'm afk

jdoubleu:
Just noticed, that [this part](https://github.com/woodpecker-ci/woodpecker/blob/52d3652f2e714eec09b7e787bb7455a940b91140/server/remote/gitlab/gitlab.go#L285-L289) has not been removed:
```golang
// TODO(648) remove when woodpecker understands nested repos
if strings.Count(repo.FullName, "/") > 1 {
    log.Debug().Msgf("Skipping nested repository %s for user %s, because they are not supported, yet (see #648).", repo.FullName, user.Login)
    continue
}
```

It was introduced by [!656](https://github.com/woodpecker-ci/woodpecker/pull/656/files).

Is this correct?

Issue #276, opened 2021-08-23
Enhance config linter


anbraten:
Woodpecker should properly lint and evaluate the ci file/files and show errors with the specific position to users. 

- [x] return linter errors to UI #460
- [x] add way to let linter return warnings like deprecations / bad habits #922
- [x] lint every pipeline config before executing with json schema (currently only some checks)
- [x] add cli command to execute pipeline linter (currently only json schema check)
- [ ] add endpoint to be able to paste in config code and highlight the result in UI #892
- [ ] show json schema errors of executed pipelines in UI by highlighting the wrong part
- [x] allow to lint using multiple workflow files (needed for #2685)

With #89 we could show the errors at their specific position in the config. Maybe somehow like:

![Screenshot from 2021-08-24 00-53-36](https://user-images.githubusercontent.com/6918444/130529580-a578c52c-4128-415f-91ae-7889cf053d8a.png)

6543:
@anbraten would it help if we just add an endpoint, where you can send config in body and get lint error in return?

anbraten:
There are a few things I had in mind:

- lint every pipeline config before executing with json schema (currently only some checks)
- show json schema errors of executed pipelines in UI by highlighting the wrong part
- add endpoint to be able to paste in config code and highlight the result in UI
- add cli command to execute pipeline linter (currently only json schema check)
- add way to let linter return warnings like deprecations / bad habits

x80486:
> @anbraten would it help if we just add an endpoint, where you can send config in body and get lint error in return?

I think along with some integration in the UI will do it. It's always great to double check the config before pushing any related changes.

6543:
https://www.npmjs.com/package/jsonschema
or
https://www.npmjs.com/package/ajv
?

PR #1082, opened 2022-08-08
Add syntax highlighting for pipeline config

lafriks:
Screenshots:

attels

attels

### DIFF
--- a/web/src/components/atomic/SyntaxHighlight.ts
+++ b/web/src/components/atomic/SyntaxHighlight.ts
@@ -1,0 +1,34 @@
+import '~/style/prism.css';
+
+import Prism from 'prismjs';
+import * as Vue from 'vue';
+import { VNode } from 'vue';
+
+declare type Data = Record<string, unknown>;
+
+export default Vue.defineComponent({
+  props: {
+    code: {
+      type: String,
+      default: '',
+    },
+    language: {
+      type: String,
+      default: 'yaml',
+    },
+  },
+  setup(props, { attrs }: { attrs: Data }) {
+    const { h } = Vue;
+    const { code, language } = props;
+    const prismLanguage = Prism.languages[language];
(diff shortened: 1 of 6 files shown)


anbraten:
@lafriks Can we use prismjs to highlight yaml errors? I think it would be quite nice to show errors of the configuration using our editor/view in the long term.

lafriks:
@anbraten where are yaml errors displayed? you mean my other PR?

anbraten:
@lafriks Yep. The idea to show the config file and underline / highlight the incorrect part as you can see by the screenshot in #276. We could also use that to have some kind of an editor in the long term to allow users to paste and check their config without committing it.

lafriks:
I can add that later on when this is merged

lafriks:
Only drawback would be that it would require parsing yaml parser error messages

anbraten:
> I can add that later on when this is merged

It's definitely a task for a new PR. I just wanted to make sure prismjs supports additional highlighting like this underline wave other editors use to show errors in general. Couldn't find this exact use-case from a short search, but it seems primsjs supports some kind of plugin mechanism that could be used for it.

web/src/components/atomic/SyntaxHighlight.ts
```diff
@@ -0,0 +1,34 @@
+import '~/style/prism.css';
+
+import Prism from 'prismjs';
+import * as Vue from 'vue';
+import { VNode } from 'vue';
```
anbraten, line 5: I would prefer using named imports instead of a wildcard import.

```suggestion
import { h, defineComponent, VNode } from 'vue';
```

anbraten (approved):

web/src/components/atomic/SyntaxHighlight.ts
lafriks: done

PR #1083, opened 2022-08-09
Update dependencies and tiny adjustments to UI

anbraten:
- updated all web/ dependencies
- fixed icon positioning issue
- changed text color of pipeline message
- add hover underline effect to breadcrumb org name

# before
Screenshot from 2022-08-09 09-30-55
Screenshot from 2022-08-09 09-28-48

# after
Screenshot from 2022-08-09 09-30-30
Screenshot from 2022-08-09 09-30-16


6543 (approved):

lafriks:
why unplugin is needed? at least for me component.d.ts gets generated way differently that it's here.

attels

either some configuration is missing from this commit or something else. Also eslint for this file should be excluded otherwise it's complaining a lot about it :)

anbraten:
@lafriks Strange, that it is detecting additional entries for your setup. The file should be ignored for eslint and prettier as well. Maybe you need to reload your editor 🤔

lafriks:
reloading vscode does not help. Also file is generated when running `yarn run build` so that is not editor dependent

lafriks:
I resolved vscode eslint/prettier problems #1084 but generated `component.d.ts` is still an issue (generated differently)

PR #1084, opened 2022-08-09
Fix vscode eslint/prettier settings and Vue extension recommendation

lafriks:


.vscode/extensions.json
```diff
@@ -6,7 +6,7 @@
     "dbaeumer.vscode-eslint",
     "esbenp.prettier-vscode",
     "voorjaar.windicss-intellisense",
-    "johnsoncodehk.volar",
+    "Vue.volar",
```
6543, line 9: codium does only know johnsoncodehk.volar ...

6543: https://open-vsx.org/extension/johnsoncodehk/volar

lafriks: But it contains very old version (0.30).

https://github.com/johnsoncodehk/volar - starting from 0.39 version it's published under Vue.volar name:
attels

lafriks: Also there is already fix for publishing in open-vsx 10 days ago, so hopefully it will be also soon there with correct name:
https://github.com/open-vsx/publish-extensions/commit/b4d5f50237fcce8007b26dd12cbcb2a4b8785e22

6543 (approved):

PR #1095, opened 2022-08-12
Fix GitLab duplicate pipelines (#618)

### LINKED ISSUE #618: Gitlab status / pipelines are separated
Gitlab seems to have an issue which results in single Woodpecker pipelines being shown as multiple pipelines in Gitlab.

 expected:
Screenshot from 2021-12-18 23-32-52

currently (two pipelines with a single job):
Screenshot from 2021-12-18 23-33-12

Gitlab issue: https://gitlab.com/gitlab-org/gitlab/-/issues/13134

dsanader:
PipelineID is a GitLab internal ID not related to Woodpecker's build numbers. Posting a status with an wrong ID, creates a new pipeline object in GitLab.

It's not a mandatory field and it should be retrieved first through the commit status list API if we were to use it.

Since It doesn't seem to be useful in Woodpecker's case, I've simply removed it. Repo ref (commit sha) and context are enough to update pipeline statuses.

closes #618


dsanader:
I've mixed the issues ID, it relates to issue #618, not #648. My commit message is wrong.

anbraten (approved):

6543:
Please backport :)

PR #1111, opened 2022-08-15
use new multi when

anbraten:


anbraten:
we now have #770, so we should use them directly instead of writing those restrictions per step

PR #1031, opened 2022-07-18
Implement Refresher for GitLab and refactor nits

6543:
close #1019

PR #1120, opened 2022-08-24
Implement Refresher for GitLab (#1031)

6543:
Packport #1031 (without unrelated refactoring)


server/remote/gitea/gitea.go
```diff
@@ -40,6 +40,7 @@ const (
 	authorizeTokenURL = "%s/login/oauth/authorize"
 	accessTokenURL    = "%s/login/oauth/access_token"
 	perPage           = 50
+	giteaDevVersion   = "v1.18.0"
```
6543, line 43:
```suggestion
```

PR #1127, opened 2022-08-25
Remove own copy of oauth2 implementation

6543:
at some point (~7years ago) the oauth2 implementation was copied into the code-base and never touched.
We only use it for gitlab the rest is already back using std.

the only reason that make std complicated is, that you can not specify a transport that easy with it anymore

TODO:
- [x] figure out how to alter transport for std oauth2


6543:
PS: tested... works

PR #1129, opened 2022-08-25
Use One File for all common flags

6543:
share code for flags used by server, agent and cli

### DIFF
--- a/cli/common/flags.go
+++ b/cli/common/flags.go
@@ -14,9 +14,13 @@
 
 package common
 
-import "github.com/urfave/cli/v2"
+import (
+	"github.com/urfave/cli/v2"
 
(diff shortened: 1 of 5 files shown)


6543:
well I have to many open pulls and so I'm not mad at all, if they are implemented by others or if they get picked up and finished.

thanks for cleaning up :heart:

6543:
PS: I leave them open sometimes just as notes ... yes I could write a ticket but sometimes a code diff is just more descriptif (for developers ;) )
PS:PS: for newer wip pulls just ask first if you want to progress on them, for old-long not touched ones just go on

_just not close things if there indicate a issue/task or have code to pick from, until code is picked or an issue is created from it_

PR #1130, opened 2022-08-25
CI: add branches filter

6543:


6543:
expect: exec for pull
get: 204 !

anbraten:
similar to #1111 at least that was my intention in there as well

6543:
yes but this looks like it is broken ... on gitea it seems to work :O

anbraten:
@6543 I added some `root.when` filters and replaced the `branches` filter.

anbraten (approved):

6543:
@anbraten we don't get any builds lately ... I'll revert it partially and we need a more clever solution

6543:
7fede8ea1dce752745ffd8c55e180e222b69dc4c

anbraten:
@6543 Yip, saw that as well. I will try to find a solution for the root level when filters. Currently they are reporting pending state back although they are not running at all

PR #1143, opened 2022-08-30
use yaml alias and adjust build targets for docker images

6543:
- use yaml alias for plaftorms settings
- add linux/s390x for agent/cli release
- add linux/riscv64
- remove darwin as mac use linux in hypervisor for docker


6543:
blocked by https://github.com/techknowlogick/xgo/issues/168

PR #1145, opened 2022-08-31
Use versioned docs

### LINKED ISSUE #1051: Versioned docs
Maybe we should start versioning our docs: https://docusaurus.io/docs/versioning to solve issues like #1048 

Alternative: Add info notes to all new features and write down which version will first implement them.

anbraten:
closes #1051 

Idea is we always serve the latest stable and the next version.

Default is latest:
Screenshot from 2022-08-31 17-38-04

You can switch to `next` version:
Screenshot from 2022-08-31 17-37-52

# TODO
- [x] downgrade version `0.15.x` docs to old state
- [x] update paths `/docs/` to the file paths so docusaurus can properly generate links with the corresponding versions.


anbraten:
> Would it be possible to checkout the release-0.15 branch and build the docs from there instead of adding everything twice?

We could add jobs which checkout that data. Maybe sub-modules, but that could end in some pretty complicated logic 🤔

anbraten:
That sounds like an interesting idea. I would like to merge this PR for general versioning first and try a bit with loading the "old" docs via release branch etc in a followup.

6543 (approved):

PR #1158, opened 2022-09-03
Improve logging

anbraten:
# changes
- change default log level from `warn` to `info`
- add log info message at server / agent start with version info
- cleanup server startup function
- change some level of some log calls from `info` to `debug`

### DIFF
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -37,6 +37,7 @@ import (
 	"github.com/woodpecker-ci/woodpecker/pipeline/backend"
 	"github.com/woodpecker-ci/woodpecker/pipeline/rpc"
 	"github.com/woodpecker-ci/woodpecker/shared/utils"
+	"github.com/woodpecker-ci/woodpecker/version"
 )
 
 func loop(c *cli.Context) error {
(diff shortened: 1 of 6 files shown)


cmd/agent/agent.go
```diff
@@ -69,7 +70,7 @@ func loop(c *cli.Context) error {
 		)
 	}
 
-	zerolog.SetGlobalLevel(zerolog.WarnLevel)
+	zerolog.SetGlobalLevel(zerolog.InfoLevel)
```
lafriks, line 73: Ideally this should be settable by env variable

anbraten: You can set it. This is the value it uses in case you haven't specified it.

cmd/agent/agent.go
```diff
@@ -69,7 +70,7 @@ func loop(c *cli.Context) error {
...
 	if zerolog.GlobalLevel() <= zerolog.DebugLevel {
```
6543, line 74: shouldn't this be after line 76 ?

anbraten: Good point that's pretty strange 🤔 Line 72-75 should be in general behind 85?

anbraten: changed it

6543 (approved):

PR #1167, opened 2022-09-03
Partial-Backport of pipeline from current master

6543:


.woodpecker/test.yml
```diff
@@ -1,97 +1,84 @@
...
+    image: *golang_image
+    group: prepare
+    commands:
+      - go mod vendor
+    when:
+      path: *when_path
+
   dummy-web:
-    image: golang:1.16
+    image: *golang_image
```
6543, line 24:
```suggestion
    image: *golang_image
    pull: true
```

6543: we should at least once try to pull the golang image (on master too ...)

anbraten (approved):

PR #1184, opened 2022-09-13
Hide example of YAML override and extension

6543:
see #1192 for more infos


6543:
we have to look into https://github.com/go-yaml/yaml if it does parse this correctly!

the fix of this pull was done by error reports from https://jsonformatter.org/yaml-parser

[codeberg.org/6543/go-yaml2json](http://codeberg.org/6543/go-yaml2json) does not support it, witch does utilize github.com/go-yaml/yaml, so it could potentially be an upstream issue

6543:
https://github.com/yaml/yaml/issues/35

6543:
issue tracked at #1192 ... will disable example for now

anbraten (approved):

PR #1185, opened 2022-09-14
Add dedicated DroneCI env compatibility layer

6543:
for more info see `https://github.com/drone-plugins/drone-plugin-lib/pull/37/files`
and c79d49c862ad800f30abcd4fc9b543f99a8e70f9

### DIFF
--- a/pipeline/frontend/drone_compatibility.go
+++ b/pipeline/frontend/drone_compatibility.go
@@ -1,0 +1,58 @@
+// Copyright 2022 Woodpecker Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


pipeline/frontend/drone_compatibility.go
```diff
@@ -0,0 +1,59 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package frontend
+
+// setDroneEnviron set  dedicated to DroneCI environment vars as compatibility layer
```
6543, line 17:
```suggestion
// setDroneEnviron set dedicated to DroneCI environment vars as compatibility layer
```

6543:
> Maybe, we can look if we can extend this later to other CI systems like GH Actions, Circle CI or whatever else

as long as there plugins are handled via environment vars and do run in a docker container ... yes we can

Issue #267, opened 2021-08-17
Allow users to register own agents


anbraten:
It would be nice to add my own agent to some kinda public woodpecker instance. 
- There should be an option for the server to allow user-registered agents. 
- There should be to kinds of agents: shared agents (probably provided by the woodpecker admin), user-agents
- There has to be some way of regiserting new agents without the need to share a single secret (like currently implemented) with evil users
- User-agents should be assignable to own repos, but not to all repos (securtiy risk, if secrets / sensitive code gets loaded to some strangers agent)
- Agents should have a setting to disallow privilged / trusted builds
- There should be some way to invite agents to other repos (not directly owner of repo like a repo from some org) 

# TODO
- [ ] move label filtering to server to prevent agents from taking pipelines they should not get
- [ ]

anbraten:
From: https://codeberg.org/Codeberg-CI/feedback/issues/94#issuecomment-690822

> I like that invitation idea. Using some kind of invitations and having a two way approval is important as the agent owner does not want some stranger to execute anything on his agent and on the other hand we need to protect orgs / other users from automatically scheduling pipelines on some strangers system potentionally doing harmful stuff with their secrets / code, etc.

lafriks:
Maybe agent could authenticate with per-user secret that would be stored in user table. This way we could automatically detect that this agent is user owned

anbraten:
That's already possible after #1189 gets merged. It allows to connect an agent with a previously created token and the token gets assigned to an user (for the beginning only admin users) which acts as an owner of that agent.

For sharing agents with others it would be somehow needed that the agents owner is able to send an invitation to another org / user, so they can accept it.

PR #1189, opened 2022-09-14
Store an agents list and add agent heartbeats

### LINKED ISSUE #252: Show agents and some information about them in webinterface

### LINKED ISSUE #536: add agent heartbeat
agents should report back that they are still alive. if one fail without gracefull cancle running jobs etc, we can recover based on that after a ttl.

technical idear: each agent create a uuid on start that will be stored into db with timestamp of last heartbeat. if a agent start exec a new job the agent uuid is stored with it. if agent haeartbeat did not happen the timeToLive will (re-)set all "running" jobs as pendig so a other/new agent can pick up that work.

anbraten:
closes #536
closes #252 

prepares #267 

# TODO

- [x] update agent model and add datastore functions 
- [x] update grpc proto and add register agent and heartbeat
- [x] add `RegsiterAgent` and agent heartbeat calls to agent
- [x] fix agent not re-trying to register / doing the heartbeat
- [x] find way to store agent-id of system agens (currently all agents) somehow so the agent can be matched to the db entry
- [x] add api endpoint to get all agents as admin
- [x] show agents in admin UI
- [x] allow to change name of an agent in admin UI
- [x] fix system agents updating the same (first) agent containing the system token
  - return agent-id on register
  - use agent-id for heartbeat?

Screenshot from 2022-09-30 18-54-02

image


server/model/agent.go
```diff
@@ -14,18 +14,24 @@
...
-// TODO: check if it is actually used or just some relict from the past
-
 type Agent struct {
-	ID       int64  `xorm:"pk autoincr 'agent_id'"`
-	Addr     string `xorm:"UNIQUE VARCHAR(250) 'agent_addr'"`
-	Platform string `xorm:"VARCHAR(500) 'agent_platform'"`
-	Capacity int64  `xorm:"agent_capacity"`
-	Created  int64  `xorm:"created 'agent_created'"`
-	Updated  int64  `xorm:"updated 'agent_updated'"`
+	ID          int64 `xorm:"pk autoincr"`
```
6543, line 18: this will need a migration: drop old table and create new one with sync2

lafriks:
Where does agent stores it's ID between restarts?

anbraten:
> Where does agent stores it's ID between restarts?

It does not yet. For user-registered users I would like to get the db entry by searching using the token. For "system"-level agents (the type we currently already have), they get a generated id on the first start. This one would be needed to be saved to a file for example.

anbraten:
Or do you have a nice idea?

6543:
I like the agent's to be stateles ... we could do generate an UUID for each run but use some .env var to send a user defined var for reidentification ?

anbraten:
IMO our agents are still somehow stateless if we save that uuid to a file. The id will be only used for system level agents to reidentify and if it does not exists it will simply be a new agent. In the long term we could think about removing support for the shared agent secret completely.

lafriks:
File probably would be ok, otherwise there is no really way to reidentify same agent and for each agent restart there will be new record in database. Also for future it would be nice if agent could unregister itself but this is probably for future PRs so that for auto-scaling agents would not create a large list of agents that won't be coming back.

lafriks:
In docker if agents are scaled in docker and multiple agent instances are run in same docker host... they would have same persistent volume attached that could be a problem as they would get same file with same UUID

6543:
> File probably would be ok, otherwise there is no really way to reidentify same agent and for each agent restart there will be new record in database. Also for future it would be nice if agent could unregister itself but this is probably for future PRs so that for auto-scaling agents would not create a large list of agents that won't be coming back.

well I would not implement autoscaling by spowning new threads ... and instead refactor the loop in the agent implementation to only have one main loop and use go routines for tasks

and with a headbeat I would delete agent from db if it timed out based on mentioned UUID

lafriks:
> well I would not implement autoscaling by spowning new threads ... and instead refactor the loop in the agent implementation to only have one main loop and use go routines for tasks

By auto-scaling I meant firing up/down new VMs etc

> and with a headbeat I would delete agent from db if it timed out based on mentioned UUID

This would probably best option as I don't see how uuid persistence could be achieved reliably on docker

6543:
this timeout also can then mark procs marked as running at that specific agent as pending again so a other can pick it up ...

this will fix an issue where running steps on an agent if the agent get killet atm stay at running forever until the pipeline timeout do cancle them and then they are failed but well they dont have to

cmd/agent/agent.go
```diff
@@ -48,9 +49,11 @@ func loop(c *cli.Context) error {
 		hostname, _ = os.Hostname()
 	}
 
+	platform := runtime.GOOS + "/" + runtime.GOARCH
```
6543, line 52: on agent start can we print out some info ... so it's nice for debuging ... what ID, platform, hostname ,etc ...

cmd/agent/agent.go
```diff
@@ -111,8 +114,7 @@ func loop(c *cli.Context) error {
 		c.String("server"),
 		transport,
 		grpc.WithPerRPCCredentials(&credentials{
-			username: c.String("grpc-username"),
-			password: c.String("grpc-password"),
+			token: c.String("grpc-token"),
```
6543, line 117: how did it even worked before?

cmd/agent/agent.go
anbraten: done

6543:
uh nice! give me some *minutes to review

pipeline/rpc/proto/woodpecker.proto
```diff
@@ -113,5 +94,36 @@ message LogRequest {
...
+message AuthRequest {
+  string agent_token = 1;
+  int64  agent_id    = 2;
+}
+
+message AuthReply {
+  string status        = 1;
+  int64  agent_id      = 2;
+  string access_token  = 3;
+}
```
6543, line 129: editorconfig ;)

-> LN

server/store/datastore/migration/008_recreate_agents_table.go
```diff
@@ -0,0 +1,30 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package migration
+
+import (
+	"github.com/woodpecker-ci/woodpecker/server/model"
```
6543, line 18: format

6543:
please resolve ;)

6543:
uh and of the rpc move ... we might move the rpc stuff into it's own root directory, as it's a shared thing between agent and server but nothing else should use/touch it ...

6543:
and we need to update https://woodpecker-ci.org/docs/next/development/architecture acordingly

anbraten:
> uh and of the rpc move ... we might move the rpc stuff into it's own root directory, as it's a shared thing between agent and server but nothing else should use/touch it ...

I would leave that for another PR. Should I undo the move of the agent code, to keep changes to a minimum?

cmd/agent/agent.go
```diff
@@ -148,6 +167,29 @@ func loop(c *cli.Context) error {
 		return err
 	}
 
+	agentID, err = client.RegisterAgent(ctx, platform, engine.Name(), version.String(), parallel)
+	if err != nil {
+		return err
+	}
+
+	log.Debug().Msgf("agent registered with id %d", agentID)
```
lafriks, line 175: To be more consistent 
```suggestion
	log.Debug().Msgf("Agent registered with ID %d", agentID)
```

server/grpc/auth_server.go
```diff
@@ -0,0 +1,64 @@
+package grpc
+
+import (
+	"context"
+
+	"github.com/rs/zerolog/log"
```
lafriks, line 6:
```suggestion
	"github.com/rs/zerolog/log"

```

server/grpc/jwt_manager.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Verify verifies the access token string and return a user claim if the token is valid
+func (manager *JWTManager) Verify(accessToken string) (*AgentTokenClaims, error) {
+	token, err := jwt.ParseWithClaims(
+		accessToken,
+		&AgentTokenClaims{},
+		func(token *jwt.Token) (interface{}, error) {
+			_, ok := token.Method.(*jwt.SigningMethodHMAC)
+			if !ok {
+				return nil, fmt.Errorf("unexpected token signing method")
```
lafriks, line 56:
```suggestion
				return nil, errors.New("unexpected token signing method")
```

server/grpc/jwt_manager.go
```diff
@@ -0,0 +1,72 @@
...
+			return []byte(manager.secretKey), nil
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("invalid token: %w", err)
+	}
+
+	claims, ok := token.Claims.(*AgentTokenClaims)
+	if !ok {
+		return nil, fmt.Errorf("invalid token claims")
```
lafriks, line 68:
```suggestion
		return nil, errors.New("invalid token claims")
```

server/grpc/rpc.go
```diff
@@ -382,6 +384,40 @@ func (s *RPC) Log(c context.Context, id string, line *rpc.Line) error {
...
+}
+
+func (s *RPC) ReportHealth(ctx context.Context, status string) error {
+	agent, err := s.getAgentFromContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	if status != "I am alive!" {
+		return fmt.Errorf("Are you alive?")
```
lafriks, line 413:
```suggestion
		return errors.New("Are you alive?")
```

server/grpc/rpc.go
```diff
@@ -438,3 +474,23 @@ func (s *RPC) notify(c context.Context, repo *model.Repo, pipeline *model.Pipeli
 	}
 	return nil
 }
+
+func (s *RPC) getAgentFromContext(ctx context.Context) (*model.Agent, error) {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return nil, fmt.Errorf("metadata is not provided")
```
lafriks, line 481:
```suggestion
		return nil, errors.New("metadata is not provided")
```

server/grpc/rpc.go
```diff
@@ -438,3 +474,23 @@ func (s *RPC) notify(c context.Context, repo *model.Repo, pipeline *model.Pipeli
...
+	}
+
+	values := md["agent_id"]
+	if len(values) == 0 {
+		return nil, fmt.Errorf("agent_id is not provided")
```
lafriks, line 486:
```suggestion
		return nil, errors.New("agent_id is not provided")
```

server/grpc/rpc.go
```diff
@@ -438,3 +474,23 @@ func (s *RPC) notify(c context.Context, repo *model.Repo, pipeline *model.Pipeli
...
+	}
+
+	_agentID := values[0]
+	agentID, err := strconv.ParseInt(_agentID, 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("agent_id is not a valid integer")
```
lafriks, line 492:
```suggestion
		return nil, errors.New("agent_id is not a valid integer")
```

agent/rpc/auth_client_grpc.go
```diff
@@ -0,0 +1,44 @@
+package rpc
+
+import (
+	"context"
+	"time"
+
+	"github.com/woodpecker-ci/woodpecker/pipeline/rpc/proto"
+	"google.golang.org/grpc"
```
lafriks, line 8:
```suggestion
	"google.golang.org/grpc"

	"github.com/woodpecker-ci/woodpecker/pipeline/rpc/proto"
```

agent/rpc/auth_interceptor.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (interceptor *AuthInterceptor) refreshToken() error {
+	accessToken, _, err := interceptor.authClient.Auth()
+	if err != nil {
+		return err
+	}
+
+	interceptor.accessToken = accessToken
+	log.Printf("token refreshed: %v", accessToken)
```
lafriks, line 96:
```suggestion
	log.Printf("Token refreshed: %v", accessToken)
```

lafriks (changes requested):

lafriks (approved):

PR #1196, opened 2022-09-18
 Remove old CLI docs generation code 

qwerty287:
Removes code that's not longer required because we use a newer version of urfave/cli.

### DIFF
--- a/.woodpecker/docs.yml
+++ b/.woodpecker/docs.yml
@@ -5,10 +5,14 @@ variables:
     - ".woodpecker/docs.yml"
 
 pipeline:
+  build-cli:
+    image: golang:1.18
+    commands:
+      - make docs
+
   build:
     image: *node_image
     commands:
-      - apk add golang
       - cd docs/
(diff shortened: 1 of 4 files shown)


6543 (approved):

anbraten (approved):

anbraten (approved):

anbraten:
Should we add this command to the ci / docs generation so it would be updated automatically? :thinking:

anbraten:
Ahh nice.

anbraten:
Or we make it a separate ci step

anbraten (approved):

anbraten:
Strange, quite some failing pipelines recently.

6543:
The generated stuff could be invalid to editorconfig ... we should exclude it from tests

6543:
https://github.com/woodpecker-ci/woodpecker/blob/master/.ecrc

Does not include it ... if it pass then well its good :)

Issue #1193, opened 2022-09-17
Add option to specify healthcheck port for agent


tandav:
### Clear and concise description of the problem

I see that now port 3000 is hardcoded for local agent:

https://github.com/woodpecker-ci/woodpecker/blob/531bdf5bd2d1bb0f6194011de2dce21a02991d70/cmd/agent/agent.go#L92

### Suggested solution

add option to specify port in `.env` file like this:

```
WOODPECKER_AGENT_PORT=3001
```

anbraten:
Just for my interest: Why do you need to change that port of the healthcheck. Isn't the health-endpoint only used in container systems (docker, kubernetes) where you map the port internally?

tandav:
I want to run local agent on a machine where port 3000 is already taken (by gitea in my case). 
When I run agent I get an error:

```sh
./woodpecker-agent
{"level":"error","time":"2022-09-17T20:02:01+03:00","message":"can not listen on port 3000: listen tcp :3000: bind: address already in use"}
```

anbraten:
For local deployments you can just disable the healthcheck endpoint with `WOODPECKER_HEALTHCHECK=false`

tandav:
thank you, I'll close the issue

6543:
I still think it should not be hardcoded

PR #1197, opened 2022-09-18
Make healthcheck port configurable

qwerty287:
Closes https://github.com/woodpecker-ci/woodpecker/issues/1193

### DIFF
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -90,7 +90,7 @@ func loop(c *cli.Context) error {
 
 	if c.Bool("healthcheck") {
 		go func() {
-			if err := http.ListenAndServe(fmt.Sprintf(":%s", c.String("healthcheck-port")), nil); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/agent/agent.go
```diff
@@ -89,7 +90,7 @@ func loop(c *cli.Context) error {
 
 	if c.Bool("healthcheck") {
 		go func() {
-			if err := http.ListenAndServe(":3000", nil); err != nil {
+			if err := http.ListenAndServe(fmt.Sprintf(":%s", c.String("healthcheck-port")), nil); err != nil {
```
6543, line 93:
```suggestion
			if err := http.ListenAndServe(fmt.Sprintf(":%d", c.Int("healthcheck-port")), nil); err != nil {
```
?

6543 (approved):

lafriks:
imho healthcheck should also use addr format to allow specify also IP address to listen on as currently it will listen on all interfaces otherwise

PR #1212, opened 2022-09-27
Set a default value for `build-event` flag of `cli exec` command

sergiofenoll:
Otherwise `cli exec` does not do anything due to the default constraints applied to a pipeline (i.e., some kind of build event is expected).


6543:
@sergiofenoll can you commit changes of "make docs" ?

PR #1218, opened 2022-09-27
Make exec run be detected as WoodpeckerCI

6543:
and exclude docs/docs/40-cli.md from git as it will be generated by CI

e.g. checkout to https://github.com/adamdecaf/cidetails/pull/3 and exec `woodpecker-cli exec --build-event push --backend-engine docker` will fail atm


anbraten (approved):

6543:
if we build docs/docs/40-cli.md in pipeline anyway befor publishing, we might just exclude it from git?

6543:
ok will start with docs for now

6543:
1sec I need to update the when filter a bit

6543:
NOW it should be fine

PR #1219, opened 2022-09-27
Add option to ignore failures on steps

### LINKED ISSUE #834: Ignore failure in certain steps

For specific steps in our pipeline, we want to allow failure, eg overambitious linting or spellchecks, and still make the whole pipeline successful.
Alternatively we could probably isolate this check in a separate pipeline and have that fail. But for both cases I could not find a solution.

a similar feature in github actions:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

Looks like it is already supported to run a specific step on failure (or on failure and success), but I'd like my whole pipeline to show up as succesful if this particular step fails
https://github.com/woodpecker-ci/woodpecker/blob/master/pipeline/samples/sample_4/pipeline.yml#L20-L21

### Suggested solution

allow skipping the exitcode of a specific pipeline step:

add an option on-error/on-failure

```
 check precommit:
   image: xxx/venv-3.9
   on-error: ignore
   commands:
     - tox -e pre-commit
```
 looks like drone also has a failure:ignore option

### Alternative

_No response_

### Additional context

_No response_

### Validations

### LINKED ISSUE #1181: conditional step event failure aways triggered
### Component

server

### Describe the bug

conditional step event failure always triggered  even when there is no error

```yml
    when:
      status:
        - failure
```
is triggered even when no jobs fails

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-6433dfea"}
```

### Additional context

image

image

### Validations

sergiofenoll:
closes #1181
closes #834 

Adds `ignore_failure` to pipeline steps. When it's set to true, if the step fails the following steps continue to execute as if no failure had occurred.

---

failure enums idea:
* fail (default) = if other steps run in parallel, wait for them and then let workflow fail
* cancel = if other steps run in parallel, kill them
* ignore = we mark the step as failed but it wont have any impact


6543:
first impression looks good

lets let the ci run over itselve ;)
and then I'll have a look

6543:
with shell magic you normaly can ignore errors right now too ...
but plugins dont ... and this could be a simplification

In any case if we add it documentation is missing (https://github.com/woodpecker-ci/woodpecker/blob/master/docs/docs/20-usage/20-pipeline-syntax.md)

6543:
just exec `make format` and commit changes - we ensure proper code format via ci

sergiofenoll:
> with shell magic you normaly can ignore errors right now too ... 

You're right, but at work we wanted a simple equivalent to Drone's [`failure: ignore`](https://docs.drone.io/pipeline/docker/syntax/steps/#failure), so I wanted to have a try at implementing it :smile: 

If the addition is okay as-is I'll try to add the docs tomorrow morning (thanks for mentioning it!)

6543:
it does only add to pipeline config ... and dont move much code so why not

the only question would be: show error in UI and a ignored lable or not
not -> as is now
show in ui -> need bit more code (can be added later anyway)

pipeline/pipeline.go
```diff
@@ -100,7 +100,16 @@ func (r *Runtime) Run() error {
 			return ErrCancel
 		case err := <-r.execAll(stage.Steps):
 			if err != nil {
-				r.err = err
+				shouldIgnoreFailure := false
+				for _, step := range stage.Steps {
+					if step.IgnoreFailure {
+						shouldIgnoreFailure = true
```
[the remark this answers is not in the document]
anbraten: I think so as well. It has to be somewhere inside `execAll` only allowing the steps to pass which have `ignore_failure` set.

pipeline/frontend/yaml/container.go
```diff
@@ -42,6 +42,7 @@ type (
 		ExtraHosts    []string               `yaml:"extra_hosts,omitempty"`
 		Group         string                 `yaml:"group,omitempty"`
 		Image         string                 `yaml:"image,omitempty"`
+		IgnoreFailure bool                   `yaml:"ignore_failure,omitempty"`
```
anbraten, line 45: Or should we use `failure: ignore` like drone?

anbraten (commented):
We need to update the schema as well.

pipeline/pipeline.go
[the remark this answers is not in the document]
sergiofenoll: The distinction between `Stage` and `Step` is a little unclear to me. While working on this I used the following pipeline for testing purposes:

```yml
pipeline:
  lint:
    image: danlynn/ember-cli:3.24.0
    commands:
      - npm run lint:js
      - npm run lint:hbs
      - npm run test:ember
    ignore_failure: true
  dummy:
    image: debian
    commands:
      - echo "This is just a dummy step :)"
```

At first I expected that pipeline to be parsed as a single stage with two steps, but it actually turned into two stages, each with one step. That's why I added the code at the stage level in the end.

Do you have an example of when a stage has multiple steps?

pipeline/frontend/yaml/container.go
lafriks: depends if there could be other options beside ignore/fail?

pipeline/pipeline.go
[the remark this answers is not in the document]
sergiofenoll: If I pass `--log-level=debug` to the `cli`, this it says it's executing 2 stages:

```
^^/C/woodpecker >>> go run ./cmd/cli/ --log-level=debug exec ../test-woodpecker-project/.woodpecker.yml
2:18PM ??? cli/common/zerologger.go:18 > LogLevel = debug
2:18PM DBG pipeline/pipeline.go:73 > Executing 2 stages, in order of: CLI=exec
2:18PM DBG pipeline/pipeline.go:83 > stage CLI=exec Stage=woodpecker_stage_0 Steps=woodpecker_step_0
2:18PM DBG pipeline/pipeline.go:83 > stage CLI=exec Stage=woodpecker_stage_1 Steps=woodpecker_step_1
2:18PM DBG pipeline/pipeline.go:164 > Prepare CLI=exec Step=woodpecker_step_0
2:18PM DBG pipeline/pipeline.go:188 > Executing CLI=exec Step=woodpecker_step_0
[lint:L0:0s] + npm run lint:js
[lint:L1:0s] 
[lint:L2:0s] > test-woodpecker-project@0.0.0 lint:js /woodpecker/src
[lint:L3:0s] > eslint . --cache
[lint:L4:0s] 
[lint:L5:0s] + npm run lint:hbs
[lint:L6:1s] 
[lint:L7:1s] > test-woodpecker-project@0.0.0 lint:hbs /woodpecker/src
[lint:L8:1s] > ember-template-lint .
[lint:L9:1s] 
[lint:L10:1s] 
[lint:L11:1s] + npm run test:ember
[lint:L12:2s] 
[lint:L13:2s] > test-woodpecker-project@0.0.0 test:ember /woodpecker/src
[lint:L14:2s] > ember test
[lint:L15:2s] 
[lint:L16:5s] Building
[lint:L17:5s] Environment: test
[lint:L18:5s] building... 
[lint:L19:15s] cleaning up
[lint:L20:15s] cleaning up...
[lint:L21:15s] Built project successfully. Stored in "/tmp/tests-dist-2022828-64-8s85qv.klmto".
[lint:L22:18s] ok 1 Chrome 87.0 - [4 ms] - ember-qunit: Ember.onerror validation: Ember.onerror is functioning properly
[lint:L23:18s] 
[lint:L24:18s] 1..1
[lint:L25:18s] # tests 1
[lint:L26:18s] # pass  1
[lint:L27:18s] # skip  0
[lint:L28:18s] # todo  0
[lint:L29:18s] # fail  0
[lint:L30:18s] 
[lint:L31:18s] # ok
2:19PM ERR pipeline/pipeline.go:235 > process logging failed error="io: read/write on closed pipe" CLI=exec
2:19PM DBG pipeline/pipeline.go:194 > Complete CLI=exec Step=woodpecker_step_0
2:19PM DBG pipeline/pipeline.go:164 > Prepare CLI=exec Step=woodpecker_step_1
2:19PM DBG pipeline/pipeline.go:188 > Executing CLI=exec Step=woodpecker_step_1
[dummy:L0:0s] + echo "This is just a dummy step :)"
[dummy:L1:0s] This is just a dummy step :)
2:19PM ERR pipeline/pipeline.go:235 > process logging failed error="io: read/write on closed pipe" CLI=exec
2:19PM DBG pipeline/pipeline.go:194 > Complete CLI=exec Step=woodpecker_step_1
```

pipeline/frontend/yaml/container.go
6543: :point_up: 

you normaly have three options, fail hard (cancle running steps), fail (let already running steps be finished **we do this atm**), ignore

fail hard will only have an efect if failing step is in a group (parallel exec)

lafriks: Maybe we can have:

failure:
* fail (default)
* cancel
* ignore

sergiofenoll: The different ways to handle a failed step make sense to me, I'll have a go at implementing it :)

pipeline/pipeline.go
```diff
@@ -100,7 +100,16 @@ func (r *Runtime) Run() error {
 			return ErrCancel
 		case err := <-r.execAll(stage.Steps):
 			if err != nil {
-				r.err = err
+				shouldIgnoreFailure := false
+				for _, step := range stage.Steps {
+					if step.IgnoreFailure {
```
6543, line 105: this is the wrong place i would say

pipeline/pipeline.go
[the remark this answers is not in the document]
6543: > AFAIK a `Stage` is the whole pipeline, and a `Step` just one step (running in one container). You have 2 steps in your example, `lint` and `dummy` (and `clone` OFC).

we should rename it, to have one name schema ....!

pipeline/pipeline.go
6543: look at L141 (`execAll()`)

lafriks (approved):

PR #1223, opened 2022-09-29
Sort secret lists and events

### LINKED ISSUE #1176: Very very minor UI nitpick: Secret events are sorted by last change instead of static order
### Component

web-ui

### Describe the bug

image

In the picture, you can see that the labels are differently sorted, making it harder to compare in a highly complex setup (unlike mine). These should be sorted the same way as they are sorted in the settings menu. Maybe these could also get color-coded in the future.

### System Info

```shell
Using a relatively recent next version. So it's not relevant here.
```

### Additional context

_No response_

### Validations

6543:
close  #1176

### DIFF
--- a/server/model/const.go
+++ b/server/model/const.go
@@ -25,6 +25,12 @@ const (
 	EventManual WebhookEvent = "manual"
 )
 
+type WebhookEventList []WebhookEvent
+
+func (wel WebhookEventList) Len() int           { return len(wel) }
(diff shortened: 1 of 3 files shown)


anbraten:
That's just for sorting the secrets. Its not sorting the events array inside the json field.

6543:
sorting events array in backend is not trivial ... we either sort on each secret update/create in backend or do it in frontend

lafriks (approved):

PR #1225, opened 2022-09-29
docs: add alternative naming options

### LINKED ISSUE #1119: Docu: steps can be map **or list**
we support
```yaml
pipeline:
  echo:
    image: debian
    commands:
      - echo hello
```
and
```yaml
pipeline:
  - name: echo
    image: debian
    commands:
      - echo hello
```

keep in mind name is optional if not added step will be numerated

-> #826

jonatan5524:
Added naming option to pipeline syntax with a list and name keyword

resolves #1119

### DIFF
--- a/docs/docs/20-usage/20-pipeline-syntax.md
+++ b/docs/docs/20-usage/20-pipeline-syntax.md
@@ -20,6 +20,24 @@ pipeline:
 ```
 
 In the above example we define two pipeline steps, `frontend` and `backend`. The names of these steps are completely arbitrary.
+Another way to name a step is by using the name keyword:
+
+```yaml
+pipeline:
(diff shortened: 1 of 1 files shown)


docs/docs/20-usage/20-pipeline-syntax.md
```diff
@@ -20,6 +20,24 @@ pipeline:
 ```
 
 In the above example we define two pipeline steps, `frontend` and `backend`. The names of these steps are completely arbitrary.
+Another way to name a step is by using the name keyword:
```
6543, line 23:
```suggestion

Another way to name a step is by using the name keyword:
```

6543 (approved):

PR #1231, opened 2022-10-04
gitea: fetch all branches

ennetech:
Both repo branch view and manual branch selector were limited to the stock pageSize of gitea client.

I've added a loop to fetch all branches from gitea

### DIFF
--- a/server/remote/gitea/gitea.go
+++ b/server/remote/gitea/gitea.go
@@ -462,15 +462,28 @@ func (c *Gitea) Branches(ctx context.Context, u *model.User, r *model.Repo) ([]s
 		return nil, err
 	}
 
-	giteaBranches, _, err := client.ListRepoBranches(r.Owner, r.Name, gitea.ListRepoBranchesOptions{})
-	if err != nil {
-		return nil, err
+	branches := make([]string, 0)
+
+	page := 1
+
+	for page > 0 {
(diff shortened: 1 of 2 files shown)


server/remote/gitea/gitea.go
```diff
@@ -462,15 +462,28 @@ func (c *Gitea) Branches(ctx context.Context, u *model.User, r *model.Repo) ([]s
...
+		if err != nil {
+			return nil, err
+		}
+		if len(giteaBranches) > 0 {
+			for _, branch := range giteaBranches {
+				branches = append(branches, branch.Name)
+			}
+			page++
+		} else {
+			page = -1
```
6543, line 483: use break ?

ennetech: solved

ennetech:
@qwerty287 i'll add the todo, do you prefer it in the base interface or just in the function modified?

6543:
well just make a func that get a func as argument in common ... witht is you dont need to create special interfaces etc ...

just some func as arg and generics:
```go
func pagginate[T type](get func(page) ([]*T, error)) ([]*T, error)
```

?

6543 (approved):

6543:
@qwerty287 did you have something like this in mind: #1236 ?

6543:
@ennetech :point_up: that's a nice example how to use generics and pass func as arguments in go

6543:
Yes if we start supporting pagination ... we do not need that func in most cases anymore

PR #1236, opened 2022-10-06
Introduce and use Pagination helper func

6543:
as per https://github.com/woodpecker-ci/woodpecker/pull/1231#issuecomment-1268485496


6543:
Sure ... this pull just add a helper ...

... PS: test failure is related, i have to look into this

6543:
bug fixed, the slice did contained empty but  initialized items ... ( `make(T, n)` vs `make(T, 0, n)` ) 

@qwerty287

PR #1240, opened 2022-10-07
Replace yarn with pnpm

lukashass:
Should resolve startup issues in gitpod and be a lot faster :rocket:


6543:
https://rushjs.io/pages/maintainer/package_managers/ ...

I'm neither against nor for it

anbraten (approved):

Issue #665, opened 2022-01-06
Add token Auth


6543:
Add option for token generation to auth.

So you can do more ganular access control

anbraten:
link #368

ambroisie:
Noting something which was discussed on chat: this would be useful for writing a NixOS test and/or other kind of integration testing.

gzm0:
We need this and I would like to contribute this. I have sketched out a rough plan below.

I'd appreciate some early feedback from the maintainers of whether the plan looks in line with what you had in mind.

## 1. Support multiple tokens per user

- Tokens are still personal and full scope.
- Should be revokeable individually.

**Open Question**
How to build token revocation? Currently the user hash is reset to revoke the token (this even revokes the user's session tokens).
This would not work for individual token revocation. Should we just build a `token` table and a `token_id` in the JWT that we check for revocation? (we'll need a `token` table anyways so we can display the tokens to the user).

## 2. Support scoped tokens

Restrict the scope of tokens. Probably on 2 axes:
- resource (repositories or organizations)
- action (probably: admin, pull, push for starters, can add more over time for more granularity).

For backwards compatibility, missing scope value in JWT means "unrestricted".

For user tokens, we need to check both the upstream permissions from the forge and the token permissions.

## 3. Add machine accounts

- Machine accounts are never tied to a forge.
- Machine accounts cannot log in.
- Admins can create tokens for machine accounts (probably we should force explicit action scope selection).
- Permission checks are solely based on the token permissions.

qwerty287:
I'm against adding something like machine accounts but what I could imagine is adding something like repo tokens instead, i.e. tokens linked to a single repo, but that requires refactoring the whole auth system.

Options 1 (which can be extended by 2 later) is probably more straightforward to add.

gzm0:
Oh, sorry. Maybe I wasn't clear. The numbers in my original comment were not meant to be options, but steps.

The primary need for us is to authorize the autoscaler. It needs system wide permissions to manage system wide agents.

Currently I'm using my personal token, but in an organizational context, that's not ideal. If my user account gets disabled, autoscaling will fail.

The concept of "machine account" really is meant to serve as a "vessel" for tokens without user association (certain things track owners / creators, it feels easier to just create a "user" that tracks this metadata).

Hope that clarifies.

qwerty287:
Ok, so yes I agree that your points 1 and 2 are good ideas.

[...]

PR #1248, opened 2022-10-09
Create test repo on Gitpod start

lukashass:
Creates a testing repository with a preexisting woodpecker config.

The only steps left to see a running pipeline in the UI are now:

1. Click `Add repository`
2. Enable repository
3. Click `Run pipeline`


.gitpod.yml
```diff
@@ -40,6 +40,19 @@ tasks:
...
+        -H 'accept: application/json' \
+        -H 'Content-Type: application/json' \
+        -H "Authorization: token ${GITEA_TOKEN}" \
+        -d '{ "auto_init": false, "name": "woodpecker-test", "private": true, "template": false, "trust_model": "default" }'
+      cd contrib/woodpecker-test-repo
+      git init
+      git checkout -b main
+      git remote add origin http://woodpecker:${GITEA_TOKEN}@localhost:3000/woodpecker/woodpecker-test.git
+      git add .
+      git commit -m "Initial commit"
```
anbraten, line 54:
```suggestion
      git commit -m ":tada: Initial commit"
```

.gitpod.yml
```diff
@@ -40,6 +40,19 @@ tasks:
...
+      git push -u origin main
```
anbraten, line 55: Would be nice to automatically activate the repo in Woodpecker as well. 😉

lukashass: @anbraten This would be nice indeed. I had look and this seems quite difficult right now since there is no easy way to programmatically authenticate with woodpecker.

anbraten: Fair. #665 and a command like giteas `admin user generate-access-token -u woodpecker --raw` would probably help, but we can ignore it for now.

anbraten (approved):

PR #1261, opened 2022-10-14
Hide icon in plugin detail page if not set

6543:
fix: 
image


anbraten (commented):
I would prefer to use our generic plugin icon as we do it in the list again.

6543:
I tryed that as well but it did not scale :/ ... @anbraten you could have a try if you want ...

lukashass:
I already fixed this in when working on #1247. Just forgot to push it :see_no_evil:

PR #1284, opened 2022-10-18
Set drone environment from pipeline metadata not already set envs

6543:
so we dont have to worry at wicht time what env-var is already set and it's easy to refactor


6543:
-> #1790

PR #1285, opened 2022-10-18
Rename multi pipelines to workflows

anbraten:


6543:
can we start with mentioning: "a piepeline do at least have one workflow. if you have multible one they will be named after the filenames..."  or so

just make clear that it always exist but is only noticable if there are multible one

lafriks (approved):

PR #1305, opened 2022-10-21
Change locale in user settings

### LINKED ISSUE #1109: Allow user to set language

As a user I may want to select a language in Woodpecker which isn't the language my Browser is set to.

### Suggested solution

Add some language selection in the user details screen `/user` (clicking on the user icon).

### Alternative

_No response_

### Additional context

follow up of #258 

### Validations

lukashass:
Closes #1109


web/src/views/User.vue
```diff
@@ -27,48 +29,61 @@
...
-    const cliDownload = 'https://github.com/woodpecker-ci/woodpecker/releases';
+const localeOptions = computed(() =>
+  availableLocales.map((availableLocale) => ({
+    value: availableLocale,
+    text: new Intl.DisplayNames(availableLocale, { type: 'language' }).of(availableLocale) || availableLocale,
+  })),
+);
 
-    return { token, usageWithShell, usageWithCurl, usageWithCli, cliDownload, address };
+const storedLocale = useLocalStorage('woodpecker:locale', locale.value);
```
lafriks, line 77: we should come up with naming for local/session storage names.
Currently we have `last-badge-type`, `log-auto-scroll` (without `woodpecker:` prefix)

web/src/views/User.vue
```diff
@@ -27,48 +29,61 @@
...
-    const token = ref<string | undefined>();
+onMounted(async () => {
+  token.value = await apiClient.getToken();
+});
 
-    onMounted(async () => {
-      token.value = await apiClient.getToken();
-    });
+// eslint-disable-next-line no-restricted-globals
+const address = `${location.protocol}//${location.host}`;
```
lafriks, line 54: address should also have `${location.port}`

anbraten: incase its not 443

lukashass: Description from `lib.dom.d.ts`:

```ts
    /**
     * Returns the Location object's URL's host and port (if different from the default port for the scheme).
     *
     * Can be set, to navigate to the same URL with a changed host and port.
     */
    host: string;
```

web/src/views/User.vue
lukashass: I would prefer a prefix, mostly to avoid conflict when developing on localhost.

web/src/views/User.vue
anbraten: In that case add a comment please.

6543:
please resolve conflict

web/src/views/User.vue
6543: is this resolved ?!?

web/src/views/User.vue
6543: status ?!?

web/src/views/User.vue
6543: to have a single source of truth why not add the save func to web/src/utils/locale.ts ?

6543:
@lukashass please address codeComments 🚀

web/src/views/User.vue
lukashass: I agree the locale configs are a bit all over the place, but I'm not really sure how to improve this at the moment, as it is not a simple save function.

I think we should tackle this after #1345.

web/src/views/User.vue
lukashass: Added a comment.

6543:
@lukashass please dont tuch translation files!!!

if it's an issue it should be configured within weblate!

-> [3c94ba1](https://github.com/woodpecker-ci/woodpecker/pull/1305/commits/3c94ba16316114a51dda20efcbd0052a1d0d15a9)

6543 (changes requested):
as per comment else lgtm

lukashass:
> @lukashass please dont tuch translation files!!!
> 
> if it's an issue it should be configured within weblate!
> 
> -> [3c94ba1](https://github.com/woodpecker-ci/woodpecker/pull/1305/commits/3c94ba16316114a51dda20efcbd0052a1d0d15a9)

It is an issue indeed, since we are passing the locale identifier i.e. the file name to `Intl.DisplayNames`.

https://www.unicode.org/reports/tr35/#Unicode_locale_identifier treats `_` and `-` as equivalent but `_` errors in firefox at least.

6543:
@lukashass well can you reverte this non the less - we will fix it in the weblate config

6543 (approved):

PR #1304, opened 2022-10-21
Make pipeline steps collapsible

lukashass:
Closes #948 

## TODO

- [x] style step button

image

image

image

image

PR #1319, opened 2022-10-22
Don't show workflow name if it's only one

qwerty287:
Was introduced by https://github.com/woodpecker-ci/woodpecker/pull/1304.

If there's only one workflow, don't make it collapsible and don't show the name of it.


lukashass:
`mt-1` of the steps has to be disabled for the first one in this case, as well:

image

6543:
I have no strong preference for/against it ... so lgtm

6543:
@qwerty287 

`src/components/repo/pipeline/PipelineProcList.vue(93,84): error TS2532: Object is possibly 'undefined'.`

6543 (approved):

PR #1321, opened 2022-10-22
Add pull request labels as environment variable

### LINKED ISSUE #1308: Add env var with labels from PR
It would be nice to have an env var like `CI_COMMIT_PULL_REQUEST_LABELS` containing a list of all set labels of a PR as a comma separated string. This would allow a user to filter steps and pipelines with https://woodpecker-ci.org/docs/next/usage/pipeline-syntax#evaluate to only be executed if a specific label is set or is not set. 

For example the quite time consuming task of building docker images for PRs could be skipped by default and only be activated if the corresponding label was set by a maintainer.

Later with the support of label changes triggering pipelines like suggested in https://github.com/woodpecker-ci/woodpecker/issues/286#issuecomment-1116699790 a pipeline could immediately be executed in case a label got added.

lukashass:
Closes #1308 

## TODO

- [x] get labels from
  - [x] github
  - [x] gitea
  - [x] gitlab
- [x] use labels in [evaluate](https://woodpecker-ci.org/docs/next/usage/pipeline-syntax#evaluate)
- [x] update docs


6543:
I'm not sure if we should call it `CI_COMMIT_PULL_REQUEST_LABELS` as there also exist labels for milestones and issues in forges ...

docs/docs/20-usage/20-pipeline-syntax.md
```diff
@@ -462,6 +462,13 @@ when:
   - evaluate: 'not (CI_COMMIT_MESSAGE contains "please ignore me")'
 ```
 
+Run on pull requests with the label `deploy`:
+
+```yaml
+when:
+  - evaluate: 'CI_COMMIT_PULL_REQUEST_LABELS contains "deploy"'
```
lukashass, line 469: As `CI_COMMIT_PULL_REQUEST_LABELS` is currently a string this only checks if the string contains `deploy`.

An exact match on a label would be better. Something clunky like:

`any([CI_COMMIT_PULL_REQUEST_LABELS], {value == "deploy"})`

Any ideas?

server/model/pipeline.go
```diff
@@ -52,6 +52,7 @@ type Pipeline struct {
 	Files               []*File           `json:"files,omitempty"         xorm:"-"`
 	ChangedFiles        []string          `json:"changed_files,omitempty" xorm:"json 'changed_files'"`
 	AdditionalVariables map[string]string `json:"variables,omitempty"     xorm:"json 'additional_variables'"`
+	Labels              []string          `json:"labels,omitempty"        xorm:"-"`
```
anbraten, line 55: I guess we need to save them otherwise we can't restart a pipeline.

pipeline/frontend/metadata.go
```diff
@@ -55,19 +55,20 @@ type (
...
+		Timeout  int64    `json:"timeout,omitempty"`
+		Status   string   `json:"status,omitempty"`
+		Event    string   `json:"event,omitempty"`
+		Link     string   `json:"link,omitempty"`
+		Target   string   `json:"target,omitempty"`
+		Trusted  bool     `json:"trusted,omitempty"`
+		Commit   Commit   `json:"commit,omitempty"`
+		Parent   int64    `json:"parent,omitempty"`
+		Cron     string   `json:"cron,omitempty"`
+		Labels   []string `json:"labels,omitempty"`
```
anbraten, line 71: On this level they should either be called `CI_LABELS` instead of `CI_COMMIT_PULL_REQUEST_LABELS` or we need to move that property into the Commit part.

docs/docs/20-usage/20-pipeline-syntax.md
anbraten: expr has proper support for more complex data structures as our current string => string map. We could for example allow the user to directly access the metadata object. Problem I noticed when trying it, is that we would need to document it somehow.

If we stick to the env vars I would prefer something simple like the current approach and not introduce json strings which then need to be parsed again. Using `,` inside a label seems to be pretty uncommon to me.

pipeline/frontend/metadata.go
```diff
@@ -147,19 +148,20 @@ func (m *Metadata) Environ() map[string]string {
...
+		"CI_COMMIT_SOURCE_BRANCH":       sourceBranch,
+		"CI_COMMIT_TARGET_BRANCH":       targetBranch,
+		"CI_COMMIT_LINK":                m.Curr.Link,
+		"CI_COMMIT_MESSAGE":             m.Curr.Commit.Message,
+		"CI_COMMIT_AUTHOR":              m.Curr.Commit.Author.Name,
+		"CI_COMMIT_AUTHOR_EMAIL":        m.Curr.Commit.Author.Email,
+		"CI_COMMIT_AUTHOR_AVATAR":       m.Curr.Commit.Author.Avatar,
+		"CI_COMMIT_TAG":                 "", // will be set if event is tag
+		"CI_COMMIT_PULL_REQUEST":        "", // will be set if event is pr
+		"CI_COMMIT_PULL_REQUEST_LABELS": "",
```
anbraten, line 164:
```suggestion
		"CI_COMMIT_PULL_REQUEST_LABELS": "", // will be set if event is pr
```

docs/docs/20-usage/20-pipeline-syntax.md
```diff
@@ -462,6 +462,13 @@ when:
   - evaluate: 'not (CI_COMMIT_MESSAGE contains "please ignore me")'
 ```
 
+Run on pull requests with the label `deploy`:
```
anbraten, line 465: We could be more specific that this is the label of the PR here.

anbraten (commented):
Some nits. Nice work in general!

pipeline/frontend/metadata.go
6543: I'm for `CI_LABELS` as we already store them in all structs internaly this way ...

server/model/pipeline.go
6543: xorm:"json 'labels'"

server/remote/gitea/helper.go
```diff
@@ -229,3 +230,11 @@ func matchingHooks(hooks []*gitea.Hook, rawurl string) *gitea.Hook {
 	}
 	return nil
 }
+
+func convertLabels(from []*gitea.Label) []string {
+	var labels []string
```
6543, line 235:
```suggestion
	labels := make([]string, len(from))
```
we already know how much mem we have to allocate

server/remote/gitea/helper.go
```diff
@@ -229,3 +230,11 @@ func matchingHooks(hooks []*gitea.Hook, rawurl string) *gitea.Hook {
...
+	for _, label := range from {
+		labels = append(labels, label.Name)
+	}
```
6543, line 238:
```suggestion
	for i, label := range from {
		labels[i] = label.Name
	}
```

server/remote/github/convert.go
```diff
@@ -164,3 +164,13 @@ func convertRepoHook(eventRepo *github.PushEventRepository) *model.Repo {
 	}
 	return repo
 }
+
+// covertLabels is a helper function used to convert a GitHub label list to
+// the common Woodpecker label structure.
+func convertLabels(from []*github.Label) []string {
+	var labels []string
```
6543, line 171: similar ...

docs/docs/20-usage/20-pipeline-syntax.md
anbraten: By having a look at the docs. This seems to be already pretty clear.

pipeline/frontend/metadata.go
anbraten: @6543 The `Labels` property was just added in this PR. What other labels do you plan to store instead of pull-request labels in there? The only other thing I can think of atm is issue labels, but we currently have no pipeline event for issues.

anbraten (approved):

docs/docs/20-usage/20-pipeline-syntax.md
anbraten: Would you both be fine with keeping this super simple for now by sticking to the string divided by `,` approach?

lukashass: @anbraten What about surrounding the string with `,` as well? i.e. `,ready for review,deploy,`

This would allow `contains ",deploy,"` to work every time.

pipeline/frontend/metadata.go
6543: @anbraten **issues** also can have labels but for now we dont support any event in that direction ...

docs/docs/20-usage/20-pipeline-syntax.md
6543: I'm fine with as is ... and the eval will have more info already and so should handle strange cases if you need to

PR #1325, opened 2022-10-22
Fix redirects

qwerty287:
Not a very elegant solution, but it should work.

### DIFF
--- a/server/model/feed.go
+++ b/server/model/feed.go
@@ -19,24 +19,24 @@ package model
 //
 // swagger:model feed
 type Feed struct {
-	Owner    string `json:"owner"                   xorm:"repo_owner"`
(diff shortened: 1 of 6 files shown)


web/src/router.ts
```diff
@@ -112,6 +112,20 @@ const routes: RouteRecordRaw[] = [
       {
         path: 'build/:pipelineId',
```
anbraten, line 113: Did you tried `build/:rest(.*)`?

anbraten: Couldn't we use a function with doing `to.path.replace(/^build\//, 'pipeline/');`?

6543 (approved):
Lgtm, if there is  better way we can still switch to that later

anbraten (approved):

PR #1329, opened 2022-10-23
Allow to change directory for steps

qwerty287:
Add `directory` YAML key that changes the workdir. Can replace a `cd` before your commands start or make it possible to run plugins in a subdirectory.

### DIFF
--- a/pipeline/schema/schema.json
+++ b/pipeline/schema/schema.json
@@ -215,6 +215,9 @@
         "environment": {
           "$ref": "#/definitions/step_environment"
         },
+        "directory": {
+          "$ref": "#/definitions/step_directory"
+        },
         "secrets": {
           "$ref": "#/definitions/step_secrets"
         },


pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -68,7 +68,7 @@ func (c *Compiler) createProcess(name string, container *yaml.Container, section
 	}
 
 	if !detached || len(container.Commands) != 0 {
-		workingdir = path.Join(c.base, c.path)
+		workingdir = path.Join(c.base, c.path, container.Directory)
```
6543, line 71: can we move this out to it's own function?

pipeline/frontend/yaml/container.go
```diff
@@ -37,6 +37,7 @@ type (
 		Tmpfs         []string               `yaml:"tmpfs,omitempty"`
 		DNS           types.Stringorslice    `yaml:"dns,omitempty"`
 		DNSSearch     types.Stringorslice    `yaml:"dns_search,omitempty"`
+		Directory     string                 `yaml:"directory,omitempty"`
```
6543, line 40: we need docs

pipeline/frontend/yaml/compiler/convert.go
6543: just in the same code-file for now but that will make it easyer to maintain the code

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -184,3 +184,7 @@ func (c *Compiler) createProcess(name string, container *yaml.Container, section
 		IpcMode:      ipcMode,
 	}
 }
+
+func (c *Compiler) stepWorkdir(container *yaml.Container) string {
+	return path.Join(c.base, c.path, container.Directory)
```
lafriks, line 189: Shouldn't it use filepath.Join?

Also currently it's not possible to provide absolute container directory

6543: But users will use them!

6543: and path trafesal is not an issue as you alwas could do it via commands anyway ...
so check if it is absolute the use this else use filepath.join i would say

lafriks: > An absolute path wasn't the idea behind it.

Yes I know this was just a suggestion that it would be nice to have otherwise it could be confusing for users

6543:
Uh and dont forget to update the json schema ...

pipeline/schema/schema.json
```diff
@@ -467,6 +467,10 @@
       "items": { "type": "string" },
       "minLength": 1
     },
+    "step_directory": {
+      "description": "Read more: https://woodpecker-ci.org/docs/usage/pipeline-syntax#directory",
+      "type": "string"
```
6543, line 472: You have to reference it into step (start is in line 200)

6543 (approved):

PR #1330, opened 2022-10-23
Update pipeline steps and logs background

dvjn:
Why?
- Better contrast
- Looks modern

**Original** | **New**
:--------:|:-------:
image | image


lafriks (approved):

6543 (changes requested):
I like the old one, the new one do have to less contrast so you dont see the different elements difided anymore

6543:
yes you have more contrast for text but you loose the other one :/

6543:
image

if we go light with the elements, we might go dark with the background ...

6543:
197409473-4b97a2b8-a220-4019-ab33-abd2407d590a
?

-> the greay is just to demonstrate it ... but there should be better colors

dvjn:
I've increased the shadow to match all the other cards in the UI.

If contrast between the card and background is still lacking, it will call for a bigger change on all the views and cards, and should be a separate issue/pr.

dvjn:
New Screenshot:
image

6543 (approved):
thanks - good work

6543:
> If contrast between the card and background is still lacking, it will call for a bigger change on all the views and cards, and should be a separate issue/pr.

yes we should do that

PR #1338, opened 2022-10-25
Allow to run cron manually

### LINKED ISSUE #1154: Add Button to run cron manually now
build upon #934
- exec as if nextSchedule would been reached
- not alter nextSchedule for cron

qwerty287:
Closes #1154


6543:
if repo is gated you still have to approve it, just a nit but it would nice that in this case it does run directly

6543 (approved):

6543:
> if repo is gated you still have to approve it, just a nit but it would nice that in this case it does run directly

never mind also for "manual pipeline" so it is it's own issue

Issue #1138, opened 2022-08-29
placeholder


6543:

lafriks:
There are also some forges missing org membership/org admin right check

6543:
uh right the docs there also need adjustments

mzampetakis:
@6543 Is there any specific requirement for getting pr file changes for bitbucket or is this supported through basic forge's functionality (.e.g Dir(), BranchHead() etc)?

6543:
It's just basic forge functionality

E.g.:

https://github.com/woodpecker-ci/woodpecker/blob/c37fab3dd74aae269f16258eec763a7e0ab699e7/server/forge/gitea/gitea.go#L497

mzampetakis:
Hmmm, it seems that bitbucket is still missing this functionality from https://github.com/woodpecker-ci/woodpecker/blob/main/server/forge/bitbucket/convert.go#L151

henkka:
As part of our organization migrating to Woodpecker CI, I will work on the Bitbucket Datacenter support for `when.path` filter

PR #1342, opened 2022-10-26
Support changed files for Gitea PRs

qwerty287:
#1138

TODOs:
 - [x] add tests (just use existing and extend) for this testcase
 - [x] ignore error if gitea version is to low
 - [x] change docs acordingly


server/remote/gitea/gitea.go
```diff
@@ -540,3 +557,39 @@ func getStatus(status model.StatusValue) gitea.StatusState {
...
+	user, err := _store.GetUser(repo.UserID)
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := c.newClientToken(ctx, user.Token)
+	if err != nil {
+		return nil, err
+	}
+	return common.Paginate(func(page int) ([]string, error) {
```
6543, line 582: check if the error is "to low gitea version" and return no errror an no changed files -> as we have it now

6543:
TODOs:
 - [ ] add tests (just use existing and extend) for this testcase
 - [ ] ignore error if gitea version is to low
 - [ ] change docs acordingly

server/remote/gitea/gitea.go
6543: asap we have init the client we can do ... but in terms of api cals there wont be any difference

6543: (the sdk do check on it's own before calling the endpoint)

6543: It's done on client init if you dont tell to do otherwhise

docs/docs/20-usage/20-pipeline-syntax.md
```diff
@@ -412,8 +412,7 @@ when:
 
 :::info
 Path conditions are applied only to **push** and **pull_request** events.
-It is currently **only available** for GitHub, GitLab.
-Gitea only supports **push** at the moment ([go-gitea/gitea#18228](https://github.com/go-gitea/gitea/pull/18228)).
+It is currently **only available** for GitHub, GitLab and Gitea.
```
6543, line 415: could you clarify that 'sed s/Gitea/Gitea[1]/g' ...

[1] all gitea version >= 1.18.0 ...

server/remote/gitea/gitea.go
```diff
@@ -540,3 +557,45 @@ func getStatus(status model.StatusValue) gitea.StatusState {
...
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := c.newClientToken(ctx, user.Token)
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.Contains(client.CheckServerVersionConstraint("1.18.0").Error(), "does not satisfy version constraint") {
```
6543, line 583:
```suggestion
	if client.CheckServerVersionConstraint("1.18.0") != nil {
```

server/remote/gitea/gitea.go
```diff
@@ -540,3 +557,45 @@ func getStatus(status model.StatusValue) gitea.StatusState {
...
+		// version too low
+		return []string{}, nil
```
6543, line 585:
```suggestion
		log.Debug ....
		return []string{}, nil
```

6543:
thanks :heart: now only tests are missing

I think you can just look into github remote as there already should be some in that you can copy/use as template

server/remote/gitea/gitea.go
```diff
@@ -33,6 +33,7 @@ import (
 
 	"code.gitea.io/sdk/gitea"
 	"github.com/rs/zerolog/log"
+	"github.com/woodpecker-ci/woodpecker/server/store"
```
lafriks, line 36: this should be moved to import group below

docs/docs/20-usage/20-pipeline-syntax.md
```diff
@@ -412,8 +412,7 @@ when:
 
 :::info
 Path conditions are applied only to **push** and **pull_request** events.
-It is currently **only available** for GitHub, GitLab.
-Gitea only supports **push** at the moment ([go-gitea/gitea#18228](https://github.com/go-gitea/gitea/pull/18228)).
+It is currently **only available** for GitHub, GitLab and Gitea (versions newer than 1.18.0)
```
lafriks, line 415:
```suggestion
It is currently **only available** for GitHub, GitLab and Gitea (version 1.18.0 and newer)
```

6543 (approved):

6543:
added tests by using the mockup store

lafriks (approved):

6543:
We cant as we use gin framework

We would have to patch that upstream

PR #1350, opened 2022-10-27
Always use rounded quadrat user avatars

qwerty287:
To stay consistent with the header.


6543:
well I'm personally for quadrat avatars with rounded corners ... but I agree, it should be consistent

6543:
image

so quadrat with rounded corners winns

6543 (approved):

PR #1353, opened 2022-10-27
Proof of concept for a file caching system

qwerty287:
This is some try to implement a file caching system and make it possible to cache files between pipelines, not only between steps.

### How does this work

You can define a new yaml object `caches` that allows to define a cache name and a mount point in docker. It will be added as a docker volume then, with a "safe" location on the host system to prevent access outside the cache.

First: What do you think about this solution in general?

If it's fine, we could look somewhat more into how this might work or how we can improve it. Also, I'm not sure if this solution has any drawbacks (mainly regarding security).

Please also note that I did not really test this (only few debugging Printlns, but "real" caching requires Docker I don't want to install on my machine because it takes too much space :D).

### TODOs / ideas

- [ ] use different directory for PRs
- [ ] allow to disable cache for specific events
- [ ] show caches in web UI and allow clearing and downloading of them / their contents
- [ ] quotas / size and lifetime limits (global and per-repo)
- [ ] auto-clean if unused
- [ ] allow to define pre- / post-run-"scripts" (e.g. for S3 support that downloads the cache content to a temporary directory, mounts it as volume and uploads changes afterwards)

### DIFF
--- a/docs/docs/30-administration/10-server-config.md
+++ b/docs/docs/30-administration/10-server-config.md
@@ -294,6 +294,7 @@ Read the value for `WOODPECKER_DATABASE_DATASOURCE` from the specified filepath
 > Default: empty
 
 Token to secure the Prometheus metrics endpoint.
+Must be set to enable the endpoint.
 
 ### `WOODPECKER_PROMETHEUS_AUTH_TOKEN_FILE`
 > Default: empty
--- a/server/api/metrics/prometheus.go
+++ b/server/api/metrics/prometheus.go
@@ -17,6 +17,7 @@ package metrics
 import (
 	"errors"
 	"fmt"
+	"net/http"
 
 	"github.com/gin-gonic/gin"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -35,7 +36,7 @@ func PromHandler() gin.HandlerFunc {
 		token := server.Config.Prometheus.AuthToken
 
 		if token == "" {
-			handler.ServeHTTP(c.Writer, c.Request)
+			c.AbortWithStatus(http.StatusNotFound)
 			return
 		}
 
--- a/web/components.d.ts
+++ b/web/components.d.ts
@@ -15,6 +15,7 @@ declare module '@vue/runtime-core' {
     Checkbox: typeof import('./src/components/form/Checkbox.vue')['default']
     CheckboxesField: typeof import('./src/components/form/CheckboxesField.vue')['default']
     CronTab: typeof import('./src/components/repo/settings/CronTab.vue')['default']
+    DeployPipelinePopup: typeof import('./src/components/layout/popups/DeployPipelinePopup.vue')['default']
     DocsLink: typeof import('./src/components/atomic/DocsLink.vue')['default']
     FluidContainer: typeof import('./src/components/layout/FluidContainer.vue')['default']
     GeneralTab: typeof import('./src/components/repo/settings/GeneralTab.vue')['default']
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -41,6 +41,18 @@
             "value": "Variable value"
           }
         },
+        "deploy_pipeline": {
+          "title": "Trigger deployment event for current pipeline #{pipelineId}",
+          "enter_target": "Target deployment environment",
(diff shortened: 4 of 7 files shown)


6543:
1. how does that work with Kubernetes
2. how do we make sure there are no name colisions
3. how do we make sure you dont monunt from other repos
4. race conditions?

6543:
caching I realy like to do not in the upcomming release - PoC are welcome just not merged jet 🗡️

lafriks:
Imho cache implementation should be abstract so that also something like s3 etc could be used

lafriks:
Plugin could download data to dir/volume in pre-step and upload post step, of course it would involve some magic but could be doable

lafriks:
Plugin could download data to dir/volume in pre-step and upload post step, of course it would involve some magic but could be doable. So this could even possibly work cross pipelines and with different runners (docker/ssh etc)

smainz:
Can you please elaborate on what you are trying to achieve and how this is better than just using a plugin for caching?
See: https://plugins.drone.io/plugins/s3-cache or https://plugins.drone.io/plugins/cache

If we could have volumes shared between steps and pipelines which are not backed by a dedicated host volume (e.g. a anon volume per pipeline execution) this can be built with one of the plugins.

I came here, because I am looking for a solution to have volumes per pipeline.

smainz:
S3 is only an example of where a plugin can pull / push data from/ to, you can make different plugins using sftp, scp, ... or just use commands in an image.

Pipeline execution will take place on multiple machines, so the storage must be somewhere else or you will end up with different caches on each machine / agent. Probably there needs be some kind of conflict resolution for concurrent pipeline executions using the same cache.

My idea is to separate caching from the provisioning of the volumes, so each pipeline can define the caching behavior by itself and only use the mechanisms provided by woodpecker. The responsibility of woodpecker were to provision volumes, the responsibility of the pipeline script would be to get / push data to the volume.

Something like that:
```
pipeline:

  # Mount the cache-volume and pull the data from a remote location.
  - name: Pull from cache
    image: cache-pull-sftp
    settings:
      local_dir: /opt/my-.cache
      remote_dir: ...
      user:
        from_secret: ssh_user
      key:
        from_secret: ssh_key
    volumes:
      - cache-volume:/opt/my-cache

  # Build uses the cache volume and does whatever it needs to do
  - name: Build
    image: ...
    volumes:
      - cache-volume:/wherever/i/need/my/cache/to/be/for/this/step
   commands:
    - ...

  # Push some artefacts to the cache storage and do some conflict resolution, Handle race conditions.
  # Most likely this will be a set of commands instead of a plugin so you can implement the conflict resolution
  - name: Push to cache resolution
  - name: Push to cache
    image: cache-push-sftp
    settings:
      local_dir: /opt/my-.cache
      remote_dir: ...
      user:
        from_Secret: ssh_user
      key:
        from_secret: ssh_key  
    volumes:
      - cache-volume:/opt/my-.cache
    when:
      - status:
        - success

# declare ephemeral, anonymous volumes which are exclusive for this pipeline execution
# ephemeral volumes should be created before a pipeline is run, mounted to the steps where it is defined
# and be deleted after the pipeline is executed.
volumes:
  - name: cache-volume
```

If volumes were shared between multiple pipelines in a workflow (hope i got the terminology right) one could share the cache between these too using the depends_on syntax where one pipeline would only restore the data from remote storage.

If there was variable resolution in the names of the volumes, you could build different caches for different "things" in a matrix, but this is an extra.

```

  - name: Pull from cache
    image: cache-pull-sftp
    settings:
      ...
    volumes:
      - cache-volume_$ARCH:/opt/my-.cache

matrix:
  ARCH:
    - amd64
    - arm64
  
volumes:
  - chach_volume_$ARCH
```

Does this make sense?

smainz:
I think, managing a cache of artifacts (think node_modules, .m2, ...) is nothing one can do in a generic way. Some knowledge of the build process is mandatory for this.
So it should be the responsibility of the user, to define what / how to cache something during his build. In the first place it is him who defines the pipeline.

Do we understand the same thing when talking about a cache? I want to cache libraries downloaded from some repository, but I do not want to cache forever but have the option to discard the cache.

ATM (using drone) I just delete the content of the storage and am done.

PR #1356, opened 2022-10-27
#1310 Migrate sql querys to xorm builder

### LINKED ISSUE #1310: migrate more sql querys to xorm
migrate code like e.g. https://github.com/woodpecker-ci/woodpecker/pull/1296/files#diff-a3003b46215d34362e75d779f7e5b3fe234547af38afc3d21b7402f3a256636bL51-L61

to use the xorm builder or xorm instead

bitethecode:
close  #1310


server/model/feed.go
```diff
@@ -19,24 +19,24 @@ package model
...
+	Commit   string `json:"commit,omitempty"        xorm:"feed_pipeline_commit"`
+	Branch   string `json:"branch,omitempty"        xorm:"feed_pipeline_branch"`
+	Ref      string `json:"ref,omitempty"           xorm:"feed_pipeline_ref"`
+	Refspec  string `json:"refspec,omitempty"       xorm:"feed_pipeline_refspec"`
+	Remote   string `json:"remote,omitempty"        xorm:"feed_pipeline_remote"`
+	Title    string `json:"title,omitempty"         xorm:"feed_pipeline_title"`
+	Message  string `json:"message,omitempty"       xorm:"feed_pipeline_message"`
+	Author   string `json:"author,omitempty"        xorm:"feed_pipeline_author"`
+	Avatar   string `json:"author_avatar,omitempty" xorm:"feed_pipeline_avatar"`
+	Email    string `json:"author_email,omitempty"  xorm:"feed_pipeline_email"`
```
anbraten, line 41: What a crazy model. I guess we should clean it up at some point as well.

6543: @anbraten well it act's like a `view`

6543: and I prefixed it, so if we refactor stuff it wont break

anbraten: Sure. I was mainly talking about all the different properties as I guess most of them are not even part of our UI atm.

6543: well in an dedicated refactor we can remove all not needed once and expose those who do make sense to the ui

Issue #745, opened 2022-02-04
RFC: Establish some terminology


anbraten:
We currently mix some terms in the documentation and in the code. The goal of this issue is to define some terminology which is understandable by new users as well as by experts. Please feel free to comment below to suggest terms or to share your opinions. The current draft will be at the end of this description:

---
# pipeline
## 1. pipeline
Triggered by events like webhooks, crons or manual events

formerly known as build

**PRs targetting renaming** (renaming done with them)
- #1224
- #1296

## 2. workflow
formerly known as multi-pipeline pipeline

### matrix
A pipeline which is executed with multiple options (a matrix) creates multiple `workflows` (see 2.)

partly known as axes / axis

### workspace

Through the time of a workflow a workspace is shared between all steps. This folder normally contains the repository data and all data that was generated by previous steps of the same workflow.

## 3. step
A single job with a series of commands executed in a sequence.

executed steps are sometimes called procs, jobs or tasks in code

**PRs targetting renaming** (renaming done with them)
- #1331 

# components

## agent
Runs pipelines and sends results back to server

## server
Head of woodpecker deployment which serves the ui, talks to remotes and sends pipeline execution tasks to agents

## cli
A cli tool which can interact with a server and can run additional commands like linting, running pipelines without a server (locally), ...

# misc
## container

formerly known as docker container

## container image
formerly known as docker image

## forge
formerly known as SCM, VCS or remote

See https://en.wikipedia.org/wiki/Forge_(software)

**PRs targetting renaming** (renaming done with them)
- #1357

# Extendable parts

## plugin
A step of a pipeline can use a plugin to run an automated series of commands which can be adjusted by plugin settings

## Extension / Addon
A remote service (normally http) which can be configured to replace some internal service. #783

anbraten:
So sth like pipeline => ... => (multiple) matrix => step ?

6543:
well technicall matrix wil generate ... | should we still differentiate between them?

anbraten:
What do you think about naming the second level `stage` like drone, gitlab, jenkins?

![image](https://user-images.githubusercontent.com/6918444/193096878-575bf4c5-71c0-4213-ac0c-0d9862f25e4f.png)

![image](https://user-images.githubusercontent.com/6918444/193097506-fd04e5de-c2cf-4ed4-960f-b5b29b85299d.png)

So it would be a pipeline containing multiple stages like `binaries` containing steps like `clone`?

[...]

PR #1357, opened 2022-10-27
Rename `remote` to `forge`

qwerty287:
As of #745


cli/exec/flags.go
```diff
@@ -123,8 +123,8 @@ var flags = []cli.Flag{
 		Name:    "repo-link",
 	},
 	&cli.StringFlag{
-		EnvVars: []string{"CI_REPO_REMOTE"},
-		Name:    "repo-remote-url",
+		EnvVars: []string{"CI_REPO_CLONE_URL"},
```
6543, line 126:
```suggestion
		EnvVars: []string{"CI_REPO_CLONE_URL", "CI_REPO_REMOTE"},
```

6543:
just a question: as "remote" never did pop up any questions, do we need to rename all "remote" stuff as forge?

I mean we could have support for non-forge code hosts in future and so the name would be wrong ...

In my opinion we just have to address the name "forge" in the docs ... if the remote is a forge (github, gitlab, gitea, ...)
but can just all it remote in code

anbraten:
As @qwerty287 said adding non-forge code systems is very unlikely a thing being added in the near future considering the amount of work, but I guess  even if someone is doing that it wouldn't still be to problematic to have that system called forge as well. In general I am mainly interested in consistency whatever the term would be.

cli/exec/flags.go
```diff
@@ -123,8 +123,8 @@ var flags = []cli.Flag{
 		Name:    "repo-link",
 	},
 	&cli.StringFlag{
-		EnvVars: []string{"CI_REPO_REMOTE"},
-		Name:    "repo-remote-url",
+		EnvVars: []string{"CI_REPO_CLONE_URL", "CI_REPO_REMOTE"},
+		Name:    "repo-clone-url",
```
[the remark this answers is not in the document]
anbraten: https://github.com/woodpecker-ci/woodpecker/issues/276

anbraten (approved):

6543:
well no compains from my side

6543:
Lint error

server/api/cron.go
```diff
@@ -150,8 +150,9 @@ func PatchCron(c *gin.Context) {
 		return
 	}
 	if in.Branch != "" {
-		// check if branch exists on remote
-		_, err := remote.BranchHead(c, user, repo, in.Branch)
+		// check if branch exists on forge
+
```
6543, line 154:
```suggestion
```

6543 (approved):

PR #1362, opened 2022-10-28
Lazy load locales

### LINKED ISSUE #1345: Lazy load locales

Currently all locales are loaded immediately. This unnecessarily bloats the bundle size.

### Suggested solution

Lazy load locales: https://vue-i18n.intlify.dev/guide/advanced/lazy.html

### Alternative

_No response_

### Additional context

_No response_

### Validations

lukashass:
Closes #1345 

`index.js` size reduction: `538.71 KiB` -> `382.74 KiB`

Also sets correct html `lang` attribute.

## TODO

- [x] check for alternatives for vite plugin to load supported locales https://github.com/intlify/bundle-tools/issues/205


web/src/views/RepoAdd.vue
```diff
@@ -78,7 +78,7 @@ export default defineComponent({
     const { doSubmit: activateRepo, isLoading: isActivatingRepo } = useAsyncAction(async (repo: Repo) => {
       repoToActivate.value = repo;
       await apiClient.activateRepo(repo.owner, repo.name);
-      notifications.notify({ title: i18n.t('repo.enabled.success'), type: 'success' });
+      notifications.notify({ title: i18n.t('repo.enable.success'), type: 'success' });
```
6543, line 81: nice catch!

6543:
to test if key is right on linux via cli:
```sh
cat web/src/assets/locales/en.json | jq .repo.enable.success
```
if you get your value the key is right ;)

lukashass:
I couldn't get lazy loading of dayjs and TimeAgo locales to work, so I would suggest to try that later.

lafriks (approved):

6543:
Thanks :)

PR #1368, opened 2022-10-30
Use pipeline wrapper and improve scaffold UI

dvjn:
- Add support for full width views in scaffold component
- Put pipeline's router-views inside scaffold, which makes spacing around each view consistent (and each view doesn't have to handle its margins relative to the page)
- Add support for full width in fluid container, to make spacing of full width views (pipeline logs) consistent with all other pages


web/src/components/layout/scaffold/Scaffold.vue
```diff
@@ -42,6 +46,8 @@ const props = withDefaults(defineProps<Props>(), {
   // eslint-disable-next-line vue/no-boolean-default
   enableTabs: false,
   activeTab: '',
+  // eslint-disable-next-line vue/no-boolean-default
+  noFluidWrapper: false,
```
anbraten, line 50: I think you can simply leave this out. The default is always false.

web/src/components/layout/scaffold/Tab.vue
```diff
@@ -24,7 +24,11 @@ onMounted(() => {
     id: props.id || props.title.toLocaleLowerCase().replace(' ', '-') || tabs.value.length.toString(),
     title: props.title,
   };
-  tabs.value.push(tab.value);
+
+  // don't add tab if tab id already present
```
anbraten, line 28:
```suggestion
  // don't add tab if tab id is already present
```

web/src/components/layout/scaffold/Scaffold.vue
anbraten, line 50:
```suggestion
  fluid: true,
```

I would just call it `fluid`.

Anyways if you have a boolean prop it will automatically be set to true.

dvjn: Typescript is not able to infer that :(
This shows red squiggly line in vscode everywhere where it is expecting a boolean. (because it infers `boolean | undefined`)

dvjn: How about `fluidContent`?

lukashass: @dvjn `undefined` comes from the `?` marking `noFluidWrapper` as optional.

https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters

dvjn: Yes, but it needs to be optional, otherwise I'll have to pass it everywhere I create the component (or it'll give lint warnings there).

dvjn: What I originally meant to say is `defineProps` is not properly typed. In JS it converts all the `boolean | undefined` to `boolean` (by setting it to false) but doesn't reflect that in the returned prop's type unless you pass a default using `withDefaults`.

dvjn: @anbraten should I mark this resolved?

6543: @lukashass it it's resolved for you just resolve it :)

anbraten (approved):

PR #1373, opened 2022-10-30
Add RBAC to helm chart

### LINKED ISSUE #1367: Helm chart add RBAC / config section for Kubernetes backend

While testing #1032 I stumbled upon the Service Account not having the required permissions to use the Kubernetes backend.

This problem applies to any next, helm chart based setup with Kubernetes backend on a Kubernetes Cluster with active RBAC (which from a security perspective should be pretty much everywhere, all major enterprise K8s solutions rke2, okd, OpenShift, cloud providers, etc. have it active by default, many, more developer focused k8s-variants micro-k8s, minikube, etc., have options to enable it).

### Suggested solution

Add a role and roleBinding in the chart to allow agents with the Kubernetes backend to spawn the needed pods, collect logs, etc.

It might be an idea to take the Kubernetes backend config into its own values section, where all the Kubernetes backend settings can be added/configured, as well as deciding whether to deploy the service account with RBAC.

### Alternative

Write an additional section into the Kubernetes / Helm deployment docs outlining that a user should create the RBAC if RBAC is in use.
Also document which API-groups resources and verbs are required (which in effect is almost as much work as directly integrating it into the chart).

### Additional context

Was unsure whether this is considered a bug or feature request...

### Validations

gapodo:
This PR fixes #1367 with the minimum needed (plus the basics of annotations and labels, since some clusters need those for extra verifications, OPA, Kyverno, etc.).

The added role is the minimum access I could get away with (tested each verb and resource individually), since the Kubernetes go library seems to use list and get even when not strictly necessary.

I've defaulted to inactive, setting the serviceAccount.rbac.create=true will create the Role and roleBinding.

The changes only affect the woodpecker-agent chart, as the woodpecker-server chart currently does nothing directly

# Tests

- [x] non default namespace (roleBindung uses namespace in a not automatically rewritten position)
- [x] rbac.create enabled and disabled (nothing changes for disabled, since the templates use a guard)
- [x] custom serviceAccount name
- [x] both roleBinding and role with no annotations, no lables, single a&l, multiple each
- [x] helm deploy to Kubernetes, with all settings mentioned above

# Documentation

Added in the comments of the values.yaml. Taking it into the docs might be helpful, but the Kubernetes section in the next docs is fairly empty, possibly open a new issue and solve when the chart for next is mostly done.


6543 (approved):
as far as I can tell looks good

charts/woodpecker-agent/values.yaml
```diff
@@ -31,13 +31,25 @@ nameOverride: ""
...
   create: true
   # Annotations to add to the service account
   annotations: {}
   # The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
   name: ""
+  rbac:
+    # If your cluster has RBAC enabled and you're using the Kubernetes agent-
+    # backend you'll need this. (this is true for almost all production clusters)
+    create: false
```
anbraten, line 44: What about setting this to `true` directly?

anbraten (approved):

charts/woodpecker-agent/values.yaml
gapodo: We could, but if I remember correctly deploying RBAC on a cluster without it enabled causes a failure, since the API-objects for role and roleBinding are non existant (could be wrong here haven't handled a RBAC less cluster in years)

anbraten: To be fair I've never worked with a cluster without, so I would be fine to risk that someone reports an issue / failure in case its not enabled.

gapodo: So... I've just proven my self wrong... pulled up a quick setup without RBAC enabled, and I could apply the role and roleBinding (without effect, but the API objects are there).

I've also just realized that I've missed a guard condition to not deploy the roleBinding and role when not creating a service account, although if we are changing the defaults, I would also go for always deploying the serviceAccount (this would give us a known set of allowed API-Requests).

gapodo: I've updated the guard condition and updated serviceAccount.rbac.create to true (also added a comment about disabling it and serviceAccount.create=true being neccessary for it to apply)

6543:
thanks :rocket:

PR #1382, opened 2022-10-31
Added array to tag for pipeline.step.when to schema

### LINKED ISSUE #1381: Conditional pipeline.step.when.tag not taking array as documented in 0.15
### Component

cli

### Describe the bug

Same as in #1379, found while fixing that part of the schema.json.

Tag in the schema currently only supports a string while the pipeline/frontend/yaml/constraints/constraints.go also accepts an array of strings.

### System Info

```shell
v0.15.4 (problem is existent on the release/v0.15 branch, so all v0.15s at least, fixed by the large rewrite applied to main)
```

### Additional context

I'll head in and fix this as well...

### Validations

gapodo:
This fixes #1381. Similar to the branch patch the backend already supports it and only the linter messes up, which should be closed by this.

Might consider refractoring the schema to use a seperate definition for the List type as it gets used in multiple places in the schema (might find more occurances of this problem)

### DIFF
--- a/pipeline/schema/schema.json
+++ b/pipeline/schema/schema.json
@@ -38,53 +38,14 @@
         "git": {
           "type": "object",
           "properties": {
-            "image": {
-              "type": "string"
-            }
+            "image": { "type": "string" }
           }
         }
       }
     },
     "branches": {
       "description": "Only include commits based on their target branch. Read more: https://woodpecker-ci.org/docs/usage/pipeline-syntax#branches",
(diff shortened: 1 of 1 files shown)


6543:
@gapodo the diff is gone so I guess it was within #1384 already ...

gapodo:
@6543 jep, this was one of the changes leading up to me redoing the schema... was unsure whether you would accept that "massive" change so left it open...

PR #1383, opened 2022-10-31
Show workflow state in ui and collapse completed workflows

### LINKED ISSUE #1371: Show workflow state in UI
image

We should show the state of a workflow similar to the state of a single step in front of the name.

In addition finished workflows (successful or failed) could be collapsed by default.

dvjn:
Fixes #1371

image


dvjn:
I need a help with one thing here, I am calculating `initiallyCollapsed` using the first state of the prop `pipeline`.
Sometimes, the initial pipeline has steps array as undefined on initial render, so when the steps are eventually populated, they are all expended (as they are not in `initiallyCollapsed`).

@anbraten or anyone else, any idea how one would tackle this?

lafriks:
Maybe add prop to pipeline with calculated value when steps are loaded first time?

anbraten:
@dvjn I added a `v-if` which should fix it. Could you pleas test it.

anbraten (approved):

dvjn:
> @dvjn I added a `v-if` which should fix it. Could you pleas test it.

Sure man, thanks :)

dvjn:
This broke UI for single workflow pipeline. Let me investigate more and fix it.

dvjn:
Reference:
image

dvjn:
@anbraten

Fixed the issue, and everything is working, but there is a small ux problem introduced. 

Even though workflows are collapsed, logs of the first step of first workflow is shown by default. We should also include showing the logs of first running step in any workflow by default in this pr.

Thoughts?

6543:
one nit: now workflow names look like steps - can we change this?

6543:
I'm no UX designer, just trained myself to see this stuff - so no no idea :sweat: 

I'll send this question into our designer channel ...

anbraten:
I think its totally fine how it is done for now that workflows and steps are similar. The user would normally not care about the details of a workflow if it passes. In the case the workflow has an error the user wants to fix it so it is important what steps are inside a workflow and by that what is a workflow and what not. 

Anyways keep it simple and lets don't change it for.

---

Regarding the pre-selected step: I would say it would be pretty cool to load the first failed step (if one failed at all). The precedence should be:
- show step selected  by url
- else show first failed step
- else show first step

6543:
well let's merge it and itterate on it

Issue #1379, opened 2022-10-31
Conditional pipeline.step.when.branch not taking array as documented in 0.15


gapodo:
### Component

cli

### Describe the bug

When using the pipeline.step.when.branch with ["main","dev","release/*"] (or without the quotes) the linter fails with `In pipeline.publish-docker-image.when.branch: Invalid type. Expected: string, given: array`, but the [docs for "Conditional Step Execution" in 0.15](https://woodpecker-ci.org/docs/usage/conditional-execution#branch) state that this should be possible.

I've checked the schema definition, and it states only string [see in schema.json L201, v0.15 branch](https://github.com/woodpecker-ci/woodpecker/blob/release/v0.15/pipeline/schema/schema.json#L201).

### System Info

```shell
woodpecker-cli v0.15.4

server and agent v0.15.5 (not really relevant it runs on there, installed via helm on Kubernetes)
```

### Additional context

Minimal replicator pipeline:
```yaml
pipeline:
  issue-replicator:
    image: busybox
    commands:
      - echo "im useless"
    when:
      branch: [main, test]
```

ran against `woodpecker-cli lint <filename>`

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

PR #1384, opened 2022-10-31
Update json schema for cli lint to cover valid cases

gapodo:
This is a follow up on #1379 and #1381, after fixing those (I noticed them while building a pipeline for one of my projects) I took a deep dive into the schema, and consolidated the functionality as far as possible.

As a reference, I matched it against pipeline/frontend/yaml/constraint//constraints.go of the release/v0.15 branch.

The .branches while not in the pipeline uses the Type (List) and functions of pipeline/frontend/yaml/constraint//constraints.go

Regarding the formatting i went with what was already there, since enabling editorconfig / formatting would have changed most of the file (not that there is too much untouched after this anyways)

### DIFF
--- a/pipeline/schema/schema.json
+++ b/pipeline/schema/schema.json
@@ -38,53 +38,14 @@
         "git": {
           "type": "object",
           "properties": {
-            "image": {
-              "type": "string"
-            }
+            "image": { "type": "string" }
           }
         }
       }
     },
     "branches": {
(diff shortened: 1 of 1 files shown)


gapodo:
Forgot to mention any input / feedback welcome, I've tested a local build, linting still works, but now accepts what the server would accept already.

6543 (approved):

Issue #1073, opened 2022-08-03
Prefer `*.yaml` over `*.yml`


qwerty287:
`*.yaml` is the recommended file extension.

jolheiser:
Rather than "prefer", I think we should possibly just handle both. 
Doing so would be backwards compatible, and it would probably be the best UX since both versions are prominent.

PR #1388, opened 2022-11-02
Support .yaml as file-ending for workflow config too

gapodo:
This implements #1073, adds .yaml to the accepted endings for woodpecker configs.

This currently adds some more lines to the duplication (tried to compensate by fixing the other duplication in the configFetcher) as the CLI and Server are still separate.

# ToDos

- [x] Add documentation
- [x] Add possibly more testing ~(currently just testing the path sanitization)~ Done do we need more?
- [x] ~Add option to disable the .yaml variant (is this really needed for next)~ -> we are only breaking in edge-cases so in accordance with the chat between @6543 and me this should be fine without
- [x] Decide whether default reading .yaml is breaking (else we might just skip the conditional) -> see above
- [x] ~Add a backport with a default off addition of .yaml~ -> only backporting bugfixes, etc. I'll do it if needed, but that will be a less invasive change

### DIFF
--- a/cli/common/pipeline.go
+++ b/cli/common/pipeline.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/urfave/cli/v2"
+
 	"github.com/woodpecker-ci/woodpecker/shared/constant"
 )
 
--- a/cli/common/pipeline.go
+++ b/cli/common/pipeline.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/urfave/cli/v2"
+	"github.com/woodpecker-ci/woodpecker/shared/constant"
 )
 
 func DetectPipelineConfig() (multiplies bool, config string, _ error) {
@@ -13,21 +14,12 @@ func DetectPipelineConfig() (multiplies bool, config string, _ error) {
 		return true, config, nil
 	}
 
-	config = ".woodpecker.yml"
-	if fi, err := os.Stat(config); err == nil && !fi.IsDir() {
-		return true, config, nil
+	for _, file := range constant.DefaultConfigFiles {
+		if fi, err := os.Stat(file); err == nil && !fi.IsDir() {
+			return true, config, nil
+		}
 	}
 
-	config = ".woodpecker.yaml"
-	if fi, err := os.Stat(config); err == nil && !fi.IsDir() {
-		return true, config, nil
-	}
-
-	config = ".drone.yml"
-	fi, err := os.Stat(config)
-	if err == nil && !fi.IsDir() {
-		return false, config, nil
-	}
 	return false, "", fmt.Errorf("could not detect pipeline config")
 }
 
--- a/cli/common/pipeline.go
+++ b/cli/common/pipeline.go
@@ -18,6 +18,11 @@ func DetectPipelineConfig() (multiplies bool, config string, _ error) {
 		return true, config, nil
 	}
 
+	config = ".woodpecker.yaml"
+	if fi, err := os.Stat(config); err == nil && !fi.IsDir() {
+		return true, config, nil
+	}
+
 	config = ".drone.yml"
 	fi, err := os.Stat(config)
 	if err == nil && !fi.IsDir() {
--- a/cli/common/pipeline.go
+++ b/cli/common/pipeline.go
@@ -3,21 +3,20 @@ package common
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/urfave/cli/v2"
 
 	"github.com/woodpecker-ci/woodpecker/shared/constant"
 )
 
-func DetectPipelineConfig() (multiplies bool, config string, _ error) {
-	config = ".woodpecker"
-	if fi, err := os.Stat(config); err == nil && fi.IsDir() {
-		return true, config, nil
-	}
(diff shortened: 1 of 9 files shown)


server/shared/configFetcher.go
```diff
@@ -134,41 +128,51 @@ func (cf *configFetcher) fetch(c context.Context, timeout time.Duration, config
 	}
 
 	config = ".woodpecker.yml"
```
6543, line 130: could we refactor it to itterate over a string list that contains the config file/folers and add it to the shared constants?

(https://github.com/woodpecker-ci/woodpecker/blob/master/shared/constant/constant.go)

cli/common/pipeline.go
```diff
@@ -18,6 +18,11 @@ func DetectPipelineConfig() (multiplies bool, config string, _ error) {
 		return true, config, nil
 	}
 
+	config = ".woodpecker.yaml"
```
6543, line 21: we should use mentioned const asap implemented ...

server/shared/configFetcher.go
gapodo: Absolutely, I had that in mind while working on it, was just unsure whether this was OK with you (tried to keep changes minimal)

6543: keep changes minimal is only of value if it need's to be backported witch we dont do here (mostly bugfixes do match this description)

if there is something you can improve while touching the code just do :)

gapodo:
@6543 should we add .drone.yaml also or should we leave as is since it's a legacy config anyways

6543:
well just enhance and refactor it now, next is a simple pull where be break the stuff

gapodo:
I'll look into changing the collection over into a one-stop shop, so we can also add the directory to the constant

server/shared/stepBuilder.go
```diff
@@ -400,6 +400,7 @@ func metadataPipelineFromModelPipeline(pipeline *model.Pipeline, includeParent b
 func SanitizePath(path string) string {
 	path = filepath.Base(path)
 	path = strings.TrimSuffix(path, ".yml")
+	path = strings.TrimSuffix(path, ".yaml")
```
6543, line 403: can we just rm leading `.` and then rm all fine endings?

server/shared/configFetcher.go
gapodo: Moved to /shared/constant

cli/common/pipeline.go
gapodo: also moved to shared/constant

server/shared/stepBuilder.go
gapodo: Would have to be the other way round (first trim (yaml|yml) then the dot since it works right to left), I'm unsure about the performance implications of string-manipulation in go... If you want it with trim . then (yaml|yml) I'll implement it

6543: hm well lets just do it as it is now ... I might push something alter ...

cli/common/pipeline.go
```diff
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/urfave/cli/v2"
+	"github.com/woodpecker-ci/woodpecker/shared/constant"
```
6543, line 8:
```suggestion

	"github.com/woodpecker-ci/woodpecker/shared/constant"
```

gapodo:
Added the "one-stop-shop" to allow the DefaultConfigFiles to also accept directories, added the tests I could think of. (.yml before .yaml outside the directory, any *.(yml|yaml) inside, the rest should already be tested via the "old" tests

gapodo:
Added docs to reflect the changes

gapodo:
@6543 is there a way for you to re-trigger codecov? I've added the tests since and hope this should fix the reduction in coverage.

cli/common/pipeline.go
gapodo: Pushed with the latest changes

docs/docs/91-migrations.md
```diff
@@ -13,6 +13,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 - Updated Prometheus gauge `build_*` to `pipeline_*`
 - Updated Prometheus gauge `*_job_*` to `*_step_*`
 - Renamed config env `WOODPECKER_MAX_PROCS` to `WOODPECKER_MAX_WORKFLOWS` (still available as fallback)
+- The pipelines are now also read from `.yaml` files, the new default order is `.woodpecker/*.yml` and `.woodpecker/*.yaml` (without ani prioritization) -> `.woodpecker.yml` ->  `.woodpecker.yml` -> `.drone.yml`
```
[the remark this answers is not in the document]
gapodo: The ani is fixed (sorry that on slipped through).

Regarding order, when reading a folder (multi pipelines / workflow) the order of reading should be irrelevant, having duplicate pipelines with different file endings would be the same as having duplicates with the same file ending currently (`build.yml` and `build.copy.yml` is the same as having `build.yml` and `build.yaml`, when `build.*` is identical).

For the effective ordering (before parsing `depends_on` / `needs`, etc.), this relies on the order (`*[]remote.FileMeta`) given by the `remote.Dir()` (which is the same behavior as previously used when using only `.yml` in a workflow).

PR #1392, opened 2022-11-03
Wait for gitea to start in gitpod

lukashass:


6543:
so the gitpod shell is always bash ... ?

PR #1395, opened 2022-11-05
Fix `local` and `ssh` backends

qwerty287:
Base64-encoded string was not decoded.


lafriks (approved):

6543:
Well good hotfix ... missed that, but we should return an unencoded func ... I send a followup

6543:
->  #1397

PR #1396, opened 2022-11-05
Add Jsonnet support

qwerty287:
Jsonnet is a templating language for JSON.

Example for a pipeline config (running with local backend):
```jsonnet
{
    pipeline: [
        {
            name: "bash-step",
            image: "bash",
            commands: ["echo Hello " + self.image + "!"],
        },
        self.pipeline[0] { name: "zsh-step", image: "zsh" },
    ]
}
```

which is evaluated into:
```json
{
   "pipeline": [
      {
         "commands": [
            "echo Hello bash!"
         ],
         "image": "bash",
         "name": "bash-step"
      },
      {
         "commands": [
            "echo Hello zsh!"
         ],
         "image": "zsh",
         "name": "zsh-step"
      }
   ]
}
```


6543:
I'm against it and for a global "compile" step that is plugin based ..

@anbraten ?

6543 (changes requested):
as per comment - we need to chat about this clearly

lafriks:
I also agree that this should be made plugable at least by clear interface than they could be either builtin or external plugins

6543:
still thanks for looking into it :)

6543:
->  #1400

PR #1398, opened 2022-11-05
Match account icon hover background shape with account icon shape

dvjn:


anbraten (approved):

anbraten (changes requested):
Cloud you add screenshots?

Why don't you want it to be round anymore?

dvjn:
@anbraten

| hover effect now | hover effect earlier |
| --- | --- |
| image | image |


> Isn't the problem that the underlying image is not rounded?

yes, but i think that was an intentional decision.

6543 (approved):

6543:
we did a poll and got "square with rounded corners", I would say this change is in the right direction, we just somehow need to adapt the "pipeline runing" indicator to match the theme too ...

dvjn:
> we just somehow need to adapt the "pipeline runing" indicator to match the theme too ...

@6543 +1

Also, then we need to convert all the icon buttons background to "square with rounded corners" for consistency sake

6543:
how many do we have of them?

dvjn:
> how many do we have of them?

a lot, but I think they originate from the same component so would be strightforward

anbraten:
I am not convinced yet. Rounded square would look quite special as almost all our icons are based on a circle: 
image

Should we really make them squares on hover?

dvjn:
or we can go the github approach of highlighting an icon button only using the foreground color

anbraten:
was implemented by #2000

PR #1402, opened 2022-11-06
Fix cut of woodpecker animation

Epsilon02:
the woodpecker svg was cut on top and bottom when moving during an active pipeline.
@6543 and me saw this, sadly I have no demo


6543:
image

6543 (approved):

Issue #1266, opened 2022-10-14
Bump Frontend Deps for release branch v0.15.x


6543:
Bump js deps in `release/v0.15` branch
```
yarn.lock (yarn)

Total: 2 (MEDIUM: 1, HIGH: 0, CRITICAL: 1)

┌──────────┬────────────────┬──────────┬───────────────────┬───────────────┬──────────────────────────────────────────────────────────────┐
│ Library  │ Vulnerability  │ Severity │ Installed Version │ Fixed Version │                            Title                             │
├──────────┼────────────────┼──────────┼───────────────────┼───────────────┼──────────────────────────────────────────────────────────────┤
│ minimist │ CVE-2021-44906 │ CRITICAL │ 1.2.5             │ 1.2.6         │ minimist: prototype pollution                                │
│          │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2021-44906                   │
├──────────┼────────────────┼──────────┼───────────────────┼───────────────┼──────────────────────────────────────────────────────────────┤
│ vite     │ CVE-2022-35204 │ MEDIUM   │ 2.6.13            │ 2.9.13        │ Vitejs Vite before v2.9.13 vulnerable to directory traversal │
│          │                │          │                   │               │ via crafted URL to...                                        │
│          │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2022-35204                   │
└──────────┴────────────────┴──────────┴───────────────────┴───────────────┴──────────────────────────────────────────────────────────────┘

```

PR #1403, opened 2022-11-06
web: remove unused @meforma/vue-toaster

6543:
CVE-2022-37601

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -19,7 +19,6 @@
   "dependencies": {
     "@intlify/vite-plugin-vue-i18n": "^6.0.3",
     "@kyvg/vue3-notification": "^2.4.1",
-    "@meforma/vue-toaster": "^1.3.0",
     "@vueuse/core": "^9.3.1",
     "ansi_up": "^5.1.0",
     "dayjs": "^1.11.5",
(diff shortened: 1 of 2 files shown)


6543:
https://avd.aquasec.com/nvd/2022/cve-2022-37601/

web/package.json
```diff
@@ -18,55 +18,60 @@
   },
   "dependencies": {
     "@intlify/vite-plugin-vue-i18n": "^6.0.3",
-    "@kyvg/vue3-notification": "^2.4.1",
+    "@kyvg/vue3-notification": "^2.6.1",
     "@meforma/vue-toaster": "^1.3.0",
```
anbraten, line 22: @6543 We don't use `@meforma/vue-toaster` at all as far as I can tell. So we can remove it and get rid of the problematic dependency completely.

6543: hmm it was added in 58838f225c8425436b83b4aff98a33ad41218e61

anbraten: I guess it was just a test and was never removed again 🙈 

If you search the code for `vue-toaster` it returns just the lines in `package.json` and `pnpm-lock.yaml`

anbraten (approved):

web/package.json
6543: I'll have a look into our v015 release branch, we might clean and bump that too ... (#1266)

anbraten: As it wasn't even imported that doesn't matter.

Issue #1328, opened 2022-10-23
different scroll areas for logview and workspace/step view [firefox]


6543:
logview does resize e.g. when you de-/collapse workspaces ...

6543:
<details>
 <summary>clarify it</summary>

atm:  
one scrowl zone:
![image](https://user-images.githubusercontent.com/24977596/197893575-555a8646-ab72-49bd-88d9-1af7efb11b0c.png)

want two zones:
- ![image](https://user-images.githubusercontent.com/24977596/197893634-f1ede2b1-da6f-43e8-8d64-5e417902f885.png)
- ![image](https://user-images.githubusercontent.com/24977596/197893663-ebd3d94a-646a-437b-9e17-544238e1eb5f.png)
</details>

xoxys:
I guess this issue was fixed already?

6543:
yes 🎉

Issue #959, opened 2022-06-04
RFC: unify pipeline states & colors and document them


anbraten:
We should unify all names, colors and meanings of pipeline states throughout Woodpecker and finally document the states with a description to our docs.

Currently the state is displayed to the user at  following places:
- badge (text with color)
- favicon (colored dot)
- UI pipelines lists (colored icon with colored bar) and 
- UI pipeline log (colored dot & colored icon at top)

The following states do currently exist in the source-code:
- skipped: Pipeline got skipped because its dependencies haven't ended successfully or not at all (exp canceled)
- pending: Pipeline is waiting for an agent to pick it up
- running: Pipeline currently being executed by an agent
- success: Pipeline ended with exit code 0
- failure: Pipeline ended with exit code < 0
- killed: Pipeline got killed (exp canceled)
- error: Pipeline could not be started (exp container image not found)
- blocked: Pipeline is waiting for the confirmation of an repo admin
- declined: Pipeline got declined by an repo admin

![image](https://user-images.githubusercontent.com/6918444/200528131-4205b6b4-a792-424f-80b0-fd7dc2ee7517.png)

PR #1409, opened 2022-11-08
Use icons for step and workflow states

### LINKED ISSUE #1393: Help color blind people by changing dots to icons

As a colorblind user, it is difficult for me to see the status of steps in pipelines.

### Suggested solution

Change the colored dot in the web-ui over to an icon. We could possibly reuse the icons used to display the status in the workflow view.

The colored dot in question:
image

This is the icon I think we could use (of course "all of them" depending on status).
image

### Alternative

Other icons or leave it be? It's not a huge inconvenience (though I'm not fully colorblind)

### Additional context

_No response_

### Validations

anbraten:
close #1393

| Old icons | New icons | New icons without colors (colorblind) |
|---|---|---|
| image | image | image |

image

# Changes
- [x] use icons for workflows and steps #1393
- [x] do not collapse failed workflows
- [x] simplify icons and their coloring #959
- [x] add popover / title to status so you can see the string version on hover
- [x] scroll workflow list #1328


anbraten:
List of all current icons:

image

Linking #959

anbraten:
Maybe as simplified new icon collections this:

image

anbraten:
@gapodo What do you think about the icons?

gapodo:
IMHO this looks great (and even makes sense in completely momochrome, which would be the most extreme form, can't really judge on that, since I am only reed green blind).

But having the icons is a huge improvement to me!

Thanks for implementing it

anbraten:
@gapodo Good catch. We definitely have to use the same colors for the icons and bars. It may even help in this case, but I guess we should not rely on that contrast as its pretty hard to maintain.

6543:
is there a new animation for running?

anbraten:
I have currently removed the animation (the rotating blue dots, the pecking woodpecker for sure not 😉). Not sure about it yet. For the workflows / steps list it was quite a lot.

6543 (approved):

PR #1410, opened 2022-11-08
CI tweaks

jsoref:
I'm trying to use `make in_docker build`

Lots of things failed.

This PR gets me passed the first couple of fails.

Env:
* macOS 12.6.1
* M1
* Rancher Desktop 1.6.1 (dockerd / moby)

Even with the PR as is, I trip on:

```
# runtime/cgo
gcc: error: unrecognized command-line option '-m64'
make: *** [Makefile:159: build-server] Error 2
make: *** [in_docker] Error 2
```


docker/Dockerfile.make
```diff
@@ -2,7 +2,14 @@
 FROM golang:1.18-alpine as golang_image
 FROM node:16-alpine
```
anbraten, line 3:
```suggestion
FROM node:18-alpine
```

docker/Dockerfile.make
```diff
@@ -2,7 +2,14 @@
...
 
-RUN apk add make gcc musl-dev
+RUN apk add --no-cache --update make gcc binutils-gold musl-dev
+
+ENV PNPM_HOME=/root/.local/share/pnpm
+ENV PATH=$PATH:$PNPM_HOME
+
+RUN apk add --no-cache --update curl && \
+    curl -fsSL "https://github.com/pnpm/pnpm/releases/latest/download/pnpm-linuxstatic-x64" -o /bin/pnpm && chmod +x /bin/pnpm && \
+    apk del curl
```
anbraten, line 12:
```suggestion
RUN corepack enable
```

As we do it in our pipelines: https://github.com/woodpecker-ci/woodpecker/blob/master/.woodpecker/web.yml#L14-L15

docker/Dockerfile.make
```diff
@@ -1,8 +1,10 @@
 # docker build --rm  -f docker/Dockerfile.server -t woodpeckerci/woodpecker-server .
 FROM golang:1.18-alpine as golang_image
-FROM node:16-alpine
+FROM node:18-alpine
 
-RUN apk add make gcc musl-dev
+RUN apk add --no-cache --update make gcc binutils-gold musl-dev
+
+RUN corepack enable
```
lafriks, line 7: This would be better added as:
```
... \
  && corepack enable
```
to previous RUN

jsoref:
Fwiw, I tried this, standalone, or w/ a simple target, it works, but used as `CGO_ENABLED=$(CAN_USE_CGO_ENABLED)`, it does not:
```Makefile
ifeq ($(shell uname)-$(shell uname -m), Darwin-arm64)
CAN_USE_CGO_ENABLED = 0
else
CAN_USE_CGO_ENABLED = 1
endif
```

The general `Darwin`/`arm64` `cgo` build failure is described here:
https://github.com/golang/go/issues/23288#issuecomment-354569738

I did a quick look at current `cgo` sources and didn't spot the broken code, but I don't have the energy to determine which version of `cgo` this build is using, and I think it's reasonable to say that the build system should strive for "working" over "being fast on some systems and totally broken on other modern systems".

Here's n expression you can run that yields something which feels like it'd be compatible w/ `CGO_ENABLED=`:
```sh
$ [ "$(uname)-$(uname -m)"==Darwin-arm64 ] && echo 0 || echo 1          
0
```

docker/Dockerfile.make
jsoref: The rest of the file favors `&& \`, so I've applied that style (personally, I agree that `\`/`&& ...` is better, but I try not to violate file style).

Makefile
```diff
@@ -154,7 +156,7 @@ build-ui: ## Build UI
 	(cd web/; pnpm install --frozen-lockfile; pnpm build)
 
 build-server: build-ui ## Build server
-	CGO_ENABLED=1 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -ldflags '${LDFLAGS}' -o dist/woodpecker-server github.com/woodpecker-ci/woodpecker/cmd/server
+	CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -ldflags '${LDFLAGS}' -o dist/woodpecker-server github.com/woodpecker-ci/woodpecker/cmd/server
```
6543, line 159: that also would drop sqlite support

Makefile
```diff
@@ -188,7 +190,7 @@ release-server-xgo: check-xgo ## Create server binaries for release using xgo
 
 release-server: ## Create server binaries for release
 	# compile
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=1 go build -ldflags '${LDFLAGS}' -o dist/server/linux_amd64/woodpecker-server github.com/woodpecker-ci/woodpecker/cmd/server
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -ldflags '${LDFLAGS}' -o dist/server/linux_amd64/woodpecker-server github.com/woodpecker-ci/woodpecker/cmd/server
```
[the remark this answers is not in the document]
jsoref: Well, right now, it's m1 or SQLite. 

If you can find a better way to get m1 and SQLite, I'm all for it. And if you need help testing a change to see if it works on m1, I'm happy to test.

Makefile
jsoref: Well, right now, it's m1 or SQLite. 

If you can find a better way to get m1 and SQLite, I'm all for it. And if you need help testing a change to see if it works on m1, I'm happy to test.

Makefile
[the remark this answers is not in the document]
jsoref: I'll check to see if this is necessary, I had assumed they were equivalent, but this appears to be a cross compilation from the perspective of an m1.

Note: if you need SQLite, then your test coverage is incomplete because this PR let me run all the tests successfully.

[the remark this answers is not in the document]
anbraten: You have to set a specific env var to test against one of the database adapters. So you can run tests against sqlite / postgres / mysql.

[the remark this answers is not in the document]
anbraten: Should we maybe have a look at [https://pkg.go.dev/modernc.org/sqlite?utm_source=godoc]?

Makefile
6543: we could make it a overwrite-able env, so by default .. and our CI do build with CGO and you still can build on m1

jsoref: Personally, I'd rather it just did the right thing when building on an m1. But my efforts to make it figure it out failed miserably, and this is really not my primary focus.

That said, an env var would be fine w/ me. Can you suggest the code to make that work?

6543 (approved):

6543:
just use `CGO_ENABLED=0 make XYZ`

PR #1413, opened 2022-11-10
Fix broken link in FAQ

xordspar0:


6543:
well I guess that's because we have versioned docs now :D

xordspar0:
Ah, I see what you mean. The docs are versioned, but the faq is not. If the faq were versioned, `next/faq` would point to `pipeline-syntax` and `/faq` would point to `conditional-execution`.

6543 (approved):
lgtm for now - once we have a new releases we just revert this ...

PR #1418, opened 2022-11-15
Rename forge-id to forge-remote-id

anbraten:
preparation for #1417 

I want to use forge-id for the id of the correspondig forge of a repo. Therefore I would like to rename `ForgeID` to `ForgeRemoteID`. (Tell me if there is a better name for it).


server/store/datastore/migration/013_rename_forge_id_to_forge_remote_id.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"xorm.io/xorm"
+)
+
+var renameForgeIDToForgeRemoteID = task{
+	name:     "rename-forge-id-to-forge-remote-id",
+	required: true,
+	fn: func(sess *xorm.Session) error {
+		if err := renameColumn(sess, "repos", "forge_id", "forge_remote_id"); err != nil {
```
anbraten, line 25: @qwerty287 Is this migration correct? I saw you added some sync in front in the 12 migration. Do I need to do this as well?

anbraten: perfect

PR #1419, opened 2022-11-15
Fix wrong drone env vars

### LINKED ISSUE #1416: DRONE_BUILD_STATUS environment variable not set to CI_BUILD_STATUS (makes plugin drillster/drone-email fail)
### Component

server, agent

### Describe the bug

While trying to port some pipelines from drone to woodpecker-ci I found out, that the [drillster/drone-email](https://github.com/Drillster/drone-email)  plugin is not working with woodpecker-ci.

Even if the build is successful, the plugin will always send failure notifications. The reason is, that the plugin checks for `DRONE_BUILD_STATUS=='success'` (`DRONE_BUILD_STATUS` is put into the context of a template as `build.status`, the templates uses this to create the message / subject).

 The reason seems is that `DRONE_BUILD_STATUS` has the value `pending`. In the [docs](https://woodpecker-ci.org/docs/usage/environment) it says:

|   NAME  |    Description  |
|----|----|
| CI_BUILD_STATUS | build status (success, failure) |

`DRONE_*` environment variables are not mentioned.

A simple `env | sort` reveals:
```text
CI_BUILD_STATUS=success
CI_JOB_STATUS=success
CI_PIPELINE_STATUS=success
CI_STEP_STATUS=success
DRONE_BUILD_STATUS=pending
```

Is it intended behavior that `CI_BUILD_STATUS` differs from `DRONE_BUILD_STATUS` or a bug?

If the DRONE_* environment variables are just a deprecated legacy, i can offer to fork the plugin and use the correct environment variable. For this it would be helpful, if someone explained to me the difference between `BUILD_STATUS`, `JOB_STATUS`, `PIPELINE_STATUS` and `STEP_STATUS` and when they are set.

### System Info

```shell
I am using a simple docker-compose config with

-compose.yml
services:
  woodpecker-server:
    image: woodpeckerci/woodpecker-server:next

...

  woodpecker-agent1:
    image: woodpeckerci/woodpecker-agent:next
```

### Additional context

_No response_

### Validations

smainz:
Provide up to date drone compatibility environment variables to each step execution.

closes #1416

Before a step is executed, some environemnt variables are updated. This PR ensures, that the updated environment variables are copied to their corresponding `DRONE_` environemt variables.

Side effect is that the `DRONE_` environemnt variables are no longer available in the metadata which should not harm as they are not used inside woodpecker.

Tested by executing this pipeline:

```
pipeline:
  test:
    image: alpine
    commands:
      - env | sort
```

with the docker agent and the woodpecler-cli

**Can someone please comment, if the chosen approach is correct?**

### DIFF
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -179,7 +179,7 @@ func (r *Runtime) execAll(steps []*backend.Step) <-chan error {
 
 			logger.Debug().
 				Str("Step", step.Name).
-				Msgf("environment after SetDroneEnviron(): %s", step.Environment)
+				Msg("Executing")
 
 			processState, err := r.exec(step)
 
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -175,9 +175,11 @@ func (r *Runtime) execAll(steps []*backend.Step) <-chan error {
 				return err
 			}
 
+			SetDroneEnviron(step.Environment)
+
 			logger.Debug().
 				Str("Step", step.Name).
-				Msg("Executing")
+				Msgf("environment after SetDroneEnviron(): %s", step.Environment)
 
 			processState, err := r.exec(step)
 
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -175,6 +175,7 @@ func (r *Runtime) execAll(steps []*backend.Step) <-chan error {
 				return err
 			}
 
+			// add compatibility for drone-ci plugins
 			SetDroneEnviron(step.Environment)
 
 			logger.Debug().
--- a/pipeline/tracer.go
+++ b/pipeline/tracer.go
@@ -21,7 +21,7 @@ func (f TraceFunc) Trace(state *State) error {
 
 // DefaultTracer provides a tracer that updates the CI_ environment
 // variables to include the correct timestamp and status.
-// TODO(bradrydzewski) find either a new home or better name for this.
+// TODO: find either a new home or better name for this.
 var DefaultTracer = TraceFunc(func(state *State) error {
 	if state.Process.Exited {
 		return nil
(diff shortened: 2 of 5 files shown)


smainz:
If someone comments I can resolve the confict by moving the line in question to the new file location

pipeline/pipeline.go
```diff
@@ -175,9 +175,11 @@ func (r *Runtime) execAll(steps []*backend.Step) <-chan error {
 				return err
 			}
 
+			SetDroneEnviron(step.Environment)
+
 			logger.Debug().
 				Str("Step", step.Name).
-				Msg("Executing")
+				Msgf("environment after SetDroneEnviron(): %s", step.Environment)
```
lafriks, line 182: I don't think this change is needed

lafriks (approved):

6543:
I would like to move it into "backend" package in this case ...
-> e.g. ssh/local wont need that

6543:
well for now... I just gona update https://github.com/woodpecker-ci/woodpecker/pull/1419/files#diff-3cbfef9e2984cb46c88d1510f85943834522f5c4c66e4999ffdc9671ee5453f1R24

smainz:
> I would like to move it into "backend" package in this case ... -> e.g. ssh/local wont need that

I am not sure what people are doing with ssh/local at all. My use case is to migrate pipelines from drone to woodpecker. Changing env vars in commands is easy, but not for plugins. 

If you want the drone env vars in ssh/local depends on the mid term goal about compatibility to drone ...

6543:
dont worry merge is on the way

pipeline/tracer.go
```diff
@@ -52,5 +52,6 @@ var DefaultTracer = TraceFunc(func(state *State) error {
 		state.Pipeline.Step.Environment["CI_BUILD_STATUS"] = "failure"
 		state.Pipeline.Step.Environment["CI_JOB_STATUS"] = "failure"
 	}
+
```
6543, line 55:
```suggestion
```

pipeline/pipeline.go
```diff
@@ -175,6 +175,9 @@ func (r *Runtime) execAll(steps []*backend.Step) <-chan error {
 				return err
```
6543, line 175: you dont need to change the default( or any other) tracer as they will be exec here at L173 ... so we always apply to it afterwards in any case :)

6543 (approved):

lafriks (approved):

PR #1423, opened 2022-11-17
Change healtcheck port into address format, redo #1197

velsinki:
As discussed in the comments in PR #1197. Also add documenation accordingly.

One thing I'm not sure about is the simple check in health.go if the address is usable in the GET request or not. From reading https://pkg.go.dev/net#Dial it seems that the only non-standard address format that would work in the `net` package but not in a GET url would likely only be `:port`, as the others listed here are actually also valid urls:

`For TCP, UDP and IP networks, if the host is empty or a literal unspecified IP address, as in ":80", "0.0.0.0:80" or "[::]:80" for TCP and UDP, "", "0.0.0.0" or "::" for IP, the local system is assumed.`

One additional thing I noticed is that while `WOODPECKER_SERVER_ADDR` and `WOODPECKER_SERVER_ADDR` use the default value format of `:PORT`, `WOODPECKER_SERVER` actually uses `localhost:9000`. I guess it makes a bit of sense, considering the server might not be local to the agent, but it looks a bit inconsistent this way. I don't think it would hurt to make the `WOODPECKER_HEALTHCHECK_ADDR` in this format too, but then it's different from the server flags again... :-)

### DIFF
--- a/cmd/agent/health.go
+++ b/cmd/agent/health.go
@@ -78,7 +78,12 @@ var counter = &agent.State{
 // handles pinging the endpoint and returns an error if the
 // agent is in an unhealthy state.
 func pinger(c *cli.Context) error {
-	resp, err := http.Get("http://localhost:3000/healthz")
+	healthcheckAddress := c.String("healthcheck-addr")
+	if healthcheckAddress[0] == ':' {
+		// this seems sufficient according to https://pkg.go.dev/net#Dial
+		healthcheckAddress = "localhost" + healthcheckAddress
+	}
+	resp, err := http.Get("http://" + healthcheckAddress + "/healthz")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


lafriks (approved):

cmd/agent/health.go
```diff
@@ -78,7 +79,12 @@ var counter = &agent.State{
 // handles pinging the endpoint and returns an error if the
 // agent is in an unhealthy state.
 func pinger(c *cli.Context) error {
-	resp, err := http.Get("http://localhost:3000/healthz")
+	healthcheckAddress := c.String("healthcheck-addr")
```
6543, line 82: i wonder if we could use url.Parse instead ...
and if host = "" use localhost

velsinki: I tried playing with `url.Parse`, but it can't even parse `0.0.0.0:3000` for example. So I don't think that's the solution, as it looks like it doesn't handle ports in combination with IP addresses that well.

6543: thanks for clarification :)

PR #1426, opened 2022-11-18
Dismiss popup with escape

lukashass:


anbraten (approved):

lukashass:
GitHub seems to be mixing up checks from #1425 and this PR. They only differ by source repository. Is that our problem or GitHubs?

PR #1432, opened 2022-11-18
Refresh token when running cron

### LINKED ISSUE #1431: The users token is not refreshed when running cron jobs -> cron jobs stop working
### Component

server

### Describe the bug

When executiing cron jobs woodpecker uses the oauth token of the user who created the cron job to authenticate against the forge. Once this token is expired woodpecker can no longer authenticate and execution of the cron jobs fails.

To fix this, woodpecker has to refresh the token using the refresh token stored in the database.

### System Info

```shell
latest next version run from docker
```

### Additional context

Will submit a PR soon.

### Validations

smainz:
Before accessing a repo during a cron job the users token is refreshed like it is done for e.g. create pipeline. User in this context means the creator of the cron job.

fixes #1431

If the scheduled delay between two executions of a cron job is larger than the exiration time of the oauth2 refresh token (provided by the forge) this will fail and the cron job will not be run. But this is the best one can do.

### DIFF
--- a/server/cron/cron.go
+++ b/server/cron/cron.go
@@ -22,6 +22,7 @@ import (
 	"github.com/robfig/cron"
 	"github.com/rs/zerolog/log"
 
+	"github.com/woodpecker-ci/woodpecker/server"
 	"github.com/woodpecker-ci/woodpecker/server/forge"
 	"github.com/woodpecker-ci/woodpecker/server/model"
 	"github.com/woodpecker-ci/woodpecker/server/pipeline"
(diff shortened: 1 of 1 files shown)


server/cron/cron.go
```diff
@@ -121,7 +122,23 @@ func CreatePipeline(ctx context.Context, store store.Store, forge forge.Forge, c
 		return nil, nil, err
 	}
 
-	commit, err := forge.BranchHead(ctx, creator, repo, cron.Branch)
+	// if the forge has a refresh token, the current access token
+	// may be stale. Therefore, we should refresh prior to dispatching
+	// the pipeline.
+	if refresher, ok := server.Config.Services.Forge.(forge.Refresher); ok {
```
6543, line 128:
```suggestion
	if refresher, ok := f.(forge.Refresher); ok {
```
shouldnt this also work?

smainz: It does :-)

Updated

smainz:
Why doesn't the local 'make lint' find what is found in the CI server?

Code_RQTtytNddt

lafriks (approved):

PR #1434, opened 2022-11-19
Point docs nav bar items to selected version

anbraten:
By this change the docs will point to the currently selected docs version. So if you have selected next for example, you will get to the next docs  when clicking on `Docs` in the nav-bar.


anbraten:
@qwerty287 Sorry for that. Mixed sth up. Should be fine now.

lafriks (approved):

lafriks:
Docs build step is failing

PR #1436, opened 2022-11-21
Remove unrelated public events from user feed

### LINKED ISSUE #971: Synchronize status between active view, the right pane, and the top (navigation) bar
### Component

web-ui

### Describe the bug

When a pipeline is triggered, if I browse over the view/screen where the steps are displayed, when the pipeline finishes, the interface "gives the impression" the pipeline is still in progress based on some indicators, specifically, there will still be a section with the `user/repository` (with the blue two-dots animation) and the circle in the top (navigation) bar.

Notice that when it finishes, a section with the `user/repository` is created as well right before the top one with the correct "finished" status.

So it's just a matter of synchronizing some elements and hiding/removing them from the DOM — my guess :thinking:

### System Info

```shell
Codeberg's Woodpecker CI
```

### Additional context

_No response_

### Validations

MFTabriz:
fixes: https://codeberg.org/Codeberg-CI/feedback/issues/28

closes #971

### DIFF
--- a/server/api/stream.go
+++ b/server/api/stream.go
@@ -85,8 +85,7 @@ func EventStreamSSE(c *gin.Context) {
 				log.Trace().Msgf("pubsub subscribe recover return: %v", obj)
 			}()
 			name := m.Labels["repo"]
-			priv := m.Labels["private"]
-			if repo[name] || priv == "false" {
+			if repo[name] {
 				select {
 				case <-ctx.Done():
 					return


MFTabriz:
Looks like the CI is failing the same way as [for the base repo](https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/pipeline/4595/20)

anbraten:
Thanks for having a look on this. 

Notes for testing:
check if the user gets pipeline updates (not only the sidebar, but most importantly new pipelines and updates to the workflows and steps)
- for own private projects
- for org private projects
- for public projects the user does not directly own, but is allowed to see
- for public projects as unauthenticated users

MFTabriz:
> Notes for testing: check if the user gets pipeline updates (not only the sidebar, but most importantly new pipelines and updates to the workflows and steps)
> 
>     * for own private projects
> 
>     * for org private projects
> 
>     * for public projects the user does not directly own, but is allowed to see
> 
>     * for public projects as unauthenticated users

Are there any automated tests for these or do we need to add them?

anbraten:
I just wrote it down for manual testing 😅There are not automatic tests for this atm. Adding them would probably be a lot of work as those would be almost integration tests.

anbraten:
I have manually tested it and my concerns have been confirmed. This change doesn't solve the problem. Instead a user without direct access does no longer get updates of projects he does not own.

MFTabriz:
> I have manually tested it and my concerns have been confirmed. **This change doesn't solve the problem**. Instead a user without direct access does no longer get updates of projects he does not own.

Thanks for checking this out. Do you mean the issue with seeing the public repos in the sidebar is still there after this patch?

anbraten:
> Thanks for checking this out. Do you mean the issue with seeing the public repos in the sidebar is still there after this patch?

That's  "fixed". But it also stops all other events of public repos. 

For example you can have a look at the https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker repo without being authenticated at all as the repo is public. With your change the user isn't getting any updates for the pipelines anymore. So the list in the middle (not the sidebar) would not update from running to successful.

anbraten:
I am currently having a look at the code. In stead of limiting the events in the backend for now I would add some kind of a view filter to the UI which hides projects you are not part of from the sidepanel and the "my" projects / repos list.

MFTabriz:
> I am currently having a look at the code. In stead of limiting the events in the backend for I would add some kind of a view filter to the UI which hides projects you are not part of from the sidepanel and the "my" projects / repos list.

I would be more in favor of filtering them somehow in the backend (for example based on the page the user is on). Sending all public CI events to all the users would create lots of unnecessary traffic specially for the servers with a many public repos.

anbraten:
> I would be more in favor of filtering them somehow in the backend (for example based on the page the user is on). Sending all public CI events to all the users would create lots of unnecessary traffic specially for the servers with many public repos.

The biggest problem with that approach is that the backend needs to know the repos a user is interested in as its not only the repos the user owns, but also public repos the user is having a look at.

MFTabriz:
> The biggest problem with that approach is that the backend needs to know the repos a user is interested in as its not only the repos the user owns, but also public repos the user is having a look at. (Possible, but requires quite some logic).

Is it possible instead of subscribing to the default Pubsub topic (`topic/events`) [here]( https://github.com/woodpecker-ci/woodpecker/blob/a884e13885b9519c6c7bf7f6e41a2f177207b50d/server/api/stream.go#L82),  to subscribe to a topic based on the page url instead? For example something with user/repo of the current page's url appended to it.

anbraten:
> Is it possible instead of subscribing to the default Pubsub topic (topic/events) [here](https://github.com/woodpecker-ci/woodpecker/blob/a884e13885b9519c6c7bf7f6e41a2f177207b50d/server/api/stream.go#L82), to subscribe to a topic based on the page url instead? For example something with user/repo of the current page's url appended to it.

That should be possible. It would require some refactors as the user needs to pass the topic name somehow.

For the beginning I have created #1453 which is a code cleanup of the Vue store and solves the problem quite easily.

MFTabriz:
> > Is it possible instead of subscribing to the default Pubsub topic (topic/events) [here](https://github.com/woodpecker-ci/woodpecker/blob/a884e13885b9519c6c7bf7f6e41a2f177207b50d/server/api/stream.go#L82), to subscribe to a topic based on the page url instead? For example something with user/repo of the current page's url appended to it.
> 
> That should be possible. It would require some refactors as the user needs to pass the topic name somehow.
> 
> For the beginning I have created #1453 which is a code cleanup of the Vue store and solves the problem quite easily.

Thanks a lot! I opened #1454 to track that issue independently.

PR #1439, opened 2022-11-22
CI: pin xgo and go to v1.18.7

6543:


.woodpecker/docker.yml
```diff
@@ -3,10 +3,10 @@ depends_on:
   - web
 
 variables:
-  - &golang_image 'golang:1.18'
+  - &golang_image 'golang:1.19.3'
```
lafriks, line 6: Why not `1.19`?

.woodpecker/docker.yml
```diff
@@ -3,10 +3,10 @@ depends_on:
...
   - &node_image 'node:18-alpine'
-  - &xgo_image 'techknowlogick/xgo:go-1.18.x'
-  - &xgo_version 'go-1.18.x'
+  - &xgo_image 'techknowlogick/xgo:go-1.19.3'
+  - &xgo_version 'go-1.19.3'
```
lafriks, line 9: Same here

.woodpecker/docker.yml
6543: more pinning ... if it goes wrong as with 1.18 again ...

6543: ... well it's an experiment for now if it works or if I need to dig into xgo :(

6543:
**so the issue is a xgo one and it happened in transition from  v1.18.7 to v1.18.8**

anbraten (approved):

6543:
we do need to find the underlying issue long term ...

Issue #1430, opened 2022-11-18
Wrong pipeline status on cancel


roggenbrot:
### Component

server

### Describe the bug

When a started pipeline is canceled, the pipeline status is always `Failed`

It seems that the pipeline status is always based on the the step state

When a step is canceled, the step state reported back by the killed is always failed.

This is caused by 

https://github.com/woodpecker-ci/woodpecker/blob/1786c0554a3f347924a0a1d36c8b2684eb45ac86/agent/runner.go#L365

that is called even if cancel is set

I think `Done` should just be called if cancel is not set

```go
if canceled.IsNotSet() {
   err = r.client.Done(ctxmeta, work.ID, state)
}
```

Also the step state has to be overwritten

```go
stepState := rpc.State{
  Step:     state.Pipeline.Step.Alias,
  Exited:   state.Process.Exited,
  ExitCode: state.Process.ExitCode,
  Started:  time.Now().Unix(), // TODO do not do this
  Finished: time.Now().Unix(),
}
if canceled.IsSet() {
  stepState.ExitCode = 137
  stepState.Error = "Step canceled"
} else if state.Process.Error != nil {
  stepState.Error = state.Process.Error.Error()
}

```

since the error reported seems to be always caused by the step getting killed

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-05d2f493"}
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

roggenbrot:
Duplicate #833

Issue #833, opened 2022-03-07
Cancle a Running Pipeline update status as failed


6543:
I expect the status to be cancled after a cancle, but when already running it shows failed

roggenbrot:
I can submit an PR based on the changes suggested in #1430

6543:
working fix:  #6011 just needs polishing

PR #1441, opened 2022-11-23
Refactor agent: split code in subfunctions

6543:
logs of a killed pipeline are stored, with this pull

### DIFF
--- a/agent/runner.go
+++ b/agent/runner.go
@@ -148,8 +148,8 @@ func (r *Runner) Run(runnerCtx context.Context) error {
 	var uploads sync.WaitGroup
 	err = pipeline.New(work.Config,
 		pipeline.WithContext(workflowCtx),
-		pipeline.WithLogger(r.createLogger(logger, ctxmeta, &uploads, work)),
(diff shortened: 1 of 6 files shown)


6543:
towards  #1430,  #833 ...

6543:
I try to catch the issue (https://github.com/6543-forks/woodpecker/compare/refactor_agent...6543-forks:woodpecker:fix_cancle-status-overwride_WIP?expand=1)  but did not had success for now :/

at least now the logs of a killed pipeline are stored

agent/runner.go
```diff
@@ -150,221 +146,56 @@ func (r *Runner) Run(ctx context.Context) error {
...
-			state.ExitCode = 1
-			state.Error = err.Error()
-		}
-		if canceled.IsSet() {
-			state.ExitCode = 137
+
+	if canceled.IsSet() {
+		state.Error = ""
+		state.ExitCode = 137
+	} else {
```
lafriks, line 167: would be less levels to if `} else if err != nil {` would be used

lafriks (approved):

PR #1442, opened 2022-11-23
router: create apiBase

6543:
https://github.com/woodpecker-ci/woodpecker/pull/1442/files?diff=unified&w=1


server/router/api.go
```diff
@@ -24,168 +24,172 @@ import (
...
 	}
 
-	e.GET("/api/signature/public-key", session.MustUser(), api.GetSignaturePublicKey)
-
 	// TODO: remove /hook in favor of /api/hook
 	e.POST("/hook", api.PostHook)
-	e.POST("/api/hook", api.PostHook)
 
 	// TODO: move to /api/stream
 	sse := e.Group("/stream")
```
anbraten, line 195: We should simply rename those streams as well. As they are not documented we just need to make sure the frontend uses the new path.

server/router/api.go
```diff
@@ -24,168 +24,172 @@ import (
...
+				debugger.POST("/pprof/symbol", debug.SymbolHandler())
+				debugger.GET("/pprof/trace", debug.TraceHandler())
+			}
+		}
 	}
 
-	e.GET("/api/signature/public-key", session.MustUser(), api.GetSignaturePublicKey)
-
 	// TODO: remove /hook in favor of /api/hook
 	e.POST("/hook", api.PostHook)
```
anbraten, line 192: Do we already use `/api/hook` when adding a project and when repairing?

anbraten (approved):

server/router/api.go
6543: no 😆 

And I personally would not describe this as an "api" function as it's not restfull

I would rm /api/hooks instead

same goes with streams, they are not restufll so we should not prefix them with api

anbraten: Api has nothing to do with REST in my understanding. All of those routes are our interface to get data from and into the backend. Putting those urls under a common prefix makes proxying them much easier as well.

6543: what about
/api/* for all backend stuff
and 
/api/v1/* for all restfull stuff?

-> it's about make it easy to generate an openswagger docu and generate clients

else we can just ask what the generall consensus is in mainteiner channel?

anbraten: > -> it's about make it easy to generate an openswagger docu and generate clients

Swagger just ignores not specified / disabled endpoints so that should not be a problem.

6543: well I'll merge this refactor ... and we can move on i an dedicated pull for that ...

server/router/api.go
6543: something for another pull

PR #1453, opened 2022-11-27
Hide not owned repos from sidebar and repo list

### LINKED ISSUE #971: Synchronize status between active view, the right pane, and the top (navigation) bar

### LINKED ISSUE #1454: do not send all public pipeline updates to all the users
Currently, all the public pipeline updates are pushed to every user and get filtered on the client-side. This creates lots of unnecessary traffic for servers with many (authenticated + unauthenticated) users and/or active public repositories.

See #1453 and [#1436](https://github.com/woodpecker-ci/woodpecker/pull/1436#issuecomment-1328218722)

anbraten:
closes #971 , closes #1454
closes #1436 

fixes: https://codeberg.org/Codeberg-CI/feedback/issues/28

## Notes for testing:
check if the user gets pipeline updates (not only the sidebar, but most importantly new pipelines and updates to the workflows and steps)

- for own private projects
- for org private projects
- for public projects the user does not directly own, but is allowed to see
- for public projects as unauthenticated users


web/src/compositions/useEvents.ts
```diff
@@ -30,7 +29,6 @@ export default () => {
     }
     const { pipeline } = data;
     pipelineStore.setPipeline(repo.owner, repo.name, pipeline);
-    pipelineStore.setPipelineFeedItem({ ...pipeline, name: repo.name, owner: repo.owner, full_name: repoSlug(repo) });
```
anbraten, line 33: Not needed anymore as we store a single list of all pipelines.

web/src/views/repo/RepoWrapper.vue
```diff
@@ -132,7 +132,7 @@ watch([repoOwner, repoName], () => {
   loadRepo();
 });
 
-const badgeUrl = computed(() => `/api/badges/${repo.value.owner}/${repo.value.name}/status.svg`);
+const badgeUrl = computed(() => `/api/badges/${repo.value?.owner}/${repo.value?.name}/status.svg`);
```
lukashass, line 135: When `repo` is undefined the `badgeUrl` will be invalid. I would suggest hiding it in that case.

web/src/store/repos.ts
```diff
@@ -1,44 +1,54 @@
...
+    repos.set(repoSlug(repo), repo);
+    return repo;
+  }
+
+  async function loadRepos() {
+    const _repos = await apiClient.getRepoList();
+    _repos.forEach((repo) => {
+      repos.set(repoSlug(repo), repo);
+    });
+    ownedRepoSlugs.value = _repos.map((repo) => repoSlug(repo));
```
lukashass, line 42: @anbraten How is `ownedRepos` different from `repos` when both are derived from `apiClient.getRepoList()`?

web/src/store/pipelines.ts
```diff
@@ -1,99 +1,103 @@
...
+            <PipelineFeed>{
+              ...pipeline,
+              full_name: _repoSlug,
+              owner: _repoSlug.split('/')[0],
+              name: _repoSlug.split('/')[1],
+              number: _pipelineNumber,
+            },
+        );
+        return [...acc, ...repoPipelinesArray];
+      }, [] as PipelineFeed[])
```
lukashass, line 77: @anbraten Can't you pass `PipelineFeed[]` as type parameter to `reduce`?

web/src/store/repos.ts
anbraten: @lukashass Its the same in this moment. The difference is that we add repos (not owned ones) when we receive an event for it see useEvents.

web/src/store/pipelines.ts
anbraten: changed it

web/src/views/repo/RepoWrapper.vue
anbraten: done

web/src/store/repos.ts
lukashass: @anbraten What is the behavior we want eventually? Display public repos, never show public repos or let the user toggle? I feel like rest and events should return the same thing :thinking:

anbraten: @lukashass We want both. The sidepanel and the repos page should show owned repos. In addition a user and anonymous users should be able to access public repos of other users / orgs. Those public repos are added on demand to the store if the corresponding pages are opened.

lukashass, line 42:
```suggestion
    const _ownedRepos = await apiClient.getRepoList();
    _ownedRepos.forEach((repo) => {
      repos.set(repoSlug(repo), repo);
    });
    ownedRepoSlugs.value = _ownedRepos.map((repo) => repoSlug(repo));
```

@anbraten What do you think of this?

lukashass:
Tested:

- [x] for own private projects
- [x] for org private projects
- [x] for public projects the user does not directly own, but is allowed to see
- [x] for public projects as unauthenticated users


Findings:

- Org private projects do not appear in the feed
- Woodpecker admin sees all repos/pipeline updates

@anbraten

anbraten:
> when starting on route /org/repo the feed is empty and doesn't receive updates. When starting on /repos the feed shows pipelines and receives updates.

Strange. It should be have a pipeline entry to add the step updates to, but I will investigate on that.

> Woodpecker admin sees all repos/pipeline updates

Are you sure that is the case? The filter I've added only shows pipelines which match the names from `GetRepos` and that function uses `RepoList` from the store which is not checking for admin permission as far as I can tell.

@lukashass

anbraten:
@lukashass Kindly reminder

lukashass:
> Are you sure that is the case? The filter I've added only shows pipelines which match the names from GetRepos and that function uses RepoList from the store which is not checking for admin permission as far as I can tell.

@anbraten Well, just tried this again and can't reproduce. I guess I messed something up before.

The other thing is still an issue though!

anbraten:
@lukashass Thanks for testing. The problem was that the repo list was not loaded on some pages. Should be fixed now.

web/components.d.ts
```diff
@@ -77,6 +77,7 @@ declare module '@vue/runtime-core' {
     IPhXCircle: typeof import('~icons/ph/x-circle')['default']
     ISimpleIconsGitea: typeof import('~icons/simple-icons/gitea')['default']
     ITeenyiconsGitSolid: typeof import('~icons/teenyicons/git-solid')['default']
+    ITeenyiconsRefreshOutline: typeof import('~icons/teenyicons/refresh-outline')['default']
```
lukashass, line 80: @anbraten What's this change about?

anbraten: @lukashass Its an automatically created file adding typings for all components. It has some issues detecting the icon components, so each time you open a page or so with an icon not being indexed before it adds that component. We just approve it in reviews. 🙈

lukashass (approved):

PR #1415, opened 2022-11-11
Add button in web-ui to trigger the deployment event

smainz:
This PR adds a button to the web ui to trigger the deployment event on the current pipeline.

closes #1412

The button is only active, if the current pipeline is in the state success.

I would like to get some feedback if the path I ave taken to implement this was the correct one (my first ever contact with Vue):

- add deployPipeline( ... ) to WoodpeckerClient similair to what is done in client.go
- Copy the ManualPipelinePopup.vue and add a param id for the current pipeline id, change the text ids
- Add a button into PipelineWrapper to open the new popup
- Add texts for locales en and de

Is there anything I have forgotten to change?

PR #1455, opened 2022-11-27
Consistent and more descriptive naming of parameters in index.ts

smainz:
Parameter naming has been made more descriptive and consistent:
- Change `number: number` to `pipelineNumber: number`.
- Rename `pipeline`  to `pipelineNumber` when it refered to a pipeline number.

(Child of #1415)

There are still inconsistencies in the code, some methods declare the pipelineNumber as number, others as string.
Which direction to go?


web/src/assets/locales/de.json
```diff
@@ -77,6 +77,7 @@
         "cancel": "Abbrechen",
         "cancel_success": "Pipeline abgebrochen",
         "canceled": "Dieser Schritt wurde abgebrochen.",
+        "deploy": "Deployment starten",
```
lafriks, line 80: New translation keys must be added to en.json otherwise they will be removed by weblate and everyone expects to translate from English to other languages

lafriks (changes requested):

6543:
please resolve conflicts @smainz  :)

web/src/assets/locales/de.json
lafriks, line 80: Translations must be done in weblate not directly in code

lafriks (changes requested):

smainz:
Sorry for being away for so long, but  I was busy doing other things.

6543 (approved):

PR #1463, opened 2022-11-29
Manually skip workflow

### LINKED ISSUE #1354: Allow to manually skip (waiting) workflows
Add a button to skip a workflow that is currently waiting for e.g. `depends_on` or agent capacities.

the-writer-dev:
Trying to close #1354
 
@qwerty287 I had problems with the previous branch so moving to this one with resolution of your comments. Sorry for any potential confusion!

### DIFF
--- a/server/api/pipeline.go
+++ b/server/api/pipeline.go
@@ -445,11 +445,11 @@ var deleteStr = `[
 
 func PostSkipWorkflow(c *gin.Context) {
 	var (
-		_store     = store.FromContext(c)
-		repo       = session.Repo(c)
-		user       = session.User(c)
-		num, _     = strconv.ParseInt(c.Params.ByName("number"), 10, 64)
-		stepPid, _ = strconv.Atoi(c.Params.ByName("stepPid"))
+		_store        = store.FromContext(c)
+		repo          = session.Repo(c)
+		user          = session.User(c)
+		num, _        = strconv.ParseInt(c.Params.ByName("number"), 10, 64)
+		workflowId, _ = strconv.Atoi(c.Params.ByName("workflowId"))
 	)
 
 	pl, err := _store.GetPipelineNumber(repo, num)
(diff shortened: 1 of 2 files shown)


bitethecode:
@qwerty287 Thanks for the comment! I will be working on them sooner

bitethecode:
@qwerty287 Those should be resolved! 

Screen Shot 2022-12-14 at 1 21 24 PM

lafriks:
would be nice if it would show confirmation dialog for workflow skip button

bitethecode:
@lafriks Thanks for the comment! I will make sure to update that

bitethecode:
Screen Shot 2022-12-15 at 12 22 48 PM

@lafriks does this look good?

lafriks:
This also definetly helps, what I was more thinking about confirmation dialog with buttons yes/no or something like that to confirm before skipping steps

bitethecode:
@lafriks Hmm.. I follow cancel flow. And it is not popping any dialog.. I wonder we want to throw a dialog for any mutating operations? From my understanding, skip workflow is way less impactful than cancel.

lafriks:
That's definitely not requirement, just would be nice to have imho

Cancel button also would be nice to have that

bitethecode:
@qwerty287 I agree..

lafriks:
I have no problem with that 👍🏻

gapodo:
> That's definitely not requirement, just would be nice to have imho
> 
> Cancel button also would be nice to have that

I'm with you on having those (I'd almost tend to see it as a requirement), just thinking about some long-running pipelines (stuff like integration testing, which takes potentially hours and "must" be run in the same workflow as the release build, or large builds that cannot be resumed part way), also re-running on accidentally canceled tags may cause troubles (i.e. things already uploaded to a, then canceling either the step or the entire pipeline and on a rerun running into conflics with leftovers from the last tagged run).

server/api/pipeline.go
```diff
@@ -442,3 +442,26 @@ var deleteStr = `[
 		"out": "logs purged by %s on %s\n"
 	}
 ]`
+
+func PostSkipStep(c *gin.Context) {
```
6543, line 446: step or workflow :D ...

please make a consistent naming acording to #745

bitethecode: Okay.. which one do you think making more sense?

6543: well the issue is about workflows that are pending

6543:
please resolve conflict :)

anbraten:
closing as #1731 seems to be a newer version of this one

PR #1470, opened 2022-12-09
disable metrics access if no token is set

anbraten:
backport of #1469


lafriks (approved):

gapodo:
IMHO, this should be considered a breaking change (it changes default behavior and cannot be overridden currently)

6543 (approved):

winterqt:
Out of curiosity, anyone know why [this pipeline](https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/pipeline/4708/36) failed for this commit? The next commit at this tag that actually ran pipelines succeeded, with no related changes, so...

6543:
database was to slow to drop index before test did try to create new one

PR #1475, opened 2022-12-11
Secrets encryption in database

### LINKED ISSUE #101: Store secrets encrypted

antomy-gc:
closes #101

Added secrets encryption in database

- Google TINK or simple AES as encryption mechanisms
- Keys rotation support on TINK
- Existing SecretService is wrapped by encryption layer
- Encryption can be enabled and disabled at any time


lafriks:
Would be nice if encryption/decryption could be made more generic (using `Encryption` as interface not type) so that other type of encryption could be also implemented (like for example simple symmetric encryption with static secret).
While TINK is cool but it's also not for everyone imho and in some scenarios simpler solution would be nice where you don't need external tools to generate keys etc.

antomy-gc:
I\`ve already seen some encryption code in server, so `Encryption` interface seems to be a really good idea. But AEAD encryption requires additional piece of data (AAD) on each encrypt and decrypt operation, while simple symmetric algorithms doesn\`t, so I\`m not sure if we can generalize this two approaches.

As for the keys, we need extra tools, such as openssl, to generate them anyway, but tink allows us to avoid the human factor in the keys lifecycle. 
Actually we could easily add some features like "generate keys on first startup" or "rotate every N days", but i think server administrator should perform these operations manually to avoid data loss or corruption.

lafriks:
Other implementations can just return empty data or empty functions to support that interface.

For asymmetric encryption it would require OpenSSL to generate pub/priv key but for symmetric entryption secret can be any random word :)

antomy-gc:
Interface - ok, sounds good.

Yes, you can specify "password" as an encryption key, but it doesn\`t seem to be secure :) 
And nothing will prevent you from typing "passwor" instead of "password" or copying not only part of secret string, while encrypting all the secrets

lafriks:
> Yes, you can specify "password" as an encryption key, but it doesn`t seem to be secure :)

it's still better than no encryption. And as I said previously everything depends on usecase. There is no point in having super secure encryption that is not very easy to use and because of that no one are using. Having also easy to use encryption as alternative means more adoption and more secure deployments.

> And nothing will prevent you from typing "passwor" instead of "password" or copying not only part of secret string, while encrypting all the secrets

That's not really an argument

antomy-gc:
I would really want to avoid weak algorithms because of the nature of encrypted data: common use case is storing very sensitive, sometimes administrative tokens as secrets. And if the point is easy setup i think we could discuss the setup flow instead of agorithms.

So what do you think about adding "generate-keys" startup flag, which could create secret file without using additional tools? 
It will make initial setup easy:
1.Launch server with this flag
2.Backup keyset file if needed

We could even do it without additional flag:
if keyset filepath provided, file not exists and server is not encrypted we will generate own keyset

lafriks:
I would not say that symmetric encryption algorithm means weak encryption. There is not really much difference from security perspective to be honest between your implementation and good symmetric encryption if attacker gets to the secret keystore.

antomy-gc:
Yes, it just provides good enough implementation and enforces user to use secure enough key material. And the last one is the main encryption security factor.

We have the classic "chicken or the egg" dilemma when talking about "what if attacker gets my master-key": we wold like to encrypt master key too, but then we will have another unencrypted master key:) 
HashiCorp Vault solves this dilemma using external authorities, such as cloud providers. Using tink, potentially we could also enable user to store keyset encrypted, but i think it\`s a very-future optional features which are not essential right now. 
And until we\`ll be able to provide third-party-encrypted secret keys, the only thing we can is hardening attacker\`s life by adding security layers

lafriks:
That was not my point!

My point was that asymmetric encryption with secure enough secret for **this** use case will be as secure as your solution but way easier to use.

I'm not saying that we can not have both - that's why I said about interface to be able to implement both

antomy-gc:
Sorry, i didn\`t seen your last message and updated my previous one. And it seems to be a great misunderstanding then. It\`s absolutely not a problem for me to realize both approaches. So i`ll be waiting for your comment on this part of my previous:

Anyway, i\`m going to refactor my implementation of encryption now, so tell me if my vision fits yours:
- Create Encryption interface, which includes "encrypt string" and "decrypt string" methods
- Implement "tink" and "simple symmetric" variants of encryption
- Allow user to choose one of them or "none" on startup
- support keys rotation for tink
- Create SecretStore wrappers, which will add encryption to "store" an "get" operations
- Pass suitable wrapper or unwrapped storage to existing SecretService on server setup

antomy-gc:
@lafriks could you review pls?

cmd/server/flags.go
```diff
@@ -525,4 +525,22 @@ var flags = []cli.Flag{
 		Hidden:  true,
 		// TODO(485) temporary workaround to not hit api rate limits
 	},
+	//
+	// secrets encryption in DB
+	//
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_ENCRYPTION_KEY"},
```
lafriks, line 532: it should be possible to set both secret as direct env or `WOODPECKER_ENCRYPTION_KEY_PATH` that would read it from file so that docker secrets can be used for this

cmd/server/server.go
```diff
@@ -18,6 +18,8 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
+	"github.com/woodpecker-ci/woodpecker/server/plugins/encrypted_secret_store"
```
lafriks, line 21: please sort imports (this should be in lower group for local imports)

server/model/encryption.go
```diff
@@ -0,0 +1,27 @@
...
+	Build()
+}
+
+type EncryptionServiceBuilder interface {
+	WithClients(clients []EncryptionClient) EncryptionServiceBuilder
+	Build() EncryptionService
+}
+
+type EncryptionService interface {
+	Encrypt(plaintext string, associatedData string) string
```
lafriks, line 15: does associatedData change for each call? imho it should be just:

```go
Encrypt(plaintext string) string
Decrypt(ciphertext string) string
```

antomy-gc: yes, the common practise is using object id to encrypt its fields

cmd/server/server.go
antomy-gc: fixed

server/plugins/encryption/aes.go
```diff
@@ -0,0 +1,78 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package encryption
+
+import (
+	"crypto/cipher"
```
lafriks, line 18: please sort imports in all files, it should be in groups (separated by empty line):
* go bultin packages
* external packages
* project local packages

lafriks: **offtopic** @6543 @anbraten why do we have different order used that in other projects I have seen... usually it's:
* go bultin packages
* project local packages
* external packages

antomy-gc: ok, i will check all changed files. Unfortunately my IDE by default attempts to make it one big group, not 3 different

antomy-gc: done

6543: @lafriks good question, I guess because of default gofumpt behaviour?

docs/docs/30-administration/40-encryption.md
```diff
@@ -0,0 +1,64 @@
...
+
+By default, Woodpecker does not encrypt secrets in its database. You can enable encryption 
+using simple AES key or more advanced [Google TINK](https://developers.google.com/tink) encryption.
+
+## Common
+
+### Enabling secrets encryption
+
+To enable secrets encryption and encrypt all existing secrets in database set
+`WOODPECKER_ENCRYPTION_KEY`, `WOODPECKER_ENCRYPTION_KEY`  or `WOODPECKER_ENCRYPTION_TINK_KEYSET_PATH` environment
```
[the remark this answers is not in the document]
antomy-gc: fixed

docs/docs/30-administration/40-encryption.md
```diff
@@ -0,0 +1,64 @@
...
+You will need plaintext AEAD-compatible Google TINK keyset to encrypt your data.
+
+To generate it and then rotate keys if needed, install `tinkey`([installation guide](https://developers.google.com/tink/install-tinkey))
+
+Keyset contains one or more keys, used to encrypt or decrypt your data, and primary key ID, used to determine which key 
+to use while encrypting new data.
+
+Keyset generation example:
+```shell
+tinkey create-keyset --key-template AES256_GCM --out-format json --out keyset.json`
```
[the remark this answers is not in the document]
antomy-gc: fixed

6543:
please merge latest master into & fix the lint issue :)

antomy-gc:
done

server/plugins/encryption/aes.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (svc *aesEncryptionService) Encrypt(plaintext, _ string) string {
+	msg := []byte(plaintext)
+	chainSize := svc.blockSize()
+	infoBlock := svc.newSizeInfoChunk(len(msg))
+	msg = svc.alignDataByChainSize(msg)
+	encrypted := make([]byte, len(infoBlock)+len(msg))
+	err := svc.encode(encrypted[0:len(infoBlock)], infoBlock)
+	if err != nil {
+		log.Fatal().Err(err).Msgf("encryption error")
```
lafriks, line 42: this should be returned as error and handled as error in upper layer

server/plugins/encryption/aes.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	for n := 0; n < len(msg)/chainSize; n++ {
+		dst, src := encrypted[(n+1)*chainSize:(n+2)*chainSize], msg[n*chainSize:(n+1)*chainSize]
+		err = svc.encode(dst, src)
+		if err != nil {
+			log.Fatal().Err(err).Msgf("encryption error")
```
lafriks, line 49: same here

server/plugins/encryption/aes.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}
+	return base64.StdEncoding.EncodeToString(encrypted)
+}
+
+func (svc *aesEncryptionService) Decrypt(ciphertext, _ string) string {
+	ct, err := base64.StdEncoding.DecodeString(ciphertext)
+	if err != nil {
+		log.Fatal().Err(err).Msgf("encryption error: Base64 decryption failed")
```
lafriks, line 58: and here

server/plugins/encryption/aes_builder.go
```diff
@@ -0,0 +1,56 @@
...
+		store:   c.store,
+		clients: c.clients,
+	}
+	svc.initClients()
+	svc.loadCipher([]byte(c.key))
+	err := svc.validateCipher()
+	if err == encryptionNotEnabledError {
+		svc.enable()
+	} else if err == encryptionKeyInvalidError {
+		log.Fatal().Err(err).Msg("Error initializing AES encryption")
```
lafriks, line 53: this should also return error

lafriks (changes requested):
Error handling should be reworked so that actual caller can handle error appropriate. Currently all errors are hidden in internal functions and can not be handled correctly. For example returning correct error to UI when saving secret or returning error to pipeline when reading secrets for pipeline

antomy-gc:
> Error handling should be reworked so that actual caller can handle error appropriately. Currently all errors are hidden in internal functions and can not be handled correctly. For example returning correct error to UI when saving secret or returning error to pipeline when reading secrets for pipeline

The main idea was to use `Fatal()` and crash server in case of any errors which could not be properly handled and recovered. 
I agree that such methods as `encrypt` or `decrypt` should just return errors to caller because it doesn\`t seem to be `Fatal` situation. 
But i still would like to log and crash server in case of encryption errors on server startup or especially keys rotation

6543:
On server init if there is an error we do stop (gracefull error handling)

If its while securestore is already init and we try to use it ... we should not crash but log the error

So in any case i would say fatal is just ugly and dont help.
(Dont get me wrong I alsp prevere a fail-hard on startup)

If you think this error should not only be backpropagated but also logged ... just log errorlevel ...

lafriks:
yes it's definitely should not crash server if there is error saving secret to database for example

6543:
☝️ well I know tests are anoying :D ... adding some would still help

6543:
oh you did ... only tink has no coverage

6543:
@lafriks we should add a linter that prevent panic / or log.fatal - beside if explicite excluded from linting

antomy-gc:
> On server init if there is an error we do stop (gracefull error handling)
> If its while securestore is already init and we try to use it ... we should not crash but log the error
> So in any case i would say fatal is just ugly and dont help. (Dont get me wrong I also prevere a fail-hard on startup)
> If you think this error should not only be backpropagated but also logged ... just log errorlevel ...

Sure, fatals on encryption/decryption is a bad idea. Will fix. 
And i used fatal in securestore init because the only reason of this error is wrong coding/usage. I think we shouldnt allow server to launch if it contains incorrect code inside. 
Developer can ignore "error" messages while implementing some feature, but one cant ignore fatal. 
But we could move this fatal from store wrapper implementation to the encryption builder

antomy-gc:
> oh you did ... only tink has no coverage

aes has basic coverage because of custom implementation. 
tink has none, because we just use external library without any complicated logic. 

anyway, i could add some, if needed

6543:
The fatal issue would be important, more tests is a bonus

server/plugins/encryption/wrapper/store/secret_store_wrapper.go
```diff
@@ -32,60 +33,87 @@ func NewSecretStore(secretStore model.SecretStore) *EncryptedSecretStore {
...
 	}
+	return nil
 }
 
-func (wrapper *EncryptedSecretStore) _save(secret *model.Secret) {
+func (wrapper *EncryptedSecretStore) _save(secret *model.Secret) error {
 	err := wrapper.store.SecretUpdate(secret)
 	if err != nil {
-		log.Fatal().Err(err).Msg("Storage error: could not update secret in DB")
+		log.Err(err).Msg("Storage error: could not update secret in DB")
```
lafriks, line 115: this will be logged already down the code in caller

server/plugins/encryption/encryption.go
```diff
@@ -62,56 +65,17 @@ func (b builder) Build() {
 	keyType := b.detectKeyType()
 
 	if !enabled && (disableFlag || keyType == keyTypeNone) {
-		noEncryptionBuilder{}.WithClients(b.clients).Build()
+		_, err := noEncryptionBuilder{}.WithClients(b.clients).Build()
+		if err != nil {
+			log.Fatal().Err(err).Msg("failed initializing server in unencrypted mode")
```
lafriks, line 70: this should also return error and log error in: https://github.com/inDriver/woodpecker-encrypted/blob/39287b9932837a0be0c424005430f3f8e95034d6/cmd/server/server.go#L267

server/plugins/encryption/encryption.go
```diff
@@ -62,56 +65,17 @@ func (b builder) Build() {
...
-	return b.serviceBuilder(keyType).WithClients(b.clients).Build()
-}
-
-func (b builder) isEnabled() bool {
-	_, err := b.store.ServerConfigGet(ciphertextSampleConfigKey)
-	if err != nil && !errors.Is(err, types.RecordNotExist) {
-		log.Fatal().Msgf("Failed to read server configuration: %s", err)
+		err := svc.Disable()
+		if err != nil {
+			log.Fatal().Err(err).Msg("failed disabling encryption")
```
lafriks, line 78: same here

server/plugins/encryption/wrapper/store/secret_store_wrapper.go
antomy-gc: updated logging in wrapper

server/plugins/encryption/encryption.go
antomy-gc: will fix now. and found another 2 fatals in wrapper

antomy-gc:
about testing: i didnt find an easy way to add tests from store to wrapper and as for me it doesnt seem to have any profit to add tink tests

server/plugins/encryption/tink.go
```diff
@@ -0,0 +1,61 @@
...
+// limitations under the License.
+
+package encryption
+
+import (
+	"encoding/base64"
+	"fmt"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/google/tink/go/tink"
```
6543, line 22:
```suggestion
	"github.com/google/tink/go/tink"

```

server/plugins/encryption/tink_state.go
```diff
@@ -0,0 +1,146 @@
...
+		keysetFilePath:    svc.keysetFilePath,
+		primaryKeyID:      "",
+		encryption:        nil,
+		store:             svc.store,
+		keysetFileWatcher: nil,
+		clients:           svc.clients,
+	}
+	err := newSvc.loadKeyset()
+	if err != nil {
+		return fmt.Errorf("failed rotating TINK encryption: %w", err)
```
6543, line 60: same string is used a lot, we might consider a private const value

6543:
& please merge latest master into this branch

server/plugins/encryption/wrapper/store/secret_store_wrapper.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/rs/zerolog/log"
+
+	"github.com/woodpecker-ci/woodpecker/server/model"
+)
+
+type EncryptedSecretStore struct {
+	store      model.SecretStore
+	encryption model.EncryptionService
+}
```
6543, line 30:
```suggestion
}

// ensure wrapper match interface
var _ model.SecretStore = new(EncryptedSecretStore)
```

just a nit to make sure thing work, and you will get an error in IDe asap you miss something or on compiletime afterwards ...

server/plugins/encryption/tink_state.go
antomy-gc: replaced all of messages with constants. i actually like this approach, it allows to easily find similar/duplicated logs on different abstraction layers

6543 (approved):

lafriks:
Please fix lint errors

6543:
@antomy-gc `server/plugins/encryption/tink_keyset_watcher.go:51: (*github.com/rs/zerolog.Event).Msgf format %w reads arg #1, but call has 0 args`

server/plugins/encryption/tink_keyset_watcher.go
```diff
@@ -0,0 +1,61 @@
...
+		select {
+		case event, ok := <-svc.keysetFileWatcher.Events:
+			if !ok {
+				log.Fatal().Msg(errMessageTinkKeysetFileWatchFailed)
+			}
+			if (event.Op == fsnotify.Write) || (event.Op == fsnotify.Create) {
+				log.Warn().Msgf(logTemplateTinkKeysetFileChanged, event.Name)
+				err := svc.rotate()
+				if err != nil {
+					log.Fatal().Err(err).Msgf(errTemplateFailedRotatingEncryption)
```
6543, line 51:
```suggestion
					log.Fatal().Err(err).Msgf(errTemplateFailedRotatingEncryption, err)
```

6543:
@antomy-gc please merge master :)

6543 (approved):

antomy-gc:
@lafriks @6543 any updates?)

lafriks:
Why are you using so custom AES encryption? At least code wise it seems quite hard to follow it through to validate it. Why not use tested standard implementation, for example AES-256-CFB, ex: https://go.dev/src/crypto/cipher/example_test.go#L187

lafriks:
also also maybe hmac-sha256 can be added to this where sha256(additional data+secret) would be hmac secret to be completely safe to be able to validate it

lafriks:
Or AES-256-GCM that provides ability to use additional data for validation: https://go.dev/src/crypto/cipher/example_test.go#L18 (aad is nil in example but it can be provided for Seal/Open as last parameter)

antomy-gc:
review pls.
updated aes to use google implementation from [example](https://go.dev/src/crypto/cipher/example_test.go#L18)

effectiveKey := sha256(userKey) is used as key to avoid incorrect user key length
bcrypt(effectiveKey) is used as keyID to detect invalid keys on startup

docs/docs/30-administration/40-encryption.md
```diff
@@ -0,0 +1,64 @@
...
+To disable secrets encryption and decrypt database you need to start server with valid
+`WOODPECKER_ENCRYPTION_KEY` or `WOODPECKER_ENCRYPTION_TINK_KEYSET_FILE` environment variable set depending on
+enabled encryption method, and `WOODPECKER_ENCRYPTION_DISABLE` set to true.
+
+After secrets was decrypted server will proceed working in unencrypted mode. You will not need to use "disable encryption"
+variable or encryption keys to start server anymore.
+
+
+## AES
+Simple AES encryption.
```
6543, line 27: AES **GCM** ...

why did you choose GCM over CBC @antomy-gc ?

antomy-gc: https://thehftguy.com/2020/04/20/what-aes-ciphers-to-use-between-cbc-gcm-ccm-chacha-poly/
>AES-GCM (Galois Counter Mode)
The most widely used block cipher worldwide.
Mandatory as of TLS 1.2 (2008) and used by default by most clients.
RFC 5288 year 2008 https://tools.ietf.org/html/rfc5288AES-GCM (Galois Counter Mode)

> AES-CBC
First historic block cipher for AES.
CBC mode is insecure and must not be used. It’s been progressively deprecated and removed from SSL libraries.
Introduced with TLS 1.0 year 2002. Superseded by GCM in TLS 1.2 year 2008. Removed in TLS 1.3 year 2018.
RFC 3268 year 2002 https://tools.ietf.org/html/rfc3268

Google also recommends AEAD with GCM key as suitable and secure enough for most uses.
https://developers.google.com/tink/aead

lafriks: If I remember correctly Hashicorp Vault also uses AES GCM

lafriks (approved):
To be honest I'm not big fan of log/error messages being moved out to constants as it complicates looking at code and log entries to follow the code path but if everyone else are ok with that I won't block.

Otherwise looks good from my side, thanks for your work

docs/docs/30-administration/40-encryption.md
antomy-gc: can i resolve it and merge?

antomy-gc: @6543 can we resolve it and merge?

6543:
@antomy-gc big thanks! :)

6543:
we something went horible wrong ... -> https://github.com/woodpecker-ci/woodpecker/pull/1549 will disable it in code for now

PR #1510, opened 2022-12-30
Allow adding additional labels/annotations to kubernetes worker pods

smuth4:
The issue I'm actually trying to fix is that the Istio service mesh injects sidecar containers into the worker pods which A) confuses the log fetcher since it doesn't know which container is correct and B) doesn't seem to properly clean up pods (since the sidecar runs forever). I fixed A because it's simple and non-impactful, but even if B could be fixed I'm sure there are other landmines waiting.

This solution should also be general enough to be useful for other edge cases. I personally don't need annotations, just figured it was easy to add while I was mucking around in the right areas.

Example agent environment configuration using the new value:
```yaml
  - env:
    - name: WOODPECKER_BACKEND
      value: kubernetes
    - name: WOODPECKER_BACKEND_K8S_NAMESPACE
      value: default
    - name: WOODPECKER_BACKEND_K8S_POD_LABELS
      value: '{"sidecar.istio.io/inject":"false"}'
```

### DIFF
--- a/pipeline/backend/kubernetes/kubernetes.go
+++ b/pipeline/backend/kubernetes/kubernetes.go
@@ -249,7 +249,7 @@ func (e *kube) Tail(ctx context.Context, step *types.Step) (io.ReadCloser, error
 	<-up
 
 	opts := &v1.PodLogOptions{
-		Follow: true,
+		Follow:    true,
 		Container: podName,
 	}
 
--- a/pipeline/backend/kubernetes/kubernetes.go
+++ b/pipeline/backend/kubernetes/kubernetes.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"github.com/woodpecker-ci/woodpecker/pipeline/backend/types"
+	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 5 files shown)


lafriks (approved):

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -147,9 +152,21 @@ func (e *kube) Setup(ctx context.Context, conf *types.Config) error {
 
 // Start the pipeline step.
 func (e *kube) Exec(ctx context.Context, step *types.Step) error {
-	pod := Pod(e.config.Namespace, step)
+	labels := make(map[string]string)
```
6543, line 155: could you deserialize this on `configFromCliContext` instead on first exec

so we fail on agent start instead if a step should be exec

6543 (changes requested):

pipeline/backend/kubernetes/kubernetes.go
smuth4: Done

6543 (approved):

6543:
thanks :)

PR #1515, opened 2022-12-31
Rework status constraint logic for successes

### LINKED ISSUE #1181: conditional step event failure aways triggered

smuth4:
Since "success" and "failure" are the only two possible values, and "success" is considered to be included by default, the existing code can also be simplified a little.

This has the side effect of ignoring the "exclude" part of the constraint completely. I put it in the tests just to make sure the workaround in https://github.com/woodpecker-ci/woodpecker/issues/1181#issuecomment-1347253585 continues to work as expected, but couldn't think of any legitimate use cases for it.

Fixes #1181

### DIFF
--- a/pipeline/frontend/yaml/compiler/convert.go
+++ b/pipeline/frontend/yaml/compiler/convert.go
@@ -134,11 +134,10 @@ func (c *Compiler) createProcess(name string, container *yaml.Container, section
 		cpuSet = c.reslimit.CPUSet
 	}
 
-	// all constraints must exclude success.
-	onSuccess := container.When.IsEmpty() ||
-		!container.When.ExcludesStatus("success")
+	// all constraints must not include success.
+	onSuccess := container.When.IncludesStatusSuccess()
 	// at least one constraint must include the status failure.
-	onFailure := container.When.IncludesStatus("failure")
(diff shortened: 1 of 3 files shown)


lafriks (approved):

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -134,11 +134,10 @@ func (c *Compiler) createProcess(name string, container *yaml.Container, section
 		cpuSet = c.reslimit.CPUSet
 	}
 
-	// all constraints must exclude success.
-	onSuccess := container.When.IsEmpty() ||
-		!container.When.ExcludesStatus("success")
+	// all constraints must not include success.
```
6543, line 137:
```suggestion
	// at least one constraint must include the status success.
```

?

6543 (approved):

Issue #999, opened 2022-06-22
Autoscaler


kdumontnu:
### Clear and concise description of the problem

As a potential user of Woodpecker I would really like to be able to provision my agent servers on an "as-needed" basis, without having to support Kubernetes. 

### Suggested solution

Support a Woodpecker-Autoscaler image, which will accept user credentials to spin-up and shut down agent instances as necessary. 

# Roadmap: 
- [x] #1333
- [x] allow to set no-schedule flag to agents to prevent them from grabbing new workflows #1567 
- [x] list workflows executed by an agent

Support major cloud providers:
- [x] Hetzner
- [ ] AWS
- [ ] GCP
- [ ] #1334 https://github.com/woodpecker-ci/autoscaler/pull/16
- [ ] vultr https://github.com/woodpecker-ci/autoscaler/pull/17
- [ ] Azure
- [ ]  #1335
- [ ]  #1336
- [x] Linode https://github.com/woodpecker-ci/autoscaler/pull/15

### Alternative

I believe Kubernetes is currently proposed as the alternative to autoscaler for woodpecker, but requires all of the infrastructure (and cost) associated with that.

### Additional context

I know this has been discussed at many points before on discord, but I'm not sure if it's been determined that it's out of scope or not. It would be helpful to track that discussion here. 

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

anbraten:
Some questions we have to answer in general:
- Is there maybe a go package Woodpecker could use or would it require to write some extendable system to support all kind of hosters? 
- Which metrics does it need to decide about rescaling.

mr337:
I'm not for sure how relevant this is, but BuildKite [1] has something similar for AWS but leverages external tooling to do so. The BuildKite service has an API that an AWS Lambda polls every minute and asks for pending jobs. If there are pending jobs it increase the count on an autoscale group. A new instance is fired, agent connects, and puck up the pending job. That handles the autoscaling increment operation.

The other part is if the agent is idle for X minutes it will then shut itself down, as part of shutting down it will decrement the autoscale group. 

I think this design is pretty nice as it has the benefits of:
[...]

PR #1517, opened 2023-01-02
Add my small autoscaler project to awesome list

ljoonal:
I've made a small project (as in, very minimal implementation that pretty much only addresses my own needs currently) that creates and removes VPS server based on the current CI needs, with support for Woodpecker.

And also noticed now that picus seems to do about the same, just didn't have the magic word I was ctrl+f'ing for.
Additionally my project isn't setup to be convenient to others (using my own git & container registry, neither of which has the best of uptime).
So I'm not 100% if this fits on this list?
But thought I'd at least open a PR and see, as I personally think that multiple implementations for the same thing is healthy 😄


6543:
well I would say better two autoscaler implementations to rever to than none ...

#999 is planed to be build in at some point ... but I dont see any reason to not let peopel have own projects too :)

Issue #1508, opened 2022-12-29
Agents: pull all default defined images on start / or new release  ?


6543:
so we wont need that https://floss.social/@WoodpeckerCI/109595047113683395 in the future

anbraten:
We please should use fixed tagging of such images instead as discussed earlier.

6543:
it we also pin the patch version ... we would have to release a new woodpecker version just to bump a bugfix ...

pining majour and minor still is usefull as that way we make sure it will exactly as intended with no features/enhancements or even breaking changes added

gapodo:
I have a question and a thought on this topic...

What is concidered a default defined image? just the git-plugin, all images in the default escalate list,...?

"Forced pulling" is IMHO generally a bad idea...
* you may introduce new bugs, which then "just appear" in a users' setup after a reboot
* Pulling on startup would lead to increased startup times (esp. if the images aren't needed for a pipeline, or images are pulled blocking)
* potentially waste space (depending on which images you pull, and which images the user uses)
* Risk users running into API-Limits (Pull limits) because of reboots, agent restarts and crashes
* You'd potentially override "manipulated" images, when a user did something to them and just tagged them the same, as to not change the configs (I know one should not do this, but trust me there are setups, where this is easier for an urgently needed fix)
* Introduce a new difference between the backends (does it also force pull for Kubernetes, which in some cases will just fail, as it may be prevented by cluster wide policies? how does it deal with podman,...)
* The "solution" is IMHO the wrong approach as changing released Tags after the fact calls for trouble... (e.g. I'm proxying all images (like many companies do), and prohibit tags from ever changing once they are present on that cache...)

If it's an "included" (hardcoded) image, the fix should be a new patch release (it's a fix after all).

lafriks:
@gapodo it's common practice to use docker image versioning schema:
* `:2` -> latest 2.x.y version (will change when new minor/patch version change)
* `:2.0` -> latest 2.0.x version (will change when patch version change)
* `:2.0.1` -> specific version (this should never change)

gapodo:
> @gapodo it's common practice to use docker image versioning schema:
> * `:2` -> latest 2.x.y version (will change when new minor/patch version change)
> * `:2.0` -> latest 2.0.x version (will change when patch version change)
> * `:2.0.1` -> specific version (this should never change)

I'm not disputing that this is commonly done (though highly frowned upon by many of the big players in the container field).

[...]

PR #1526, opened 2023-01-03
Pin official default images

6543:
address https://github.com/woodpecker-ci/woodpecker/issues/1508#issuecomment-1368007494


docs/docs/30-administration/10-server-config.md
```diff
@@ -203,7 +203,7 @@ Always use authentication to clone repositories even if they are public. Needed
 List of event names that will be canceled when a new pipeline for the same context (tag, branch) is created.
 
 ### `WOODPECKER_DEFAULT_CLONE_IMAGE`
-> Default: `woodpeckerci/plugin-git:latest`
+> Default is defined in [shared/constant/constant.go](https://github.com/woodpecker-ci/woodpecker/blob/release/v0.15/shared/constant/constant.go)
```
6543, line 206:
```suggestion
> Default is defined in [shared/constant/constant.go](https://github.com/woodpecker-ci/woodpecker/blob/master/shared/constant/constant.go)
```

shared/constant/constant.go
```diff
@@ -14,12 +14,13 @@
...
-	"plugins/docker",
-	"plugins/gcr",
-	"plugins/ecr",
-	"woodpeckerci/plugin-docker",
-	"woodpeckerci/plugin-docker-buildx",
+	"docker.io/plugins/docker:20.14.0",
+	"docker.io/plugins/gcr:20.14.0",
+	"docker.io/plugins/ecr:20.14.0",
+	"docker.io/woodpeckerci/plugin-docker-buildx:2.1.0",
+	// "docker.io/woodpeckerci/plugin-docker",
```
anbraten, line 23: I am not sure if this list support tags and if we should pin those images as they are used for checking the permissions and not for fetching images.

lafriks: This will be hard to manage to have to count all allowed tags for each image

lafriks: Also I don't see how that improves security as if you can push new tag you can also push malicious tag over existing version tag. For security reasons we would have to pin to image tag sha256 than

shared/constant/constant.go
```diff
@@ -14,12 +14,13 @@
...
 package constant
 
+// PrivilegedPlugins can be changed by 'WOODPECKER_ESCALATE' at runtime
 var PrivilegedPlugins = []string{
 	"plugins/docker",
 	"plugins/gcr",
 	"plugins/ecr",
-	"woodpeckerci/plugin-docker",
 	"woodpeckerci/plugin-docker-buildx",
+	// "woodpeckerci/plugin-docker",
```
lafriks, line 23:
```suggestion
```

lafriks (approved):

PR #1535, opened 2023-01-12
WIP: add support for Forgejo

earl-warren:


6543:
I don't see any need for it right now, as woodpecker use the api and this dont differ from gitea ...

and as long as it's a soft fork it should be able to use the same api calls etc...

---
Beside adding documentation and alternative config env vars, what would be the benefit to add this code?

jolheiser:
I realize I'm going to have some bias here, but as of now the fork is entirely the same. 
Is there a compelling reason (at least currently) to copy/paste the entire Gitea structure for this?

I won't approve or deny because as above, I may bring bias to this situation, but until the API has diverged I would consider this noise.

gapodo:
> I realize I'm going to have some bias here, but as of now the fork is entirely the same.
> Is there a compelling reason (at least currently) to copy/paste the entire Gitea structure for this?

I'm going to have some bias as well (Forgejo contributor and somewhat unhappy with how the Gitea thing went down), so take it with a pinch of salt, there are two major reasons I currently see for separating:
1) People that are unwilling to contribute things benefitting Gitea (not one of them, but I understand the grudge held by some)
2) Not breaking setups once there is some divergence

For 1, there is no solution except "total separation" which causes code duplication. For 2 I think we could for now "just" add the `FOGREJO_` flags (alias and extra ENV) to the `GITEA_` flags, allowing users to set up their CI with `FORGEJO_` now and move these flags once there is divergence between the 2 APIs, allowing for a non-breaking upgrade path once the differences affect the woodpecker-ci integration.

jolheiser:
As I thought about it, your second idea came across my mind as well, and I think that would suffice.

With my woodpecker maintainer hat on, I am largely disinterested in Gitea/Forgejo politics.
Copy/pasting entire swathes of code because of grudges just adds unnecessary complexity at this stage.
If/when the API changes, I think probably the easiest path is to embed the gjtea client in a potential forgejo client, such that it can proxy any funcs that haven't changed at that time to the gitea handler. 
This would be the easiest to maintain imo, and allows for a quicker iteration on both rather than complete separation from the get-go.

earl-warren:
I understand and respect the reservations of the maintainers.

6543:
I'm for mentioning it in the docs and provide an alias for config ... (to make it clear we do support ...)

PR #1544, opened 2023-01-14
fix & cleanup encryption code

### LINKED ISSUE #1541: Secrets encryption can not be disabled
`... By default, Woodpecker does not encrypt secrets in its database. You can enable encryption using simple AES key or more advanced Google TINK encryption...`

but right now we just get an error on start

rev #1475
cc @antomy-gc

anbraten:
close #1541


antomy-gc:
sorry for that bug, i missed return statement in encryption.go / Build()
i see you removed disable flag, is it still supposed to support disable?

PR #1546, opened 2023-01-15
detached steps can have ports

earl-warren:
Signed-off-by: Earl Warren <contact@earl-warren.org>

### DIFF
--- a/pipeline/schema/schema.json
+++ b/pipeline/schema/schema.json
@@ -243,6 +243,14 @@
           "type": "string",
           "enum": ["fail", "ignore"],
           "default": "fail"
+        },
+        "ports": {
+          "description": "expose ports to which other steps can connect to",
+          "type": "array",
+          "items": {
+            "oneOf": [{ "type": "number" }, { "type": "string" }]
+          },
+          "minLength": 1
         }
       }
     },


6543:
one note: we did not specify for normal steps witch hostname name-convention is used if any ...

so for detatched steps to have a port ... that's normaly what I would call a service ... but I can see value in a service that depend on a build-step before exec and so that would be a detatched step, depending on other steps ...

this need two todo's solved first:
 - [ ] write down the convention how hostnames are created derived from steps
 - [ ] implement the convention for docker and kubernetis backend
 - [ ] write convention and examples how to use into docs

cc @woodpecker-ci/maintainers

6543:
info for @earl-warren : indocumented features like that are considered to be removed or declared properly ... until then it's an unofficial feature that can break at any time

6543:
(lets let it stay to get consensus ...)

earl-warren:
I won't be following this but anyone is welcome to push commits.

lafriks:
> one note: we did not specify for normal steps witch hostname name-convention is used if any ...
> 
> so for detatched steps to have a port ... that's normaly what I would call a service ... but I can see value in a service that depend on a build-step before exec and so that would be a detatched step, depending on other steps ...
> 
> this need two todo's solved first:
> 
>     * [ ]  write down the convention how hostnames are created derived from steps
> 
>     * [ ]  implement the convention for docker and kubernetis backend
> 
>     * [ ]  write convention and examples how to use into docs
> 
> 
> cc @woodpecker-ci/maintainers

maybe we could have service property `depends_on: [ stepx ]` and then service would be started only after specific steps are done thus no additional concept needs to be made on how to connect to them

6543:
@lafriks nice idea, that would solve a lot...

pipeline/schema/schema.json
```diff
@@ -243,6 +243,14 @@
           "type": "string",
           "enum": ["fail", "ignore"],
           "default": "fail"
+        },
+        "ports": {
```
anbraten, line 247: Maybe we should have two step types (normal & service) based on `detached: true` using [oneOf](https://json-schema.org/understanding-json-schema/reference/combining.html#oneof)

anbraten:
@6543 @lafriks `depends_on` is in general a nice approach for defining dependencies (instead of group) and running steps in parallel. Main problem is that it would be a breaking change at least for normal steps.

lafriks:
> @6543 @lafriks `depends_on` is in general a nice approach for defining dependencies (instead of group) and running steps in parallel. Main problem is that it would be a breaking change at least for normal steps.

In this case I meant this to add this in service section not in pipeline steps. Services currently does not have group imho, so this won't be breaking

PR #1564, opened 2023-01-25
fix SetupConsoleLogger which is a noop

earl-warren:
When App.Before is called the options are not parsed yet, use Command.Before instead.

### DIFF
--- a/cmd/cli/app.go
+++ b/cmd/cli/app.go
@@ -65,7 +65,9 @@ func newApp() *cli.App {
 			Out: os.Stderr,
 		},
 	)
-	app.Before = common.SetupConsoleLogger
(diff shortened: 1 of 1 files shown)


cmd/cli/app.go
```diff
@@ -65,7 +65,6 @@ func newApp() *cli.App {
 			Out: os.Stderr,
 		},
 	)
-	app.Before = common.SetupConsoleLogger
```
lafriks, line 68: Imho it would be better to loop app.Commands and set it there than for each command adding it manually

earl-warren: much better indeed

lafriks (approved):

PR #1568, opened 2023-01-30
Show platform, backend and capacity as badges in agent list

lafriks:
Screenshots:

attels

attels


web/src/components/atomic/Badge.vue
```diff
@@ -0,0 +1,22 @@
+<template>
```
anbraten, line 1: @lafriks Isn't there already a similar component for the badges in the pipeline side panel for matrix pipelines? Would be nice to re-use them. I am also not so sure if blue is the best fitting color.

anbraten (commented):
Nice idea.

web/src/components/atomic/Badge.vue
lafriks: @anbraten had not used matrix pipelines didn't know there was already badges. Updated badge style to match one in pipelines and made it to also use new component

anbraten: thanks

anbraten (approved):

PR #1569, opened 2023-01-30
allow custom dnsConfig in agent deployment

vquiering:
At the moment there is no way to set a custom dns config inside the agents deployment. When using alpine images a deployment pipeline it is impossible to overwrite just using the pipeline yaml. Therefor this adds the possibility to add the config using the values.yaml of the helm chart.

There is no breaking change.


charts/woodpecker-agent/templates/deployment.yaml
```diff
@@ -91,3 +91,8 @@ spec:
       - name: sock-dir
         emptyDir: {}
       {{- end }}
+      {{- with .Values.dnsConfig }}
+      dnsConfig:
```
anbraten, line 95: Should we add those settings to the server chart as well?

vquiering: Since I didn't need it there I didn't add the option. Added it now though.

anbraten (approved):

PR #1578, opened 2023-02-04
Update 15-agent-config.md

inflatador:
Document DOCKER_HOST environment variable as described at https://github.com/woodpecker-ci/woodpecker/pull/763


lafriks (approved):

anbraten:
That's the wrong version I guess. It was added to 0.15

lafriks:
It is used in 0.15

lafriks:
Though it should be added also for next

PR #1597, opened 2023-03-03
docs: add note about image cleanup

### LINKED ISSUE #1532: [Docs] Mention that agents dont delete images and an demo howto add clean task
agents dont clean pulled images, so sysadmin has to deal with it.

we should mention that and add a hint how to do so for docker:
```
sudo crontab -e

> @daily               docker system prune --all -f --volumes
```

hay-kot:
Closes #1532

Adds notes about images not being cleaned up by wood-pecker and provides examples on how you can automate the cleanup. Also provides a warning that these are destructive actions. 

CleanShot 2023-03-03 at 09 36 35@2x

### DIFF
--- a/docs/docs/30-administration/15-agent-config.md
+++ b/docs/docs/30-administration/15-agent-config.md
@@ -132,26 +132,3 @@ See [Docker backend configuration](backends/docker/#configuration)
 ### `WOODPECKER_BACKEND_SSH_*`
 
 See [SSH backend configuration](backends/ssh/#configuration)
-
-
-## Image Cleanup
-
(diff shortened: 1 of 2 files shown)


docs/docs/30-administration/15-agent-config.md
```diff
@@ -132,3 +132,26 @@ See [Docker backend configuration](backends/docker/#configuration)
 ### `WOODPECKER_BACKEND_SSH_*`
 
 See [SSH backend configuration](backends/ssh/#configuration)
+
+
+## Image Cleanup
```
anbraten, line 137: @hay-kot Not sure if this is the best place to put it. I would move it to the docker backend page as it is only valid / needed for that backend.

hay-kot: Should be good now 👍

anbraten (approved):

PR #1602, opened 2023-03-04
wip: render button as a tag (need feedback)

### LINKED ISSUE #1259: repo list allow to open in new tab
repo list should also be as able as build list to open in a new tab

hay-kot:
Closes #1259

I've adjusted the button component to conditionally render the proper element given the props. I did run into one oddity. If you specify both the `:href="..."` and `:to="..."` the Router Links will not work, hence the btnAttrs work around to avoid the issue.

Additionally, this PR currently causes a recursive render because the Button component and the DOM button element share the same name, so if we conditionally render based on the string `button` it tries to render the Button component and then we hit a max recursion limit. To resolve this we need to **rename the button component**. 

What should I rename it to? `BaseButton` would be my suggestion, but I wanted some feedback for I make that big of a change.

### DIFF
--- a/web/src/components/atomic/Button.vue
+++ b/web/src/components/atomic/Button.vue
@@ -1,6 +1,7 @@
 <template>
-  <button
-    type="button"
+  <component
+    :is="to === null ? 'button' : httpLink ? 'a' : 'router-link'"
+    v-bind="btnAttrs"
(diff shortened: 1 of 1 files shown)


6543:
BaseButton sounds fine to me

cc @anbraten

Issue #1483, opened 2022-12-13
Confirm secret deletion


gapodo:
### Clear and concise description of the problem

I've just accidentally deleted a secret (which I do not have access to the values) by clicking in the wrong row, there was not confirmation asked and it's gone now.

This is Hypothetical, as I actually have the value and wanted to delete it, but I noticed how easy it is to delete... (and I'm working in projects where I admin the CI but don't have access to all secrets, think things like release signing keys)

### Suggested solution

In the web-ui we could add a confirmation popup (modal?), at least showing the name of the secret to be deleted and ask the user to click on `Delete` or `cancel` (which should be spaced apart a bit for better safety on touch devices / for people with problems with their fine-motor skills).

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

PR #1604, opened 2023-03-05
feat: UI should confirm secret deletion

alexmt:
Closes https://github.com/woodpecker-ci/woodpecker/issues/1483

PR adds a confirmation dialog that warns user before deleting a secret. 

https://user-images.githubusercontent.com/426437/222985225-9e14ddac-6c5a-4aed-ac89-cc6562180ac6.mov

Using just basic [window.Confirm](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) since it is used in [other](https://github.com/search?q=repo%3Awoodpecker-ci%2Fwoodpecker+confirm%28&type=code) places. Any objections to switch to proper dialog component in a separate PR?

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -125,6 +125,7 @@
                 "show": "Show secrets",
                 "name": "Name",
                 "value": "Value",
+                "delete_confirm": "Do you really want to delete this secret?",
                 "deleted": "Secret deleted",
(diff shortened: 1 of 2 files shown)


lafriks (approved):

lafriks:
I would definitely like to switch to proper confirmation dialog

anbraten (approved):

Issue #1606, opened 2023-03-08
`depends_on` ignored if server is restarted


alexef:
### Component

server

### Describe the bug

assume a multi pipeline set up, with:
- pipeline a
- pipeline b, depends_on: [pipeline a]

while a is running and b is pending, restart the drone server

what is expected: b stays in pending until a completes
what happens: b immediately starts running, ignoring a which is also still running

### System Info

```shell
{
  "source": "https://github.com/woodpecker-ci/woodpecker",
  "version": "v0.14.2"
}

I know this is an older version, I will now try to reproduce the issue on the latest release.
```

### Additional context

maybe related to: https://github.com/woodpecker-ci/woodpecker/issues/200

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

alexef:
Update: we (cc: @lukaszgyg) were able to reproduce the issue (same steps) also on latest stable: 
```
{
  "source": "https://github.com/woodpecker-ci/woodpecker",
  "version": "0.15.6"
}```

alexef:
By checking the code (it's been a while since I touched it), it seems like we have two representation of a Task, with a slight difference:
- queue.Task (in memory, has Dependencies and DepStatus)
- model.Task (in datastore, has only Dependencies)

the [task.ShouldRun](https://github.com/woodpecker-ci/woodpecker/blob/e79ad0082687ea9aa1ecc7847f69e2c851752dc4/server/queue/queue.go#L42) method will check DepStatus. However, when we load tasks in memory (at start up) we [don't fill the DepStatus property](https://github.com/woodpecker-ci/woodpecker/blob/e79ad0082687ea9aa1ecc7847f69e2c851752dc4/server/queue/persistent.go#L39). When this is empty, then ShouldRun returns true.

[...]

PR #1610, opened 2023-03-09
Persist `DepStatus` of tasks

lukaszgyg:
This PR is going the add the task DepStatus persistence support. This will allow the server to properly restore the state after restart. This change should fix the issue https://github.com/woodpecker-ci/woodpecker/issues/1606


server/store/datastore/task_test.go
```diff
@@ -47,6 +49,9 @@ func TestTaskList(t *testing.T) {
 	if got, want := list[0].Data, "foo"; string(got) != want {
 		t.Errorf("Want task data %s, got %s", want, string(got))
 	}
+	if got, want := list[0].DepStatus, map[string]string{"test": "dep"}; !reflect.DeepEqual(got, want) {
```
lafriks, line 52: probably this would be better:

```go
require.EqualValues(t, map[string]string{"test": "dep"}, list[0].DepStatus)
```

anbraten (approved):

server/store/datastore/task_test.go
```diff
@@ -39,19 +38,10 @@ func TestTaskList(t *testing.T) {
...
-	}
-	if got, want := list[0].Data, "foo"; string(got) != want {
-		t.Errorf("Want task data %s, got %s", want, string(got))
-	}
-	if got, want := list[0].DepStatus, map[string]string{"test": "dep"}; !reflect.DeepEqual(got, want) {
-		t.Errorf("Want task data %s, got %s", want, got)
-	}
+	assert.Len(t, list, 1, "Expected one task in list")
+	assert.EqualValues(t, "some_random_id", list[0].ID)
+	assert.EqualValues(t, "foo", list[0].Data)
```
lafriks, line 43:
```suggestion
	assert.Equal(t, "some_random_id", list[0].ID)
	assert.Equal(t, "foo", list[0].Data)
```

lukaszgyg: As the `Data` attribute is type byte array the `assert.Equal` will not work in that case. See the unit test output:
```
=== RUN   TestTaskList
    task_test.go:43: 
        	Error Trace:	/Users/lukasz/projects/woodpecker/server/store/datastore/task_test.go:43
        	Error:      	Not equal: 
        	            	expected: string("foo")
        	            	actual  : []uint8([]byte{0x66, 0x6f, 0x6f})
        	Test:       	TestTaskList
--- FAIL: TestTaskList (0.00s)
```
So, I just changed the assert for `ID` which will work with `Equal`.

lukaszgyg: jap, the string casting will work as well

lafriks (approved):

lafriks:
Please rebase with base

alexef:
I'm guessing we should also backport this to the release/0.15 branch. Will prepare the pr

Issue #1615, opened 2023-03-12
 woodpecker-cli exec .woodpecker.yaml: "io: read/write on closed pipe"


alexmt:
### Component

cli

### Describe the bug

Woodpecker CLI logs the error after each step execution.

Sample pipeline:

```
pipeline:
  step1:
    image: golang
    commands:
      - echo step1
  step2:
    image: golang
    commands:
      - echo step2
```

Result:

```
woodpecker exec .woodpecker.yaml                                                                        (k3d-akuity/akuity-platform)
[step1:L0:0s] + echo step1
[step1:L1:0s] step1
8:04PM ERR process logging failed error="io: read/write on closed pipe" CLI=exec
[step2:L0:0s] + echo step2
[step2:L1:0s] step2
8:04PM ERR process logging failed error="io: read/write on closed pipe" CLI=exec
```

### System Info

```shell
woodpecker-cli version 0.15.6
```
```


### Additional context

_No response_

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

PR #1616, opened 2023-03-12
fix: docker backend should not close 'engine.Tail' result

alexmt:
Closes https://github.com/woodpecker-ci/woodpecker/issues/1615

The error described in https://github.com/woodpecker-ci/woodpecker/issues/1615 is happening because `Tail` method of the docker backend closes the instance of `io.ReadCloser` it returns in `defer` function. As a result anything that try to read data returned by `Tail` method eventually will attempt to read from closes reader and get an error:

https://github.com/woodpecker-ci/woodpecker/blob/2171212c5a31bfffc8f882716bdd503c65413eee/pipeline/backend/docker/docker.go#L229

The fix is just don't close returned reader and let the consumer of `Tail` method do it. Good thing is that `Tail` is used only in one place and reader is correctly closed:

https://github.com/woodpecker-ci/woodpecker/blob/2171212c5a31bfffc8f882716bdd503c65413eee/pipeline/pipeline.go#L231-L237

Example of `woodpecker exec` output using pipeline from https://github.com/woodpecker-ci/woodpecker/issues/1615 with the fix: 

```
woodpecker exec .woodpecker.yaml
[step1:L0:0s] + echo step1
[step1:L1:0s] step1
[step2:L0:0s] + echo step2
[step2:L1:0s] step2
```

### DIFF
--- a/pipeline/backend/docker/docker.go
+++ b/pipeline/backend/docker/docker.go
@@ -226,7 +226,6 @@ func (e *docker) Tail(ctx context.Context, step *backend.Step) (io.ReadCloser, e
 		_, _ = stdcopy.StdCopy(wc, wc, logs)
 		_ = logs.Close()
 		_ = wc.Close()
-		_ = rc.Close()
 	}()
 	return rc, nil
 }


lafriks (approved):

6543:
backport -> #1620

Issue #1617, opened 2023-03-12
`woodpecker exec` intermittently switches between local and docker backends


alexmt:
### Component

cli

### Describe the bug

The `woodpecker exec` command is not consistent when run on the machine with docker installed. . It intermittently switches between local and docker backends.

# Steps to reproduce:

To reproduce try to execute the following pipeline multiple times:

.woodpecker.yaml
```
pipeline:
  step1:
    image: golang
    commands:
      - echo step1
```

```
while true; do  woodpecker exec .woodpecker.yaml; done
[step1:L0:0s] + echo step1
[step1:L1:0s] step1
[step1:L0:0s] + echo step1
[step1:L1:0s] step1
[step1:L0:0s] + echo step1
[step1:L1:0s] step1
[step1:L0:0s] + echo step1
[step1:L1:0s] step1
[step1:L0:0s] Backend engine error while running step: exec: "golang": executable file not found in $PATH1:48PM FTL error running cli error="exec: \"golang\": executable file not found in $PATH"
[step1:L0:0s] + echo step1
[step1:L1:0s] step1
[step1:L0:0s] Backend engine error while running step: exec: "golang": executable file not found in $PATH1:48PM FTL error running cli error="exec: \"golang\": executable file not found in $PATH"
[step1:L0:0s] Backend engine error while running step: exec: "golang": executable file not found in $PATH1:48PM FTL error running cli error="exec: \"golang\": executable file not found in $PATH"
```

Periodically woodpecker chooses to use the local backend.

# Expected Behavior:

CLI should consistently choose the same backend. IMHO docker backend should take priority over local backend.

### System Info

```shell
woodpecker-cli --version                                                                                    
woodpecker-cli version 0.15.6
```
```


### Additional context

_No response_

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
[...]

PR #1618, opened 2023-03-12
fix: backend auto-detection should be consistent

alexmt:
Closes https://github.com/woodpecker-ci/woodpecker/issues/1617

The `woodpecker exec` auto-detects the backend by iterating over a map of backends. However, since Go 1 the runtime randomizes map iteration order, so a random backend might be chosen during each execution.

PR changes to auto-detection to iterate over the backends list with predefined priority: `docker`, `local`, `ssh`, `kubernetes`.

### DIFF
--- a/pipeline/backend/backend.go
+++ b/pipeline/backend/backend.go
@@ -11,19 +11,20 @@ import (
 	"github.com/woodpecker-ci/woodpecker/pipeline/backend/types"
 )
 
-var engines map[string]types.Engine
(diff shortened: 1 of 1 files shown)


lafriks (approved):

alexmt:
Thanks for approving @lafriks ! Pushed [commit](https://github.com/woodpecker-ci/woodpecker/pull/1618/commits/99b810983ef3f26a623ad671a50549056af10fd9) that fixed linter error

PR #1632, opened 2023-03-16
Add queue details UI for admins

anbraten:
# Changes
- Adds an admin view to see the whole work-queue of the server. 
- The admin can also pause / resume the queue. 
- The view is reloading data every 5 seconds automatically.
- The task model from queue got removed in favor of the one from models.

image


server/router/api.go
```diff
@@ -145,8 +145,8 @@ func apiRoutes(e *gin.Engine) {
 		{
 			queue.Use(session.MustAdmin())
 			queue.GET("/info", api.GetQueueInfo)
-			queue.GET("/pause", api.PauseQueue)
-			queue.GET("/resume", api.ResumeQueue)
+			queue.POST("/pause", api.PauseQueue)
```
6543, line 148: is breaking but :+1:

anbraten: I will add a note to the migration list

6543: need to adjust woodpecker-go too

6543: they are not used there so its fine & thanks for the docs change

anbraten: seems to be only having the info endpoint atm

6543 (approved):

PR #1634, opened 2023-03-16
Add users UI for admins

anbraten:
# Changes
- [x] add UI to create, view, update (currently nothing to update) and remove users
- [x] remove active flag as it wasn't used

image


server/api/users.go
```diff
@@ -58,7 +58,8 @@ func PatchUser(c *gin.Context) {
 		c.AbortWithStatus(http.StatusNotFound)
 		return
 	}
-	user.Active = in.Active
+
+	// TODO: currently nothing is updatable
```
6543, line 62: in should be used and only the  `in.id = user.id` to make sure id is not changed ...
or if we allow partial changes... make a check if a field in `in` has a value and replace it here in `user` ...

PS: securety: we make sure only allowe user can change protected fields like Admin etc ...

anbraten: Sure. Currently users are updated by the oauth systems, so in 99% there currently is no need to update an user at all.

server/model/user.go
```diff
@@ -56,9 +56,6 @@ type User struct {
 	// the avatar url for this user.
 	Avatar string `json:"avatar_url" xorm:" varchar(500) 'user_avatar'"`
 
-	// Activate indicates the user is active in the system.
-	Active bool `json:"active" xorm:"user_active"`
```
6543, line 60: migration to rm this field?

server/api/users.go
anbraten: That's currently nothing as isAdmin is set by WOODPECKER_ADMINS and login, email, avatar are received from the forgea

server/model/user.go
anbraten: will add one

anbraten: done

server/api/users.go
lafriks: Would be nice if it was possible to set admin option from UI and env variable would be used for initial user creation

anbraten: Yeah definitely. I think we could add some code that sets `isAdmin` at server start for the listed users, but via the UI those setting could be changed later on.

anbraten: I have adjusted the code and now the admin flag is set to true when a user logs in to the UI.

lafriks (approved):

web/src/components/admin/settings/AdminUsersTab.vue
```diff
@@ -0,0 +1,161 @@
...
+
+        <InputField :label="$t('admin.settings.users.admin.admin')">
+          <Checkbox
+            :model-value="selectedUser.admin || false"
+            :label="$t('admin.settings.users.admin.placeholder')"
+            @update:model-value="selectedUser!.admin = $event"
+          />
+        </InputField>
+
+        <div class="flex gap-2 justify-center">
```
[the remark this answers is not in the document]
anbraten: moved it back to left side for now.

PR #1640, opened 2023-03-17
UI improvements

anbraten:
closes #1482
closes #1619

Cut long commit messages:
image

Decrease icon size:
image

Show duration for workflows (#1482):
image

PR #1641, opened 2023-03-18
Add PR pipeline list

### LINKED ISSUE #1619: PR pipelines are visible in the branches tab

If someone opens a PR to a branch (for example master), you can find the relevant pipeline under the target branch's section. For example:
image

This could cause some great confusion on which pipelines are run from commits verified by trusted people (people with push rights to the repo), and which are not. (I realize there's the PR icon on the pipeline, but most people won't notice it/won't know what it means.)

If something like #1577 merged, this could even lead to dangerous code being distributed to the target users. 
Example of abuse:
Users are instructed to download the artifact from the latest pipeline on branch `master`. 
A malicious person opens a PR to `master` containing malicious code.
Now users download the malicious binary when downloading the last artifact on branch `master`.

### Suggested solution

Somehow separate PRs (not sure how exactly ATM).

### Alternative

_No response_

### Additional context

_No response_

### Validations

qwerty287:
Instead of viewing PR pipelines in the branches lists, add a separate list for them. The API endpoint for PRs supports pagination (thus I added a lot of pagination-related stuff), the UI doesn't yet though.

wp

Closes #1619 

Extends this part of #1640


anbraten (approved):

pipeline/rpc/proto/woodpecker.pb.go
```diff
@@ -1326,31 +1327,33 @@ func file_woodpecker_proto_rawDescGZIP() []byte {
 	return file_woodpecker_proto_rawDescData
 }
 
-var file_woodpecker_proto_msgTypes = make([]protoimpl.MessageInfo, 22)
```
6543, line 1329: please do `go generate ./...`

server/forge/bitbucket/bitbucket.go
```diff
@@ -302,6 +303,10 @@ func (c *config) BranchHead(_ context.Context, _ *model.User, _ *model.Repo, _ s
 	return "", forge_types.ErrNotImplemented
 }
 
+func (c *config) PullRequests(_ context.Context, _ *model.User, _ *model.Repo, _ *model.PaginationData) ([]*model.PullRequest, error) {
+	return nil, errors.New("Bitbucket does not support pull requests yet")
```
6543, line 307: forge_types.ErrNotImplemented would be the way to go :laughing:

6543: #1652 ;)

PR #1643, opened 2023-03-18
Various fixes and improvements

qwerty287:
- allow repo names to be case-insensitive
- improve backend error handling on DB get errors (record not found -> 404, else -> 500)
- replace magic numbers of http response codes
- unify the look and feel of cancel / save buttons on forms and view them in one line


server/api/cron.go
```diff
@@ -97,13 +97,13 @@ func PostCron(c *gin.Context) {
 		Branch:    in.Branch,
 	}
 	if err := cron.Validate(); err != nil {
-		c.String(400, "Error inserting cron. validate failed: %s", err)
+		c.String(http.StatusBadRequest, "Error inserting cron. validate failed: %s", err)
```
lafriks, line 100: Validation should result in 422 error (400 should be when data is unparsable etc)

lafriks (changes requested):

lafriks (approved):

PR #1644, opened 2023-03-18
Drop coding support

qwerty287:
Coding support is likely broken and nobody will ever fix it. Also it looks like nobody wants to use it, otherwise we would have get some bug reports.


anbraten (approved):

lafriks:
Lint failures

6543 (approved):

anbraten:
@qwerty287 Could you add a note to the migration guide in the docs. That it got removed.

Issue #1131, opened 2022-08-25
Support SSH auth with key and password


qwerty287:
### Clear and concise description of the problem

Currently only one is supported because we didn't how goph handles this, but I opened https://github.com/melbahja/goph/issues/36 and now they responded: https://github.com/melbahja/goph/issues/36#issuecomment-1227478797. Also see https://github.com/woodpecker-ci/woodpecker/pull/861#discussion_r841204609 for the original discussion.

### Suggested solution

If both WOODPECKER_BACKEND_SSH_PASSWORD and WOODPECKER_BACKEND_SSH_KEY are set, use the method that was suggested in the issue above.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

Issue #896, opened 2022-04-29
Add flags for backend options


anbraten:
We added some backend options like `WOODPECKER_BACKEND_SSH_ADDRESS` in #893 and #861 which are accessed by  `os.Getenv` directly. As the backend can be executed by the agent and by the cli exec command those settings need to be  available for both and should be listed in the documentations of both systems. 

To properly use them with urfave cli we should think about placing a flags.go file into each backend package and include this in the cli exec and agent options. This would allow us to see those options in the help of both tools and auto document them.

PR #1645, opened 2023-03-18
Various enhancements in configuration

qwerty287:
- backends: move to cli flags instead of os.Getenv
- ssh: support 2fa with key and password
- allow to set grpc jwt secret (solves todo)
- allow to set default and max timeout (solves todo)

Closes https://github.com/woodpecker-ci/woodpecker/issues/896
Closes https://github.com/woodpecker-ci/woodpecker/issues/1131


cmd/agent/flags.go
```diff
@@ -108,7 +108,50 @@ var flags = []cli.Flag{
 		Value:   "auto-detect",
 	},
 
-	// TODO: add flags of backends
+	// backend docker
```
lafriks, line 111: can these flags be moved to engine (function in interface that would return them)? And than append them by looping all engines and calling that func?

6543: I had a similar idea what if we could move backend flag declaration into pipeline/backend/x/flags.go ...

but if we want to have a simple codebase for now we have to do it as is ... and I might come up with an elegant idea later :)

cmd/server/flags.go
```diff
@@ -120,6 +126,18 @@ var flags = []cli.Flag{
 		Usage:   "The default docker image to be used when cloning the repo",
 		Value:   constant.DefaultCloneImage,
 	},
+	&cli.Int64Flag{
+		EnvVars: []string{"WOODPECKER_DEFAULT_PIPELINE_TIMEOUT"},
+		Name:    "default-pipeline-timeout",
+		Usage:   "The default time for a repo in minutes before a pipeline gets killed",
```
6543, line 132: minutes seconds ...?

6543, line 132:
```suggestion
		Usage:   "The default time in minutes for a repo in minutes before a pipeline gets killed",
```

pipeline/backend/docker/docker.go
```diff
@@ -41,9 +41,6 @@ type docker struct {
 	volumes    []string
 }
 
-// make sure docker implements Engine
```
6543, line 44: why rm?

pipeline/backend/local/local.go
```diff
@@ -47,9 +47,6 @@ type local struct {
 	workingdir string
 }
 
-// make sure local implements Engine
```
6543, line 50: why rm?

pipeline/backend/ssh/ssh.go
```diff
@@ -29,9 +28,6 @@ func (c readCloser) Close() error {
 	return nil
 }
 
-// make sure local implements Engine
```
6543, line 32: same

cmd/agent/flags.go
anbraten: Do we need to know which backend is active / selected to get those flags. Couldn't we just get a list of all available options from each backend and add them to the list.

6543: if we just import things ...
we get:
 - import circles and
 - https://woodpecker-ci.org/docs/next/development/architecture get violated

6543 (approved):

PR #1647, opened 2023-03-18
Fix linter

anbraten:


lukashass (approved):

woodpecker-go/woodpecker/client.go
```diff
@@ -530,9 +530,9 @@ func (c *client) post(rawurl string, in, out interface{}) error {
 }
 
 // helper function for making an http PUT request.
-func (c *client) put(rawurl string, in, out interface{}) error {
-	return c.do(rawurl, "PUT", in, out)
-}
+// func (c *client) put(rawurl string, in, out interface{}) error {
```
6543, line 533: i would just delete helper functions that are that easy ...

PR #1628, opened 2023-03-16
Golang 1.20

6543:
Address most of #1573

PR #1352, opened 2022-10-27
Only inject netrc to trusted clone plugins

6543:

PR #1650, opened 2023-03-19
Update Go dependencies and minimal Go version to 1.20

lafriks:
I did not update docker dependencies to v23.0.1 as I did not know implications of that

Merge in changes and replaces #1628


6543 (approved):

pipeline/stepBuilder.go
```diff
@@ -279,9 +279,9 @@ func (b *StepBuilder) toInternalRepresentation(parsed *yaml.Config, environ map[
 		compiler.WithSecret(secrets...),
 		compiler.WithPrefix(
 			fmt.Sprintf(
-				"wp_%d_%d",
+				"wp_%s_%d",
```
lafriks, line 282: @6543 I changed to ULID so that volumes could be easily sortable by age

6543: oh not bad

6543: I did saw that you changed the lib and it looked nice but I did not notice that property

lafriks:
Ok, I have no idea why migration tests are failing 😕

6543:
well that's a special case where the DBMS do not clean up the old index after table drop ... before we do a create table again ...

6543:
on old golang it just happend 1 in 100 cases ... so I just restarted pipeline if it happens but now it looks like we need a better solution

lafriks:
@6543 what do you think about disabling that test for postgres and mysql until we find solution to make it stable?

6543:
I'm against it as we dont have to rush to get go1.20

6543:
and we would have missed e.g. https://github.com/woodpecker-ci/woodpecker/pull/1352/commits/7030cc19749771d4ca0b9b1dc45f55096a4d5920 ...

lafriks:
> and we would have missed e.g. [7030cc1](https://github.com/woodpecker-ci/woodpecker/commit/7030cc19749771d4ca0b9b1dc45f55096a4d5920) (#1352) ...

wouldn't that be catch by sqlite migration test?

lafriks:
Looks to me this should be good to go now

6543 (approved):

PR #1654, opened 2023-03-19
gofump don't touch code generated files

6543:
`make format` did touch code generated files witch creates a mess from time to time ...

this change it ...


6543:
It's not well documented ... so I did open an upstream issue here first: https://github.com/mvdan/gofumpt/issues/265

lafriks (approved):

PR #1655, opened 2023-03-19
Update frontend dependencies

lafriks:
Also fixes bug that wrong message is displayed when updating user


6543 (approved):

web/src/components/admin/settings/AdminUsersTab.vue
```diff
@@ -118,14 +118,18 @@ const { doSubmit: saveUser, isLoading: isSaving } = useAsyncAction(async () => {
 
   if (isEditingUser.value) {
     await apiClient.updateUser(selectedUser.value);
+    notifications.notify({
```
anbraten, line 121: @lafriks Thanks for fixing those messages. Could you maybe also correct the spacing of the admin badge / buttons for me 😅 : 

image

lafriks: they do look ok to me:
attels

PR #1659, opened 2023-03-20
Use full qualifyer for images and user quay.io if possible

6543:
& [bump alpine to alpine:3.17](https://github.com/woodpecker-ci/woodpecker/pull/1659/commits/588f493c47f4660d72ba76c9f1df3ea05f37bbd5)


shared/constant/constant.go
```diff
@@ -16,10 +16,11 @@ package constant
...
 var PrivilegedPlugins = []string{
-	"plugins/docker",
-	"plugins/gcr",
-	"plugins/ecr",
-	"woodpeckerci/plugin-docker-buildx",
+	"docker.io/plugins/docker",
+	"docker.io/plugins/gcr",
+	"docker.io/plugins/ecr",
+	"docker.io/woodpeckerci/plugin-docker-buildx",
+	"quay.io/woodpeckerci/plugin-docker-buildx",
```
lafriks, line 23: maybe add also codeberg?
```suggestion
	"quay.io/woodpeckerci/plugin-docker-buildx",
	"codeberg.org/woodpecker-plugins/docker-buildx",
```

.woodpecker/test.yml
```diff
@@ -46,12 +46,13 @@ pipeline:
...
 
   lint-editorconfig:
-    image: mstruebing/editorconfig-checker
+    image: docker.io/mstruebing/editorconfig-checker
     group: test
 
   securitycheck:
     group: test
-    image: aquasec/trivy:latest
+    pull: true
```
anbraten, line 54: Is this really needed as it creates a lot of traffic?

.woodpecker/test.yml
```diff
@@ -103,7 +104,7 @@ pipeline:
 
   codecov:
     pull: true
```
anbraten, line 106: Should probably be removed as well.

.woodpecker/test.yml
6543: other option would be to add a cron job to keep the latest tagged images up to date ...

lafriks: Pull won't repull if actual latest image is already cached just check if one exists, while it is some additional traffic I don't think it's that big

pat-s:
Using the FQ for `docker.io `is quite uncommon as its the default anyhow. 

quay for rate limiting or what's the motivation?

6543:
yes it was created while Docker Inc. try to remove OpenSource orgs

I still think we should still provide an other option as docker-hub. but it's not as urgent anymore

6543:
we first need ->  #1995

pat-s:
> I still think we should still provide an other option as docker-hub

Nothing to say against this but I think it's not necessary to add `docker.io` everywhere explicitly.

xoxys:
Conflicts resolved in https://github.com/woodpecker-ci/woodpecker/pull/2692

PR #1661, opened 2023-03-21
fix(backend/kubernetes): Ensure valid naming of name field

NeilHanlon:
- Kubernetes v1.26 on VKE causes error when creating persistent volume claim because of uppercase characters in name field

This patch is trivial just in order to get it working - happy to implement differently.

The error in question:

```
The PersistentVolumeClaim "wp-01G1131R63FWBSPMA4ZAZTKLE-0-clone-0" is invalid: metadata.name: Invalid value: "wp-01G1131R63FWBSPMA4ZAZTKLE-0-clone-0": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```


pipeline/backend/kubernetes/utils.go
```diff
@@ -10,8 +12,16 @@ import (
...
 
+var dnsPattern = regexp.MustCompile(`[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`)
+
 func dnsName(i string) string {
-	return strings.Replace(i, "_", "-", -1)
+	res := strings.Replace(strings.ToLower(i), "_", "-", -1)
+
+	match := dnsPattern.MatchString(res)
+	if !match {
+		log.Panic("name not valid kubernetes DNS name")
```
6543, line 22: dont Panic make the dnsName func return an error to report up the chain ...

and create a new error type .. `var Name = errors.New(...)`

pipeline/backend/kubernetes/utils.go
```diff
@@ -10,8 +12,16 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+var dnsPattern = regexp.MustCompile(`[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`)
```
6543, line 15:
```suggestion
var (
  dnsPattern = regexp.MustCompile(`[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`)
  ErrDnsPatternInvalid = errors.New("name not valid kubernetes DNS name")
)
```
:point_up: this + `make format`

pipeline/backend/kubernetes/utils.go
6543, line 22:
```suggestion
		return "", ErrDnsPatternInvalid
```

6543:
also adding a smal unit test might be nice :)

if you need help just tell

NeilHanlon:
Thank you for the comments! I've introduced a new commit which addresses the review items.

lafriks:
ULID that is generated it volume name can be lowercased as ULIDs are case insensitive

PR #1672, opened 2023-03-23
Global and organization registries

lafriks:
Adds way to add server wide and organization wide registries

* [x] Database and store changes
* [ ] Server API changes
* [ ] Organization settings UI
* [ ] Server settings UI


lafriks:
Should be ready for review now

lafriks:
lint failure not related to my changes

```
server/store/datastore/migration/common.go:198:64: `alterColumnNull` - `null` always receives `false` (unparam)
func alterColumnNull(sess *xorm.Session, table, column string, null bool) error {
```

server/router/api.go
```diff
@@ -61,11 +61,18 @@ func apiRoutes(e *gin.RouterGroup) {
...
 					org.DELETE("", session.MustAdmin(), api.DeleteOrg)
 					org.GET("", api.GetOrg)
+
 					org.GET("/secrets", api.GetOrgSecretList)
 					org.POST("/secrets", api.PostOrgSecret)
 					org.GET("/secrets/:secret", api.GetOrgSecret)
 					org.PATCH("/secrets/:secret", api.PatchOrgSecret)
 					org.DELETE("/secrets/:secret", api.DeleteOrgSecret)
+
+					org.GET("/registry", api.GetOrgRegistryList)
```
anbraten, line 71: Most other endpoints  are using the plural version. Therefore I would use `registries` here as well.

server/store/datastore/migration/032_registries_add_user.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"src.techknowlogick.com/xormigrate"
+	"xorm.io/xorm"
+)
+
+// Registry represents a docker registry with credentials.
+type RegistryV032 struct {
+	ID       int64  `json:"id"       xorm:"pk autoincr 'id'"`
+	OrgID    int64  `json:"org_id"   xorm:"NOT NULL DEFAULT 0 UNIQUE(s) INDEX 'org_id'"`
```
anbraten, line 25: Shouldn't this be the old version without OrgID?

web/src/lib/api/types/registry.ts
```diff
@@ -3,4 +3,5 @@ export interface Registry {
   address: string;
```
anbraten, line 3:
```suggestion
  address: string;
  repo_id: string;
  org_id: string;
```

web/src/views/admin/AdminSettings.vue
```diff
@@ -9,6 +9,9 @@
     <Tab id="secrets" :title="$t('secrets.secrets')">
       <AdminSecretsTab />
     </Tab>
+    <Tab id="registries" :title="$t('registries.registries')">
+      <AdminRegistriesTab />
```
anbraten, line 13: Component needs to be imported

web/src/views/repo/RepoWrapper.vue
```diff
@@ -9,8 +9,10 @@
       <span class="flex">
         <router-link :to="{ name: 'org', params: { orgId: repo.org_id } }" class="hover:underline">{{
           repo.owner
+          /* eslint-disable-next-line @intlify/vue-i18n/no-raw-text */
         }}</router-link>
-        {{ `&nbsp;/&nbsp;${repo.name}` }}
+        &nbsp;/
```
anbraten, line 14: had a space after the `/` before?!

cli/registry/registry.go
```diff
@@ -30,3 +37,47 @@ var Command = &cli.Command{
...
+		return true, -1, -1, nil
+	}
+
+	repoIDOrFullName := c.String("repository")
+	if repoIDOrFullName == "" {
+		repoIDOrFullName = c.Args().First()
+	}
+
+	orgIDOrName := c.String("organization")
+	if orgIDOrName == "" && repoIDOrFullName == "" {
```
anbraten, line 52: What do you think about using sub-commands like `woodpecker-cli org regsitry ...` to have a similar "path" like the UI where you also go to org settings to change an orgs registry. This would also heavily improve navigation through shell completion IMO.

(I know this is done for secrets like this and would need to be refactored as well)

server/store/datastore/migration/032_registries_add_user.go
lafriks: No, it does creates new table version

server/router/api.go
lafriks: At first I started to use registries but later changed to registry as repo registry endpoint was already using `registry`

web/src/lib/api/types/registry.ts
lafriks: these are not used in UI anyway

web/src/views/admin/AdminSettings.vue
lafriks: nice catch, thanks 👍🏻

cli/registry/registry.go
lafriks: this should probably be part of bigger refactoring than

server/router/api.go
lafriks: @anbraten should I rename existing repo reigstry endpoints as well than?

anbraten: I would suggest we rename all to `/registries` and leave the `/registry` routes with a TODO to remove them in 3.x

server/store/datastore/migration/032_registries_add_user.go
anbraten: AFAIK normally we sync the old state of the model in migrations to ensure the are synced. Then the migration is executed and afterwards the new model is syned in https://github.com/woodpecker-ci/woodpecker/blob/e5f3e67bf282bd938b5ab39295f738e2fefa7d4a/server/store/datastore/migration/migration.go#L119

cli/registry/registry.go
anbraten: Would like to have it done in that way directly to not have a breaking change. Might be best to extract the cli commands to a separate PR.

server/store/datastore/migration/032_registries_add_user.go
lafriks: seems hard to catch errors this way to be sure it's working when upgrading from different versions but ok, removed sync as I do not depend on it that much

cli/registry/registry.go
lafriks: @anbraten refactored CLI commands to subcommands

cli/repo/registry/registry.go
```diff
@@ -0,0 +1,59 @@
...
+		repoIDOrFullName = c.Args().First()
+	}
+
+	if repoIDOrFullName == "" {
+		if err := cli.ShowSubcommandHelp(c); err != nil {
+			return -1, err
+		}
+
+		return -1, fmt.Errorf("missing arguments")
+	}
```
anbraten, line 51: This should be part of `internal.ParseRepo` I guess as it has logic to determine the repo from the git remote inside.

anbraten (approved):

PR #1676, opened 2023-03-23
web: use i18n-t to avoid v-html directive

janydoe:
Use i18n-t component to avoid v-html directive
and to make localization easier


lafriks (approved):

lafriks:
please fix formatting https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/pipeline/5285/29

PR #1686, opened 2023-04-01
Set 'HOME' during local pipeline step

### LINKED ISSUE #1685: HOME unset in local pipeline step
### Component

agent

### Describe the bug

When running a local pipeline step, `echo $HOME` shows nothing.

I have a tool (`nix` in this instance) which needs to write to `HOME`, and therefore fails if it is unset as it tries to write to `/`.

Doing `HOME=$(mktemp -d) my command here...` however, works.

I think `HOME` should be set to a sensible value: either the `PWD` (i.e: the CI workspace) or a temporary directory (as-if created by `mktemp -d`).

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-e7ca2802"}
```

### Additional context

_No response_

### Validations

ambroisie:
close #1685


lafriks:
I don't think it is correct to set working dir as that could pollute it and have undesired side effects

6543:
I think both use-cases are valid ... so a flag would be best

6543:
well I think it also address https://github.com/woodpecker-ci/woodpecker/pull/1686/files#diff-a04c642ebae3a1409872e78fbebc32223305acdfb5ec9a67edf544176b000d82R89

but we have to:
 make the current workdir as pipeline-base-dir and use subfolders for work and home dir

lafriks (approved):

PR #1691, opened 2023-04-03
Access repos by their IDs

### LINKED ISSUE #648: support nested repositories (GitLab)

### LINKED ISSUE #1295: Access projects by their id
We currently use the repository owner and repository name in api calls and in the UI as part of the url:

```
https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker
https://ci.woodpecker-ci.org/api/repos/woodpecker-ci/woodpecker/branches
```

Although this is nice for humans it has some limitations:

- Gitlab has support for nested repositories / sub-projects which would require multiple levels #648 
- Woodpecker is limited to a single remote as the remote is currently not part of the url and therefore it currently can not by design support multiple forges at once #138

# Current proposal

Refactor the UI and api urls to use the id of the Woodpecker project instead.

# Problems

This would be a breaking change and we would need to thing about adding redirects.

anbraten:
closes #1295 
closes #648

# TODO
- [x] add new routes with `:repoID`
- [x] load repo in middleware using `:repoID` if present
- [x] update UI routes `:owner/:name` to `:repoID`
- [x] load repos using id in UI
- [x] add lookup endpoint `:owner/:name` to `:repoID`
- [x] redirect `:owner/:name` to `:repoID` in UI
- [x] use badge with `:repoID` route in UI
- [x] update `woodpecker-go`
- [x] check cli
- [x] add migrations / deprecation notes
- [x] check if #648 got solved directly
- [x] Test
  - [x] create repo
  - [x] repo pages
  - [x] ui redirects
  - [x] forge status links


server/api/user.go
```diff
@@ -70,20 +70,23 @@ func GetRepos(c *gin.Context) {
...
 			c.String(http.StatusInternalServerError, "Error fetching repository list. %s", err)
 			return
 		}
+
 		var repos []*model.Repo
 		for _, r := range _repos {
 			if r.Perm.Push {
-				if active[r.FullName] {
+				if active[r.ForgeRemoteID] != nil && active[r.ForgeRemoteID].IsActive {
+					// TODO: should we merge other fields as well?
```
anbraten, line 88: @qwerty287 I had to change this matching a bit to be independent of the full-name and copied the existing id in case the repo already exists. Should we merge all other "internal"  values which are not set by the remote as well?

lafriks:
I think UI routes with org/name should NOT be deprecated and still used as primary (at least for gitea/github) as they makes it so much easier to use for multiple reasons:
* You can easily enter them manually if you know them or just change git domain to woodpecker and get to the build list
* In my company we heavily use README etc templates and badge URL can be pregenerated in readme as org/name is same as new repositories even before enabling it in woodpecker

6543:
☝️

anbraten:
I see you point and also thought how it could still be possible to access repos via their name, so I have added the redirect using the lookup endpoint. 
On the other hand I can not image that a lot of users are really entering the woodpecker repo url on their own. I would rather say 99% is using links from the status checks of the forge etc, the base url of woodpecker and navigating via UI or the browser history. 

There are some problems with the named routes which apply for all forges which are really annoying IMO. When renaming a repo / moving the org, the url changes and therefore all pointing urls in the status checks for example are broken, we would not be able to determine different forges and repos with special names containing slashes like gitlab subprojects are not supported.

lafriks:
I have no problem with gitlab behaving differently as it also uses repo id's in some of their own urls and I don't use it so I don't mind 😆 
I'm also fine with changing things on repo if I rename it. But currently our company uses a lot of templates like:
image

This would be very breaking change for me it these routes would be deprecated 😞

anbraten:
@lafriks Would leaving the badge endpoint supporting both versions and the UI redirect work for you (as it currently is solved in this PR)?

lafriks:
Why not leave it to work both ways without redirect? But that wont be deal breaker for me

anbraten:
I've tried supporting both at the beginning, but it became a mess and was pretty hard to handle. Biggest issue is the slash between owner and repo-name in the route which makes parameter matching in go gin and vue pretty difficult.

lafriks:
Maybe they could use different routes: `/<owner>/<repo>` and `/repos/<id>`?

anbraten:
Tried that at the beginning. One issue was that the links of the repo views need to distinguish pointing to the owner-repo or the repo-id route, so we would need to duplicate those views / add some routing helper or always use the repo-id in links again which would result in almost the same way as directly redirecting

server/api/user.go
anbraten: @qwerty287 Changed it to use the existing repo entry with updated values using our existing `repo.Update(from)` function

anbraten:
@lafriks I changed some parts and removed some deprecation plannings. Would these changes be fine for you:
- api routes via id only + lookup endpoint to find repo id
- badges via id and owner/name combination
- UI via id, owner/name combination will be redirected

server/store/datastore/pipeline.go
```diff
@@ -73,7 +73,7 @@ func (s storage) GetPipelineLastBefore(repo *model.Repo, branch string, num int6
 }
 
 func (s storage) GetPipelineList(repo *model.Repo, p *model.ListOptions) ([]*model.Pipeline, error) {
-	var pipelines []*model.Pipeline
+	pipelines := make([]*model.Pipeline, 0)
```
6543, line 76:
```suggestion
	pipelines := make([]*model.Pipeline, 0, 16)
```

6543 (approved):

PR #1692, opened 2023-04-03
Add backend label to agents

### LINKED ISSUE #708: Add condition to run pipeline only on agents with a specific backend
@ambroisie:

> I would really like to be able to select which backend is used on a pipeline basis instead, like one can on Drone.

maybe something like:

```yml
labels:
  backend: local

pipeline:
  [...]
```

## To be able to do this we can add auto-generated labels for agents such as:

- backend type (local, docker, ...)
- arch
- os

anbraten:
closes #708 

Allows users to filter for a specific backend type.


6543 (approved):
:heart:

PR #1697, opened 2023-04-05
Nomad Engine

cullenmcdermott:
This is still pretty WIP but I figured I would go ahead and create a draft PR for this. Basic functionality seems to work, however the code needs cleaned up quite a bit still.


pat-s:
@cullenmcdermott Are you still aiming to continue here?

6543:
just match the interface and look at the "reference impl. (docker)"

anbraten:
Before we continue I would like to wait for more requests by other users. Nomad seems TBH not to be the most used orchestrator and we should first of all make sure there is a real need for this before spending too much time implementing it.

anbraten (changes requested):
Before we continue I would like to wait for more requests by other users. Nomad seems TBH not to be the most used orchestrator and we should first of all make sure there is a real need for this before spending too much time implementing it.

cullenmcdermott:
> Before we continue I would like to wait for more requests by other users. Nomad seems TBH not to be the most used orchestrator and we should first of all make sure there is a real need for this before spending too much time implementing it.

Sounds good. I'll keep an eye on this and will be happy to continue working on it if there's more desire for this one.

livioribeiro:
I am deploying a nomad cluster at the company I work for and this would be very nice for our ci/cd

pat-s:
> Since it seems there are a few people thinking this would be useful and we decided to include new backends if they are not too specific - which isn't the case here - I think it's fine to integrate this.

Are you basing your decision on the 4 upvotes from the comment above?

I'd be more hesitant here. Nomand is built on top of k8s and hence essentially a special k8s backend - in WP terms. It will always be very niche and the amount of work to maintain it must currently come from outside since I don't think any contributor/maintainer is actively using it. Which is why I am not sure if this should be included/supported officially.

pat-s:
How should this PR be handled going forward? I'd actually like to close it as it is stale and I don't see who would continue/maintain this here. Opinions?

PR #1707, opened 2023-04-18
Use arg for docker image versions

6543:
just wrote down my idea ...

-> allow us to bump the go/alpine version just by changing a single line in .woodpecker/docker.yml


docker/Dockerfile.make
```diff
@@ -1,5 +1,5 @@
-# docker build --rm  -f docker/Dockerfile.server -t woodpeckerci/woodpecker-server .
-FROM golang:1.20-alpine as golang_image
+ARG GO="1.20"
```
6543, line 1:
```suggestion
ARG GO="latest"
```

docker/Dockerfile.cli.alpine.multiarch
```diff
@@ -1,4 +1,6 @@
-FROM --platform=$BUILDPLATFORM golang:1.20 AS build
+ARG GO="latest"
+ARG ALPINE="3.16"
```
6543, line 2: should we use latest in alpine too and only pin by ci config?

pat-s:
I like the approach. However, I would go with the latest versioned releases as otherwise the version bumps slide in unnoticed and might break things.
With proper versions they at least have to go through CI once. 

Alpine 3.18
Go 1.20?

(can't push to your fork)

6543:
@pat-s I have to infite you to my repo - it's an github limitation :/

pat-s:
> I have to infite you to my repo - it's an github limitation :/

Two questions: 

1. Why do you use a fork?
2. The option "allow edits by maintainers" usually grants write access to branches of PRs

Thanks for the invite to the fork but this shouldn't be necessary 😄

6543:
It's github's fault ... well I can't change that

.woodpecker/docker.yml
```diff
@@ -16,6 +16,7 @@ variables:
   - &platforms_server 'linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le,linux/riscv64'
   - &platforms_preview 'linux/amd64'
   - &platforms_alpine 'linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le'
+  - &build_args 'GO=1.20 ALPINE=3.16'
```
6543, line 19:
```suggestion
  - &build_args 'GO=1.20 ALPINE=3.18'
```

docker/Dockerfile.make
6543: latest does not work ...

docker/Dockerfile.server.multiarch
```diff
@@ -1,4 +1,5 @@
-FROM --platform=$BUILDPLATFORM golang:1.20 AS certs
+ARG GO="latest"
```
pat-s, line 1:
```suggestion
ARG GO="1.20"
```

pat-s:
I thought about the 'fork' approach again but I don't really want to add more git remotes and then always having to take to push to the correct remote 😅 

For collab purposes I think it would be better to either have "allow edits from maintainers" (but this requires a personal account with permissions and no org) or a direct branch-based PR in this repo.

PR #1713, opened 2023-04-22
Fully support `.yaml`

qwerty287:
Follow-up to https://github.com/woodpecker-ci/woodpecker/pull/1388
Closes https://github.com/woodpecker-ci/woodpecker/issues/1073

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -81,7 +81,7 @@
                 "success": "Repository settings updated",
                 "pipeline_path": {
                     "path": "Pipeline path",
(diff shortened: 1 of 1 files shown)


6543 (changes requested):
I acknowledge the consensus, that we should support yaml too ...

But I dont see why we should do a breaking change in the order and make yaml default over yml ?!?

6543:
@anbraten your opinion?

lafriks:
> I acknowledge the consensus, that we should support yaml too ...
> 
> But I dont see why we should do a breaking change in the order and make yaml default over yml ?!?

I don't think it's being breaking for many as there is already something very wrong if you have both `.woodpecker.yml` and `.woodpecker.yaml` in project

6543 (approved):

PR #1720, opened 2023-04-29
Implement YAML Map Merge, Overrides, and Sequence Merge Support

### LINKED ISSUE #1192: Support YAML override and extension
you used lib does not provide this by default: https://github.com/yaml/yaml/issues/35

so we need to look into writing an extention for it if posible or add a post-proccessor

6543:
close  #1192


6543:
https://ci.codeberg.org/test_it/CI_Sequence_Merges/pipeline/2

https://ci.codeberg.org/test_it/CI_Map_Merges_Overwrites

That's done already ;)

Issue #1668, opened 2023-03-22
Add Manual workflow Trigger


tiny6996:
### Clear and concise description of the problem

I would like to be a able to run only certian workflows in a deployment without updating code in a repository.

### Suggested solution

Add a an option in the yaml file or a UI option that requires each workflow be triggered manually.

### Alternative

* skip workflows as described in #1354
* skip 

### Additional context

I would like to see manual workflow steps added so I can have a 1 work flow per ansible role I would like to deploy. Then start a build manually to deploy that role whenever without updating git. This would function to manual steps in bitbucket pipelines https://bitbucket.org/blog/pipelines-manual-steps-confidence-deployment-pipeline

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

6543:
hmm you can already skip by filter out event "manual" for all workflows not intended to run

I don't think we should make a selector for workflows ... if at all, an option to specify the config path (witch would result in same functionality)

6543:
-> #1722

anbraten:
You can pass variables already using the manual workflow. Such a variable could be the ansible role / filter. Or you could pre-define all ansible tasks you want to execute and add when filters like:

```yml
when:
  - evaluate: 'MY_ROLE == "openssh"'
```

tiny6996:
How would I define `MY_ROLE` I updated one of my files in the .woodpecker folder and I am getting an error saying 

```
Execution error
unknown name ROLES (1:1) | ROLES contains "kics" | ^
```

when roles defined as a secret and and as a variable when I run the pipeline manually? I am guess [this](https://woodpecker-ci.org/docs/next/usage/pipeline-syntax#evaluate) section of the docs is the right place to look

```
clone:
  git:
    image: woodpeckerci/plugin-git
    settings:
      skip_verify: true
pipeline:
  KICS Scan:
    image: checkmarx/kics:latest
    commands:
      - kics scan --no-progress -p . --disable-secrets --report-formats json --output-name kics-results
    when:
[...]

PR #1722, opened 2023-04-29
WIP: Add option to specify event and configuration path when creating pipeline via API and CLI

6543:
Partially address  #1668 (UI missing)


server/forge/configFetcher_test.go
```diff
@@ -526,6 +527,7 @@ func TestFetchFromConfigService(t *testing.T) {
 				&model.User{Token: "xxx"},
 				repo,
 				&model.Pipeline{Commit: "89ab7b2d6bfb347144ac7c557e638ab402848fee"},
+				"",
```
6543, line 530:
```suggestion
				tt.repoConfig,
```

anbraten (changes requested):
I am kind of against "faking" pipeline events / configurations! This would for example allow you to "create" a pull-request without actually doing it. Even if this is only doable by an admin its confusing IMO to create events this way!

anbraten:
Closing as this introduces risks as discussed in the maintainer channel recently.

PR #1725, opened 2023-04-30
Add UI to use new manual pipeline create options

6543:
based on  #1722

image
image


web/src/components/layout/popups/ManualPipelinePopup.vue
```diff
@@ -72,9 +82,17 @@ const repo = inject('repo');
 
 const router = useRouter();
 const branches = ref<{ text: string; value: string }[]>([]);
-const payload = ref<{ branch: string; variables: Record<string, string> }>({
+const events = ref<{ text: string; value: string }[]>(
+  Object.values(WebhookEvents).map((e) => ({
+    text: e,
```
6543, line 87: do we have a func that can convert e into localized text ?

anbraten:
see https://github.com/woodpecker-ci/woodpecker/pull/1722#issuecomment-2009742016

PR #1731, opened 2023-05-01
Manually skip workflow and its depending ones

### LINKED ISSUE #1354: Allow to manually skip (waiting) workflows

qwerty287:
Supersedes #1463 
Closes #1354


anbraten (changes requested):
Could you explain your use-case again. One issue I see with allowing such things is that you could bypass the CI. Especially if working in Teams this is a real issue as admins / users could skip workflows needed to ensure quality without editing the config and therefore without sticking to flows like reviews. For example in the Woodpecker project this way a maintainer could easily bypass a check without fixing the real issue.

6543:
well only repo admins should be allowed to skip workflows ... and normal users with rite acces should just be able to cancle ...

6543:
or we just report cancled and all is fine?

server/pipeline/pipelineStatus.go
```diff
@@ -60,3 +60,10 @@ func UpdateToStatusKilled(store model.UpdatePipelineStore, pipeline model.Pipeli
 	pipeline.Finished = time.Now().Unix()
 	return &pipeline, store.UpdatePipeline(&pipeline)
 }
+
+func UpdateToStatusSkipped(store model.UpdatePipelineStore, pipeline model.Pipeline, reviewer string) (*model.Pipeline, error) {
```
6543, line 64: never used?

server/queue/queue.go
```diff
@@ -83,6 +83,8 @@ type Queue interface {
 	// EvictAtOnce removes a pending task from the queue.
 	EvictAtOnce(c context.Context, id []string) error
 
+	EvictCurrent(c context.Context, id string, exitStatus model.StatusValue) error
```
6543, line 86: please add code comment of what it does here too :)

6543:
Well lets just mae it a server config option and let id be disabled by default

6543:
well it depends if the cancel let depending pipelines will going to run or if they are sciped

lafriks:
Question is if part of the pipeline is skipped and when it finishes does whole pipeline status remains as canceled or success?

server/api/pipeline.go
```diff
@@ -576,3 +576,38 @@ var deleteStr = `[
 		"out": "logs purged by %s on %s\n"
 	}
 ]`
+
+// PostSkipWorkflow
+//
+//	@Summary	Skip a workflow
```
6543, line 582:
```suggestion
//	@Summary	Skip a workflow and it's depending ones
```

6543 (approved):

6543:
also tested - works as proposed

server/model/step.go
```diff
@@ -63,7 +63,7 @@ func (p *Step) Running() bool {
 
 // Failing returns true if the process state is failed, killed or error.
 func (p *Step) Failing() bool {
-	return p.State == StatusError || p.State == StatusKilled || p.State == StatusFailure
+	return p.State == StatusError || p.State == StatusKilled || p.State == StatusFailure || p.State == StatusSkipped
```
lafriks, line 66: Wouldn't this also make skipped by condition pipelines mark as failed?

lafriks (changes requested):

6543:
ok what's the current review proccess of you?
cc @anbraten @lafriks

server/store/datastore/workflow.go
```diff
@@ -69,3 +69,11 @@ func (s storage) WorkflowUpdate(workflow *model.Workflow) error {
 	_, err := s.engine.ID(workflow.ID).AllCols().Update(workflow)
 	return err
 }
+
+func (s storage) WorkflowFind(pipeline *model.Pipeline, pid int) (*model.Workflow, error) {
+	wf := &model.Workflow{
+		PipelineID: pipeline.ID,
+		PID:        pid,
+	}
+	return wf, wrapGet(s.engine.Get(wf))
```
6543, line 78: explizite when statement instead of implizite query

6543 (changes requested):

server/store/datastore/workflow.go
6543: and tests for that feature

anbraten:
I am still unsure how we can prevent pipeline bypassing after adding this feature. In case of Github we would have to define all required status-checks (which for example in the woodpecker projects we are not atm). However that would be possible, but in case of Bitbucket and Gitlab we don't have such options:
image

Sending a failed status-check in case of skipping a workflow, would be solving this, but I guess that would not help with the intention of this PR.

anbraten:
Currently it creates pending status-checks for all workflows after pushing. To merge the PR there has to be at least one status-check so you can not bypass even if Woodpecker would be offline etc. If I currently cancel a complete pipeline it is marked as failed and therefore it prevents the merge.

anbraten:
Could we maybe add an `optional` setting to workflows which would allow skipping a workflow and normally it would not be allowed to skip any workflows?

anbraten:
> Ah, I see. You're right, sending failed status would solve it, but that's not the idea. I think it's fine though to send a failed status for every skipped pipeline (which would also include pipelines that didn't start when canceling). What do you think about this?

Sending failed status for skipped workflows would be fine for the security aspect. Could you explain how you plan to use the skipping of workflows then? 

So for example with Woodpecker you would start a pipeline by opening a PR. You wait for the tests, but the docs and UI workflows you don't care about and would skip it. Push changes to the go code until you have finished it and at the end you would run all workflows to be able to merge?

PR #1732, opened 2023-05-01
Identify users using their remote ID

qwerty287:
do not use login name as primary ID when logging in

This can be extended later to support changing usernames etc.


server/store/datastore/user.go
```diff
@@ -16,16 +16,31 @@ package datastore
...
+	"xorm.io/xorm"
 )
 
 func (s storage) GetUser(id int64) (*model.User, error) {
 	user := new(model.User)
 	return user, wrapGet(s.engine.ID(id).Get(user))
 }
 
+func (s storage) GetUserRemoteID(remoteID model.ForgeRemoteID, login string) (*model.User, error) {
+	sess := s.engine.NewSession()
```
anbraten, line 28: Why do you need a new session here?

anbraten (commented):
Nice change.

anbraten (approved):

PR #1737, opened 2023-05-04
Kubernetes: Add `nodeSelector` support

### LINKED ISSUE #1736: Kubernetes: `platform` does not guarantee correct scheduling
### Component

agent

### Describe the bug

When setting `platform: <any arch>` with the Kubernetes backend, the resulting jobs are not placed on the respective nodes.

For this to happen the platform arch must most likely be translated into a `nodeSelector` for the upcoming pod but this field is not set.

When running a mixed-arch cluster, this makes WP practically unusable :(

### System Info

```shell
k8s 1.26

Version: `next-049c2622`
```

### Additional context

related to #platform

### Validations

pat-s:
fix #1736 

This is a very naive and initial attempt to add support for a mapping of the `platform` setting in the WP YAML to the `nodeSelector` field in the respective pod spawned by the WP agent.

My go knowledge is very basic and hence everything in here could be simply wrong.

~~Something that's still missing is the logic how to read the `platform` input from the YAML and pass it to the k8s backend in `kubernetes.go`. All other vars there are being read from the CLI or env vars but none are from the YAML.
@6543 @anbraten I guess it might be quite easy for you to add the last steps for this? Feel free to push to this branch!~~


pat-s:
@anbraten @6543 It feels like I am not far away from a working solution but I don't know how I get the `platform` information from the pipeline definition into the `pod` struct. Can you help me here?

pat-s:
It's working as desired. It might not be complete and the `NodeSelector` object can certainly be filled with more possible inputs but that might be something for additional PRs.

It solves the urgent need for me to run WP on a mixed-arch cluster.

pat-s:
@anbraten Looks like an approval is needed for auto-merge :)

anbraten (approved):

PR #1747, opened 2023-05-12
Improve `skip_clone` documentation

6543:
address https://discord.com/channels/838698813463724034/838698813463724037/1106545516457754665

### DIFF
--- a/docs/docs/20-usage/20-pipeline-syntax.md
+++ b/docs/docs/20-usage/20-pipeline-syntax.md
@@ -727,6 +727,15 @@ pipeline:
   ...
 ```
 
+## `skip_clone`
+
+If you don't specify (a) clone step(s), a default one is added.
+So if you like to not have one at all, just specify:
+
+```yaml
+skip_clone: true
+```
+
 ## `when` - Global pipeline conditions
 
 Woodpecker gives the ability to skip whole pipelines (not just steps #when---conditional-execution-1) based on certain conditions by a `when` block. If all conditions in the `when` block evaluate to true the pipeline is executed, otherwise it is skipped, but treated as successful and other pipelines depending on it will still continue.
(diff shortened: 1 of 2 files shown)


docs/docs/20-usage/20-pipeline-syntax.md
```diff
@@ -727,6 +727,15 @@ pipeline:
   ...
 ```
 
+## `skip_clone`
+
+If you don't specify (a) clone step(s), a default one is added.
+So if you like to not have one at all, just specify:
```
6543, line 733:
```suggestion
If you don't specify clone steps, a default one is added.
So if you like to not have one at all, just specify:
```
?

anbraten:
```suggestion
By default Woodpecker is automatically adding a clone step. This clone step can be configured by the [clone](#clone) property. If you do not need a `clone` step at all you can skip it using:
```

anbraten: @qwerty287 As far as I know you could add steps to the pipeline area with one of the trusted clone images which would get the netrc credentials passed which is possible since #1352 (pretty cool IMO and somehow makes the clone area obsolete to be fair)

docs/docs/20-usage/25-workflows.md
```diff
@@ -124,19 +124,7 @@ depends_on:
...
-depends_on:
-  - deploy
-
-runs_on: [ success, failure ]
-+skip_clone: true
-```
+:::info
+Some workflows don't need the source code, like creating a notification on failure.
+Read more about `skip_clone` at [pipeline syntax](./20-pipeline-syntax.md#skip_clone)
+:::
```
anbraten, line 130: Do we even need this anymore after adding the `skip_clone` section? Maybe we can simply move the example to that section as well.

6543: I'd like the reminder here as most wont think of that while creating notification workflows i would say

anbraten: Never mind, missed that its a completely different file

PR #1751, opened 2023-05-13
Always send a status back to forge

qwerty287:


server/grpc/rpc.go
```diff
@@ -275,6 +275,8 @@ func (s *RPC) Init(c context.Context, id string, state rpc.State) error {
 		}
 	}
 
+	s.updateForgeStatus(c, repo, currentPipeline, step)
```
anbraten, line 278: This would set the status to pending? Would the changed status in line 297 be send to the forge as well?

anbraten (approved):

server/pipeline/cancel.go
```diff
@@ -84,20 +84,22 @@ func Cancel(ctx context.Context, store store.Store, repo *model.Repo, pipeline *
...
 
-	killedBuild, err := UpdateToStatusKilled(store, *pipeline)
+	killedPipeline, err := UpdateToStatusKilled(store, *pipeline)
 	if err != nil {
 		log.Error().Err(err).Msgf("UpdateToStatusKilled: %v", pipeline)
 		return err
 	}
 
-	steps, err = store.StepList(killedBuild)
+	updatePipelineStatus(ctx, pipeline, repo, user)
```
6543, line 93: is it intentional to use pipeline and not killedPipeline ?

PR #1752, opened 2023-05-13
Drop Gogs support

qwerty287:
Gogs support is broken (and we won't fix it because we don't care about it...) because it does not support OAuth, at least after we introduced the new Vue UI.

See:
https://github.com/woodpecker-ci/woodpecker/blob/77d830d5b56f85084a74fe2396689cef8b2cdee2/server/forge/gogs/gogs.go#L84

This route is not present in the new UI.


6543:
well it also lacks of api's we need, to make a equal integration to other forges

anbraten (approved):

6543 (approved):

PR #1753, opened 2023-05-14
Move securitychecks into own workflow

6543:
- [x] activate backend again
- [x] move

followups:
- make trivy server work
- create notifications


.woodpecker/securityscan.yml
```diff
@@ -0,0 +1,31 @@
...
+    group: check
+    image: *trivy_plugin
+    settings:
+      skip-dirs: web/,docs/
+
+  check docs:
+    group: check
+    image: *trivy_plugin
+    settings:
+      skip-dirs: node_modules/,plugins/woodpecker-plugins/node_modules
```
lafriks, line 21: For consistency
```suggestion
      skip-dirs: node_modules/,plugins/woodpecker-plugins/node_modules/
```

pat-s:
Sufficient to only run on `main`?

6543:
well we can do that once we get notification if something is wrong

pat-s:
I don't quite understand your reply. Can you rephrase? 😅 

My thinking was that a PR shouldn't introduce a sec risk usually - and these usually come by some existing old library in which an issue get's detected.
It's also a resources question: is it really needed to run this WF (for a possible minor gain) in every PR? Hence my proposal above.

6543:
I dont think 14sec are a burden

6543:
> Can you rephrase?

the idea of mine was, to let it only run as cron job and create a discord&matrix notification if it do detect something.
I don't want to do this atm as we would only spontaneously notice it.

But while reading your replay, I realized that we should still run on pulls too!

yes it's rare that there are sec issues newly introduced in pulls, but for <=15sec, it it worth to just scan for them.

being as secure as possible should be one of our goals.

It should just not block pulls, and this new workflow will do exactly that :)

.woodpecker/securityscan.yml
```diff
@@ -0,0 +1,31 @@
...
+      skip-dirs: web/,docs/
+
+  check docs:
+    group: check
+    image: *trivy_plugin
+    settings:
+      skip-dirs: node_modules/,plugins/woodpecker-plugins/node_modules/
+      dir:  docs/
+      # TODO currently it is not fixable so just do not block currently
+      exit-code: 0
```
6543, line 24:
```suggestion
```

PR #1763, opened 2023-05-16
Adding Android CI example to awesome woodpecker

dessalines:
This shows an example of how to use woodpecker to lint and build android projects.


6543:
@dessalines moved your entry and changed some bits ... if you are fine with it ... we can merge

dessalines:
Sure, go right ahead.

anbraten (approved):

Issue #567, opened 2021-12-04
Breaking changes before releasing 0.16.0


6543:
from #393

- pipeline config changes
  - [ ] remove support for `root.pipeline.[step].commands` string values in favor of array lists -> need more discussion
  - [ ] move `root.platform`, `root.branches`, `root.labels`,  `root.runs_on`,  `root.depends_on` to `root.when.xxx` to match step filters #283
  - [ ] make `root.when` and `root.pipeline.[step].when` an array of the current with items using the current settings to allow ORs
        -> non breaking
  - [ ] remove `root.Cache` and functionality in favor of plugins? -> need more discussion

LamaAni:
Hi, whats the ETA on 0.16?

6543:
This is an opensource project, work is done in free time. So we can only guess when v1.0.0 will be released. This can happen in 2Months or in 5 ...

LamaAni:
Hum. Thank you. Though I was asking when for for the 0.16.0, or the root.when feature, which I understand is the next candidate. That would be good for me. 

In any case, I would resort to building from source and PR back anything I got.

LamaAni:
Hi Looks GOOD!

LamaAni:
Can we change pipeline to steps?

6543:
we can but it would break **all** existing configs, what's the benefit? 

you can add it to #829 as comment, but for now I don't see a benefit - if it would technically improve a ting or make something possible that was not before it should be considered - but just for the sake of taste we should not break things - call it backwards compatibility

anbraten:
> All types should echo the yaml structure, and no/minimal "special" handling.
> (For example when is list. Always).

I really like that idea of having just one way of doing things. Having the option to use arrays or just a single string confuses be quite often and makes things like json-schema a hell to write.

LamaAni:
It would. When we are following the yaml config we can do the following things:
1. Add default definition to a server, E.G. Allow a "default pipeline" that will be merged.
2. Easier for new people to understand. 
3. Allow anchors to be used in the yaml config anywhere -> leads to more configurable yaml.

In general, the idea of version is good since it allows one to slowly migrate to a more "readable" yaml. Following standards, IMO, is the best way to expand use. 

Finally, Once proper yaml exists, we can in general start considering dynamic pipelines.

LamaAni:
Oh another thing, one can see an example of over complication by not following yaml standards in the Containers unmarshel,
```golang
[...]

PR #1767, opened 2023-05-19
Allow setting resources for kubernetes on a per-step basis

### LINKED ISSUE #1809: document Kubernetes backend
https://woodpecker-ci.org/docs/next/administration/backends/ does not contain a Kubernetes section!

smuth4:
This add a simple implementation of requests/limits for individual steps. There is no validation of what the resource actually is beyond checking that it can successfully be converted to a Quantity, so it can be used for things other than just memory/CPU.

close #1809


pipeline/backend/types/step.go
```diff
@@ -33,4 +33,5 @@ type Step struct {
 	NetworkMode  string            `json:"network_mode,omitempty"`
 	IpcMode      string            `json:"ipc_mode,omitempty"`
 	Sysctls      map[string]string `json:"sysctls,omitempty"`
+	Resources    Resources         `json:"resources,omitempty"`
```
anbraten, line 36: I think its not the best idea to add new properties on this step level, because they are pretty backend specific. I would suggest using a more generic approach for setting backend specific options on a per step basis. We could either use the existing env vars with prefixes like `WOODPECKER_K8S_RESOURCES_REQUEST_CPU=1000m` for this or add a generic key-value map like the plugin-settings property or a property called backend with key-value string-string mappings to the step. This way other backends could also specify further per step options in the future without bloating the base schema to much.

docs/docs/30-administration/80-kubernetes.md
```diff
@@ -1,6 +1,6 @@
 # Kubernetes
 
-Woodpecker does support Kubernetes as a backend.
+Woodpecker does support Kubernetes as a backend. See the [Kubernetes backend configuration](backends/kubernetes/#configuration) for backend-specific options.
```
anbraten, line 3: The url should use the path to the md file as link.

docs/docs/30-administration/15-agent-config.md
```diff
@@ -133,6 +133,10 @@ See [Docker backend configuration](backends/docker/#configuration)
 
 See [SSH backend configuration](backends/ssh/#configuration)
 
+### `WOODPECKER_BACKEND_K8S_*`
+
+See [Kubernetes backend configuration](backends/kubernetes/#configuration)
```
anbraten, line 138: The url should use the path to the md file as link.

anbraten (dismissed):

anbraten:
In general a really nice, clean and thoughtful PR. Thanks for that.

pipeline/frontend/yaml/container.go
```diff
@@ -62,11 +67,13 @@ type (
 		Sysctls       types.SliceorMap       `yaml:"sysctls,omitempty"`
 		When          constraint.When        `yaml:"when,omitempty"`
 		Settings      map[string]interface{} `yaml:"settings"`
+		Resources     Resources              `yaml:"resources,omitempty"`
 	}
 )
 
 // UnmarshalYAML implements the Unmarshaler interface.
 func (c *Containers) UnmarshalYAML(value *yaml.Node) error {
+	fmt.Println("UnmarshalYAML")
```
6543, line 76: rm

pat-s:
Ah great! I was just playing around with this a few days ago without success...

Would it be hard to add a global resource definition? Which is overwritten by definitions in individual steps?

pipeline/backend/types/step.go
smuth4: I prefer the plugin property settings style as well (I was simply copying the existing docker backend settings). Environment variables could still be used as an additional modifier later down the line if desired, but they're not as suitable for structured data like this.

smuth4:
> Ah great! I was just playing around with this a few days ago without success...
> 
> Would it be hard to add a global resource definition? Which is overwritten by definitions in individual steps?

It certainly sounds like a good idea, how does a top-level `backend_options` sound @6543? It might not be a good idea to try and over-generalize the solution with some sort of top>step merge (since you might want to handle things like arrays differently depending on context), but the structure could at least be the same.

```yaml
---
backend_options:
  kubernetes:
    resources:
      request:
        cpu: 1000m
        memory: 1024Mi
pipeline:
  my-step:
    image: golang
    backend_options:
      kubernetes:
        resources:
          request:
            cpu: 2000m # Requests 2000m CPU
                       # Requests 1024 Mi memory
```

6543:
I would not add a top level definition feature as you can do that already with: https://woodpecker-ci.org/docs/next/usage/advanced-yaml-syntax#map-merges-and-overwrites

if it's an agent config that do overwrite step config ... if step.limit>agent.limit that's of course another request and valid

pat-s:
> I would not add a top level definition feature as you can do that already with: [woodpecker-ci.org/docs/next/usage/advanced-yaml-syntax#map-merges-and-overwrites](https://woodpecker-ci.org/docs/next/usage/advanced-yaml-syntax#map-merges-and-overwrites)

Are you sure this is possible? To my knowledge this is not parsed to `pod.go` and the object created there. This is the nr.1 feature I am still waiting for (would take any workaround that works!). AFAICS the resources are currently hardcoded to 1 CPU, 2 GB mem.

smuth4:
> I would not add a top level definition feature as you can do that already with: https://woodpecker-ci.org/docs/next/usage/advanced-yaml-syntax#map-merges-and-overwrites

That's what I get for working with tools that scaffold YAML for me all day, I completely blanked on the native features. :disappointed: 

> if it's an agent config that do overwrite step config ... if step.limit>agent.limit that's of course another request and valid

I thought about adding this, but kubernetes already has native tools for setting default/limit ranges (although only on a per-namespace basis): https://kubernetes.io/docs/concepts/policy/limit-range/

> > I would not add a top level definition feature as you can do that already with: [woodpecker-ci.org/docs/next/usage/advanced-yaml-syntax#map-merges-and-overwrites](https://woodpecker-ci.org/docs/next/usage/advanced-yaml-syntax#map-merges-and-overwrites)
> 
> Are you sure this is possible? To my knowledge this is not parsed to `pod.go` and the object created there. This is the nr.1 feature I am still waiting for (would take any workaround that works!). AFAICS the resources are currently hardcoded to 1 CPU, 2 GB mem.

I tested it and it should work,  e.g.
```yaml
variables:
  - &resources
    requests:
      memory: "4Gi"
      cpu: "250m"
    limits:
      memory: "8Gi"
      cpu: "1000m"
pipeline:
  build:
    image: golang
    backend_options:
      kubernetes:
        # Gets 4Gi & 250m requests, 2000m limits
        resources:
          <<: *resources
          limits:
            cpu: "2000m"
```

pat-s:
> I tested it and it should work, e.g.

But does it work without explicit mapping? I.e. usually I would like to write it as follows

```yml
variables:
  - &resources
    requests:
      memory: "4Gi"
      cpu: "250m"
    limits:
      memory: "8Gi"
      cpu: "1000m"
pipeline:
  build:
    image: golang
```

and this should result in the step honoring the top-level resources *unless* they are explicitly set/overwritten in the respective step.

Imagine a pipeline with many steps: it should be possible that each step follows a global resource definition without a redundant resources mapping within each step.

@6543 Given that other top-level arguments like `platform` exist, I think it would be "cleaner" to have `resources` being an "official" top level setting rather than forcing users to invoke it via map-merges?

zc-devs:
@smuth4, great work and thank you for this desired functionality!

As I understand, this configuration is per step or per pipeline. Could we add per agent configuration (like I've done here 8eb240be3fa35c71bf9bbd6b2408c008191bf0ff)?

smuth4:
> @smuth4, great work and thank you for this desired functionality!
> 
> As I understand, this configuration is per step or per pipeline. Could we add per agent configuration (like I've done here [8eb240b](https://github.com/woodpecker-ci/woodpecker/commit/8eb240be3fa35c71bf9bbd6b2408c008191bf0ff))?

I understand the desire for the feature, but can you explain the use case which requires it to be part of the woodpecker code? Specifically, I'm trying to avoid re-implementing [LimitRanges](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/), which are very powerful and native to kubernetes, but do require application at the namespace level.

zc-devs:
@smuth4, yeah, you're right. Cool feature.

docs/docs/30-administration/22-backends/40-kubernetes.md
```diff
@@ -0,0 +1,66 @@
...
+
+### `WOODPECKER_BACKEND_K8S_POD_LABELS`
+> Default: empty
+
+Additional labels to apply to worker pods. Must be a YAML object, e.g. `{"example.com/test-label":"test-value"}`.
+
+### `WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS`
+> Default: empty
+
+Additional labels to apply to worker pods. Must be a YAML object, e.g. `{"example.com/test-annotation":"test-value"}`.
```
zc-devs, line 41: Annotations?

smuth4: Ah, thanks!

6543:
> @6543 Given that other top-level arguments like `platform` exist ...

well it does exist but I consider this an legacy and we do move to labels ... witch can filter backend, platform, ...

so there might come the time where it will be deprecated in favour of a simpler and more generic solution

but that's something for the #567 to talk about :D

6543 (approved):

docs/docs/30-administration/22-backends/40-kubernetes.md
```diff
@@ -0,0 +1,66 @@
+# Kubernetes backend
+
+:::caution
+Kubernetes support is still experimental and not all pipeline features are fully supported yet.
+
+Check the [current state](https://github.com/woodpecker-ci/woodpecker/issues/9#issuecomment-483979755)
```
6543, line 6:
```suggestion
Check the [current state](https://github.com/woodpecker-ci/woodpecker/issues/1513)
```

6543 (approved):

PR #1775, opened 2023-05-25
PipelineLog.vue: Make it possible to link to a log line

runephilosof-karnovgroup:
Set id attribute of lines to make it possible to link to a pipeline line, like it was in version 0.9. For instance https://woodpecker.example.com/org/repo/1#L7


runephilosof-karnovgroup:
I haven't actually run the code and checked that it works.

anbraten (approved):

runephilosof-karnovgroup:
The reason why I did not run the code was that the project seems to be missing a setup guide to make it easy for new contributors.
Also, the docker-compose.example.yml does not work out of the box.

6543:
Also the ci build preview imges and push them to dockerhub ... if you prefer that :)

PR #1779, opened 2023-05-27
Add alpine matrix builds

### LINKED ISSUE #1778: Provide images the two latest alpine releases

~~The bump to alpine 3.18 in #1765 caused the side-effect of WP requiring the docker version on the host to be 23.x as otherwise an incompatibility with Docker API 1.42 is thrown.~~

~~Many package managers don't provide docker 23.x yet which makes staying on WP next impossible atm.~~

Provide images for the last two alpine versions - in case something is up with latest.

### Suggested solution

~~It would be great if alpine 3.17 could be added to the distribution matrix for better distro support until all those package managers have catched up with the docker daemon versions.~~

### Alternative

~~Downgrade to alpine 3.17 until the situation has improved.~~

### Additional context

_No response_

### Validations

pat-s:
fix #1778 

Not sure if I got everything right for the workflow to succeed, feel free to tweak.

Also includes some whitespace and quotes harmonizations which my editor first changes and I then realized that the mappings are not consistent across the file.

The biggest issue with this PR is that when building with multiple alpine versions, the tags `next-alpine` and `alpine-latest` are not unique anymore and would probably need to to be removed as I don't see how there could be a logic to only apply them for the respective latest alpine version builds?

### DIFF
--- a/.woodpecker/docker.yml
+++ b/.woodpecker/docker.yml
@@ -11,6 +11,7 @@ variables:
   - &platforms_server 'linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le,linux/riscv64'
   - &platforms_preview 'linux/arm/v6,linux/arm64/v8,linux/amd64,linux/riscv64,windows/amd64'
   - &platforms_alpine 'linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le'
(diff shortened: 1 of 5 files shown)


anbraten:
As you noted in https://github.com/woodpecker-ci/woodpecker/issues/1778#issuecomment-1567869717 this would not solve the docker compatibility issue. As having multiple alpine versions complicates maintenance, I would close this PR for now.

PR #1781, opened 2023-05-27
support custom .JS and .CSS files for custom banner messages (white-labeling)

nitram509:
This PR introduces two new server configuration options, for providing a custom .JS and .CSS file.
These can be used to show custom banner messages, add environment-dependent signals, or simply a corporate logo.

### Motivation (what problem I try to solve)

I'm operating Woodpecker in multiple k8s clusters for different environments.
When having multiple browser tabs open, I prefer strong indicators for each environment.
E.g. a red "PROD" banner, or just a blue "QA" banner.
Also, we sometimes need to have the chance for maintenance, and instead of broadcasting emails,
I prefer a banner message, stating something like: "Heads-up: there's a planned downtime, next Friday, blabla...".

Also, I like to have the firm's logo visible, which makes Woodpecker look more like an integral part of our platform.

### Implementation notes

* Two new config options are introduced ```WOODPECKER_CUSTOM_CSS_FILE``` and ```WOODPECKER_CUSTOM_JS_FILE```
* I've piggy-bagged the existing handler for assets, as it seemed to me a minimally invasive approach
* the option along with an example is documented
* a simple unit test for the Gin-handler ensures some regression safety
* no extra dependencies are introduced

### Visual example

The documented example will look like this.

Screenshot 2023-05-27 at 17 00 44

### Areas of uncertainty 

This is my first contribution to Woodpecker and I tried my best to align with your conventions.
That said, I found myself uncertain about these things and would be glad about getting feedback.

* The handler tests are somewhat different than the other ones because I wanted to keep them simple - I hope that still matches your coding guidelines
* caching the page sometimes will let the browser not recognize changes and a user must reload. I'm not fully into the details of how caching is implemented and neither can judge if it's a real problem. Another pair of eyes would be good.

Any feedback is welcome.


server/web/web.go
```diff
@@ -49,13 +50,33 @@ func New() (*gin.Engine, error) {
...
+	var serveFileOrEmptyContent = func(w http.ResponseWriter, r *http.Request, localFileName string) {
+		if len(localFileName) > 0 {
+			http.ServeFile(w, r, localFileName)
+		} else {
+			// prefer zero content over sending a 404 Not Found
+			http.ServeContent(w, r, localFileName, time.Now(), bytes.NewReader([]byte{}))
+		}
+	}
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.RequestURI == "/assets/custom.js" {
```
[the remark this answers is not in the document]
nitram509: thank you for that feedback, will update soon.

docs/docs/30-administration/10-server-config.md
```diff
@@ -105,6 +105,47 @@ or generate a random one like this:
...
+Showing custom banner message in the web UI, or a corporate logo can be done by providing
+custom .JS and .CSS files. These files must be present in the server's filesystem.
+They can be backed in a Docker image or mounted from a ConfigMap inside a Kubernetes environment.
+
+```text
+WOODPECKER_CUSTOM_CSS_FILE=/usr/local/www/woodpecker.css
+WOODPECKER_CUSTOM_CSS_FILE=/usr/local/www/woodpecker.js
+```
+
+These are examples, how to place a corporate logo in the top navigation bar of Woodpecker.
```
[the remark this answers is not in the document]
nitram509: did improve wording and hope its getting clearer now.

6543:
resolved conflict :)

6543:
we got lint issues :(

nitram509:
✅ did fix the 2 linter and the 1 prettier issue the pipelines reported.

---

That said, there is one linter issue left, which originates from code I did not touch.
I was unable to fix it, because I don't know, what's the linter's expected style == what rule is broken?
Here's the issue:
```shell
$> make lint

func setupEvilGlobals(c *cli.Context, v store.Store, f forge.Forge) {
                         ^
make: *** [lint] Error 1
```
Any ideas?

nitram509:
> @nitram509 You removed an import in `cmd/server/server.go` which is required https://github.com/woodpecker-ci/woodpecker/pull/1781/files#diff-f63ad7661b197b38fdeecb70af480024e91189d9283e5bc7d8fb20f0f0b158cc

Ooh, no. Thank you for pointing this out and apologies for the inconvenience.
I did switch my laptop this morning and had to set up a completely fresh workspace.
During IntelliJ index creation I did the refactoring and likely I was not patient enough.
Regression suites to the rescue :)
Will fix this right away.

PS: also I'm monitoring what other issues in the pipeline occur, to fix e.g. complaining linters etc.

cmd/server/flags.go
```diff
@@ -72,6 +72,16 @@ var flags = []cli.Flag{
 		Name:    "server-key",
 		Usage:   "server ssl key path",
 	},
+	&cli.StringFlag{
```
6543, line 75: I wonder if we could use env vars with value too instead of require a file ...

nitram509: Transporting CSS or JS code via a ENV VAR would IMHO require the content to be BASE64 encoded, to avoid special character issues. And that would be cumbersome.

I'm not sure about the need to do so, since e.g. in a Kubernetes environment, one can mount custom CSS and JS files from a ConfigMap, see https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#populate-a-volume-with-data-stored-in-a-configmap.
This trick works easily with large CSS or JS files as well (not just the single-line data in the documented example).

For any Woodpecker users, provisioning non-Kubernetes systems, I would assume some mechanics for provisioning custom files are present as well (thinking of puppet, chef, ansible).

This makes me wonder, in which scenario Woodpecker users/admins would require the ability to provision custom CSS and JS as ENV VAR without file mapping?

6543: well that's a good argument against my idea yes

6543 (approved):

PR #1791, opened 2023-05-31
Remove unused file system api

anbraten:


server/store/datastore/migration/migration.go
```diff
@@ -51,7 +51,6 @@ var allBeans = []interface{}{
 	new(model.Pipeline),
 	new(model.PipelineConfig),
 	new(model.Config),
-	new(model.File),
```
6543, line 54: should we drop that table too?

anbraten: Yes, I will create a migration. It is empty for all system as it wasn't used anyways.

6543:
@anbraten updated proto file for you

server/store/datastore/migration/migration.go
anbraten: done

6543 (approved):

PR #1793, opened 2023-05-31
Buildin Env Vars, use _LINK for all links/urls

6543:
non breaking as no tagged release do have it & not completely removed jet


6543:
ok let's do a vote: https://framadate.org/mOJABOjBGDEVPsyv

6543:
-> #1794

PR #1796, opened 2023-05-31
Remove MSSQL specific code and cleanups

qwerty287:
We don't have MSSQL support

### DIFF
--- a/server/store/datastore/migration/000_legacy_to_xorm.go
+++ b/server/store/datastore/migration/000_legacy_to_xorm.go
@@ -86,7 +86,7 @@ var legacy2Xorm = task{
 
 		{ // recreate build_config
 			type BuildConfig struct {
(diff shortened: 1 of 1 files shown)


server/store/datastore/migration/000_legacy_to_xorm.go
```diff
@@ -86,13 +86,13 @@ var legacy2Xorm = task{
 
 		{ // recreate build_config
 			type BuildConfig struct {
-				ConfigID int64 `xorm:"NOT NULL 'config_id'"` // xorm.Sync2() do not use index info of sess -> so it try to create it twice
+				ConfigID int64 `xorm:"NOT NULL 'config_id'"` // xorm.Sync() do not use index info of sess -> so it try to create it twice
```
anbraten, line 89:
```suggestion
				ConfigID int64 `xorm:"NOT NULL 'config_id'"` // xorm.Sync() do not use index info of sess -> so it tries to create it twice
```

anbraten (approved):

Issue #1370, opened 2022-10-30
Branches api endpoint returns 500 if user is not authenticated


anbraten:

6543:
well that's what tests are for, to make sure bugs are fixed once and for all :laughing:

anbraten:
`{"level":"error","error":"Error #01: GET https://api.github.com/repos/woodpecker-ci/woodpecker/branches?page=1&per_page=50: 401 Bad credentials []\n","ip":"","latency":130.351425,"method":"GET","path":"/api/repos/woodpecker-ci/woodpecker/branches","status":500,"time":"2023-06-02T06:38:07Z","user-agent":"","time":"2023-06-02T06:38:07Z"}`

PR #1803, opened 2023-06-02
Do not use oauth client without token

qwerty287:
Closes https://github.com/woodpecker-ci/woodpecker/issues/1370


server/forge/github/github.go
```diff
@@ -394,16 +394,29 @@ func (c *client) newConfig(req *http.Request) *oauth2.Config {
...
-	)
-	tc := oauth2.NewClient(ctx, ts)
-	if c.SkipVerify {
-		tc.Transport.(*oauth2.Transport).Base = &http.Transport{
-			Proxy: http.ProxyFromEnvironment,
-			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
-			},
+	var tc *http.Client
+	if token != "" {
```
anbraten, line 398: Why could the token be empty?

anbraten: I think that could be an issue if the repo is private and the woodpecker project is public. Could we pass the token of the project / repo instead?

server/forge/common/utils.go
```diff
@@ -37,3 +43,20 @@ func ExtractHostFromCloneURL(cloneURL string) (string, error) {
...
+	if u != nil {
+		return u.Token
+	}
+
+	_store, ok := store.TryFromContext(ctx)
+	if !ok {
+		log.Error().Msg("could not get store from context")
+		return ""
+	}
+	user, err := _store.GetUser(r.UserID)
```
6543, line 57: hmm we might als add a nil check for r

6543: I added a simple check that in best case will never be exec, in worst case we have a unique error msg we can trace down

PR #1825, opened 2023-06-05
remove undocumented docker specific option

6543:


gapodo:
Hi,

short note from my side, removin this will likely result in it being requested for readdition soon(tm), as esp. with K8s there is often a need to label containers for them to be able to run (depends on the k8s setup, but things like network restrictions,... are often applied via labels (or annotations)).

Though this could/should potentially be a seperate issue / rfc / FR as to how we handle backend specific options?

6543:
well in that case it will be a k8s specific backend option yes

6543:
@gapodo https://woodpecker-ci.org/docs/next/administration/backends/kubernetes#resources

gapodo:
> @gapodo https://woodpecker-ci.org/docs/next/administration/backends/kubernetes#resources

Oh, thx, sorry had some troubles keeping up with all the changes... Since it's cared for by the new backend_options (where it should be implemented in that case), you can disregard my note.

PR #1834, opened 2023-06-06
WIP: add version to workflow configs

6543:


6543:
// TODO: should we require a version number -> in therms of UX we should not, in terms of strong typisation we should

PR #1835, opened 2023-06-07
wip wip

6543:
make it able to have pipeline/frontend/yaml/types to be used outside of this repo for serialization etc ..

### DIFF
--- a/pipeline/frontend/yaml/types/base/bool.go
+++ b/pipeline/frontend/yaml/types/base/bool.go
@@ -1,6 +1,7 @@
 package base
 
 import (
+	"fmt"
 	"strconv"
 
 	"gopkg.in/yaml.v3"
@@ -26,6 +27,11 @@ func (b *BoolTrue) UnmarshalYAML(value *yaml.Node) error {
 		return err
 	}
 	return nil
+}
+
+// MarshalText implements custom Yaml marshaling.
+func (b BoolTrue) MarshalText() (text []byte, err error) {
+	return []byte(fmt.Sprint(b.Bool())), nil
 }
 
 // Bool returns the bool value.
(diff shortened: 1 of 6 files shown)


6543:
was ready just *mostly ... not all types where covered jet :D

pipeline/frontend/yaml/types/base/slice.go
```diff
@@ -44,6 +44,16 @@ func (s *StringOrSlice) UnmarshalYAML(unmarshal func(any) error) error {
 	return errors.New("failed to unmarshal StringOrSlice")
 }
 
+// MarshalYAML implements custom Yaml marshaling.
+func (s StringOrSlice) MarshalYAML() (any, error) {
+	if len(s) == 0 {
+		return "", nil
```
6543:
```suggestion
		return nil, nil
```

6543:
... so we could in the long run let woodpecker do configuration migrations by itself ...

6543:
... this also makes sure we can have more coverage about parsing configs (even if codecouverage is not happy because new code was added :D) ...

pipeline/frontend/yaml/types/base/bool.go
```diff
@@ -42,7 +42,16 @@ func (b *BoolTrue) UnmarshalYAML(value *yaml.Node) error {
...
+	return b.Bool(), nil
+}
+
 // Bool returns the bool value.
 func (b BoolTrue) Bool() bool {
 	return !b.value
 }
+
+func ToBoolTrue(v bool) BoolTrue {
+	return BoolTrue{value: !v}
```
qwerty287, line 56: I don't think this is correct. `BoolTrue` just means the default is true. So `var: false` still sets `false`.

6543: no its correct the implementation is kind a cursed ...

it uses the fact that golang serializes and inutializes bool to false by default and just always work with the inferted

6543: see Bool() at line 52(new file) and line 37

6543:
I agre that the TrueBool is not nice and inexplizit ... after this pull i'll migrate to OptionalBool type witch let us allow to do the same but explizite 

https://github.com/6543/go-optional/issues/3

PR #1850, opened 2023-06-12
01-getting-started.md: GCC go does not work

runephilosof-karnovgroup:
Guide developers around GCC go for easier onboarding.


6543:
go-delve/delve is not used

PR #1658, opened 2023-03-20
Do not store inactive repos without any resources

qwerty287:
Do not sync repos with forge if the repo is not necessary in DB.

In the DB, only repos that were active once or repos that are currently active are stored. When trying to enable new repos, the repos list is fetched from the forge instead and displayed directly. In addition to this, the forge func `Perm` was removed and is now merged with `Repo`.

Solves a TODO on RepoBatch.

PR #1853, opened 2023-06-12
sync repo permissions on login

### LINKED ISSUE #1666: Updating user permissions on login
https://codeberg.org/Codeberg-CI/feedback/issues/66  
https://github.com/woodpecker-ci/woodpecker/issues/485  
https://github.com/woodpecker-ci/woodpecker/pull/1658#discussion_r1143782278

### LINKED ISSUE #1851: Missing Permissions for users of same org
### Component

server

### Describe the bug

Scenario:
User A and B are members of the same org on gitea.
Both have admin permissions to the repo in question.

When User A activates the repo in woodpecker, i would expect User B to be able to see the repository in their feed.

I think #1658 change the behavior here and this used to be the case.

When manually navigating as User B to the repo in woodpecker the server logs
`Error fetching permission for <user> <repo>. sql: no rows in result set`
and the repo becomes permanently visible in their feed.

### System Info

```shell
Woodpecker next-b5b3b957
```

### Additional context

_No response_

### Validations

bkleiner:
if indeed a bug, fixes #1851 after re-log

close #1666


server/api/login.go
```diff
@@ -145,6 +145,30 @@ func HandleAuth(c *gin.Context) {
 		return
 	}
 
+	repos, err := _store.RepoListAll(true)
```
6543, line 148: well instead of itterate over all repos, wouldn't it be better to get the list a user has acces to from the forge?

in any case we need to be carefull as souch a large-scale-sync do hit api rate limits fast!!!

bkleiner: > well instead of itterate over all repos, wouldn't it be better to get the list a user has acces to from the forge?

right - switched to that way of handling it, makes it much simpler too.

> in any case we need to be carefull as souch a large-scale-sync do hit api rate limits fast!!!

yeah about that, the `user_synced` attribute was dropped recently. https://github.com/woodpecker-ci/woodpecker/pull/1658/files and i couldn't find a good other attribute this could tie to.
Should we reintroduce it?

6543: well in this regards I think we should just encourage to not logout and in all the time ... the session can stay logged in quite some time D:

bkleiner: fair enough - imho it would be nice if these were refreshed without a re-log (we have _a lot_ of traffic on our ci)
but thats outside of the scope of this PR anyway so  ¯\_(ツ)_/¯

server/api/login.go
```diff
@@ -145,6 +145,41 @@ func HandleAuth(c *gin.Context) {
...
+	}
+
+	for _, forgeRepo := range repos {
+		dbRepo, err := _store.GetRepoForgeID(forgeRepo.ForgeRemoteID)
+		if err != nil && errors.Is(err, types.RecordNotExist) {
+			continue
+		}
+		if err != nil {
+			log.Error().Msgf("cannot list repos for %s. %s", u.Login, err)
+			c.Redirect(http.StatusSeeOther, "/login?error=internal_error")
```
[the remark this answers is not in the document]
bkleiner: i guess this relates to line 148 and following, the db query here should always succeed.
i really don't know about swallowing _all_ errors. you would want to let users know that it can't list the repos from the forge, wouldn't you?
So _imho_ if one wanted to ignore rate-limit errors here, it should check only for that. (which would require re-working _all_ the forge clients)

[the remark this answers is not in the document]
bkleiner: that's completely besides what i saying but ok...

PR #1869, opened 2023-06-17
Added Woodpecker Host Config used for Webhooks

zc-devs:
When SCM and Woodpecker are deployed in the same cluster, I want that SCM requests Woodpecker directly by internal URL.

Webhook-host-diagram
 
This PR adds optional URL used for Webhook calls.

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -341,7 +341,11 @@ func setupEvilGlobals(c *cli.Context, v store.Store, f forge.Forge) {
 	server.Config.Server.Key = c.String("server-key")
 	server.Config.Server.AgentToken = c.String("agent-secret")
 	server.Config.Server.Host = c.String("server-host")
-	server.Config.Server.WebhookHost = c.String("server-webhook-host")
+	if c.IsSet("server-webhook-host") {
+		server.Config.Server.WebhookHost = c.String("server-webhook-host")
+	} else {
+		server.Config.Server.WebhookHost = c.String("server-host")
+	}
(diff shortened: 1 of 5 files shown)


server/api/repo.go
```diff
@@ -108,9 +108,13 @@ func PostRepo(c *gin.Context) {
 		return
 	}
 
+	wpHost := server.Config.Server.WebhookHost
+	if wpHost == "" {
+		wpHost = server.Config.Server.Host
```
6543, line 113: We should move that check to config init

cmd/server/server.go
```diff
@@ -341,6 +341,7 @@ func setupEvilGlobals(c *cli.Context, v store.Store, f forge.Forge) {
 	server.Config.Server.Key = c.String("server-key")
 	server.Config.Server.AgentToken = c.String("agent-secret")
 	server.Config.Server.Host = c.String("server-host")
+	server.Config.Server.WebhookHost = c.String("server-webhook-host")
```
6543, line 344: If set ... else set conf with host conf

docs/docs/30-administration/10-server-config.md
```diff
@@ -131,6 +131,13 @@ Server fully qualified URL of the user-facing hostname.
 
 Example: `WOODPECKER_HOST=http://woodpecker.example.org`
 
+### `WOODPECKER_WEBHOOK_HOST`
+> Default: empty
+
```
6543, line 136: Default is HOST config env

6543 (approved):

PR #1873, opened 2023-06-19
Use id to access orgs

### LINKED ISSUE #1743: Access orgs by id
Similar to #1295 it would be great to access orgs by their id. This would solve renaming issues

anbraten:
closes #1743 

fixes: setting secrets for own user namespace

# TODO
- [x] create org on repo activation
- [x] add `org_id` property to repo
- [x] load secrets using `org_id`
- [x] change org api endpoints to use `org_id`
- [x] change org-secrets endpoints to use `org_id`
- [x] change user is member (membership / service) endpoint to use `org_id`
- [x] support personal (org = user)
- [x] add api endpoint to get repos of `org` (=> separate PR)
- [x] add migration to create orgs for existing repos and set `repo.org_id`
- [x] add migration to set `org_id` for secrets
- [x] adjust UI
- [x] test adding repo
- [x] test loading org
- [x] test org secrets CRUD


server/router/api.go
```diff
@@ -46,13 +46,15 @@ func apiRoutes(e *gin.Engine) {
 			users.DELETE("/:login", api.DeleteUser)
 		}
 
-		orgBase := apiBase.Group("/orgs/:owner")
+		apiBase.GET("/orgs/lookup/*org_full_name", api.LookupOrg) // TODO: check if this public route is a security issue
```
6543, line 49: if LookupOrg does call the gitea api with no user auth ... yes it is an issue, lookup from DB should be fine

server/store/datastore/org.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (s storage) OrgCreate(org *model.Org) error {
+	_, err := s.engine.Insert(org)
+	return err
+}
+
+func (s storage) OrgFind(id int64) (*model.Org, error) {
+	org := &model.Org{
+		ID: id,
+	}
```
6543, line 29: explizite filter

server/store/datastore/org.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (s storage) OrgDelete(id int64) error {
+	return wrapDelete(s.engine.ID(id).Delete(new(model.Org)))
+}
+
+func (s storage) OrgFindByName(name string) (*model.Org, error) {
+	org := &model.Org{
+		Name: name,
+	}
+	return org, wrapGet(s.engine.Get(org))
```
6543, line 46: explizite filter

server/store/datastore/migration/021_add_orgs.go
```diff
@@ -0,0 +1,96 @@
...
+
+		orgs := make(map[string]*model.Org)
+		for _, repo := range repos {
+			orgName := repo.Owner
+
+			// create org if not already created
+			if _, ok := orgs[orgName]; !ok {
+				org := &model.Org{
+					Name:   repo.Owner,
+					IsUser: false, // TODO: should we get this info from the forges?
```
6543, line 69: i would not call the forge, migrations should be selve-contained as posible

we need to add a check&fix on our lookups

server/api/repo.go
```diff
@@ -114,6 +114,31 @@ func PostRepo(c *gin.Context) {
 		sig,
 	)
 
+	// find org of repo
```
6543, line 117: wee need to check if it's not an org but user repo

server/api/org.go
```diff
@@ -15,41 +15,121 @@
...
 
-	perm, err := server.Config.Services.Membership.Get(c, user, owner)
+	org, err := _store.OrgFind(orgID)
 	if err != nil {
-		c.String(http.StatusInternalServerError, "Error getting membership for %q. %s", owner, err)
+		c.String(http.StatusInternalServerError, "Error getting org %d. %s", orgID, err)
+		return
+	}
+
+	if (org.IsUser && org.Name == user.Login) || user.Admin {
```
6543, line 93: we should handle the case here where the org is actually a user but migration got it wrong

server/store/datastore/migration/021_add_orgs.go
6543: https://github.com/woodpecker-ci/woodpecker/pull/1873/files#r1258472267

server/router/api.go
anbraten: That would be the best as orgs could be private and this endpoint would allow enumeration otherwise.

server/api/repo.go
anbraten: Why? Its intended to also save orgs for users as it should be possible to set for exp secrets for your user-org as well.

6543:
migration finaly works :laughing: ... dont forget to drop the index & constrains first befor droping the column :D ...

6543 (approved):
lgtm to code that I did not touch :D

server/store/datastore/org_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// create two more orgs and repos
+	some_user := &model.Org{Name: "some_other_u", IsUser: true}
+	assert.NoError(t, store.OrgCreate(some_user))
+	assert.NoError(t, store.OrgCreate(&model.Org{Name: "some_other_org"}))
+	assert.NoError(t, store.CreateRepo(&model.Repo{UserID: 1, Owner: "some_other_u", Name: "abc", FullName: "some_other_u/abc", OrgID: some_user.ID}))
+	assert.NoError(t, store.CreateRepo(&model.Repo{UserID: 1, Owner: "some_other_u", Name: "xyz", FullName: "some_other_u/xyz", OrgID: some_user.ID}))
+	assert.NoError(t, store.CreateRepo(&model.Repo{UserID: 1, Owner: "renamedorg", Name: "567", FullName: "renamedorg/567", OrgID: orgOne.ID}))
+
+	// get all orgs
```
6543, line 67:
```suggestion
	// get all repos for specific org
```

6543:
Ok ironed out edgecases and fixed a bug while writing tests - also tested it against a database of a big installation.

This is ready :)

6543:
@anbraten you can not lgtm your own pull ... but you could create a comment that sates the same

so if you have time, could you review my changes :)

server/api/org.go
```diff
@@ -15,41 +15,149 @@
...
+
+		_ = c.AbortWithError(http.StatusInternalServerError, err)
+		return
+	}
+
+	// don't leak private org infos
+	if org.Private {
+		user := session.User(c)
+		if user == nil {
+			c.String(http.StatusUnauthorized, "User not authorized")
```
lafriks, line 140: We would still leak info that org/user exists. Should return not found (same status/msg as above)

server/api/org.go
```diff
@@ -15,41 +15,149 @@
...
+			return
+		}
+
+		if !user.Admin && org.Name != user.Login {
+			c.String(http.StatusForbidden, "User not authorized")
```
lafriks, line 145: Same here

server/api/org.go
```diff
@@ -15,41 +15,149 @@
...
+		} else if !user.Admin {
+			perm, err := server.Config.Services.Membership.Get(c, user, org.Name)
+			if err != nil {
+				log.Error().Msgf("Failed to check membership: %v", err)
+				c.String(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
+				return
+			}
+
+			if perm == nil || !perm.Member {
+				c.String(http.StatusForbidden, "User not authorized")
```
lafriks, line 156: Same here

anbraten:
lgtm apart from remaining threads. One thing I am not sure about is if the catch orgs for users which wont login themself to woodpecker. Exp I am a member of a repo from `6543/example`, but he doesn't use / login to woodpecker. In that case we would not create that org I guess.

6543:
In that edgecase the oser is handled like an org ... witch does not conflict in any way ... tested it also

lafriks (approved):

Issue #1648, opened 2023-03-19
Agent just stoped working


6543:
error code:
```
rpc error: code = Unknown desc = sql: no rows in result set
rpc error: code = Unknown desc = sql: no rows in result set
rpc error: code = Unknown desc = sql: no rows in result set
rpc error: code = Unknown desc = sql: no rows in result set
...
```

version 1b43b0bf20c03e63fb270453d2ffa5afed456fb2

anbraten:
It was running in between, but now its throwing that error again.

mvdkleijn:
I appear to have the same problem.

Clean setup of 0.15.7, no errors apart from the "no rows" one.

Docker-compose based, clean sqlite db, server and agent in same compose file using same agent secret.

I basically appear to be running a server and agent successfully but I haven't been able to run a pipeline due to this issue.

Note that the webhook is created correctly in the repo and arrives at the server causing this error.

mvdkleijn:
I looked at the sqlite DB... seems kinda empty to me?!? @6543 Is the DB supposed to be empty like this?

This is on a clean install using docker-compose with one server, one agent, one user, couple of repos from an external gitea instance.

```
sqlite> .tables
agents        config        migrations    registry      senders     
build_config  files         perms         repos         tasks       
builds        logs          procs         secrets       users       
sqlite> select * from agents;
sqlite> select * from build_config;
sqlite> select * from builds;
sqlite> select * from config;
sqlite> select * from files;
sqlite> select * from logs;
sqlite> select * from migrations;
name                                
------------------------------------
xorm                                
alter-table-drop-repo-fallback      
drop-allow-push-tags-deploys-columns
fix-pr-secret-event-name            
alter-table-drop-counter            
alter-table-logs-update-type-of-data
sqlite> select * from perms;
perm_user_id  perm_repo_id  perm_pull  perm_push  perm_admin  perm_synced  created     updated   
------------  ------------  ---------  ---------  ----------  -----------  ----------  ----------
1             1             1          1          1           1680032950   1680036562  1680036562
1             2             1          1          1           1680032950   1680036562  1680036562
1             3             1          1          1           1680032950   1680036562  1680036562
1             4             1          1          1           1680032950   1680036562  1680036562
1             5             1          1          1           1680032950   1680036562  1680036562
1             6             1          1          1           1680032950   1680036562  1680036562
1             7             1          1          1           1680032950   1680036562  1680036562
[...]

PR #1888, opened 2023-06-26
Store agent ID in config file

ttomasini:
This PR solves the following TODO in `/cmd/agent/agent.go`:
https://github.com/woodpecker-ci/woodpecker/blob/1fa021273126ddf2700a8dea0a3e990ede344a90/cmd/agent/agent.go#L112

It ensures that the agent ID is persisted in a file and can then, for example, also be bind-mounted in a Docker container. The new ~`WOODPECKER_AGENT_ID_CONFIG_PATH`~ `WOODPECKER_AGENT_ID_FILE` environment variable makes it possible to configure the file path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,6 @@ __debug_bin
 web/dist/**
 !web/dist/.gitkeep
 web/node_modules/
-web/package-lock.json
 web/*.log
 web/.env
 
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ __debug_bin
 web/dist/**
 !web/dist/.gitkeep
 web/node_modules/
+web/package-lock.json
 web/*.log
 web/.env
 
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -50,7 +50,7 @@ import (
 )
 
 func run(c *cli.Context) error {
-	agentIdConfigPath := c.String("agent-id-config-path")
+	agentIDConfigPath := c.String("agent-id-config-path")
 	hostname := c.String("hostname")
 	if len(hostname) == 0 {
 		hostname, _ = os.Hostname()
@@ -111,7 +111,7 @@ func run(c *cli.Context) error {
 	}
 	defer authConn.Close()
 
-	agentID := readAgentId(agentIdConfigPath)
+	agentID := readAgentID(agentIDConfigPath)
 	agentToken := c.String("grpc-token")
 	authClient := agentRpc.NewAuthGrpcClient(authConn, agentToken, agentID)
 	authInterceptor, err := agentRpc.NewAuthInterceptor(authClient, 30*time.Minute)
@@ -180,7 +180,7 @@ func run(c *cli.Context) error {
 		return err
 	}
 
-	writeAgentId(agentID, agentIdConfigPath)
+	writeAgentID(agentID, agentIDConfigPath)
 
 	labels := map[string]string{
 		"hostname": hostname,
@@ -285,30 +285,32 @@ func stringSliceAddToMap(sl []string, m map[string]string) error {
 	return nil
 }
 
-func readAgentId(agentIdConfigPath string) int64 {
-	const defaultAgentIdValue = int64(-1)
+func readAgentID(agentIDConfigPath string) int64 {
+	const defaultAgentIDValue = int64(-1)
 
-	rawAgentId, fileErr := os.ReadFile(agentIdConfigPath)
+	rawAgentID, fileErr := os.ReadFile(agentIDConfigPath)
 	if fileErr != nil {
-		return defaultAgentIdValue
+		log.Debug().Err(fileErr).Msgf("could not open agent-id config file from %s", agentIDConfigPath)
+		return defaultAgentIDValue
 	}
 
-	strAgentId := strings.TrimSpace(string(rawAgentId))
-	agentId, err := strconv.ParseInt(strAgentId, 10, 64)
-	if err != nil {
-		return defaultAgentIdValue
+	strAgentID := strings.TrimSpace(string(rawAgentID))
+	agentID, parseErr := strconv.ParseInt(strAgentID, 10, 64)
+	if parseErr != nil {
+		log.Warn().Err(parseErr).Msg("could not parse agent-id config file content to int64")
+		return defaultAgentIDValue
 	}
 
-	return agentId
+	return agentID
 }
 
-func writeAgentId(agentID int64, agentIdConfigPath string) {
-	currentAgentId := readAgentId(agentIdConfigPath)
+func writeAgentID(agentID int64, agentIDConfigPath string) {
+	currentAgentID := readAgentID(agentIDConfigPath)
 
-	if currentAgentId != agentID {
-		err := os.WriteFile(agentIdConfigPath, []byte(strconv.FormatInt(agentID, 10)+"\n"), 0644)
+	if currentAgentID != agentID {
+		err := os.WriteFile(agentIDConfigPath, []byte(strconv.FormatInt(agentID, 10)+"\n"), 0o644)
 		if err != nil {
-			log.Warn().Err(err).Msgf("could not write agent-id config file to %s", agentIdConfigPath)
+			log.Warn().Err(err).Msgf("could not write agent-id config file to %s", agentIDConfigPath)
 		}
 	}
 }
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -49,9 +49,8 @@ import (
 	"github.com/woodpecker-ci/woodpecker/version"
 )
 
-const configPath = "/agent-id.conf"
-
 func run(c *cli.Context) error {
+	agentIdConfigPath := c.String("agent-id-config-path")
 	hostname := c.String("hostname")
 	if len(hostname) == 0 {
 		hostname, _ = os.Hostname()
@@ -112,7 +111,7 @@ func run(c *cli.Context) error {
 	}
 	defer authConn.Close()
 
-	agentID := readAgentId()
+	agentID := readAgentId(agentIdConfigPath)
 	agentToken := c.String("grpc-token")
 	authClient := agentRpc.NewAuthGrpcClient(authConn, agentToken, agentID)
 	authInterceptor, err := agentRpc.NewAuthInterceptor(authClient, 30*time.Minute)
@@ -181,7 +180,7 @@ func run(c *cli.Context) error {
 		return err
 	}
 
-	writeAgentId(agentID)
+	writeAgentId(agentID, agentIdConfigPath)
 
 	labels := map[string]string{
 		"hostname": hostname,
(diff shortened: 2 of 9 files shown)


asdil12 (approved):

.gitignore
```diff
@@ -25,6 +25,7 @@ __debug_bin
 web/dist/**
 !web/dist/.gitkeep
 web/node_modules/
+web/package-lock.json
```
6543, line 28: ?

cmd/agent/agent.go
```diff
@@ -280,3 +284,31 @@ func stringSliceAddToMap(sl []string, m map[string]string) error {
...
+	const defaultAgentIdValue = int64(-1)
+
+	rawAgentId, fileErr := os.ReadFile(agentIdConfigPath)
+	if fileErr != nil {
+		return defaultAgentIdValue
+	}
+
+	strAgentId := strings.TrimSpace(string(rawAgentId))
+	agentId, err := strconv.ParseInt(strAgentId, 10, 64)
+	if err != nil {
```
6543, line 298: log error

cmd/agent/agent.go
```diff
@@ -280,3 +284,31 @@ func stringSliceAddToMap(sl []string, m map[string]string) error {
 	}
 	return nil
 }
+
+func readAgentId(agentIdConfigPath string) int64 {
+	const defaultAgentIdValue = int64(-1)
+
+	rawAgentId, fileErr := os.ReadFile(agentIdConfigPath)
+	if fileErr != nil {
```
6543, line 292: log err

cmd/agent/flags.go
```diff
@@ -67,6 +67,12 @@ var flags = []cli.Flag{
 		Name:    "hostname",
 		Usage:   "agent hostname",
 	},
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_AGENT_ID_CONFIG_PATH"},
+		Name:    "agent-id-config-path",
+		Usage:   "agent-id config file path",
+		Value:   "/etc/agent-id.conf",
```
6543, line 74:
```suggestion
		Value:   "/etc/woodpecker/agent-id.conf",
```
?

6543: also we might move this value outside into a global defined var ... so we can change it on build time (differ it for docker builds and binarys etc...)

ttomasini:
- [ ] add new env variable to documentation: `cs/docs/30-administration/15-agent-config.md`

cmd/agent/flags.go
ttomasini: > also we might move this value outside into a global defined var ... so we can change it on build time (differ it for docker builds and binarys etc...)

I've been thinking about this a bit, and I'm not sure if it doesn't cause more problems than it ends up simplifying. If the default value is changed at build time, it no longer matches the documentation, nor does it match the default value in the flags.go file. This could confuse users. But at the end of the day, I guess it doesn't make a difference where the agent-id conf file is in the Docker container.

So my suggestion would be to use the path you suggested for all applications and not introduce a globally defined var. What do you think?

cmd/agent/agent.go
anbraten: Should we make it debug as it will otherwise popup each time at the first start for normal runs.

zc-devs:
Do we really need a config file? Why? Why a file per option? May be one (few) yaml config(s) is better?

My proposals:
1. Introduce env var `WOODPECKER_AGENT_ID`. It is more convenient to configure via env vars, at least in Kubernetes. It is more consistent with current configuration also.
2. Use UUID instead of Int.
3. Drop config. If not, use one yaml config. Or leave as is.

If we want to automatically generate ID by agent, then
1. Try load ID from `WOODPECKER_AGENT_ID`.
2. If doesn't exist, generate new UUID and store in file. In this case separate file could be better.

zc-devs:
> Perhaps hostname is more readable.

There is `Name (name of agent)` field. So hostname suits better here, I think...

ttomasini:
> Do we really need a config file? Why? Why a file per option? May be one (few) yaml config(s) is better?
> 
> My proposals:
> 
> 1. Introduce an env var `WOODPECKER_AGENT_ID`. It is more convenient to configure via env vars, at least in Kubernetes. It is more consistent with current configuration also.
> 2. Use UUID instead of Int. Perhaps hostname is more readable.
> 3. Drop config. If not, use one yaml config. Or leave as is, but consider to rename file var to `WOODPECKER_AGENT_ID_FILE`.
> 
> If we want to automatically generate an ID by agent, then
> 
> 1. Try to load the ID from `WOODPECKER_AGENT_ID`.
> 2. If doesn't exist, generate new UUID and store it in the file. In this case the separate file could be better.

Hey @zc-devs 

I fully agree with most of your points. The main reason I chose this implementation it this way is backwards compatibility.

Currently the server generates the agent ID for the agent and so far the server can only handle int64 values. It was important to me to improve the current situation with as little effort as possible. This includes the choice of a simple configuration format. Especially since the 1.0.0 release is not far away, I didn't want to introduce any new complex changes here.

But I think, especially for the next new and breaking release, everything you suggest (especially the use of UUIDs) is highly preferable over the current implementation.

anbraten:
`WOODPECKER_AGENT_ID` is not necessary as you could already set the specific agent-token instead. The id file would be only interesting for agents using the system-token. The issue is that the system-token normally always creates a new agent in the DB and returns the id to the agent. This way the agent could store it to a file and at a later point access the server again using the system-token & agent-id combination. If using a specific agent-token the id can be directly identified by the server as the token would be unique.

zc-devs:
> Currently the server generates the agent ID for the agent

Ohh... That's sad...

Then env var should be renamed in consistent with current [state/docs](https://woodpecker-ci.org/docs/next/administration/server-config#handling-sensitive-data-in-docker-compose-and-docker-swarm):

> As it may be difficult to use docker secrets for environment variables woodpecker allows to read sensible data from files by providing a *_FILE option of all sensible configuration variables (WOODPECKER_AGENT_SECRET_FILE, WOODPECKER_DATABASE_DATASOURCE_FILE, WOODPECKER_ENCRYPTION_KEY_FILE ...)

zc-devs:
@anbraten thanks for great explanation. Things became much more clear. I'll add agent registration process in docs later, if no one do it ahead.

> The id file would be only interesting for agents using the system-token. The issue is that the system-token normally always creates a new agent in the DB and returns the id to the agent. This way the agent could store it to a file and at a later point access the server again using the system-token & agent-id combination.

Yeah, I see
Screenshot 2023-06-29 121043

> WOODPECKER_AGENT_ID is not necessary as you could already set the specific agent-token instead. If using a specific agent-token the id can be directly identified by the server as the token would be unique.

That is good news. I'll try it.

In [#1648](https://github.com/woodpecker-ci/woodpecker/issues/1648#issuecomment-1509054480) I forgot to add `WOODPECKER_AGENT_SECRET` (__system token__) in the server configuration. So, the agent could not register on the server. When I added __specific agent token__ manually, the agent registered successfully.

Eventually I realized mistake and added system token in the server configuration. It works also - I have duplicates now :)

6543:
please update branch - so ci runs again as expected :)

cmd/agent/flags.go
6543: well agents did not had a state before ... so we should just go with  `/etc/woodpecker/`* ...

docs/docs/30-administration/15-agent-config.md
```diff
@@ -80,6 +80,11 @@ Disable colored debug output.
 
 Configures the agent hostname.
 
+### `WOODPECKER_AGENT_ID_FILE`
```
6543, line 83: we should add to all examples the volume `/etc/woodpecker/` so it is persistent

docker/Dockerfile.agent.alpine.multiarch
```diff
@@ -6,13 +6,15 @@ ARG TARGETOS TARGETARCH
...
     make build-agent
+RUN mkdir -p /etc/woodpecker
 
 FROM alpine:3.18
 RUN apk add -U --no-cache ca-certificates
 ENV GODEBUG=netdns=go
 EXPOSE 3000
 
 COPY --from=build /src/dist/woodpecker-agent /bin/
+COPY --from=build /etc/woodpecker /etc
```
6543, line 17:
```suggestion
RUN mkdir -p /etc/woodpecker
```

docker/Dockerfile.agent.alpine.multiarch
```diff
@@ -6,13 +6,15 @@ ARG TARGETOS TARGETARCH
 RUN --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/go/pkg \
     make build-agent
+RUN mkdir -p /etc/woodpecker
```
6543, line 9:
```suggestion
```

docker/Dockerfile.agent.multiarch
```diff
@@ -15,6 +16,7 @@ EXPOSE 3000
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 # copy agent binary
 COPY --from=build /src/dist/woodpecker-agent /bin/
+COPY --from=build /etc/woodpecker /etc
```
6543, line 19:
```suggestion
RUN mkdir -p /etc/woodpecker
```

docker/Dockerfile.agent.multiarch
```diff
@@ -6,6 +6,7 @@ ARG TARGETOS TARGETARCH
 RUN --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/go/pkg \
     make build-agent
+RUN mkdir -p /etc/woodpecker
```
6543, line 9:
```suggestion
```

docker/Dockerfile.agent.multiarch
ttomasini: It's not possible to run mkdir in a scratch image since there is not even a BusyBox binary. To be consistent, I did the same in the alpine Dockerfile.

docker/Dockerfile.agent.multiarch
ttomasini: See comment above.

docs/docs/30-administration/15-agent-config.md
ttomasini: I added it as an example for the other file config of the agent: `WOODPECKER_AGENT_SECRET_FILE`

ttomasini:
> please update branch - so ci runs again as expected :)

```log
github.com/woodpecker-ci/woodpecker/cmd/server imports
	github.com/gin-gonic/gin imports
	github.com/gin-gonic/gin/binding imports
	github.com/go-playground/validator/v10 imports
	github.com/gabriel-vasile/mimetype: github.com/gabriel-vasile/mimetype@v1.4.2: reading https://proxy.golang.org/github.com/gabriel-vasile/mimetype/@v/v1.4.2.zip: 403 Forbidden
```

anbraten (approved):

ttomasini:
I am not sure what is happening here but it looks like the pipeline has gone through successfully but github is still waiting for feedback:
https://ci.woodpecker-ci.org/repos/3780/pipeline/6298/1
image

It is a bit but not fully simiar to https://github.com/woodpecker-ci/woodpecker/issues/1895, any ideas?

PR #1901, opened 2023-06-29
Optimize pipeline and switch to "steps"

6543:


.woodpecker/binaries.yml
```diff
@@ -2,6 +2,9 @@ depends_on:
   - test
   - web
 
+when:
```
6543, line 5: binaries and docker do quite the same ... and it's artifacts are only used on releases

anbraten (approved):

PR #1909, opened 2023-07-01
Refactored K8s Pod node selector

zc-devs:
- using existing map instead of an array
- applying var name convention
- using arch constant


zc-devs:
Pod config:
```
...
  containers:
    - name: wp-01h49cmzpzapdq3r1bkqbrttxa-0-step-0
      image: alpine
      ...
      imagePullPolicy: Always
      ...
  nodeSelector:
    kubernetes.io/arch: amd64
  ...

```

pipeline/backend/kubernetes/pod.go
```diff
@@ -92,18 +93,19 @@ func Pod(namespace string, step *types.Step, labels, annotations map[string]stri
...
 	labels["step"] = podName
 
-	var platform string
-	for _, e := range mapToEnvVars(step.Environment) {
-		if e.Name == "CI_SYSTEM_ARCH" {
-			platform = e.Value
-			break
-		}
+	var nodeSelector map[string]string
+	if platform, exist := step.Environment["CI_SYSTEM_ARCH"]; exist {
```
6543, line 97: step.platform ?

6543: todo for me: check if this was not a bug in the first place :)

zc-devs: > step.platform

There is no such field in step model.

Also there is deprecation warning in [environment.go](https://github.com/woodpecker-ci/woodpecker/blob/ae2bfe666f08d065ee4de0a22a8f56f747df9445/pipeline/frontend/metadata/environment.go#L116):
```
// DEPRECATED
"CI_SYSTEM_ARCH": m.Sys.Platform, // TODO: remove after v1.0.x version
```

6543 (approved):

Issue #1912, opened 2023-07-01
Respecting Kubernetes Default image pull policy


zc-devs:
### Clear and concise description of the problem

When I set up pipeline like
```
pipeline:
  test:
    image: alpine
    commands:
      - echo Hello
```
or like
```
pipeline:
  test:
    image: alpine:latest
    commands:
      - echo Hello
```
I'm expecting [behavior](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting):
> if you omit the imagePullPolicy field, and the tag for the container image is :latest, imagePullPolicy is automatically set to Always;
> if you omit the imagePullPolicy field, and you don't specify the tag for the container image, imagePullPolicy is automatically set to Always;

In other words I'm expecting [checking for a new image always](https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy), not sticking with current cached image.

I know that there is the [statement in docs](https://woodpecker-ci.org/docs/next/usage/pipeline-syntax#image)
> Woodpecker does not automatically upgrade container images. Use `pull: true`.

and  `ImagePullPolicy` [is always set](https://github.com/woodpecker-ci/woodpecker/blob/3d435a9cb73cd8f49f030d58568cae7890f2c5d2/pipeline/backend/kubernetes/pod.go#L46)... So, technically all works as intendent... But I didn't set image pull policy explicitly and expected default Kubernetes behavior.

Is it good idea to align with Kubernetes [Default image pull policy](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting)? At least it shouldn't be so difficult to implement. What do you think?

### Suggested solution

If step's `pull: false` or is not set, then do not fill `v1.Container.ImagePullPolicy` field.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

zc-devs:
[...]

PR #1914, opened 2023-07-01
Do not set Pod's Image pull policy if not explicitly set

zc-devs:
Address #1912.

Test pipeline:
```
pipeline:
  test:
    image: alpine
    commands:
      - apk add bash
      - bash -c 'if [ -v TESTVAR ]; then echo TESTVAR exist and value is $TESTVAR; fi'
```

Pod config:
```
...
  containers:
    - name: wp-01h49cmzpzapdq3r1bkqbrttxa-0-step-0
      image: alpine
      ...
      imagePullPolicy: Always
      ...
```


pipeline/backend/kubernetes/pod.go
```diff
@@ -43,7 +43,7 @@ func Pod(namespace string, step *types.Step, labels, annotations map[string]stri
 		}
 	}
 
-	pullPolicy := v1.PullIfNotPresent
+	var pullPolicy v1.PullPolicy
```
6543, line 46: Pro: the policy can be set by cluster
Con: if policy is by default no-pull and image dont existvlocally step will fail

zc-devs: > if policy is by default no-pull and image dont exist locally step will fail

Do you mean `Never` by `no-pull`?

I didn't find how default pull policy can be changed. If you know, please, refer to it.

There may be some admission controller like [AlwaysPullImages](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#alwayspullimages), but for `Never`. In this case administrator explicitly sets this and know what he's doing. He have to populate cache with desired image. As in general, not only with woodpecker.

zc-devs: Maybe there Kuberenetes implementations exist, which doesn't follow official docs...

Anyway, suppose cluster have different default image pull policy. In this case we can introduce configuration parameter `WOODPECKER_BACKEND_K8S_DEFAULT_POD_IMAGE_PULL_POLICY`.

Then if step's `pull` is not defined, look into `WOODPECKER_BACKEND_K8S_DEFAULT_POD_IMAGE_PULL_POLICY`. If `WOODPECKER_BACKEND_K8S_DEFAULT_POD_IMAGE_PULL_POLICY` is not defined, then `pullPolicy == nil` and using cluster default.

6543 (approved):

PR #1921, opened 2023-07-03
Added agent registration docs

zc-devs:
As [was promised](https://github.com/woodpecker-ci/woodpecker/pull/1888#issuecomment-1612720214).

Feel free to make it better.


6543:
thanks :)

PR #1930, opened 2023-07-04
Add Kubernetes and Dockle Plugins

zguig52:
Add Kubernetes Deployments and StatefulSet update and Dockle Scan Plugins.

For Kubernetes plugin, I based on the Drone unmaintened Kubernetes plugin and took the statefulset management evolutions. I added sync/wait and force redeploy capabilities + updates dependencies

For Dockle plugin, I took example on Trivy plugin.


6543 (approved):
thanks for also catching the wrong verified

PR #1938, opened 2023-07-07
Make secrets with "/" in name editable / deletable

### LINKED ISSUE #1937: Cannot remove/delete/trash a secret
### Component

web-ui

### Describe the bug

Create a secret (I guess the name is important for this bug).

Then try to remove it with the trash button.

image

Note : I managed to remove another secret. I think that there is a problem with the name of the secret I cannot remove.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-94f2331a"}

Backend Kubernetes
```

### Additional context

_No response_

### Validations

6543:
close #1937

### DIFF
--- a/web/src/lib/api/index.ts
+++ b/web/src/lib/api/index.ts
@@ -145,11 +145,13 @@ export default class WoodpeckerClient extends ApiClient {
   }
 
   updateSecret(repoId: number, secret: Partial<Secret>): Promise<unknown> {
-    return this._patch(`/api/repos/${repoId}/secrets/${secret.name}`, secret);
+    const secretName = encodeURIComponent(secret.name ?? '');
+    return this._patch(`/api/repos/${repoId}/secrets/encodeURIComponent(${secretName}`, secret);
   }
 
   deleteSecret(repoId: number, secretName: string): Promise<unknown> {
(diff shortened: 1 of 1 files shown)


web/src/lib/api/index.ts
```diff
@@ -145,11 +145,13 @@ export default class WoodpeckerClient extends ApiClient {
   }
 
   updateSecret(repoId: number, secret: Partial<Secret>): Promise<unknown> {
-    return this._patch(`/api/repos/${repoId}/secrets/${secret.name}`, secret);
+    const secretName = encodeURIComponent(secret.name ?? '');
+    return this._patch(`/api/repos/${repoId}/secrets/encodeURIComponent(${secretName}`, secret);
```
anbraten, line 149:
```suggestion
    return this._patch(`/api/repos/${repoId}/secrets/${secretName}`, secret);
```

anbraten (commented):
Should we migrate to id based api as well?

6543:
:point_up: sory was a relict that should not be in there - thanks

6543:
> Should we migrate to id based api as well?

well I would move this breaking refactor if so in the next release!

lafriks (approved):

PR #1957, opened 2023-07-09
Remove unused step options

6543:
remove them as they did never had any functionality


pipeline/frontend/yaml/types/container.go
```diff
@@ -38,26 +38,21 @@ type (
...
+		DNS          base.StringOrSlice  `yaml:"dns,omitempty"`
+		ExtraHosts   []string            `yaml:"extra_hosts,omitempty"`
+		IpcMode      string              `yaml:"ipc_mode,omitempty"`
+		MemLimit     base.MemStringOrInt `yaml:"mem_limit,omitempty"`
+		MemSwapLimit base.MemStringOrInt `yaml:"memswap_limit,omitempty"`
+		NetworkMode  string              `yaml:"network_mode,omitempty"`
+		Networks     Networks            `yaml:"networks,omitempty"`
+		ShmSize      base.MemStringOrInt `yaml:"shm_size,omitempty"`
+		Sysctls      base.SliceOrMap     `yaml:"sysctls,omitempty"`
+		Tmpfs        []string            `yaml:"tmpfs,omitempty"`
```
anbraten, line 55: Crazy so all left options are actually supported?

anbraten (approved):

pipeline/frontend/yaml/types/container.go
6543: I did not follow them ... some are still dead code but need more refactoring - so for the next pull

Issue #557, opened 2021-11-30
prevent service name to be equal to step name


anbraten:
![Unbenannt](https://user-images.githubusercontent.com/6918444/144145897-c17d0299-aad0-4594-8ee2-f19b825c8a21.PNG)

postgres and mysql jobs are skipped somehow

anbraten:
Issue detected: Service name is not allowed to be equal to one step name.
TODO: add config check to prevent that.

6543:
instead of preventing it we should fix the underling bug

anbraten:
```diff
// add services steps
if len(conf.Services.Containers) != 0 {
	stage := new(backend.Stage)
	stage.Name = fmt.Sprintf("%s_services", c.prefix)
	stage.Alias = "services"

	for i, container := range conf.Services.Containers {
		if !container.Constraints.Match(c.metadata) {
			continue
		}

+		name := fmt.Sprintf("%s_services_%d", c.prefix, i)
		step := c.createProcess(name, container, "services")
		stage.Steps = append(stage.Steps, step)
	}
	config.Stages = append(config.Stages, stage)
}
```

Hmm, the compiler should not be the problem.

6543:
without looking at the coude, i guess it's an issue with the gRPC api

6543:
https://github.com/woodpecker-ci/woodpecker/blob/master/pipeline/rpc/proto/woodpecker.proto#L92-L95

look's like the name is actually used as ID for gRPC - we just have to change that

anbraten:
linking #983

6543:
-> #1802

Issue #269, opened 2021-08-20
UI: separate services and build-jobs 


anbraten:
It would be nice if a user could somehow notice that a build-job is of type service.

6543:
indicate it by different icons?

6543:
idea: instead of a dot -> add a gear (not spinning if cancled/pending/...  & spinning if state = running)

dvjn:
Currently, I couldn't find a way to differentiate a proc between service and build-job in the front-end.
This will require changes in server api first.

6543:
good point!

6543:
there is now a `type` field in the step struct. so only the ui part is missing :)

Issue #1959, opened 2023-07-09
docker: use uuid instead of name as identifyer


6543:
https://github.com/woodpecker-ci/woodpecker/blob/3d0338315f14a33a9b9a205168eb5a71579fc591/pipeline/backend/docker/docker.go#L197

https://github.com/woodpecker-ci/woodpecker/blob/3d0338315f14a33a9b9a205168eb5a71579fc591/pipeline/backend/docker/docker.go#L219

...

will make sure,  #557 stays closed

PR #1960, opened 2023-07-09
remove prefix from services

6543:
test  #557

should be merged if  #269 got addressed


6543:
https://ci.woodpecker-ci.org/repos/3780/pipeline/6440/25 - looks still broken :D

6543:
it does not crash anymore but we still will likely need  #1959

PR #1964, opened 2023-07-09
Fix netrc passing to clone image

### LINKED ISSUE #1837: Git clone fails due to trusted protection for netrc
### Component

server, agent

### Describe the bug

Seems like the normal clone step fails on next-101e684059 when the repo is not trusted due to the "Only inject netrc credentials into trusted containers" option that is now set by default

Is that a bug or am I doing something wrong ?

### System Info

```shell
next-101e684059
```

### Additional context

My CLone config:
```
clone:
  git:
    image: woodpeckerci/plugin-git
    settings:
      tags: true
    when:
      event: tag
```
### Validations

anbraten:
closes #1837

### DIFF
--- a/pipeline/frontend/yaml/utils/image.go
+++ b/pipeline/frontend/yaml/utils/image.go
@@ -1,4 +1,4 @@
-package compiler
+package utils
 
 import "github.com/docker/distribution/reference"
 
@@ -33,7 +33,7 @@ func expandImage(name string) string {
 // matchImage returns true if the image name matches
 // an image in the list. Note the image tag is not used
 // in the matching logic.
-func matchImage(from string, to ...string) bool {
+func MatchImage(from string, to ...string) bool {
 	from = trimImage(from)
 	for _, match := range to {
 		if from == trimImage(match) {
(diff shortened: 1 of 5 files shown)


pipeline/frontend/yaml/utils/image.go
```diff
@@ -33,7 +33,7 @@ func expandImage(name string) string {
 // matchImage returns true if the image name matches
```
lafriks, line 33:
```suggestion
// MatchImage returns true if the image name matches
```

pipeline/frontend/yaml/utils/image.go
```diff
@@ -45,7 +45,7 @@ func matchImage(from string, to ...string) bool {
 
 // matchHostname returns true if the image hostname
```
lafriks, line 46:
```suggestion
// MatchHostname returns true if the image hostname
```

lafriks (changes requested):

lafriks (approved):

6543 (approved):

PR #1966, opened 2023-07-09
Speedup pipeline

6543:
- only build preview images for x86
- do not run `test` and `web` on default branch

### DIFF
--- a/.woodpecker/docker.yml
+++ b/.woodpecker/docker.yml
@@ -9,7 +9,7 @@ variables:
   - &xgo_version 'go-1.20.2'
   - &platforms_release 'linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/386,linux/amd64,linux/ppc64le,linux/riscv64,linux/s390x,windows/amd64,freebsd/arm64,freebsd/amd64,openbsd/arm64,openbsd/amd64'
   - &platforms_server 'linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le,linux/riscv64'
-  - &platforms_preview 'linux/arm64/v8,linux/amd64'
+  - &platforms_preview 'linux/amd64'
   - &platforms_alpine 'linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le'
 
 steps:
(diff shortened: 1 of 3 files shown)


.woodpecker/test.yml
```diff
@@ -9,21 +10,23 @@ variables:
       - "go.*"
       # schema changes
       - "pipeline/schema/**"
+    branch:
+      exclude: ${CI_REPO_DEFAULT_BRANCH}
```
6543, line 14: uh we need to match that with push env ...

anbraten:
Could we even skip arm completely for normal PR tests as they are still pretty slow due to the emulation part. If we want to test them / provide test images we could add a step with a label filter.

.woodpecker/docker.yml
```diff
@@ -9,7 +9,7 @@ variables:
   - &xgo_version 'go-1.20.2'
   - &platforms_release 'linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/386,linux/amd64,linux/ppc64le,linux/riscv64,linux/s390x,windows/amd64,freebsd/arm64,freebsd/amd64,openbsd/arm64,openbsd/amd64'
   - &platforms_server 'linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le,linux/riscv64'
-  - &platforms_preview 'linux/arm/v6,linux/arm64/v8,linux/amd64,linux/riscv64,windows/amd64'
+  - &platforms_preview 'linux/arm64/v8,linux/amd64'
```
6543, line 12:
```suggestion
  - &platforms_preview 'linux/amd64'
```

.woodpecker/docker.yml
```diff
@@ -30,17 +30,19 @@ steps:
...
+  cross-compile-server-preview:
     image: *xgo_image
     pull: true
     commands:
       - apt update
       - apt install -y tree
       - make cross-compile-server
     environment:
-      PLATFORMS: linux|arm/v7;linux|arm64/v8;linux|amd64;linux|ppc64le;linux|riscv64
+      PLATFORMS: linux|arm64/v8;linux|amd64
```
6543, line 41:
```suggestion
      PLATFORMS: linux|amd64
```

PR #1971, opened 2023-07-10
Add agent config file

6543:
followup of  #1888

### DIFF
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -23,7 +23,6 @@ import (
 	"net/http"
 	"os"
 	"runtime"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -50,7 +49,7 @@ import (
 )
 
 func run(c *cli.Context) error {
-	agentIDConfigPath := c.String("agent-id-config-path")
+	agentConfigPath := c.String("agent-config")
 	hostname := c.String("hostname")
 	if len(hostname) == 0 {
 		hostname, _ = os.Hostname()
@@ -111,9 +110,15 @@ func run(c *cli.Context) error {
 	}
 	defer authConn.Close()
 
-	agentID := readAgentID(agentIDConfigPath)
+	agentConfig := readAgentConfig(agentConfigPath)
+
+	//deprecated
+	if agentConfig.AgentID == defaultAgentIDValue {
+		agentConfig.AgentID = readAgentID(c.String("agent-id-config-path"))
+	}
+
 	agentToken := c.String("grpc-token")
-	authClient := agentRpc.NewAuthGrpcClient(authConn, agentToken, agentID)
+	authClient := agentRpc.NewAuthGrpcClient(authConn, agentToken, agentConfig.AgentID)
 	authInterceptor, err := agentRpc.NewAuthInterceptor(authClient, 30*time.Minute)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


6543:
it will break stuff if volume got RO intentionaly set ...

cmd/agent/agent.go
```diff
@@ -111,9 +110,15 @@ func run(c *cli.Context) error {
 	}
 	defer authConn.Close()
 
-	agentID := readAgentID(agentIDConfigPath)
+	agentConfig := readAgentConfig(agentConfigPath)
+
+	// deprecated
+	if agentConfig.AgentID == defaultAgentIDValue {
+		agentConfig.AgentID = readAgentID(c.String("agent-id-config-path"))
```
zc-devs, line 117: If PR is the breaking changes, for why is this compatibility? Considering that these changes will be released in 1.0.0.

6543: Well people use next already so there is a migration path ... 

... so it is breaking, but not for the release ... thats why no mentioning in docs and changelog skip

We can remove depricated code soon too ...

zc-devs: OK, then it's not a breaking change in current implementation. It's deprecation and some warning should be logged.
image

6543: well that wont appear either ... as it never got released in the first place ... so we dont have to document the removeall after this pull

zc-devs:
And it doesn't look as refactoring for me :)

cmd/agent/agent.go
6543: I'm just "nice" with the migration path, that will stick in for a few days before removal

6543:
feature-set & behaviour is the same, just an more generic way to be future-prove on how to de-/serialize config

zc-devs:
> feature-set & behaviour is the same, just an more generic way to be future-prove on how to de-/serialize config
> 
> in this case we only have the AgentID

was replied to my deleted message

> It doesn't look as refactoring to me.

So, I agree.

6543:
hu - who deleted your message?!?

cmd/agent/agent.go
zc-devs: Can't  mark as resolved (don't know why). So, do it, if you can.

zc-devs:
Nobody, but me :(

Then I realized that it's just additional path for now -> previous behavior exists still -> may be refactoring :)

cmd/agent/agent.go
```diff
@@ -175,12 +180,12 @@ func run(c *cli.Context) error {
 		return err
 	}
 
-	agentID, err = client.RegisterAgent(ctx, platform, engine.Name(), version.String(), parallel)
+	agentConfig.AgentID, err = client.RegisterAgent(ctx, platform, engine.Name(), version.String(), parallel)
```
ttomasini, line 183: I would introduce a new variable to avoid confusion and treat `agentConfig` as an immutable since it should represent the actual config.

6543: we save it later so we have to change agentConfig anyway

zc-devs:
Maybe it's good idea to add `VOLUME /etc/woodpecker` in the Dokerfiles.

lafriks (approved):

docs/docs/30-administration/00-setup.md
```diff
@@ -70,6 +70,7 @@ services:
     depends_on:
       - woodpecker-server
     volumes:
+      - woodpecker-agent-config:/etc/woodpecker
```
jolheiser, line 73: Assuming this is meant to be a local file, it should be `./woodpecker-agent-config`.

Just wanting to clarify here prior to opening a patch for it.

6543: well it depends, I interpreted it as named volume ... same as the example show for the database for the server

jolheiser: Okay, then it should be added to the named volumes at the bottom of the example. I can send a quick change for that, just to clarify in docs

6543: @jolheiser sure go on, so we do have it in the v1.0.0 docks :D

6543: #2025

PR #1974, opened 2023-07-10
Correct api docs of PostRepo & GetRepos

6543:
... and report if needed forge_remote_id is not valid


server/api/repo.go
```diff
@@ -39,19 +39,24 @@ import (
...
 //	@Summary	Activate a repository
-//	@Router		/repos/{repo_id} [post]
+//	@Router		/repos [post]
 //	@Produce	json
 //	@Success	200	{object}	Repo
 //	@Tags		Repositories
-//	@Param		Authorization	header	string	true	"Insert your personal access token"	default(Bearer <personal access token>)
-//	@Param		repo_id			path	int		true	"the repository id"
+//	@Param		Authorization			header	string	true	"Insert your personal access token"	default(Bearer <personal access token>)
+//	@Param		forge_remote_id		query		string	true	"the repository id it has at the forge"
```
anbraten, line 47:
```suggestion
//	@Param		forge_remote_id		query		string	true	"the id of a repository at the forge"
```

server/api/repo.go
```diff
@@ -39,19 +39,24 @@ import (
...
 func PostRepo(c *gin.Context) {
 	forge := server.Config.Services.Forge
 	_store := store.FromContext(c)
 	user := session.User(c)
 
-	forgeRemoteID := model.ForgeRemoteID(c.Query("forge_remote_id"))
-	repo, err := _store.GetRepoForgeID(forgeRemoteID)
+	remoteID := model.ForgeRemoteID(c.Query("forge_remote_id"))
```
anbraten, line 53: Why the renaming?

6543: Is schorter ... and the name sounded more like id of forge to me

server/api/repo.go
```diff
@@ -61,7 +66,7 @@ func PostRepo(c *gin.Context) {
 		return
 	}
 
-	from, err := forge.Repo(c, user, forgeRemoteID, "", "")
+	from, err := forge.Repo(c, user, remoteID, "", "")
```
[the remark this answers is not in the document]
6543: well that's not the scope of this pull ... if it is an issue we should make a new pull/issue

server/api/repo.go
anbraten: It would be better matching and consistent with the query name, but whatever.

6543: @anbraten done

anbraten (approved):

PR #1985, opened 2023-07-12
Add VOLUME to dockerfiles

6543:
We need to check if it does not have any downsides ...

address: https://github.com/woodpecker-ci/woodpecker/pull/1971#issuecomment-1632406047


pat-s:
https://stackoverflow.com/a/55516433/4185785 for discussion

6543:
... "Every time you create a container from this image, docker will force that directory to be a volume. " ...

PR #1992, opened 2023-07-13
Do not use `window.WOODPECKER_*` variables directly

qwerty287:
use the config helper instead


6543 (approved):

6543:
I wonder if we could enforce this via a linter?

PR #1994, opened 2023-07-13
Drop support for Bitbucket Server

### LINKED ISSUE #1962: Drop Bitbucket Server support
We currently have support for Bitbucket cloud and bitbucket server (formerly stash). As bitbucket server will no longer be supported starting Febuary 15, 2024 and is only partly supported / implemented by Woodpecker we should drop support.

image

qwerty287:
Closes #1962


anbraten:
thanks

PR #1997, opened 2023-07-14
Use global logger for xorm logs and add options

xoxys:
Please let me know if this goes in the right direction. Needs some tests and docs.

### DIFF
--- a/server/store/common.go
+++ b/server/store/common.go
@@ -14,8 +14,14 @@
 
 package store
 
+type XORM struct {
+	Log     bool
+	ShowSQL bool
+}
+
 // Opts are options for a new database connection
 type Opts struct {
 	Driver string
 	Config string
+	XORM   *XORM
 }
--- a/server/store/common.go
+++ b/server/store/common.go
@@ -23,5 +23,5 @@ type XORM struct {
 type Opts struct {
 	Driver string
 	Config string
-	XORM   *XORM
+	XORM   XORM
 }
--- a/server/store/datastore/xorm.go
+++ b/server/store/datastore/xorm.go
@@ -21,88 +21,86 @@ import (
 	xlog "xorm.io/xorm/log"
 )
 
-func NewXORMLogger(level xlog.LogLevel) *XORMLogger {
-	return &XORMLogger{
+func newXORMLogger(level xlog.LogLevel) xlog.Logger {
+	return &xormLogger{
 		level: level,
 	}
 }
 
-// XORMLogger custom log implementation for ILogger
-type XORMLogger struct {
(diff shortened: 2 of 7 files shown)


server/store/common.go
```diff
@@ -14,8 +14,14 @@
...
+type XORM struct {
+	Log     bool
+	ShowSQL bool
+}
+
 // Opts are options for a new database connection
 type Opts struct {
 	Driver string
 	Config string
+	XORM   *XORM
```
6543, line 26:
```suggestion
	XORM   XORM
```

make sure it's not nil

server/store/datastore/xorm.go
```diff
@@ -0,0 +1,116 @@
...
+package datastore
+
+import (
+	"fmt"
+
+	"github.com/rs/zerolog/log"
+	xlog "xorm.io/xorm/log"
+)
+
+func NewXORMLogger(level xlog.LogLevel) *XORMLogger {
```
6543, line 24: unexport and return type can be xorm-logger-interface directly

server/store/datastore/xorm.go
```diff
@@ -0,0 +1,116 @@
...
+	return &XORMLogger{
+		level: level,
+	}
+}
+
+// XORMLogger custom log implementation for ILogger
+type XORMLogger struct {
```
6543, line 31: unexporte

6543:
yes that's the right direction :+1:  !

server/store/datastore/xorm.go
```diff
@@ -0,0 +1,114 @@
...
+// xormLogger custom log implementation for ILogger
+type xormLogger struct {
+	level   xlog.LogLevel
+	showSQL bool
+}
+
+// Error implement ILogger
+func (x *xormLogger) Error(v ...interface{}) {
+	if x.level <= xlog.LOG_ERR {
+		log.Error().Msg(fmt.Sprintln(v...))
```
6543, line 39: we should add some string to indicate it's from xorm

log.Str("component", "xorm")

or log.Bool("xorm": true)
....

?

6543: what do you think?

xoxys: Yeah, that makes sense to me :+1:  Will finish the PR later.

xoxys: I have used the `"component", "xorm"` format. If someone wants to parse the logs in a central log mgm system, using a consistent field to get the component name might be better than a changing bool field.

6543 (approved):

6543:
No, just change current docs ...

And on branch out, versionated docs are generated ...

Issue #1079, opened 2022-08-06
Customisable theming


tecosaur:
### Clear and concise description of the problem

It could be rather nice to be able to switch out the theme used by Woodpecker.

### Suggested solution

It would be great if one could provide an alternative set of colours/styles (via a `theme.css` file) to be used.

### Additional context

This issue has been spun off from https://github.com/woodpecker-ci/woodpecker/issues/468#issuecomment-1207185270

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

silverwind:
All colors should be extracted to CSS variables on the `:root` node. This will make designing custom color themes a breeze.

mvdkleijn:
I would also suggest that an index of sorts be made available for themes, similar to plugins.

PR #2003, opened 2023-07-15
Use consistent woodpecker color scheme

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/1079

What do you think about using a consistent `woodpecker` color scheme? Right now, the `lime` color scheme from windicss is used that does not really fit the primary color used for the documentation website. I have used the primary color `#4CAF50` from the docs and created a color palette with https://palettte.app/:

<details>
  <summary>JSON source</summary>

```Json
[
  {
    "paletteName": "New Palette",
    "swatches": [
      {
        "name": "New Swatch",
        "color": "166E30"
      },
      {
        "name": "New Swatch",
        "color": "248438"
      },
      {
        "name": "New Swatch",
        "color": "369943"
      },
      {
        "name": "New Swatch",
        "color": "4CAF50"
      },
      {
        "name": "New Swatch",
        "color": "68C464"
      },
      {
        "name": "New Swatch",
        "color": "8AD97F"
      }
    ]
  }
]
```

</details>

image

I have added this color scheme to the windicss config and replaced the use of `lime` in the UI. While `woodpecker-300` would be the primary color that is used for the docs, I currently use `woodpecke-400` as primary color for the UI to fix some contrast issues.   

image
image

What do you think? If you would like to stay with the current colors, that's fine for me, I can just use the custom CSS feature in this case.


pat-s:
I like it! Same adaption for the dark scheme?

(I personally didn't like the lime green that much so far 😅 )

6543:
I'm a dark-theme user so it doesn't bother me at all :sweat_smile:

xoxys:
> Same adaption for the dark scheme?


Yes, but there are probably still some parts missing. I didnt want to spend too much time before a decision was made.

6543:
Well if we already go down this rabbitwhile what about using css-vars so we can have  #1079

xoxys:
Yes, that would help a lot. Happy to look into it.

6543:
Or do windiCSS hve some theming feature build in?

xoxys:
Windicss is another thing to reconsider as it is sunsetting https://windicss.org/posts/sunsetting.html

xoxys:
Nevertheless, I think we can implement it that way:

```
module.exports = {
  theme: {
    colors: {
      red: {
        medium: '#e53e3e',
        dark: '#742a2a'
      },
      blue: {
        medium: '#3182ce',
        dark: '#2a4365'
      },
      'th-primary': 'var(--primary)',
      'th-secondary': 'var(--secondary)'
    }
  }
}

# style.css
:root {
  --primary: theme('colors.red.dark');
  --secondary: theme('colors.red.medium');
}
```

The `theme` function is supported by windicss as well https://windicss.org/features/directives.html#theme

Ref: https://css-tricks.com/color-theming-with-css-custom-properties-and-tailwind/#aa-using-the-config-as-a-style-guide

pat-s:
@xoxys ready?

xoxys:
Technically this will work, but I'm not happy with this approach. The problem is, it's still very hard for users to create color themes because the custom properties are generic and don't have a contextual meaning. IMO, we should only expose and use meaningful properties. Example:

```
[root]
--wp-primary
--wp-secondary

--wp-pipeline-pending
--wp-pipeline-ok
--wp-pipeline-running
--wp-pipeline-error

--wp-links

...
```

Opinions?

pat-s:
This sounds a bit like code refactoring which could go into a separate PR? I very much like the new color scheme and think it's a valuable change on it's own already.

xoxys:
I would prefer to do it together as a lot of testing is required and would need to be done twice. I have started the rework to use meaningful css properties now and can hopefully finish it over the weekend. For examples, take a look at https://github.com/woodpecker-ci/woodpecker/blob/97e2eb2b74b3b740181768a00a0f324faeb822ac/web/src/style.css

Some more impressions:

image
image
image
image

web/src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  --wp-state-error-100: theme('colors.int-wp-state-error-dark.100');
+  --wp-state-neutral-100: theme('colors.int-wp-state-neutral.100');
+  --wp-state-ok-100: theme('colors.int-wp-state-ok-dark.100');
+  --wp-state-info-100: theme('colors.int-wp-state-info-dark.100');
+  --wp-state-warn-100: theme('colors.int-wp-state-warn-dark.100');
+
+  --wp-hint-warn-100: theme('colors.int-wp-hint-warn-dark.100');
+  --wp-hint-warn-200: theme('colors.int-wp-hint-warn-dark.200');
+
+  --wp-code-100: theme('colors.int-wp-secondary.700');
```
anbraten, line 88: I like the color changes in general however we should change the naming / usage. I would highly prefer to use wp-secondary-700 directly instead of adding sth like `wp-code-100` as this works against the idea of tailwind / unocss / windicss to create content-full names instead of context-full.

xoxys: Yes, I agree that it is against the basic idea of the framework, but this makes it pretty hard for end users to create themes. We will end up with a list like this: https://github.com/woodpecker-ci/woodpecker/blob/6c5849c9420e105e7c1f0dc08c43040d37f3a55c/web/src/style.css

The end user does not know what `--wp-green-200` is used for, and I really think if we want to support custom color themes created by end users (https://github.com/woodpecker-ci/woodpecker/issues/1079) we should stick to semantic class names. If we decide to not implement this FR, I can just keep the color changes/harmonizing stuff and drop the CSS custom properties again. But it looks like there is still some discussion needed :smile: will put the work on hold for now.

xoxys: What can also lead to problems with user themes is the multiple use of generic classes. Just because someone wants to change the 
background color (e.g. gray.100), in most cases they will not want to change every not want to change every occurrence of gray.100.

xoxys:
Ready from my side.

lafriks:
imho the dark theme also needs a bit more contrast for bg/text

xoxys:
True, I have slightly adjusted some colors, and at least the Report from the Firefox Accessibility Tool looks good now.

image

6543:
hey this pull should take  #2052 into account and visa versa

6543:
e.g. https://egghead.io/blog/write-a-plugin-for-tailwind-css ... https://tailwindcss.com/docs/configuration ...

xoxys:
I had also mentioned this in https://github.com/woodpecker-ci/woodpecker/pull/2003#issuecomment-1636857393 without a reaction. As there are also some other design questions without a decision yet, I'll not put more time into this PR for now.

pat-s (approved):
> hey this pull should take https://github.com/woodpecker-ci/woodpecker/issues/2052 into account and visa versa

Can't we do it in small steps? This PR it big enough already and shouldn't do too many things at once.

I really like the current state and think it's good to merge. 
The tailwind transition should be a separate PR imo.

web/src/assets/locales/en.json
```diff
@@ -1,6 +1,6 @@
 {
     "cancel": "Cancel",
-    "login": "Login",
+    "login": "Sign in",
```
lafriks, line 3: This is an unrelated change and I don't think I like it

lafriks:
Generally I like everything except do we really need so many colors - 100..900? I know it's "tailwind way" but it makes a mess as there is no really consensus as to where each is used in the end in UI when we should really be using probably 2-3 colors at max

web/src/style.css
anbraten: Still scared / pretty sure that the new custom classes will be misused in the future and therefore mess up the code, so I would personally stick to tailwind classes. But I wont block here.

web/src/assets/locales/en.json
xoxys: Done

xoxys:
> Generally I like everything except do we really need so many colors - 100..900? I know it's "tailwind way" but it makes a mess as there is no really consensus as to where each is used in the end in UI when we should really be using probably 2-3 colors at max

Don't know how to improve it... all exposed colors are in use to properly work in light and dark mode to provide the required contrasts. Only a minimal subset is exposed in https://github.com/woodpecker-ci/woodpecker/blob/89a250040da9f826a08abb60a2f077217e7495ac/web/src/style.css

For the primary and secondary color, a broader range is defined in https://github.com/woodpecker-ci/woodpecker/blob/89a250040da9f826a08abb60a2f077217e7495ac/web/windi.config.ts to, but those are not exposed to users.

web/src/style.css
anbraten: I am referring to semantic classes like `text-wp-state-error` instead of `text-red-700`. Adding extra colors / adjusting existing one like `wp-cyan-800: #...` or `red-700` with a new hex value is not a problem IMO.

xoxys:
Will merge now, happy to add improvements/refactoring in a followup PR if needed.

PR #2008, opened 2023-07-17
Fix agent config reading error handling

lafriks:
Currently for non-existing config file it would report error instead of info message


6543 (approved):

6543:
link #1971

PR #2009, opened 2023-07-17
Remove code to read agent-id.conf

6543:
I want to have c805c87e9035898e712c918cbbdda9c4394013e5 published at least for 2 days ...
... so the migration did happen


zc-devs:
Is this breaking? :)

6543:
No related code never got released

Issue #1998, opened 2023-07-14
ui: back arrow button of pipeline view does not lead to repo 


6543:
the back button return to the previous site/url and not to the repo the pipeline do run in

regression of #1691 

![image](https://github.com/woodpecker-ci/woodpecker/assets/24977596/b492d9ce-c452-4607-8b13-a63174af0210)

pat-s:
I noticed this as well but was not motivated enough to report 😓 
There might be more "back button" issues - I often encounter situations where I would have expected something else to happen compared to what happened instead (sometimes followed by a red error message in the bottom right).

6543:
Well if you expect X and got Y ... it's bad UX design and should be fixed

pat-s:
I second your observation: 

> the back button return to the previous site/url and not to the repo the pipeline do run in

I'll try to describe other "unnatural" behavior next time I encounter it

PR #2028, opened 2023-07-23
Always use a static back route

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/1998

I don't know if there is a situation where using the browser history makes sense, or if we should always use a static route for the back buttons, but this fix at least preserves the ability to use the browser history.


6543:
no it makes no sense to use browser history

6543:
@xoxys can you refactor it to drop browser histry func

xoxys:
Ok so I will rewrite the function to always use the given route.

web/src/views/org/OrgSettings.vue
```diff
@@ -41,5 +41,5 @@ onMounted(async () => {
   }
 });
 
-const goBack = useRouteBackOrDefault({ name: 'repos-owner' });
+const goBack = useRouteBack({ name: 'repos-owner' });
```
6543, line 44: we should use the orgID route here :)

6543: yes that's also an bug :sweat:

web/src/views/repo/RepoSettings.vue
```diff
@@ -73,5 +73,5 @@ onMounted(async () => {
   }
 });
 
-const goBack = useRouteBackOrDefault({ name: 'repo' });
+const goBack = useRouteBack({ name: 'repo' });
```
6543, line 76: same here ... we should use the repoID route

web/src/views/repo/pipeline/PipelineWrapper.vue
```diff
@@ -196,5 +196,5 @@ const activeTab = computed({
   },
 });
 
-const goBack = useRouteBackOrDefault({ name: 'repo' });
+const goBack = useRouteBack({ name: 'repo' });
```
6543, line 199: same ... repoID

6543 (approved):

Issue #1943, opened 2023-07-07
Add an healthcheck to server container


6543:
like the agent:
https://github.com/woodpecker-ci/woodpecker/blob/205f8ecc1649ab48a9d5fb52b0ecd7b8c24abf4f/docker/Dockerfile.agent.multiarch#L21

ref https://docs.docker.com/engine/reference/builder/#healthcheck

TODO:
- [ ] add subcommand to server that can test against https://github.com/woodpecker-ci/woodpecker/blob/205f8ecc1649ab48a9d5fb52b0ecd7b8c24abf4f/server/router/router.go#L69
- [ ] use subcommand in container

6543:
Yes but we cant use wget and need a build in func for tje from-scratch container

6543:
but `next-45319b24cd` is not ... and so if we officially want to support it ... we have to do it that way

lafriks:
The best option would be to add a command line option that would call the health check endpoint and return the exit code based on response

6543:
That's exactly whatvi try to write 😅

PR #2030, opened 2023-07-24
Add ping command to server to allow container healthchecks

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/1943

Note: Kubernetes ignores the container `HEALTHCHECK` by default.


cmd/server/health.go
```diff
@@ -0,0 +1,43 @@
...
+
+// handles pinging the endpoint and returns an error if the
+// server is in an unhealthy state.
+func pinger(c *cli.Context) error {
+	serverAddr := c.String("server-addr")
+	if strings.HasPrefix(serverAddr, ":") {
+		// this seems sufficient according to https://pkg.go.dev/net#Dial
+		serverAddr = "localhost" + serverAddr
+	}
+	resp, err := http.Get("http://" + serverAddr + "/healthz")
```
6543, line 34: can we forward the ping to the database?

6543:
never mind :laughing:

cmd/server/health.go
6543, line 34: We need to check if woodpecker does ssl itselve and use https in this case instead

6543 (changes requested):

cmd/server/health.go
xoxys: That was my question in the initial post...

xoxys: Will fix it

6543: sure ...

6543 (approved):

6543 (approved):

xoxys:
But the ping command is executed as a standalone command. Why should I as a user set let's encrypt config flags etc. on the ping command execution just to get it to use HTTPS. Maybe a more user-friendly way would be something like `woodpecker-server ping --ping-url https://my-ci.example.com`.

If the ping command is executed on the same host in the same context as the server process with access to the same env vars, this might work, but in other situations it might not (even if this might not the desired use case).

xoxys:
However, thanks for finishing it.

6543:
I dont think we need a option to a custom url.
It should use the same config options as the main server.

users/admins should not use it like the cli tool ...

if we need such a feature, it should be added to the cli tool instead

PR #2047, opened 2023-07-27
Missed deprecation

6543:
in the refactoring we did move from CI_SYSTEM_ARCH to CI_SYSTEM_PLATFORM

but we missed to set the var in the agent pipeline tracer, so the Kubernetes backend did not work with it.

this fix it and also make it  deprecated as intended


6543:
we don't need to frontport this as  #2048 will apply same docs changes and the fix. and drop the deprecated var.

xoxys (commented):
This test data is till using `CI_SYSTEM_ARCH` instead of `CI_SYSTEM_PLATFORM`:
https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/metadata/drone_compatibility_test.go#L76

This test file is still using both `CI_SYSTEM_ARCH` and `CI_SYSTEM_PLATFORM`:
https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/metadata_test.go#L85-L86

Don't know if that matters.

xoxys (approved):
Nevermind, this is going to be fixed in https://github.com/woodpecker-ci/woodpecker/pull/2048

PR #2049, opened 2023-07-27
Fix docs nit and add rm todo for v1.2.0

6543:


xoxys:
The test repo from https://github.com/woodpecker-ci/woodpecker/tree/main/contrib/woodpecker-test-repo/.woodpecker is still using the `pipeline` syntax.

6543:
good catch - updated

6543:
image

xoxys (approved):

PR #2051, opened 2023-07-28
WIP: LXD engine + mixed LXD/Docker engine

hpidcock:
This is an experiment to see if it is possible to mix two local engines, in this case mixing docker with a new engine, lxd.

This works by leveraging a prefix on the image value, `lxd:`. If a step was to be using a lxd alpine image, it would be expressed as `lxd:alpine/3.18`. Bind mounts between lxd and docker are currently handled by creating a docker volume, then mounting that volume into lxd, not ideal, but since docker is the primary engine for woodpecker, it seems reasonable.

Example of mixing usage of both:

```
pipeline:
  test_0:
    image: 'alpine'
    commands:
      - echo "Date from docker is $(date)" > docker_date
      - cat docker_date

  test_1:
    image: 'lxd:alpine/3.18'
    commands:
      - echo "Test 1 $(cat docker_date)"
```

TODO:
- [ ] Add configuration to LXD engine
- [ ] Allow optionally selecting VM images (instead of just container images)
- [ ] Pass through limits
- [ ] Implement solo lxd engine performing checkout steps
- [ ] Add testing


pat-s:
Stale since 1,5 years, closing.

PR #2053, opened 2023-07-28
Bump default clone image version to 2.1.0

nupplaphil:
With 2.0.3, at least my environment stopped working because of the missing `CI_REPO_CLONE_URL` variable, introduced in  2.1.0


6543:
Well I should have thought about it, thanks for the fix

6543 (approved):

PR #2057, opened 2023-07-28
Move from "master" to "main" as fallback for default repo branches

6543:
… and make it configurable


anbraten:
@6543 I don't think we need to have that fallback at all. I checked gitea, github and gitlab and they all have values for the default-branch. For bitbucket I will check how we can get that value as well.

6543:
yes I would prefer to just relay on the forge

6543:
Just did not do research if forges do have a change to report it empty

6543:
->  #2065

Issue #1995, opened 2023-07-13
publish images to  quay.io too


6543:
next to docker.io push to  quay.io too

anbraten:
I wouldn't add quay.io at the moment. Docker-hub is free for us again and currently works. With the recent things going on at RedHat I am not really sure if quay.io is a better place. Adding any additional registry would currently just be maintenance overhead and additional problems.

xoxys:
What problems do you see that would be added? IMO, relying on the availability of a single public registry is a bad idea and has already caused trouble for many users in the past. Removing DockerHub in favor of Quay may be questionable, but I would like to see at least a second registry added as fallback. This could be Quay or GitHub Container registry. The maintenance overhead is also quite low to my experiences. I have set this up for a lot of repos and it "just works" reliable in the same way it works for the DockerHub upload process of your CI.

anbraten:
As long as all works its most of the time fine, until sth does not work straightforward. All those tiny things are often adding maintenance overhead in the end. Things like uploading does not work (credentials are outdated, ...), we need to update all image strings at all places in code, docs etc, users could be confused which registry to choose, ...

xoxys:
I see. In case Docker will decide tomorrow to kill all open source free projects (and we had nearly such situations already) things will become unusable immediatly.

6543:
I'll add and maintain that :)

PR #2069, opened 2023-07-31
Publish releases and branch tags to quay.io too

6543:
address  #1995


6543:
tested: https://ci.woodpecker-ci.org/repos/5586/pipeline/308/10

PR #2072, opened 2023-07-31
Use buffer for log writer

6543:
close  #2064

tested with a program that spils out 3000 lines of logfiles.
It took 2min on current v1.0.0./main and 4sec with this.

But the log output is broken and dont live stream anymore the bufio.Writer only consider the buffer when to write.
We alos sould consider linebreaks and time components too.

TODO:
- [x] create own bufio.Writer that:
  - [x] flushes if for a small duration no new data was received
  - [x] flushes if a specific amount of data was buffered
  - [ ] flushes on new lines before buff is full

For future PRs:
- [ ] create own bufio.Writer that:
  - [ ] flushes on new lines before buff is full
- [ ] make log save infrastructure not reliant on the unwritten promise that each write is it's own line
  - [ ] e.g. time stamp calculation#

---
*Sponsored by Kithara Software GmbH*

Issue #2064, opened 2023-07-29
woodpecker slowed done


nupplaphil:
### Component

agent

### Describe the bug

I upgraded to the latest woodpecker "Next" version about a few weeks ago.

Since then, I notice that woodpecker is slowed down a lot (about 5-6 times slower).

I cannot really say why, but as far as I can see, it's related to the amount of console-output.

- The `php-lint` step slowed down from 0:30 to 9-10 minutes, and it's "just" static linting the code and dumping it to the output
- The `mariadb` service is starting quite fast, no delay found
- The `phpunit` step is slowed from 2 minutes to 4-5 minutes, but has not as much output as `php-lint`, but it's FAR more complex

The .woodpecker files can be found at https://github.com/friendica/friendica/blob/develop/.woodpecker/.phpunit.yml

2 month ago, it was like this:
![image](https://github.com/woodpecker-ci/woodpecker/assets/379654/d0a311fa-74d7-4e44-a1ad-917f15bcd378)

now it's:
![image](https://github.com/woodpecker-ci/woodpecker/assets/379654/36cb4259-a54a-4b0f-b5a4-a5cac0426cd8)

I moved to an exclusive, brand new root-server with 128GB ram, i7 and only docker and woodpecker to have a look if it's an IO/CPU/.. issue
--> nope the result stays the same

### System Info

```shell
woodpecker: "next-c96fc3f2"
docker info:

Client: Docker Engine - Community
 Version:    24.0.5
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.20.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 30
  Running: 17
  Paused: 0
  Stopped: 13
 Images: 22
 Server Version: 24.0.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local lvm
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: active
[...]

PR #2074, opened 2023-07-31
Add buffer to log writer via nio

6543:
might address  #2064
close  #2072

---


6543:
me too, but I want to see if there is a significant difference

6543:
tested it is as fast as stdlib

PR #2078, opened 2023-08-01
Add `failure: ignore` to security scans

qwerty287:
Otherwise all PRs are marked as failed on github.

### DIFF
--- a/.woodpecker/securityscan.yml
+++ b/.woodpecker/securityscan.yml
@@ -13,12 +13,14 @@ steps:
   check backend:
     group: check
     image: *trivy_plugin
+    failure: ignore
     settings:
       skip-dirs: web/,docs/
 
   check docs:
     group: check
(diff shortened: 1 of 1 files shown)


xoxys (changes requested):
The reported CVEs should be fixed by updating the deps or added to the trivyignore list. Just effectively bypassing the entire security scan makes it IMO useless.

xoxys:
True, there will always be cases where the results of security scans are not accurate or can be ignored because the code base is not affected, etc. But for all these cases, we should use the methods described at https://aquasecurity.github.io/trivy/v0.43/docs/configuration/filtering/ instead of "blindly" ignoring all bugs. Otherwise, we will not notice future critical vulnerabilities and we can remove the security scan at all.

xoxys:
@qwerty287 I think we can close this one?

Issue #1739, opened 2023-05-05
User Secrets


JakobDev:
### Clear and concise description of the problem

I have multiple Personal Projects which share the same Secrets. I have to configure the Secrets for each Project. It would be better to have Secrets on a User Level.

### Suggested solution

We [already have Secrets on Global and Organisation Level](https://github.com/woodpecker-ci/woodpecker/pull/1027) Level. Why not on a User Level?

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

6543:
user level in form of "for all user repos" for `<user>/*` repos ... or for all repos where the pipeline is run because of an user action ?!?

JakobDev:
> user level in form of "for all user repos" for `<user>/*` repos 

Exact that. I have on Codeberg multiple Projects which deploy a Package to PyPI on a new Release, so it would be helpful if I can enter my credentials just once and not for each Project new.

JakobDev:
Could you create A PR, so I don't have to setup a develop environment for this?

PR #2087, opened 2023-08-02
Improve user settings

qwerty287:
Make the user settings tabbed as well, currently only holding "Settings" (language) and "API". Can be extended with user secrets (#1739) etc. later.
Screenshot 2023-08-02 at 09-56-00 Woodpecker
Screenshot 2023-08-02 at 09-56-05 Woodpecker

### DIFF
--- a/docs/src/pages/faq.md
+++ b/docs/src/pages/faq.md
@@ -16,7 +16,7 @@ The Drone CI license was changed after the 0.8 release from Apache 2 to a propri
 
 Woodpecker is having two different kinds of releases: **stable** and **next**.
 
-The **stable** releases (currently version 0.15) is a long-term supported (LTS) stable version. The stable releases are only getting bugfixes.
+The **stable** releases (currently version 1.0) are long-term supported (LTS) stable versions. The stable releases are only getting bugfixes.
 
 The **next** release contains all bugfixes and features from `main` branch. Normally it should be pretty stable, but as its frequently updated, it might contain some bugs from time to time.
 
--- a/web/src/assets/locales/de.json
+++ b/web/src/assets/locales/de.json
@@ -159,11 +159,6 @@
     "activity": "Aktivitäten",
     "add": "Repository hinzufügen",
     "branches": "Branches",
-    "build": {
-      "protected": {
-        "declined": "Diese Pipeline ist abgelehnt worden!"
-      }
-    },
     "deploy_pipeline": {
       "enter_target": "Zielumgebung des Deployments",
       "title": "Deployment für aktuelle Pipeline #{pipelineId} starten",
--- a/web/src/assets/locales/fr.json
+++ b/web/src/assets/locales/fr.json
@@ -159,9 +159,6 @@
     "activity": "Activité",
     "add": "Ajouter un dépôt",
     "branches": "Branches",
-    "build": {
-      "step_not_started": "L'étape n'a pas démarré encore."
-    },
     "deploy_pipeline": {
       "enter_target": "Environnement de déploiement ciblé",
       "title": "Déclenchement d'un événement de déploiement pour le pipeline courant #{pipelineId}",
--- a/web/src/assets/locales/lv.json
+++ b/web/src/assets/locales/lv.json
@@ -159,9 +159,6 @@
     "activity": "Aktivitāte",
     "add": "Pievienot repozitoriju",
     "branches": "Atzari",
-    "build": {
-      "step_not_started": "Šis solis vēl nav uzsākts."
-    },
     "deploy_pipeline": {
       "enter_target": "Mērķa uzstādīšanas vide",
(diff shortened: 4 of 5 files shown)


xoxys:
Thanks! I had the same fix on my list.

web/src/views/org/OrgRepos.vue
```diff
@@ -6,7 +6,7 @@
 
     <template #titleActions>
       <IconButton
-        v-if="orgPermissions.admin"
+        v-if="!org.is_user && orgPermissions.admin"
```
xoxys, line 9: What's the purpose of the changed permission check to the org views?

web/src/views/admin/Admin.vue
```diff
@@ -1,24 +0,0 @@
-<template>
```
xoxys, line 1: Is the change to the admin view and router somehow related to the restructured user settings, or just a "drive-by" cleanup?

web/src/views/org/OrgRepos.vue
xoxys: Im a bit confused. AFAIK Woodpecker threads users as orgs. If I look at a repo from my personal user space, this is handled as

xoxys: Thank you for the explanation. Hmm, as you wrote in https://github.com/woodpecker-ci/woodpecker/issues/1739 the backend is the same and as we handle users as orgs now this should IMO just work? Or what's technically missing that there are no user secrets yet? Just removing it because we might want to move user secrets to another location sounds bad to me. But I'm not that deep into the codebase yet either, sorry if the questions are bullshit :see_no_evil:

xoxys: Thanks :+1: works for me as well.

web/src/assets/locales/en.json
```diff
@@ -425,16 +425,24 @@
         }
     },
     "user": {
-        "settings": "User Settings",
+        "user_settings": "User Settings",
```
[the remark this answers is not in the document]
xoxys: Can we not just use the same structure we have for the admin settings already?

```
   "user":{
      "settings":{
         "settings":"User Settings",
         "oauth_error":"Error while authenticating against OAuth provider",
         "internal_error":"Some internal error occurred",
         "access_denied":"You are not allowed to login",
         "general":{
            "general":"General",
            "language":"Language"
         },
         "api":{
            "api":"API",
            "token":"Personal Access Token",
            "shell_setup":"Shell setup",
            "api_usage":"Example API Usage",
            "cli_usage":"Example CLI Usage",
            "dl_cli":"Download CLI",
            "shell_setup_before":"do shell setup steps before",
            "reset_token":"Reset token",
            "swagger_ui":"Sagger UI"
         }
      }
   }
```

web/components.d.ts
```diff
@@ -101,6 +101,9 @@ declare module '@vue/runtime-core' {
     Tab: typeof import('./src/components/layout/scaffold/Tab.vue')['default']
     Tabs: typeof import('./src/components/layout/scaffold/Tabs.vue')['default']
     TextField: typeof import('./src/components/form/TextField.vue')['default']
+    UserAPITab: typeof import('./src/components/user/UserAPITab.vue')['default']
+    UserGeneralTab: typeof import('./src/components/user/UserGeneralTab.vue')['default']
+    UserSettingsTab: typeof import('./src/components/user/UserSettingsTab.vue')['default']
```
xoxys, line 106: Was renamed and can be removed?

xoxys (changes requested):

xoxys:
What do you think about adding a headline to the tabs, as we have in the admin settings?
image

xoxys:
Or repo settings without a subtitle (which is not required if we don't really have to explain something)

image

web/src/components/user/UserAPITab.vue
```diff
@@ -0,0 +1,95 @@
...
+      </div>
+    </div>
+
+    <div class="mt-2 mb-4">
+      <div class="flex items-center mb-2">
+        <div class="flex items-center text-wp-text-100 font-bold">
+          <label>{{ $t('user.settings.api.token') }}</label>
+        </div>
+        <Button class="ml-4" :text="$t('user.settings.api.reset_token')" @click="resetToken" />
+      </div>
```
xoxys, line 16:
```suggestion
      <div class="flex items-center mb-2">
        <div class="flex flex-row items-center text-wp-text-100 font-bold">
          <label>{{ $t('user.settings.api.token') }}</label>
        </div>
        <Button class="ml-auto" :text="$t('user.settings.api.reset_token')" @click="resetToken" />
      </div>
```

What do you think about aligning the button to the right?

image

xoxys (approved):
I like it :heart:

web/src/assets/locales/en.json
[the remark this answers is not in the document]
lafriks: yes, lgtm except errors are not related to settings but are user related so these should not be moved into settings section

Issue #2088, opened 2023-08-02
CLI repo add: invalid character '<' looking for beginning of value


runephilosof-karnovgroup:
### Component

cli

### Describe the bug

```
user@fedora:~ $ docker run -e WOODPECKER_SERVER -e WOODPECKER_TOKEN woodpeckerci/woodpecker-cli:latest-alpine -- repo add 528940444
8:08AM FTL error running cli error="invalid character '<' looking for beginning of value"
```

### System Info

```shell
Server
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"1.0.0"}

Client
REPOSITORY                    TAG             IMAGE ID       CREATED      SIZE
woodpeckerci/woodpecker-cli   latest-alpine   0eb7866dd2b9   3 days ago   48.9MB
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

runephilosof-karnovgroup:
The UI is sending a POST request to https://woodpecker-host/api/repos?forge_remote_id=528940444

PR #2090, opened 2023-08-02
client lib: remove useles "sync" option from RepoListOpts

### LINKED ISSUE #2083: Go client RepoListOpts: Deprecate sync argument
### Component

other

### Describe the bug

RepoListOpts first argument is `sync`
https://github.com/woodpecker-ci/woodpecker/blob/e0ed1b3e4914fb0aa7cad08ff22bf72437b70da0/woodpecker-go/woodpecker/client.go#L167

Which controls the argument `flush`, sent to the server
https://github.com/woodpecker-ci/woodpecker/blob/e0ed1b3e4914fb0aa7cad08ff22bf72437b70da0/woodpecker-go/woodpecker/client.go#L169

However, the server only cares about the `all` argument and ignores the `flush`.

So deprecate the sync argument (by documenting it with a code comment?)

### System Info

```shell
github.com/woodpecker-ci/woodpecker v1.0.0
```

### Additional context

_No response_

### Validations

qwerty287:
Closes #2083 

This is breaking because I removed the old, unused `sync` parameter (#2083). Is it fine anyways to get this into 1.1.0 or should we wait for 2.0 with this?

### DIFF
--- a/woodpecker-go/woodpecker/client.go
+++ b/woodpecker-go/woodpecker/client.go
@@ -28,7 +28,7 @@ import (
 const (
 	pathSelf           = "%s/api/user"
 	pathRepos          = "%s/api/user/repos"
-	pathRepoPost       = "%s/api/repos"
+	pathRepoPost       = "%s/api/repos?forge_remote_id=%d"
 	pathRepo           = "%s/api/repos/%d"
 	pathRepoLookup     = "%s/api/repos/lookup/%s"
 	pathRepoMove       = "%s/api/repos/%d/move?to=%s"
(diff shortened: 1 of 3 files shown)


runephilosof-karnovgroup:
Alternative fix for #2088 in #2091. Basically the same, but that PR does not include the breaking change for fixing #2083.

woodpecker-go/woodpecker/client.go
```diff
@@ -164,17 +165,17 @@ func (c *client) RepoList() ([]*Repo, error) {
...
+	uri := fmt.Sprintf(pathRepos+"?all=%v", c.addr, all)
 	err := c.get(uri, &out)
 	return out, err
 }
 
 // RepoPost activates a repository.
 func (c *client) RepoPost(forgeRemoteID int64) (*Repo, error) {
 	out := new(Repo)
-	uri := fmt.Sprintf(pathRepo, c.addr, forgeRemoteID)
+	uri := fmt.Sprintf(pathRepoPost+"?forge_remote_id=%d", c.addr, forgeRemoteID)
```
runephilosof-karnovgroup, line 178: Might be the way to go. However, in 
```
pathRepoMove       = "%s/api/repos/%d/move?to=%s"
```
The query payload was included in the path variable instead of appended in the function as you did here.
I can't decide which I prefer.

6543 (approved):

6543:
We should backport it in a non breaking way :)

PR #2094, opened 2023-08-02
Lazy-load TimeAgo locales

qwerty287:
1. new translation docs
2. lazy-load TimeAgo locales (used for "x min ago" messages). This 1. reduces size and 2. provides all languages without adding them manually.
3. Remove DayJS locales, they're unused.

### DIFF
--- a/server/forge/github/github.go
+++ b/server/forge/github/github.go
@@ -354,20 +354,22 @@ func (c *client) Org(ctx context.Context, u *model.User, owner string) (*model.O
 	client := c.newClientToken(ctx, u.Token)
 
 	user, _, err := client.Users.Get(ctx, owner)
+	log.Trace().Msgf("Github user for owner %s = %v", owner, user)
 	if user != nil && err == nil {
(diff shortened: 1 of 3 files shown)


web/src/components/pipeline-feed/PipelineFeedItem.vue
```diff
@@ -10,7 +10,7 @@
           <Tooltip>
             <span>{{ since }}</span>
             <template #popper
-              ><span class="font-bold">{{ $t('created') }}</span> {{ created }}</template
+              ><span class="font-bold">{{ $t('repo.pipeline.created') }}</span> {{ created }}</template
```
lafriks, line 13: was this always wrong?

6543: Well we can at least backport it for v1.0.x

lafriks:
is `.js` extension really needed as content is actually json not javascript?

lafriks (approved):

PR #2104, opened 2023-08-03
Forge Github Org: Use `login` instead of `name`

### LINKED ISSUE #2092: Enabling second repository results in: `pq: duplicate key value violates unique constraint "UQE_orgs_name"`
### Component

server

### Describe the bug

Enabling my first repository worked fine, however the second does not.

When the UI sends a POST request to https://woodpecker-host/api/repos?forge_remote_id=526933511 it returns this error
```
pq: duplicate key value violates unique constraint "UQE_orgs_name"
```

The debug log of the server does not give any more insights:
```
{ip: 10.100.32.116, latency: 474.729763, level: debug, method: POST, path: /api/repos, status: 500, user-agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36}
```

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"1.0.0"}
```

### Additional context

_No response_

### Validations

runephilosof-karnovgroup:
For "woodpecker-ci" the `name` is "Woodpecker CI"
 and the `login` is "woodpecker-ci"

Fixes #2092

This was causing the organization lookup to fail, because it looks up using the `login`, when it did not find the organization, it would try to create it. The creation would fail, because it uses the `name`, and an organization with that `name` already exists.
Resulting in
```
pq: duplicate key value violates unique constraint "UQE_orgs_name"
```


runephilosof-karnovgroup:
The commit is based on the merge-base of main and release/v1.0 to make it easy to also merge it into release/v1.0

runephilosof-karnovgroup:
> Can you check this for Gitlab, Gitea and Bitbucket too?

Looking at the code for Gitlab it seems problematic.
While Gitea and Bitbucket seems fine.

The foolproof way would be to use the `owner` variable instead of what the forge returns.
However, we are basing `IsUser` and `Private` on what we look up in the forge, so it is important that we handle the response from the forge correct.

server/forge/github/github.go
```diff
@@ -354,20 +354,22 @@ func (c *client) Org(ctx context.Context, u *model.User, owner string) (*model.O
 	client := c.newClientToken(ctx, u.Token)
 
 	user, _, err := client.Users.Get(ctx, owner)
+	log.Trace().Msgf("Github user for owner %s = %v", owner, user)
```
runephilosof-karnovgroup, line 357: Do you like this trace logging. Or should I remove it?

runephilosof-karnovgroup:
I have fixed it locally for Gitlab as well now. It should be using `FullPath`.
I will create a new PR for fixing Gitlab.

runephilosof-karnovgroup:
On second thought, I will include the Gitlab fix here, because both concerns the same issue #2092.

6543 (approved):

6543:
Backported -> #2106

PR #2106, opened 2023-08-03
Forge Github Org: Use `login` instead of `name` (#2104)

### LINKED ISSUE #2092: Enabling second repository results in: `pq: duplicate key value violates unique constraint "UQE_orgs_name"`

6543:
Backport #2104

For "woodpecker-ci" the `name` is "Woodpecker CI"
 and the `login` is "woodpecker-ci"

Fixes #2092

This was causing the organization lookup to fail, because it looks up using the `login`, when it did not find the organization, it would try to create it. The creation would fail, because it uses the `name`, and an organization with that `name` already exists.
Resulting in:

```
pq: duplicate key value violates unique constraint "UQE_orgs_name"
```

### DIFF
--- a/server/forge/github/github.go
+++ b/server/forge/github/github.go
@@ -354,20 +354,22 @@ func (c *client) Org(ctx context.Context, u *model.User, owner string) (*model.O
 	client := c.newClientToken(ctx, u.Token)
 
 	user, _, err := client.Users.Get(ctx, owner)
+	log.Trace().Msgf("Github user for owner %s = %v", owner, user)
 	if user != nil && err == nil {
 		return &model.Org{
-			Name:   user.GetName(),
+			Name:   user.GetLogin(),
 			IsUser: true,
 		}, nil
 	}
 
 	org, _, err := client.Organizations.Get(ctx, owner)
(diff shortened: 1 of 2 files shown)


runephilosof-karnovgroup:
We should also create a migration that fixes the wrongly created organizations.
Having those organizations with the wrong name in the db is a guarantee for future problems.

PR #2107, opened 2023-08-03
Fix en.json file syntax

lafriks:
<p>Weblate could not parse the translation files while updating the translations.</p>

<p>The following occurrences were found:</p>

Filename | Error
-- | --
web/src/assets/locales/en.json | JSONDecodeError('Expecting property name enclosed in double quotes: line 445 column 11 (char 18857)')
web/src/assets/locales/en.json | JSONDecodeError('Expecting property name enclosed in double quotes: line 445 column 11 (char 18857)')


6543 (approved):

6543:
do we not have a linter for this file in place :/

lafriks:
looks like we don't have :)

6543:
I'll create a stepbfor it ...

6543:
#2107 also do fix this

6543:
#2108 also do fix this

PR #2115, opened 2023-08-04
Add opt save global log output to file

### LINKED ISSUE #1933: Add option, to log to file

# BACKGROUND
Woodpecker seems to log to `stdout` only, and thus to docker's built-in "json driver". The logs can be access like so:
- `$ docker logs -t $CONTAINER_NAME`
- `tail -f /var/lib/docker/containers/$CONTAINER_ID/$CONTAINER_ID-json.log`

# PROBLEM
When the container is recycled, the logs are lost. The workaround is to use another logging tool, but that is unnecessary complexity, especially for small deployments.

Also, in general, the logs are not in a friendly format - it's much simpler, especially in small deployments, to simply have the log in a file that can be easily manipulated. Not everyone uses large complex logging apps and servers, etc.

### Suggested solution

# SOLUTION
Many apps have an option to log to file as well. Gitea does that, for example.

It would be nice if the log could be sent to a file instead of (or in addition to) stdout. If enabled, it could be written to `/var/lib/woodpecker/app.log` by default.

Personally, I would mount it as `/var/log/woodpecker.log:/var/lib/woodpecker/app.log` so that I could have it on the host, and use logrotate on it.

### Alternative

An unnecessarily complicated environment that includes log aggregation and monitoring, and who knows what else. For small servers that is overkill and hard to manage.

### Additional context

Thank you for considering it!

### Validations

6543:
close  #1933

---


6543:
> If you send a USR1 signal to woodpecker, will it "release" the file so the host can logrotate it?

not jet that's a followup ... I need to write a wrapper for the os.FileOpen that can do so ... so that's  a followup

6543:
I just created this mentioned wrapper as lib here: https://github.com/6543/logfile-open ...
... so once merged I can create a followup, that can be tested seperate

PR #2126, opened 2023-08-07
Support user secrets

### LINKED ISSUE #1739: User Secrets

qwerty287:
Just like org secrets. Therefore, I added a column `org_id` to users which associates the user with an org, everything else is used from org secrets. If there isn't an org for the user yet, it is created which means that every user will have an org.

Closes #1739

### DIFF
--- a/docs/docs/30-administration/100-external-configuration-api.md
+++ b/docs/docs/30-administration/100-external-configuration-api.md
@@ -7,6 +7,10 @@ Every request sent by Woodpecker is signed using a [http-signature](https://data
 
 A simplistic example configuration service can be found here: [https://github.com/woodpecker-ci/example-config-service](https://github.com/woodpecker-ci/example-config-service)
 
+:::warning
+You need to trust the external config service as it is getting secret information about the repository and pipeline and has the ability to change pipeline configs that could run malicious tasks.
+:::
+
 ## Config
 
 ```shell
--- a/server/api/pipeline.go
+++ b/server/api/pipeline.go
@@ -453,7 +453,12 @@ func PostPipeline(c *gin.Context) {
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


xoxys:
Was removed in https://github.com/woodpecker-ci/woodpecker/pull/2087#discussion_r1282750711

anbraten:
> This mainly improves the UX by moving these settings to the user settings (where it's much easier to find them) and it's also possible to add secrets before you enabled any repo.

Ahh okay haven't noticed that it got removed again. Could we simply add a link from the user settings page to the own org settings? At least we should have the settings icon on the orgs repo list again linking somewhere to change the secrets as well.

anbraten:
> Can't this settings icon just link to the user settings (`/user`)?

Yes

server/store/datastore/migration/022_add_org_id.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/woodpecker-ci/woodpecker/server/model"
+)
+
+var addOrgID = task{
+	name:     "add-org-id",
+	required: true,
+	fn: func(sess *xorm.Session) error {
+		if exist, err := sess.IsTableExist("orgs"); exist && err == nil {
+			if err := sess.DropTable("orgs"); err != nil {
```
6543, line 30: that would create problems!

6543 (changes requested):

anbraten (approved):

Issue #1910, opened 2023-07-01
Speed up woodpecker pipelines


anbraten:
The pipelines executed for PRs in this repo could be improved:
- use more path filters
- only push images if requested for a PR (could be done by adding a label to the PR)
- only run build for amd64 architecture as cross-compiling is pretty slow
- ...

anbraten:
Yeah sounds fine to me

6543:
did some bits at #1966

PR #2127, opened 2023-08-07
Speed up pipelines

qwerty287:
- Use workflow-global conditions to also skip the `clone` step for `docs` and `web` workflows
- never run `test` and `web` for pushes (they're running for PRs and every change we do must pass checks on PRs, so they're unnecessary)

https://github.com/woodpecker-ci/woodpecker/issues/1910


anbraten (approved):

6543:
got reverted:  #2129

PR #2129, opened 2023-08-07
Fix workflows again

qwerty287:
Since it does not work this is mainly a revert of #2127


6543:
can you write what exactly does not work?

6543:
ah get it ... you fall into that trap :D

I get it that you wan to avoid the clone step ... well we have to do this on an other level :D

6543 (approved):

PR #1799, opened 2023-06-01
Fix UI and backend paths with subpath

qwerty287:
I'm not sure if this is an ideal fix for this, but it seems to work for me. If you have another idea just let me know.

Closes #1798 
Closes #1773

PR #2133, opened 2023-08-07
Fix UI and backend paths with subpath (#1799)

qwerty287:
Backport #1799

This also contains some refactoring etc which should be included for the fix (including a new config env var `WOODPECKER_ROOT_PATH` as alias for `WOODPECKER_ROOT_URL`). Is this fine?


6543:
removed backport lable as we should not backport that to v0.15

removed regression lable as we do have it released

6543 (approved):

6543:
```

/woodpecker/src/github.com/woodpecker-ci/woodpecker/web/src/router.ts
  135:11  error  Unexpected template string expression  no-template-curly-in-string
```

6543 (approved):

PR #2137, opened 2023-08-07
Check for correct license header

qwerty287:


.woodpecker/test.yml
```diff
@@ -20,6 +20,11 @@ variables:
       event: [pull_request, tag, deployment]
 
 steps:
+  lint-license-header:
```
6543, line 23: can we move that after vendor and add it into a group with lint-pipeline

6543:
if we did not funamentally change a file, it still has to contain the drone header!!!

6543:
that's a good question ... I would leave them and make sure new headers follow our convention?

pipeline/backend/kubernetes/pod.go
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      http://www.apache.org/licenses/LICENSE-2.0
+//     http://www.apache.org/licenses/LICENSE-2.0
```
6543, line 7: mergeconflict artifact

web/LICENSE
```diff
@@ -1,5 +1,6 @@
 Copyright 2017 Drone.IO Inc
 Copyright 2019 Laszlo Fogas
+Copyright 2020 Woodpecker Authors
```
6543, line 3: hmm anbraten did rewrote the whole web frontend

@anbraten did you use some old parts ?

else we can remove the other two rows

6543 (approved):

PR #2138, opened 2023-08-07
Use fix width for pipeline step list

xoxys:
On larger monitors using dynamic width in percent, the step list can become very wide which makes it hard to read.


6543:
could you provide some screenshots :)

xoxys:
Sorry, of course:

image
image

6543 (approved):

6543 (approved):

anbraten (commented):
Wouldn't it be better to use a max width here

xoxys:
> Wouldn't it be better to use a max width here

Yep, will create a follow-up PR.

PR #2167, opened 2023-08-08
Add changelog v1.0.2

qwerty287:
After #2164 we need a 1.0.2 unfortunately


xoxys:
1.0.1 was not released yet, wouldn't it make sense to add it to 1.0.1 instead?

Issue #1860, opened 2023-06-13
drop step.group in favour of depends_on to create a DAG


6543:
Support parallelism inside a pipeline via a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph). This requires a way to define step dependencies which can then allow the runner to build said DAG. Edit: I see [`group`](https://woodpecker-ci.org/docs/usage/pipeline-syntax#group---parallel-execution) provides something similar for static fan-out, but certainly not as flexible as defining dependencies directly.

_Originally posted by @silverwind in https://github.com/woodpecker-ci/woodpecker/issues/567#issuecomment-1282954563_

6543:
what about `depends_on` we already use it for workflow dep.-DAG ?

6543:
**one big downside! and what esp. makes it breaking!!!:**

now we assume that steps need to be run in sequence if not told otherwise (because it is what you normally want)
if we change this ... it would be the exact other way around

so what I think is **really** important: by default it should run sequential (that's just good UX-Design)

so what about a workflow config: `parallel: false` that if enabled would create a step dag ?!?

6543:
so I propose:
 - pipeline/backend typse -> always use a DAG
 - pipeline/frontend types (yaml) -> use a dag if specified and if not it creates a simpjle dag
   -> `(step1->step2->(group:step3 & step 4)-> step5...)` witch could still encounter the groups feature...

6543:
we should explicite write that down as one of our core values:
- `by default the config should work intuetive for beginners and not for advanced users - KISS` &
- `the default should be the so that the normal pipelines dont need to configure almost anything`

silverwind:
In drone, it is also sequential until you add the first `depends_on` to a step, at which point it becomes a DAG.

6543:
That sounds nice as long as you dont have a workflow with steps that **all** can run paralel

6543:
I would call that not easy to lint, parse and sanitize ...

Also it would increase yaml config more ...

The more i think and talk about this with others the more i think we should have a workflow flag "parallel: true" and depends_on ... for the DAG

lumpi101:
One more idea: to execute steps in parallel or not can be decided on whether the steps in the yaml are defined as a list or a dictionary.
- A list is sequential by nature and thus means that all listed steps are executed sequentially in that order.
[...]

PR #2184, opened 2023-08-09
WIP: Represent step dependency as dag in backend

6543:
depends on #2183 

address  backend to be prepared for #1860


6543:
well it was still WIP - and we do NOT in the backend.

but I guess I'll create an issue to track this first

6543:
#3381

PR #2212, opened 2023-08-15
Move hook and events-stream routes to use `/api` prefix

anbraten:
- move hook and events-stream routes to use `/api` prefix
- rename build to pipeline


6543 (approved):

6543:
non breaking, it is done at #2639

PR #2213, opened 2023-08-15
Remove plugin-only option from secrets

anbraten:
- remove plugin-only option from secrets
The images list of secrets is currently pretty use-less if not enabled with the plugin-only option in combination, therefore the plugin-only option will now be automatically set, if at least one image was listed

- rename secrets json variables (images and events) to their proper plural


cli/secret/secret_add.go
```diff
@@ -46,17 +46,13 @@ var secretCreateCmd = &cli.Command{
 			Usage: "secret value",
 		},
 		&cli.StringSliceFlag{
-			Name:  "event",
```
6543, line 49: could we use alias to still have old one too?

anbraten: I marked this PR as breaking and added a migration note to the docs page.

anbraten:
added a migration

lafriks:
Migration should also remove image value if plugin only is not selected otherwise it will stop functioning if image is set but it's not plugin

anbraten:
> Migration should also remove image value if plugin only is not selected otherwise it will stop functioning if image is set but it's not plugin

Actually while writing the migration note and thinking about this again, I would say it more secure to fail in that case than potentially still leaking the secret. As the user would notice the error of the missing secret I guess its fine to limit existing image filters to plugins.

docs/docs/91-migrations.md
```diff
@@ -8,6 +8,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 - Dropped deprecated `pipeline:` keyword in favor of `steps:` in pipeline config
 - Dropped deprecated `branches:` filter in favor of global [`when.branch`](./20-usage/20-workflow-syntax.md#branch-1) filter
 - Deprecated `platform:` filter in favor of `labels:`, [read more](./20-usage/20-workflow-syntax.md#filter-by-platform)
+- Secrets `event` property was renamed to `events` as its a list of events. The new property `events` has to be used in the api and as cli argument. The old property `event` was removed.
```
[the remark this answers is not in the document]
anbraten: added

PR #2214, opened 2023-08-15
Remove repo pass netrc option from repos

anbraten:
Not trusted clone plugins should never get our credentials as an attacker could simply replace the clone step by a malicious one and therefore gather netrc credentials.


6543:
before we remove that option, we nee to make sure why it was added in the first place and if this can be solved in an other way now :)

anbraten:
@6543 It was added as fallback to allow custom clone images which are not listed as trusted images.

anbraten:
closing as of #2601

PR #2215, opened 2023-08-15
Bump default git clone plugin

6543:
- and add alternate quay registry to trusted clone image
- add alternate registry for buildx plugin


xoxys (approved):

6543:
->  #2220

PR #2216, opened 2023-08-16
Move "skip ci" logic into global pipeline conditions

6543:
... and make custom errors follow std err conventions

this fix a 500 response if the whole pipeline is filtered out

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,6 +16,7 @@
   "prettier.configPath": "./web/.prettierrc.js",
   "prettier.ignorePath": "./web/.prettierignore",
   "cSpell.words": [
+    "Curr",
     "doublestar",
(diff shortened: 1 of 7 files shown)


lafriks (approved):

6543:
hook gets ignored ... but it does not return a 204 to the webhook as expected ... need to have a look at it again :/

6543:
fixed and found a bug that has to be backported 😅

6543:
bugfix part was backported at  #2230

6543:
@lafriks rerequested review because of to much changes :S

lafriks (approved):

xoxys:
@6543 this seems to be broken now? https://ci.thegeeklab.de/repos/13/pipeline/14

PR #2230, opened 2023-08-16
Return 204 not 500 on filtered pipeline

6543:
the error check always failed, as the custom errors did not have the `Is(error) bool` interface implemented

backport bugfix part of  #2216

### DIFF
--- a/server/pipeline/errors.go
+++ b/server/pipeline/errors.go
@@ -18,22 +18,37 @@ type ErrNotFound struct {
 	Msg string
 }
 
-func (e ErrNotFound) Error() string {
+func (e *ErrNotFound) Error() string {
 	return e.Msg
+}
+
+func (e *ErrNotFound) Is(target error) bool {
+	_, ok := target.(*ErrNotFound) //nolint:errorlint
+	return ok
 }
 
 type ErrBadRequest struct {
 	Msg string
 }
 
-func (e ErrBadRequest) Error() string {
+func (e *ErrBadRequest) Error() string {
 	return e.Msg
+}
+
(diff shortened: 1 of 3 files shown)


server/pipeline/errors.go
```diff
@@ -18,22 +18,37 @@ type ErrNotFound struct {
 	Msg string
 }
 
-func (e ErrNotFound) Error() string {
+func (e *ErrNotFound) Error() string {
```
lafriks, line 21:
```suggestion
func (e ErrNotFound) Error() string {
```

server/pipeline/errors.go
```diff
@@ -18,22 +18,37 @@ type ErrNotFound struct {
...
 	return e.Msg
 }
 
+func (e *ErrNotFound) Is(target error) bool {
+	_, ok := target.(*ErrNotFound) //nolint:errorlint
+	return ok
+}
```
lafriks, line 28:
```suggestion
func (e ErrNotFound) Is(target error) bool {
	_, ok := target.(ErrNotFound)
	if !ok {
		_, ok = target.(*ErrNotFound)
	}
	return ok
}
```

server/pipeline/errors.go
lafriks: this way it will work for both non-pointer and pointer variables

lafriks (approved):

lafriks:
I had noticed this bug for quite some time but had no time to look into it so thanks for fixing it ❤️

6543:
applied the changes here to the related pull for main too ...

PR #2242, opened 2023-08-18
Skip settings proxy config with WithProxy if its empty

### LINKED ISSUE #2208: Ability to disable HTTPS_PROXY environment variable always being passed through.

One of the libraries in my project attempts to use a proxy if HTTPS_PROXY is set as a variable, even if it is just blank.

### Suggested solution

Possible solutions discussed in the matrix room by 6543:

"Well we just either need to add a option to 
https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/yaml/compiler/option.go#L219
and expose it as flag or add a check to 
https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/yaml/compiler/option.go#L228
If all 3 are "" and if so ignore it
I'm for second one ..."

### Alternative

_No response_

### Additional context

_No response_

### Validations

6543:
close #2208

### DIFF
--- a/pipeline/frontend/yaml/compiler/option.go
+++ b/pipeline/frontend/yaml/compiler/option.go
@@ -230,13 +230,11 @@ type ProxyOptions struct {
 // and NO_PROXY environment variables added by default to every
 // container in the pipeline.
 func WithProxy(opt ProxyOptions) Option {
-	// if options are all empty we don't set any env var
 	if opt.HTTPProxy == "" &&
 		opt.HTTPSProxy == "" &&
 		opt.NoProxy == "" {
 		return noopOption()
 	}
-	// else we pass on as upper- and lower-case
 	return WithEnviron(
 		map[string]string{
 			"no_proxy":    opt.NoProxy,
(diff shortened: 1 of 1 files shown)


pipeline/frontend/yaml/compiler/option.go
```diff
@@ -226,6 +230,13 @@ type ProxyOptions struct {
 // and NO_PROXY environment variables added by default to every
 // container in the pipeline.
 func WithProxy(opt ProxyOptions) Option {
+	// if options are all empty we don't set any env var
+	if opt.HTTPProxy == "" &&
+		opt.HTTPSProxy == "" &&
+		opt.NoProxy == "" {
+		return noopOption()
+	}
+	// else we pass on as upper- and lower-case
```
anbraten, line 239: What do you mean by upper- and lower-case? I would just leave the comment away is quite self explaining already.

6543, line 239:
```suggestion
```

pipeline/frontend/yaml/compiler/option.go
```diff
@@ -226,6 +230,13 @@ type ProxyOptions struct {
 // and NO_PROXY environment variables added by default to every
 // container in the pipeline.
 func WithProxy(opt ProxyOptions) Option {
+	// if options are all empty we don't set any env var
```
6543, line 233:
```suggestion
```

anbraten (approved):

PR #2244, opened 2023-08-18
Handle parsed hooks that should be ignored (#2243)

lafriks:
Currently it would fail with panic:

```
runtime error: invalid memory address or nil pointer dereference
...
/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/forge/gitea/gitea.go:492 (0xdfb32e)
```


6543 (commented):
lgtm 😆

Issue #2312, opened 2023-08-21
Don't fail on missing CI config


xoxys:
### Component

server

### Describe the bug

Not necessarily all branches of a repository are intended to have a CI config. But branches without a CI config will always create an error in the UI. 

![image](https://github.com/woodpecker-ci/woodpecker/assets/3391958/c33ec141-ecf7-40c3-9ec8-c67f847d3649)

In the past, I was able to work around this issue by adding `[skip ci]` to all commits to branches without a CI config. This is now also broken, see https://github.com/woodpecker-ci/woodpecker/pull/2216.

My suggestion would be to simply ignore missing CI configurations, or at least not create a failed pipeline in the UI. Alternatively, users should have the option to ignore branches. A common example would be `github-pages` branches, where only page artifacts are published and no CI configuration.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-06d1f3c92e"}
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

6543:
Hmm you could add a noop ignore all pipeline to those branches ...

lafriks:
problem with this is that it can give correct hint when you have misnamed pipeline definition file or forgot to commit it. imho it would be better to have option to configure this. At least in gitea there is option in webhooks to configure to not send events on specific branches, looks like GH does not have it

6543:
and why we **me** and @anbraten don't use a lot of configs via the repo settings is, that the pipeline config should be a single source of truth and represent the state of the pipeline. if we move to much outside of that config it does not do this anymore ... and so it's not versioned/documented etc...  anymore

xoxys:
[...]

PR #2313, opened 2023-08-21
Skip pipeline if config not found

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/2312

- missing pipeline config for hook events is ignored
- other events like restarts, manual run, etc. still throwing an error to the user

image


xoxys:
Don't know it that's the best way to achieve this, but it seems to be working.

6543 (changes requested):
that's not how we should handle it ...

i would propose a custom error and check for the error in error handling at the server/api package

anbraten:
Could you add a note to the migration docs. Otherwise lgtm

xoxys:
@anbraten Sure, done.

anbraten (approved):

lafriks:
maybe we can skip pipelines with missing config only for non-default branches?

xoxys:
Can we please just merge it for now. We agreed in the issue already that there are multiple options to improve it.

xoxys:
Argh... @6543 this needs an approval from you as you requested the change...

server/pipeline/create.go
```diff
@@ -85,10 +85,12 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline
 
 	if configFetchErr != nil {
 		log.Debug().Str("repo", repo.FullName).Err(configFetchErr).Msgf("cannot find config '%s' in '%s' with user: '%s'", repo.Config, pipeline.Ref, repoUser.Login)
+		err := &ErrConfigNotFound{Msg: fmt.Sprintf("pipeline definition not found in %s", repo.FullName)}
 		pipeline.Started = time.Now().Unix()
 		pipeline.Finished = pipeline.Started
 		pipeline.Status = model.StatusError
-		pipeline.Error = fmt.Sprintf("pipeline definition not found in %s", repo.FullName)
+		pipeline.Error = err.Msg
+		return nil, err
```
lafriks, line 93: Why not just add?

```suggestion
		if pipeline.Branch != repo.Branch {
			return nil, err
		}
```

xoxys: Why not just let the user control it? Why should we assume that the default branch requires a CI config for every user or setup? We can add options to disable/enable this feature later right now irgnoring it and logging an error in the backend is good enough IMO.

lafriks: Not for my use case, this will increase our support team work on finding out why pipelines are not working for new projects because even with current error that config was not found there were some developers that asked for help. Now it will be even harder to find that as developers won't see anything and it will just not work silently

xoxys: I understand your point but I still disagree. However not my decision, Ill leave this PR to someone else to pick up or make a decision.

lafriks: But from what you wrote previously this would work also for you so this would be good middle ground that would be good and not too breaking change and later on as you said this could be improved for other cases if needed

PR #2321, opened 2023-08-23
Fix mime type for custom.js and costom.css

6543:


6543:
yes :/

6543:
the func is not exec  at all ... so I expect it to be a bigger bug

PR #2325, opened 2023-08-25
close sidebar on outside click

xoxys:
[record.webm](https://github.com/woodpecker-ci/woodpecker/assets/3391958/f797be2c-09a9-4699-be85-7d5925e146ed)


xoxys:
What do you mean with closed directly? Clicking on the pipeline in the sidebar will not close the sidebar. Clicking outside will use the same method `setUserConfig` to change the user config, as clicking on the icon in the header does.

xoxys:
I don't get why you want to remove this feature. Removing it from the user store entire means reloading the page with a currently open sidebar breaks, and the sidebar is closed after the reload.

xoxys:
I never thought about the use-case of keeping the sidebar open all the time. Personally, I find it annoying that I had to use the icon in the header to close it. If the sidebar is so important to keep it always open, I would prefer that the pipeline not overlay other content, but that's a personal preference.

xoxys:
What's missing for this PR?

PR #2340, opened 2023-08-28
Show that repo is disabled

qwerty287:
And allow to show badges for disabled repos

Fixes 3rd checkbox from https://github.com/woodpecker-ci/woodpecker/discussions/2162#discussion-5498655

Screenshot 2023-08-28 at 11-13-23 Woodpecker


xoxys:
Why is there still an "Activate" button for disabled repos?

xoxys:
Now Im confused. Disabled does not mean disabled in the forge (eg archived repos or read-only for any other reason), it means disabled in Woodpecker? IMO the "Activate" button indicates this already, why do we need an extra label for it? Does it provide any benefit?

xoxys:
Is it expected that disabled repo settings can still be changed?

web/src/views/RepoAdd.vue
```diff
@@ -13,13 +13,19 @@
...
         <span v-if="repo.active" class="ml-auto text-wp-text-alt-100">{{ $t('repo.enable.enabled') }}</span>
-        <Button
-          v-if="!repo.active"
-          class="ml-auto"
-          :text="$t('repo.enable.enable')"
-          :is-loading="isActivatingRepo && repoToActivate?.forge_remote_id === repo.forge_remote_id"
-          @click="activateRepo(repo)"
-        />
+        <div v-else class="ml-auto flex items-center">
+          <router-link
```
xoxys, line 17: If that's a link, should e format it as a link? E.g. https://github.com/woodpecker-ci/woodpecker/blob/main/web/src/components/user/UserAPITab.vue#L20

xoxys:
> Yes, this is still possible. We can discuss this though

Was just wondering. Can be covered by a separate PR as well.

web/src/views/RepoAdd.vue
xoxys: I think all text links should look the same, and as we use it for swagger, forge links (PRs, refs, etc.) in the pipeline step list already I would prefer it.

xoxys: Just adding the hover class makes it not identifiable as a link :D users should see that this text is clickable, so I would add `text-wp-link-100` as well.

xoxys: I've done a lot of back and forth with the link color, but especially blue on black becomes an accessibility issue in terms of contrast quickly. Desaturating them might cause issues again, and for the color theme PR I decided to prioritize accessibility.

anbraten:
Just thought if we maybe should remove the disable repo function completely. Honestly users could just remove their repo or  leave it activated without using it.

xoxys:
I disagree. There were various reasons why we had to temp disable a repo to stop Drone/WP for it. Deleting it incl. all settings, secrets, build history etc. is not feasible in such cases. Please keep it.

xoxys:
For the sake of completeness, removing the webhook from the vcs + the repair button in wp could be an option but I'm still against it.

xoxys:
I'll leave it up to you, I don't have a strong opinion on this, but I don't see any improvement in removing the disable button and adding a "Pause" checkbox instead.

xoxys (approved):

PR #2343, opened 2023-08-28
Fix org fetching

### LINKED ISSUE #2341: Gitea orgs not listed as orgs
`{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-479ced3b25"}`

image

image

qwerty287:
Issue: Gitea responds with a user object even if it's actually an org. By changing the order and first checking the orgs endpoint which fails for users this is fixed.

Closes #2341

### DIFF
--- a/docs/docs/20-usage/35-advanced-yaml-syntax.md
+++ b/docs/docs/20-usage/35-advanced-yaml-syntax.md
@@ -88,3 +88,8 @@ steps:
      - echo echo from second step
      - <<: *post_cmds
 ```
+
+## References
+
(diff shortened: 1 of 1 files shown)


xoxys:
Ouch... Have you reported it upstream already? Thanks for looking into this!

xoxys:
How do we handle existing orgs?

xoxys:
Gitea has also a `gitea doctor` command to manually fix some tasks, could work here as well, even if that's not fully automated.

xoxys:
At least from my side, sure. Now its "broken" anyway :smile:

xoxys (approved):

PR #2345, opened 2023-08-28
Protected view

xoxys:
Ref: https://github.com/woodpecker-ci/woodpecker/discussions/2162#discussioncomment-6666864

Adds a `PipelineInfo` view to harmonize pipeline status information and add a review link to the approval view.

image

image

image


xoxys:
This is way more complicated.

xoxys:
Can look into it later.

xoxys:
I would like to have a button in the button row that we have already "Review", "Approve", "Decline" but afaik the current buttons can't be opened in a new tab on shift + click that's why I closed this way for now. Open for ideas/improvements.

xoxys:
Yes, saw that too. If we want to fix this first, fine for me as well. If we also don't like the button approach, it doesn't help at all :upside_down_face:

xoxys:
Ok, then I'm going to work on the buttons first.

xoxys:
Using a review button now, screenshot updated.

xoxys:
@qwerty287 please take a look if you find some time

web/src/assets/locales/en.json
```diff
@@ -255,6 +255,7 @@
         "log_auto_scroll_off": "Turn off automatic scrolling"
       },
       "protected": {
+        "review": "Review changes",
         "awaits": "This pipeline is awaiting approval by some maintainer!",
```
lafriks, line 259: Can we change this also:
```suggestion
        "awaits": "This pipeline is awaiting approval by a maintainer!",
```

xoxys: Unrelated to this PR but I can add it.

lafriks: Yes I know, but that sentence always seemed kinda strange :)

lafriks (approved):

PR #2327, opened 2023-08-25
Use full width header on pipeline view and show repo name

xoxys:
Ref: https://github.com/woodpecker-ci/woodpecker/discussions/2162#discussioncomment-6724450

**After:**
image

**Before:**

PR #2348, opened 2023-08-29
Increase header padding again

qwerty287:
From #2327 (https://github.com/woodpecker-ci/woodpecker/pull/2327#pullrequestreview-1600673521)

This probably needs some discussion first, but I think it looks ugly with the current padding.

**Before**
Screenshot 2023-08-29 at 17-17-00 Woodpecker

**After**
Screenshot 2023-08-29 at 17-16-48 Woodpecker


xoxys:
However its such a micro change, works for me :)

xoxys (approved):

PR #2356, opened 2023-08-30
Truncate commit message in pipeline log view header

xoxys:
I guess a regression introduced in https://github.com/woodpecker-ci/woodpecker/pull/2327

Before:
image

After:
image


xoxys:
I wasn't sure that's why I added the comment. If that's the way to go, we should expose `commit-title` and `commit-message` as dedicated metadata on the server-side because notification plugins might want to use it as well. Nevertheless, that's IMO a separate PR we should still truncate in case of very long titles.

PR #2357, opened 2023-08-30
Check permissions on repo lookup

qwerty287:
There was no permission check when looking up repos so you were able to get basic repo information even if you're not allowed to.

This uses `session.MustPull` (and set repo/perms before) to fix this.

### DIFF
--- a/web/src/components/layout/scaffold/Header.vue
+++ b/web/src/components/layout/scaffold/Header.vue
@@ -4,7 +4,7 @@
     :class="{ 'md:px-4': fullWidth }"
   >
     <FluidContainer :full-width="fullWidth" class="!py-0">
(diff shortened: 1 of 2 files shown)


server/api/repo.go
```diff
@@ -270,28 +269,15 @@ func ChownRepo(c *gin.Context) {
 // LookupRepo
 //
 //	@Summary	Get repository by full-name
-//	@Router		/repos/lookup/{repo_full_name} [get]
+//	@Router		/repos/lookup/{owner}/{name} [get]
```
anbraten, line 272: It could also be `owner/sub-owner/name`

anbraten (approved):

anbraten (approved):

Issue #2359, opened 2023-08-30
Provide separate `commit-title` and `commit-message` in pipeline metadata


xoxys:
We should expose `commit-title` and `commit-message` as dedicated metadata on the server-side because sometimes the title is more useful/readable instead of the full message e.g. in notification plugins or web UI.

posted by @xoxys in https://github.com/woodpecker-ci/woodpecker/issues/2356#issuecomment-1699254538

anbraten:
What if we just leave new lines in the commit messages. This way it would be quite easy for the different places to decide if they just want to show the first line or more.

xoxys:
Sure, I thought it might be more handy for users this way, and it's done just once on the server-side.

anbraten:
I think the issue is however that the commit message is a one-liner atm?

anbraten:
In that case we could just show the first line normally and on hover the full message with new-lines

PR #2361, opened 2023-08-30
Only show commit title

qwerty287:
https://github.com/woodpecker-ci/woodpecker/issues/2359


xoxys:
You might want to add it to the pipeline list as well
image

anbraten:
And in

- web/src/components/pipeline-feed/PipelineFeedItem.vue

image

xoxys:
Dont know if the hover worsk everywhere. E.g. in the pipeline list there is no explicit hover for the commit message only one hover for the entire pipeline.

xoxys:
:+1: I though it might now work because there is already a hover for the entire row. Anyway, my fault, works.

PR #2417, opened 2023-09-09
Add renovate deps groups

xoxys:
I would suggest to not group major version updates for now, or if you still prefer grouping, adding a dedicated group for major version updates.

The schedule will also reduce some noise and only create dependency update PRs once per day.


xoxys:
Sure, let's start with 5?

pat-s (approved):
LGTM and probably a good approach for all the golang deps!

PR #2434, opened 2023-09-10
fix: change config.config_data column type to longblob in mysql

s3lph:
See https://github.com/woodpecker-ci/woodpecker/pull/2418#discussion_r1320832841 for context.


anbraten (approved):
awesome. thanks for the pr

lafriks (approved):

PR #2464, opened 2023-09-16
Update gitea sdk

qwerty287:
update gitea sdk to 0.16.0. Renovate didn't notice this?


anbraten (approved):

pat-s:
@qwerty287 Maybe it can't resolve Gitea releases?

Might need some DEBUG logging during the renovate execution to see what's going on.

xoxys:
If renovate is doing something wrong or is not doing anything, my advice would be to create an issue and fix it first instead of fixing it manually (at least if it's not time critical).

pat-s:
Definitely, but somebody first needs to dig deeper into the renovate execution logs. I think only the owners can do this as we are using the public hosted app and not everyone has access?

If we would invoke renovate via WP itself, we would have more options to inspect the logs within WP and trigger builds with a custom config.

I guess we should also add some instructions to the README WRT to the whole renovate setup here.

xoxys:
There is a dashboard you were you can access all logs... https://developer.mend.io/github/woodpecker-ci/woodpecker

I guess the issue was related to the commit pinning foo we had in the previous version. Fixed now and renovate should pick-up future versions again.

image

pat-s:
Ah, didn't know we had access to it. That's great, also that you found the commit pinning being a potential reason for not picking it up. 👍

PR #2467, opened 2023-09-16
Add devx lable to compose file PRs

xoxys:
As discussed in https://github.com/woodpecker-ci/woodpecker/pull/2463#issuecomment-1722164170

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -5,6 +5,7 @@
   "packageRules": [
     {
       "matchManagers": ["docker-compose"],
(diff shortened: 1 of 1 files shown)


xoxys:
Sure if you only want to add it to a particular file we can adjust the package rule.

6543 (approved):

xoxys:
> Yes, this would be the best I think because the `devx` label is only used for development environment so mainly gitpod.

Works for me. As the other compose file is not suitable for production use anyway, I've considered a dev env as well.

PR #2001, opened 2023-07-15
Clickable repo link in feed

pat-s:
Testing

PR #2468, opened 2023-09-17
Some UI enhancement

qwerty287:
- Closes #2001 
- https://github.com/woodpecker-ci/woodpecker/discussions/2162#discussioncomment-6666895
- shows prettyRef for deployments

### DIFF
--- a/web/src/components/repo/pipeline/PipelineStepList.vue
+++ b/web/src/components/repo/pipeline/PipelineStepList.vue
@@ -14,7 +14,6 @@
         v-if="pipeline.event === 'pull_request'"
         class="flex items-center space-x-1 text-wp-link-100 hover:text-wp-link-200 min-w-0"
         :href="pipeline.link_url"
-        target="_blank"
       >
         <Icon name="pull_request" />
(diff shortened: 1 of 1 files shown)


anbraten (approved):

web/src/components/repo/pipeline/PipelineStepList.vue
```diff
@@ -10,21 +10,29 @@
         </div>
         <span>{{ pipeline.author }}</span>
       </div>
-      <div class="flex space-x-1 items-center min-w-0">
+      <a
+        v-if="pipeline.event === 'pull_request'"
+        class="flex items-center space-x-1 text-wp-link-100 hover:text-wp-link-200 min-w-0"
+        :href="pipeline.link_url"
+        target="_blank"
```
xoxys, line 17: Do we need to use `target="_blank"`? As a user I would prefer to control this on my own.

xoxys: Thanks for removing it. Looks like we use it all over the place. Might need a discussion if we want to use it or not.

Issue #2099, opened 2023-08-03
Release/support cycle


qwerty287:
This was already mentioned in https://github.com/woodpecker-ci/woodpecker/issues/1780, but there wasn't really a discussion about this or even a decision.

Now that we published the first "really" stable release 1.0.0, we should think about how we proceed with new releases and how to support them. This mainly includes the question: When do we want to publish a new release, either after a specified time or after a specified number of merged PRs/features?
For point releases, we had that question internally once after a dependency update with security fixes, and we decided to publish the release immediately after merging such security-related PRs. We also need to decide if we define rules for releasing point releases if there are no security issues, just some bugs. I'd suggest here to treat critical bugs as security fixes and release them immediately, and if there was a long time (a few weeks?) no point release because there were no security updates/critical bugs, also create a release so bug fixes can be used.

Also, the second part of this issue is the support cycle: How long do we want to support releases with point releases containing bug and security fixes?

6543:
My proposal:

as for majour releases, I would say if there are breaking changes to the inner-cores of woodpecker, like the pipeline definition, we should bump.
as for minor releases, if it requires user action to e.g. server-settings ...
as for patch releases, after some time if there are unreleased changes (e.g. 1week) or if it does acumulate enouth patches to be worth releasing. or asap on sec-fixes

as for maintaining: we should officially point out, that we only fix the current release branch but wellcome backports to older ones as long as there pipeline do work.

If there is a critical bug/sec issue we should consider the last release branch too if affected

pat-s:
My 2 cents:

- I am a friend of scheduled point releases - especially when there is enough activity. Every 3 months is often a good interval. This removes the ever-returning question of "are we ready?" & "should we still wait for PR XY?". Especially due to the experiences with 1.0.0 here I am a bit afraid that without a fixed interval the release cycles might be too long 😅 
- A point release can be a major bump if there is a breaking change. Don't be afraid of breaking changes WRT to version numbers!
- The problem with the above approach is that for important patches one would need to branch-off from the most latest release and only add the patches - otherwise all features that were merged in the meantime are also in the patch release and not bundled in the scheduled point release

6543:
the last point I dont get ... as we do squash merge and cherry-picking

zc-devs:
> I am a friend of scheduled point releases - especially when there is enough activity. Every 3 months is often a good interval.

[...]

PR #2476, opened 2023-09-19
Add workflow version

qwerty287:
Closes #1834


pipeline/frontend/yaml/parse.go
```diff
@@ -61,3 +69,18 @@ func ParseString(s string) (*types.Workflow, error) {
 		[]byte(s),
 	)
 }
+
+func checkVersion(b []byte) (int, error) {
+	ver := struct {
+		Version int `yaml:"version"`
```
anbraten, line 75: Int is enough, a string for semver is not necessary, right?

pipeline/frontend/yaml/error.go
```diff
@@ -16,6 +16,11 @@ package yaml
 
 import "errors"
 
+var (
+	ErrUnsuportedVersion = errors.New("unsuported yaml version detected")
+	ErrMissingVersion    = errors.New("missing yaml version")
```
anbraten, line 21: Maybe `pipeline config` or `workflow config` instead of yaml?

```suggestion
	ErrUnsuportedVersion = errors.New("unsuported pipeline config version detected")
	ErrMissingVersion    = errors.New("missing pipeline config version")
```

pipeline/frontend/yaml/parse.go
anbraten: Should we go with version 2 directly as it is kind of already the next version?

anbraten:
Migrating from pipeline to steps is already breaking anyways so it should be no problem to enforce the user to set a version from now on.

6543:
Its not about that we can, I'm personally against it, as i have the users in mind that are no yaml/devop-gurus

lafriks:
I'm against this as this would require for us to fix pipelines in close to 200 repositories with many having more than one pipeline with no real benefit (we already spent hours on changing pipeline -> steps)

As for versioning I see only benefit if we support multiple versions at the same time otherwise what benefit does it add?

If we do support multiple versions in the future it should default at least to the first version and then if you want newer pipeline features you can specify newer version

lafriks:
It's not so easy, for us to fix it we would require the following steps:
* clone repo
* create a branch and fix pipelines in it
* create PR
* approve PR from responsible persons

And make it 200 times

But the main question is what are the benefits of adding a version number?

lafriks:
You can't really automate approvals if they require different people and need to wait for pipeline statuses.

lafriks:
> I prefer require pinning, it's more readable and we can release config changes in minors too (with a new config version, we shouldn't do that too often though)

But what exactly does that give? Who would go over every pipeline to change version numbers with every new pipeline version release? This would require changing version numbers even if the pipeline itself is not affected and is not using changed syntax/features

6543 (changes requested):
My block stays ...

6543:
In this case drop it

lafriks:
The only reasonable way to support pipeline versions if we continue to support all of them in upcoming woodpecker versions. For example if wp <= 2.x supports pipeline version 1 and for version 3.x we add pipeline version 2 - then version 3.x should also support pipeline version 1 otherwise there is no point in versioned in pipelines

anbraten:
In that case I would revert the `steps` => `pipeline` breaking change as well for the current version.

> The only reasonable way to support pipeline versions if we continue to support all of them in upcoming woodpecker versions. For example if wp <= 2.x supports pipeline version 1 and for version 3.x we add pipeline version 2 - then version 3.x should also support pipeline version 1 otherwise there is no point in versioned in pipelines

I understand it like we would have pipeline versions (at least in general) independent from woodpecker versions. So if we update anything of the pipeline config we add a new version / compiler, but leave the current one untouched.

anbraten:
What if we say not defining a version will fallback to `version 1`. The `pipeline` => `steps` change and all other changes could then maybe already be `version 2`. With `version 1` being able to support `pipeline` or `steps`.

lafriks:
I would be ok with this only if we plan to support this and further pipeline versions in parallel but do we really want this?

anbraten:
An additional idea was also to extract the compiling code at some point, so the compiler is used like a plugin and other compilers (different versions, different file formats) can also simply be used as proposed in #2254

lafriks:
> That was the idea, but "support" means only bugfixing, no new features.

By support I mean that we will support running all pipeline versions in future woodpecker versions, for example
if we add pipeline version 2 in Woodpecker 3.0 and version 3 in Woodpecker 4.0 than:
* Woodpecker 2.0 will support pipeline version 1
* Woodpecker 3.0 will support running both pipeline version 1 (either unspecified or specified as 1) **and** pipeline version 2 (explicitly specified in pipeline definition)
* Woodpecker 4.0 will support running version 1, version 2 **and** version 3
etc

pipeline/schema/schema.json
```diff
@@ -34,7 +34,8 @@
       "type": "array",
       "minLength": 1,
       "items": { "type": "string" }
-    }
+    },
+    "version": { "type": "number" }
```
lafriks, line 38:
```suggestion
    "version": { "type": "number", "default": 1 }
```

lafriks:
If that is the case we would need to add support to allow to implement this in the future (different parsers based on version)

zc-devs:
Sorry to barge into such a heated discussion, but here are my 6.5 cents.
TLDR: I agree the most with *lafriks*.

1. Cant't say that I investigated other projects hardly, but I didn't find one who requires version.

Seems _GitLab_ has no version. _GitHub_ has actions version, but it is optional.
_Docker Compose_ [has version](https://docs.docker.com/compose/compose-file/compose-versioning), but it's optional and first one by default. Version 1 was supoorted long time until 2023. Also look at terrifying compatibilities matrices.

> If we do support multiple versions in the future it should default at least to the first version and then if you want newer pipeline features you can specify newer version

2. I see some inconsistency

> we will support running all pipeline versions in future woodpecker versions

#2099:
> we only fix the current release branch but wellcome backports to older ones

3. That point would probably sound rough, but don't get it personally. I respect all of you and your time working on the project.

It took over one month to [release](https://github.com/woodpecker-ci/woodpecker/releases/tag/v1.0.3) some security fixes. Yeah, it was not *so* security and other points, but here we are. Besides, I believe it will go right in the future. And thanks *qwerty287* for release.

In regards supporting multiple pipeline versions:
> I don't think the project has grown that much for as to have work power to do that

4. I see here a case, when _two_ versions are useful - *lafriks* [case](https://github.com/woodpecker-ci/woodpecker/pull/2476#issuecomment-1733317425).

He don't want to break development process by upgrading CI. Upgrade path below might work:
* Current version 1.0. Admin reads release notes for recently released version 2.0 and see that `pipeline` is deprecated in favor `steps` and will be removed in 3.0
* He assumes that 3.0 will be released in 2-3 months and creates issues in each team's backlog to migrate to the new syntax with due date in 1 month.
* Each team migrates to the new syntax in their sprint.
* By version 3.0 all repositories are migrated, all are happy.
* Otherwise issues happened during migration will be reported here and may be even fixed by 3.0.

It's just deprecation (you know) and likely don't needs versioning.

5. 
> old versions don't get new features

> if we update anything of the pipeline config we add a new version / compiler, but leave the current one untouched

Then even [non-breaking new feature](https://github.com/woodpecker-ci/woodpecker/pull/2550) introduces new version, doesn't it?

> This would require changing version numbers even if the pipeline itself is not affected and is not using changed syntax/features

6. "different file formats" can be implemented without versioning.

---

I wonder how it will be implemented in the code.

> An additional idea was also to extract the compiling code at some point, so the compiler is used like a plugin and other compilers

So, input is some data (yaml, json, binary) and output is some model (structure). We will have compiler and model per pipeline version. Then I see 2 approaches to deal with it.

1. `if-else` / `switch` everywhere outside compiler
```go
func Pod(step Interface) (Pod, error) {
    if reflect.TypeOf(step).Name() != "StepV1" {
        ...
    } else if reflect.TypeOf(step).Name() != "StepV2" {
        ...
    } else {
        return nil, fmt.Error("Version not found")
    }
}
```

2. Using DI framework (don't know how it works in go)
```go
// pod-v1.go
func Pod(step StepV1) Pod {
    return Pod{
        Image: step.Image
    }
}
```
```go
// pod-v2.go
func Pod(step StepV2) Pod {
    return Pod{
        Image: step.Registry + "/" + step.Image + ":" + step.ImgVersion
    }
}
```
```go
gin.register("pod-v1.go", "pod1")
gin.register("pod-v2.go", "pod2")
```
```go
steps, version := compiler.compile(pipeline)
podSvc := gin.get("pod" + version)
pod := podSvc.Pod(steps[0])
```

 
Anyway, for each pipeline version we have separate codebase. I think, it will be easier to support different WP branches than support a few versions in one branch.

---

Conclusion:

> we should not introduce versioning support and just do major upgrades on breaking pipeline changes (which should be reduced to as less as possible)

or do deprecation for one version and then drop it.

zc-devs:
> My idea was easier

Yeah, I see it now. My comments become almost irrelevant.

---

> Just some background about this ...

I suspected something like that. It's understandable and covered by my `and other points`. Thanks for sharing, though.

> No, these can be added to the latest version

Good to hear it.

> I don't really get your point here. Pipeline version != release branches ...

I thought (not now) that support of many versions / compilers would be more difficult than branches. We don't support many branches because of lack of resources => we can't support versioning.

> Versioning is mainly about breaking changes (and we currently do not plan to support anything except YAML anyways) and adding a new format will never be breaking.

It's reply to this [comment](https://github.com/woodpecker-ci/woodpecker/pull/2476#issuecomment-1752507944). I thought, that introducing other formats depends on versioning. You're right, it does not.

> you could also just say to upgrade all pipelines at once just before doing the upgrade

Doesn't matter when you update pipelines, before CI upgrade or after. It's about time frame. Without deprecation you have to update pipelines as closer as can to the upgrade point and it requires time and coordination from all teams. With deprecation there is much more time to update pipelines: each team can do it in its pace within wide time frame around month, for example.

Anyway, versioning make it available too. And as it's not so hard as I imagined, I like it 👍

pat-s:
@qwerty287 Should this be moved out of the 2.0.0 milestone? Re: the recent discussions. Decide yourself :)

pipeline/frontend/yaml/parse.go
```diff
@@ -37,3 +45,19 @@ func ParseString(s string) (*types.Workflow, error) {
...
+	}{}
+	_ = xyaml.Unmarshal(b, &ver)
+	if ver.Version == 0 {
+		// default: version 1
+		return 1, nil
```
6543, line 56: can we make this var const in "shared/const" ?

.woodpecker/docs.yml
```diff
@@ -1,3 +1,5 @@
+version: 1
+
```
6543, line 2:
```suggestion
```

have some with none set to test agains

pipeline/frontend/yaml/parse.go
lafriks: Probably version should be type/const/enum

.woodpecker/docs.yml
6543: Well yes our tests should cover all usecases ... and we must test against what most users do&will use

6543 (approved):

PR #2481, opened 2023-09-20
Remove step map (require list) 

qwerty287:
Based on https://github.com/woodpecker-ci/woodpecker/pull/2480 (must wait for this one)

Closes https://github.com/woodpecker-ci/woodpecker/discussions/2475

It is necessary to use:
```yaml
steps:
    - name: step
      image: xyz
```


6543:
I'm against this

6543:
> Why? I see that this will cause a lot of migration work (we could write some standalone migrator for this), but it unifies pipeline/workflow format.

Well thats a point ... but in any case we should first depricate it anyway and not remove it, if we did decide to remove it

6543:
I'm still against it just for the sace of have ~5lines less code for yaml decoding

PR #2503, opened 2023-09-27
Add bitbucket datacenter (server) support 

langecode:
This pull-requests re-introduces the Bitbucket Server support with a more or less complete rewrite of the forge implementation. We have a lot of on-premises git repositories hosted in Bitbucket Server and need a CI solution for running that and Woodpecker looks promising.

The implementation is based on external Bitbucket Server REST client library which we are maintaining and have created in another context. Besides the original support for Bitbucket the re-implementation also adds support for handling Bitbucket pull-request events.


langecode:
This should solve #2364

mzampetakis:
This would be a great addition! 
I was not aware about the `stash` name of Bitbucket server edition. I found a bit confusing seeing some parts of the code (e.g. configuration) and others as `Bitbucket Server` I would prefer to have a common name for documentation, configuration and code. I believe it makes more sense.
What do you think?

cmd/server/flags.go
```diff
@@ -439,6 +439,52 @@ var flags = append([]cli.Flag{
 		Usage:   "gitlab skip ssl verification",
 	},
 	//
+	// Bitbucket Server (previously Stash)
```
anbraten, line 442: Is it only supporting bitbucket server or also datacenter. As server reaches EOL it would be great if it works with the datacenter version as well and I would prefer to name it `bitbucket-datacenter` then as it seems Atlassian seems to plan to sell cloud or datacenter in the future (we could still try to support bitbucket-server as long as possible with the datacenter version as well if both share the same api).

langecode:
> This would be a great addition! I was not aware about the `stash` name of Bitbucket server edition. I found a bit confusing seeing some parts of the code (e.g. configuration) and others as `Bitbucket Server` I would prefer to have a common name for documentation, configuration and code. I believe it makes more sense. What do you think?

Yeah, we pretty much forked the code before the "old" Bitbucket server was removed so it is legacy from there. But we might try to clean up the naming removing the old "stash" name now that it has been out anyway. I'll try to do a search for "stash" :-)

web/src/views/repo/RepoWrapper.vue
```diff
@@ -38,7 +38,7 @@
     <Tab id="activity" :title="$t('repo.activity')" />
     <Tab id="branches" :title="$t('repo.branches')" />
     <Tab
-      v-if="config.forge === 'gitea' || config.forge === 'github' || config.forge === 'gitlab'"
+      v-if="config.forge === 'gitea' || config.forge === 'github' || config.forge === 'gitlab'  || config.forge == 'stash'"
```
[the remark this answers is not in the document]
langecode: Wouldn't this should also show pull-requests for the Bitbucket cloud forge? Removing the condition? But I noticed I did `==` instead of `===` must be getting rusty in JS/TS :)

cmd/server/flags.go
langecode: The API on datacenter should be the same but I am pretty sure we are running the "server" edition currently not the "datacenter" - the API documentation is the same though.

docs/versioned_docs/version-0.15/30-administration/10-server-config.md
```diff
@@ -304,7 +304,7 @@ See [Gitea configuration](vcs/gitea/#configuration)
 
 See [Bitbucket configuration](vcs/bitbucket/#configuration)
 
-### `WOODPECKER_STASH_...`
+### `WOODPECKER_BITBUCKET_DC_...`
```
anbraten:
```suggestion
### `WOODPECKER_STASH_...`
```
This is a doc page of an old version and should stay untouched AFAIK.

docs/docs/91-migrations.md
```diff
@@ -23,7 +23,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 - Updated Prometheus gauge `*_job_*` to `*_step_*`
 - Renamed config env `WOODPECKER_MAX_PROCS` to `WOODPECKER_MAX_WORKFLOWS` (still available as fallback)
 - The pipelines are now also read from `.yaml` files, the new default order is `.woodpecker/*.yml` and `.woodpecker/*.yaml` (without any prioritization) -> `.woodpecker.yml` ->  `.woodpecker.yaml`
-- Dropped support for [Coding](https://coding.net/), [Gogs](https://gogs.io) and Bitbucket Server (Stash).
+- Dropped support for [Coding](https://coding.net/), [Gogs](https://gogs.io).
```
anbraten, line 26:
```suggestion
- Dropped support for [Coding](https://coding.net/), [Gogs](https://gogs.io)
- Migrated from Bitbucket Stash support to Bitbucket Datacenter / server support
```

cmd/server/setup.go
```diff
@@ -180,16 +180,16 @@ func setupGitea(c *cli.Context) (forge.Forge, error) {
 	return gitea.New(opts)
 }
 
-// setupStash helper function to setup the Stash forge from the CLI arguments.
-func setupStash(c *cli.Context) (forge.Forge, error) {
+// setupBitbucketServer helper function to setup the Bitbucket DataCenter/Server forge from the CLI arguments.
+func setupBitbucketServer(c *cli.Context) (forge.Forge, error) {
```
anbraten, line 184:
```suggestion
func setupBitbucketDatacenter(c *cli.Context) (forge.Forge, error) {
```

anbraten (commented):
Thanks for the nice work in general and for updating the name

cmd/server/flags.go
langecode: I'll check with our Atlassian team - we might actually be running the datacenter version already - the "about" page in our Bitbucket does not seem to reveal this :-)

langecode:
Should some documentation be added around migration from 0.15, i.e., you have to rename your environment variables from `STASH` to `BITBUCKET_DC` to run major version 1.

anbraten:
> Should some documentation be added around migration from 0.15, i.e., you have to rename your environment variables from `STASH` to `BITBUCKET_DC` to run major version 1.

That would be great.

langecode:
I'll take a look at the linting errors from the builds there also :)

server/forge/bitbucketserver/convert_test.go
```diff
@@ -0,0 +1,242 @@
...
+	"time"
+
+	"github.com/franela/goblin"
+	bb "github.com/neticdk/go-bitbucket/bitbucket"
+
+	"github.com/woodpecker-ci/woodpecker/server/model"
+)
+
+//nolint:misspell
+func Test_helper(t *testing.T) {
```
anbraten, line 29:
```suggestion
func TestHelper(t *testing.T) {
```

web/src/components/atomic/Icon.vue
```diff
@@ -75,6 +75,7 @@ export type IconNames =
   | 'gitea'
   | 'gitlab'
   | 'bitbucket'
+  | 'bitbucket_dc'
```
anbraten, line 78: This is just the type. You also need to add an icon above or at least add it to the exisiting bitbucket icons `v-if`

server/forge/bitbucketserver/bitbucketserver.go
```diff
@@ -0,0 +1,625 @@
+// Copyright 2022 Woodpecker Authors
```
anbraten, line 1: Could you rename this file and the folder to `bitbucketdatacenter`

cmd/server/setup.go
```diff
@@ -177,6 +180,21 @@ func setupGitea(c *cli.Context) (forge.Forge, error) {
...
+	opts := bitbucketserver.Opts{
+		URL:               c.String("bitbucket-dc-server"),
+		Username:          c.String("bitbucket-dc-git-username"),
+		Password:          c.String("bitbucket-dc-git-password"),
+		ConsumerKey:       c.String("bitbucket-dc-consumer-key"),
+		ConsumerRSA:       c.String("bitbucket-dc-consumer-rsa"),
+		ConsumerRSAString: c.String("bitbucket-dc-consumer-rsa-string"),
+		SkipVerify:        c.Bool("bitbucket-dc-skip-verify"),
+	}
+	log.Trace().Msgf("Forge (bitbucketserver) opts: %#v", opts)
```
anbraten, line 194:
```suggestion
	log.Trace().Msgf("Forge (bitbucketdatacenter) opts: %#v", opts)
```

server/forge/bitbucketserver/bitbucketserver.go
```diff
@@ -0,0 +1,625 @@
...
+	"github.com/woodpecker-ci/woodpecker/server/model"
+	"github.com/woodpecker-ci/woodpecker/server/store"
+)
+
+const (
+	requestTokenURL   = "%s/plugins/servlet/oauth/request-token"
+	authorizeTokenURL = "%s/plugins/servlet/oauth/authorize"
+	accessTokenURL    = "%s/plugins/servlet/oauth/access-token"
+
+	secret = "045dfb11b042c3c44d68274fd22338e0" // TODO: Temporary
```
mzampetakis, line 47: Should this secret be hardcoded here or is it a leftover?

langecode: Good catch - it _is_ a left-over - kind of.

There is an "extra" security feature in the Bitbucket webhook such that when you register the webhook with Bitbucket you also provide a symmetric-key. Then when Bitbucket invokes the webhook it generates a signature using the symmetric-key such that the receiver (Woodpecker in this case) can validate that it was in fact a trusted call.

It would be no problem to generate the HMAC key however the `Hook` function only gets the `context.Context` and the `http.Request` - and before validating the incoming request that is pretty much all there is.

I see that there is no similar validation in the other Git integrations. Is there anywhere to store such a thing? Basically I guess it should be generated once and probably stored somewhere in the database.

langecode: One option could be to provide it through the configuration, of course, like the keys used to register Woodpecker as "application" in Bitbucket 🤔

langecode: Actually it seems that the jwt token part of the webhook url is signed using the "repo_hash" field from the db as HMAC key. In theory it should be possible to use the repo_hash also be for the Btibucket webhook signing feature.

Problem is that the repo (or remote forge id) is part of the signed payload and it is needed to fetch the key. So to fetch the signing key it would be required to parse the payload and afterwards validated - it is a little bit backwards 🤔

langecode: I have changed the code so that the `repo.Hash` is used as HMAC key for both the jwt in the webhook url (as it is for all webhooks) and then for the Bitbucket feature of signing the webhook payload.

langecode:
Btw, the failing build seems to be a false failure as it seems to be an issue with pulling with docker hub at the time.

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
```diff
@@ -0,0 +1,171 @@
+# Bitbucket Server
```
anbraten, line 1:
```suggestion
# Bitbucket Datacenter / Server
```

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
```diff
@@ -0,0 +1,171 @@
...
+
+Woodpecker comes with experimental support for Bitbucket Server, formerly known as Atlassian Stash. To enable Bitbucket Server you should configure the Woodpecker container using the following environment variables:
```
anbraten, line 3:
```suggestion
Woodpecker comes with experimental support for Bitbucket Datacenter / Server, formerly known as Atlassian Stash. To enable Bitbucket Server you should configure the Woodpecker container using the following environment variables:
```

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
```diff
@@ -0,0 +1,171 @@
...
+
+This stores the private RSA certificate in `key.pem`. The next command generates the public RSA certificate and stores it in `key.pub`.
+
+```nohighlight
+openssl rsa -in /etc/bitbucket/key.pem -pubout >> /etc/bitbucket/key.pub
+```
+
+Please note that the private key file can be mounted into your Woodpecker container at runtime or as an environment variable
+
+Private key file mounted into your Woodpecker container at runtime as a volume.
```
anbraten, line 43:
```suggestion
### Private key file mounted into your Woodpecker container at runtime as a volume
```

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
```diff
@@ -0,0 +1,171 @@
...
++     - WOODPECKER_BITBUCKET_DC_CONSUMER_RSA=/etc/bitbucket/key.pem
+      - WOODPECKER_BITBUCKET_DC_URL=http://stash.mycompany.com
++  volumes:
++     - /etc/bitbucket/key.pem:/etc/bitbucket/key.pem
+
+  woodpecker-agent:
+    [...]
+```
+
+Private key as environment variable
```
anbraten, line 67:
```suggestion
### Private key as environment variable
```

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
```diff
@@ -0,0 +1,171 @@
...
++     - WOODPECKER_BITBUCKET_DC_CONSUMER_RSA_STRING=contentOfPemKeyAsString
+      - WOODPECKER_BITBUCKET_DC_URL=http://stash.mycompany.com
+
+  woodpecker-agent:
+    [...]
+```
+
+## Service Account
+
+Woodpecker uses `git+https` to clone repositories, however, Bitbucket Server does not currently support cloning repositories with OAuth token. To work around this limitation, you must create a service account and provide the username and password to Woodpecker. This service account will be used to authenticate and clone private repositories.
```
anbraten, line 91:
```suggestion
Woodpecker uses `git+https` to clone repositories, however, Bitbucket Server does not currently support cloning repositories with an OAuth token. To work around this limitation, you must create a service account and provide the username and password to Woodpecker. This service account will be used to authenticate and clone private repositories.
```

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Registration
+
+You must register your application with Bitbucket Server in order to generate a consumer key. Navigate to your account settings and choose Applications from the menu, and click Register new application. Now copy & paste the text value from `/etc/bitbucket/key.pub` into the `Public Key` in the incoming link part of the application registration.
```
anbraten, line 95:
```suggestion
You must register your application with Bitbucket Datacenter / Server in order to generate a consumer key. Navigate to your account settings and choose `Applications` from the menu, and click `Register new application`. Now copy & paste the text value from `/etc/bitbucket/key.pub` into the `Public Key` in the incoming link part of the application registration.
```

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
```diff
@@ -0,0 +1,171 @@
...
+
+Read the value for `WOODPECKER_BITBUCKET_DC_CONSUMER_KEY` from the specified filepath
+
+### `WOODPECKER_BITBUCKET_DC_CONSUMER_RSA`
+
+> Default: empty
+
+Configures the path to your Bitbucket Server private key file.
+
+### `WOODPECKER_BITBUCKET_DC_CONSUMER_RSA_STRING`
```
anbraten, line 137: Now by looking at it I would swap those two to have it consistent to the other options:

`WOODPECKER_BITBUCKET_DC_CONSUMER_RSA` for the string and `WOODPECKER_BITBUCKET_DC_CONSUMER_RSA_FILE` to read from a file

docs/docs/91-migrations.md
```diff
@@ -28,6 +28,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 - Renamed config env `WOODPECKER_MAX_PROCS` to `WOODPECKER_MAX_WORKFLOWS` (still available as fallback)
 - The pipelines are now also read from `.yaml` files, the new default order is `.woodpecker/*.yml` and `.woodpecker/*.yaml` (without any prioritization) -> `.woodpecker.yml` -> `.woodpecker.yaml`
 - Dropped support for [Coding](https://coding.net/), [Gogs](https://gogs.io) and Bitbucket Server (Stash).
+- Migrated from Bitbucket Stash support to Bitbucket Datacenter / server support, rename env `WOODPECKER_STASH` to `WOODPECKER_BITBUCKET_DC` and webhooks need to be re-installed (e.g., using "repair" on the project in Woodpecker)
```
anbraten, line 31: I guess we should move this to the 2.0.0 area now

go.mod
```diff
@@ -34,6 +34,8 @@ require (
 	github.com/mattn/go-sqlite3 v1.14.18
 	github.com/moby/moby v24.0.7+incompatible
 	github.com/moby/term v0.5.0
+	github.com/mrjones/oauth v0.0.0-20190623134757-126b35219450
```
anbraten, line 37: That's a pretty old one suggesting to use https://pkg.go.dev/golang.org/x/oauth2#section-readme instead

server/forge/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -0,0 +1,637 @@
...
+		keyFileBytes, err = os.ReadFile(opts.ConsumerRSA)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		keyFileBytes = []byte(opts.ConsumerRSAString)
+	}
+
+	block, _ := pem.Decode(keyFileBytes)
+	PrivateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
```
anbraten, line 104:
```suggestion
	privateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
```

server/forge/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -0,0 +1,637 @@
...
+	}
+	hl := &httpLogger{Transport: auth}
+	cl := &http.Client{
+		Transport: hl,
+	}
+
+	return bb.NewClient(c.URLApi, cl)
+}
+
+func CreateConsumer(URL, ConsumerKey string, PrivateKey *rsa.PrivateKey) *oauth.Consumer {
```
anbraten, line 620:
```suggestion
func createConsumer(URL, ConsumerKey string, PrivateKey *rsa.PrivateKey) *oauth.Consumer {
```

server/forge/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -0,0 +1,637 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to query for user: %w", err)
+	}
+
+	return convertUser(user, accessToken.Token, c.url), nil
+}
+
+// Auth is not supported.
+func (*client) Auth(_ context.Context, _, _ string) (string, error) {
+	return "", fmt.Errorf("Not Implemented")
```
anbraten, line 161: Could you explain the oauth flow in the case for BB-DC. Normally `Login()` is used to redirect the user to the forge and `Auth` is used for the callback.

go.mod
anbraten: It should be in use for gitea as well if I remember correctly

docs/docs/91-migrations.md
```diff
@@ -20,6 +20,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 - Removed `WOODPECKER_DOCS` config variable
 - Renamed `link` to `url` (including all API fields)
 - Deprecated `CI_COMMIT_URL` env var, use `CI_PIPELINE_FORGE_URL`
+- Migrated from Bitbucket Stash support to Bitbucket Datacenter / server support, rename env `WOODPECKER_STASH` to `WOODPECKER_BITBUCKET_DC` and webhooks need to be re-installed (e.g., using "repair" on the project in Woodpecker)
```
6543, line 23: We removed it in the past ... but thats long gone. So a migration should be not mentioned

server/forge/bitbucketdatacenter/internal/client.go
```diff
@@ -0,0 +1,81 @@
...
+
+package internal
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/mrjones/oauth"
```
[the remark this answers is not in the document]
langecode: Bitbucket is using OAuth 1.0 - I think it is OAuth 2.0 which is supported - this is basically kept as-is from the Bitbucket Server implementation that was deleted. We haven't rewritten that part.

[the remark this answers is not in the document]
6543: ah ok :)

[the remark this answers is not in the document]
6543: just add that as comment to document it

docs/docs/91-migrations.md
langecode: Sure we can delete it - it was merely because this PR originally referenced #2364 where someone was running the older implementation of BB support.

6543:
just apply it - beside that ACK :)

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
```diff
@@ -0,0 +1,171 @@
+# Bitbucket Datacenter / Server
+
+Woodpecker comes with experimental support for Bitbucket Datacenter / Server, formerly known as Atlassian Stash. To enable Bitbucket Server you should configure the Woodpecker container using the following environment variables:
```
anbraten, line 3:
```suggestion
:::warning
Woodpecker comes with experimental support for Bitbucket Datacenter / Server, formerly known as Atlassian Stash. 
:::

To enable Bitbucket Server you should configure the Woodpecker container using the following environment variables:
```

server/forge/bitbucketdatacenter/convert.go
```diff
@@ -0,0 +1,150 @@
...
+		Branch:    ev.PullRequest.Source.DisplayID,
+		Title:     ev.PullRequest.Title,
+		Message:   "",
+		Avatar:    bitbucketAvatarURL(baseURL, ev.Actor.Slug),
+		Author:    authorLabel(ev.Actor.Name),
+		Email:     ev.Actor.Email,
+		Timestamp: time.Time(ev.Date).UTC().Unix(),
+		Ref:       fmt.Sprintf("refs/pull-requests/%d/from", ev.PullRequest.ID),
+		ForgeURL:  fmt.Sprintf("%s/projects/%s/repos/%s/commits/%s", baseURL, ev.PullRequest.Source.Repository.Project.Key, ev.PullRequest.Source.Repository.Slug, ev.PullRequest.Source.Latest),
+		Event:     model.EventPull,
```
anbraten, line 115: Do you know if this hook also gets called when closing a PR? We recently introduced a `pull_request_closed` event in case.

anbraten (approved):

server/forge/bitbucketdatacenter/convert.go
langecode: @anbraten  That is a good question - it should be possible to get an event from Bitbucket when pull-request is either deleted, merged or declined - that would be the Bitbucket "termination" states. I do not think the webhook is registered with these now but we could add that. You think we should include this as part of this or add it in a smaller PR afterwards?

langecode: It does - the webhook is currently registeret with these events (bitbucketdatacenter.go:404)
```go
Events: []bb.EventKey{bb.EventKeyRepoRefsChanged, bb.EventKeyPullRequestFrom},
```

I think I checked with the old Bitbucket plugin as well as the current Bitbucket cloud plugin which events they react to - there is s list of the events in the Bitbucket client: https://github.com/neticdk/go-bitbucket/blob/main/bitbucket/events.go#L55

6543 (approved):

server/forge/bitbucketdatacenter/convert.go
anbraten: If we don't add the event now users will later have to click repair on all their repos to update the webhook events. Just wanted to mention that. Both ways would be fine for me.

langecode: that is true 🤔 

shouldn't be that big a change, I'll see if I can find a few hours to put it in

langecode:
Well, merging from main I see the `Login` function of the forge interface has changed to not include request/response anymore - I am not sure whether that can work with Bitbucket OAuth 1.0 but I will have to find out. It will probably take some time to sort out - unfortunately. I'll have to investigate whether it is possible - the authentication is basically kept as-is from the old implementation. I guess we will also have to consider some other CI solution instead.

langecode:
It seems Bit bucket has actually gotten OAuth 2.0 support. I'll try to change the authentication to use OAuth 2.0 instead of the current old authentication.

docs/docs/91-migrations.md
```diff
@@ -21,6 +21,7 @@ Some versions need some changes to the server configuration or the pipeline conf
...
```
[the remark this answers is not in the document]
langecode: True we are getting pretty far from the original implementation - I'll remove the migration part. I kept it because there were some issue way back that concerned the Bitbucket support gone missing in this major version - so if that guy were to actually jump to the new implementation he would have to "migrate".

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
```diff
@@ -0,0 +1,95 @@
...
++     - WOODPECKER_BITBUCKET_DC_CLIENT_SECRET=yyy
++     - WOODPECKER_BITBUCKET_DC_URL=http://stash.mycompany.com
+
+  woodpecker-agent:
+    [...]
+```
+
+## Service Account
+
+Woodpecker uses `git+https` to clone repositories, however, Bitbucket Server does not currently support cloning repositories with an OAuth token. To work around this limitation, you must create a service account and provide the username and password to Woodpecker. This service account will be used to authenticate and clone private repositories.
```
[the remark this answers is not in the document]
langecode: Depending on the configuration Bitbucket also supports git+ssh. I guess the original implementation chose http because if you have access to the API of Bitbucket you should also have access clone using http.

anbraten (approved):
thanks for the additional work
code (apart from remaining points) lgtm
just prettier, lint and `go mod tidy` left

docs/docs/30-administration/11-forges/60-bitbucket_datacenter.md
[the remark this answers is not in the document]
anbraten: If http works we can stick to it for now and adjust later on

langecode:
> thanks for the additional work code (apart from remaining points) lgtm just prettier, lint and `go mod tidy` left

Thanks - I am just in the proces of testing the OAuth 2.0 implementation against our Bitbucket on-prem - I'll do the last clean up and let you know when I have verified that it works as intended.

server/forge/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -0,0 +1,608 @@
+// Copyright 2022 Woodpecker Authors
+// Copyright 2018 Drone.IO Inc.
```
6543, line 2:
```suggestion
```

server/forge/bitbucketdatacenter/bitbucketdatacenter_test.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2023 Woodpecker Authors
```
6543, line 1:
```suggestion
// Copyright 2024 Woodpecker Authors
```

server/forge/bitbucketdatacenter/convert.go
```diff
@@ -0,0 +1,155 @@
+// Copyright 2022 Woodpecker Authors
+// Copyright 2018 Drone.IO Inc.
```
6543, line 2:
```suggestion
// Copyright 2024 Woodpecker Authors
```

server/forge/bitbucketdatacenter/convert_test.go
```diff
@@ -0,0 +1,306 @@
+// Copyright 2022 Woodpecker Authors
+// Copyright 2018 Drone.IO Inc.
```
6543, line 2:
```suggestion
// Copyright 2024 Woodpecker Authors
```

server/forge/bitbucketdatacenter/fixtures/handler.go
```diff
@@ -0,0 +1,52 @@
+package fixtures
```
6543, line 1: missing header

server/forge/bitbucketdatacenter/internal/client.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2022 Woodpecker Authors
+// Copyright 2018 Drone.IO Inc.
```
6543, line 2:
```suggestion
// Copyright 2024 Woodpecker Authors
```

server/forge/bitbucketdatacenter/internal/client_test.go
```diff
@@ -0,0 +1,39 @@
+package internal
```
6543, line 1: missing header

6543:
@langecode just noticed that the license header should be updated ... a smal nit before merging ;)

langecode:
We have confirmed that OAuth 2.0 is indeed working now. Actually quiet nice to get rid of that old authentication stuff.

6543 (approved):

PR #2515, opened 2023-10-01
fix #2514 correct mime type used also when no content sent

### LINKED ISSUE #2514: when no custom css/js is defined, wrong mimetype causes Chrome to complain
### Component

server, web-ui

### Describe the bug

Just open Chrome developer tools window and navigate to https://ci.woodpecker-ci.org.

error message
`Refused to apply style from 'https://ci.woodpecker-ci.org/assets/custom.css' because its MIME type ('text/plain') is not a supported stylesheet MIME type, and strict MIME checking is enabled.`

Screenshot 2023-10-01 at 17 04 07

### System Info

```shell
* Chrome latest version (as of 2023-10-01)
* Woodpecker next-7d2bf54685
```

### Additional context

_No response_

### Validations

nitram509:
The fix is simple, just providing a file name, so `http.ServeContent(...)` can set the correct mimeype in case the content is zero bytes.

The test was just extended.

### DIFF
--- a/server/web/web.go
+++ b/server/web/web.go
@@ -72,19 +72,19 @@ func New() (*gin.Engine, error) {
 }
 
 func handleCustomFilesAndAssets(fs *prefixFS) func(ctx *gin.Context) {
-	serveFileOrEmptyContent := func(w http.ResponseWriter, r *http.Request, localFileName string) {
+	serveFileOrEmptyContent := func(w http.ResponseWriter, r *http.Request, localFileName string, contentTypeName string) {
 		if len(localFileName) > 0 {
 			http.ServeFile(w, r, localFileName)
 		} else {
 			// prefer zero content over sending a 404 Not Found
-			http.ServeContent(w, r, localFileName, time.Now(), bytes.NewReader([]byte{}))
+			http.ServeContent(w, r, contentTypeName, time.Now(), bytes.NewReader([]byte{}))
 		}
 	}
 	return func(ctx *gin.Context) {
 		if strings.HasSuffix(ctx.Request.RequestURI, "/assets/custom.js") {
-			serveFileOrEmptyContent(ctx.Writer, ctx.Request, server.Config.Server.CustomJsFile)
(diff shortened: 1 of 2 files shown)


nitram509:
> Does this also fix/replace #2321?

Good catch. Yes, it solves the same issue.  (#2515 is a duplicate to #2321).
I'm sorry I overlooked this one, I only checked for existing issues - my apologies.
Considering I did provide a test as well, I do suggest dropping the other PR.

nitram509:
> https://ci.woodpecker-ci.org/repos/3780/pipeline/8436/25

Fixed.
Did forgot `make format` and `make lint`

nitram509:
> @nitram509 https://ci.woodpecker-ci.org/repos/3780/pipeline/8436/25 fails

@qwerty287  Apologies, there seems to be one more issue.
the test I wrote is failing. there seems to be a relation to the machine it runs on, because "it worked on my OSX, M1".
Will make the test more robust in a minute.

server/web/web.go
```diff
@@ -72,19 +72,19 @@ func New() (*gin.Engine, error) {
 }
 
 func handleCustomFilesAndAssets(fs *prefixFS) func(ctx *gin.Context) {
-	serveFileOrEmptyContent := func(w http.ResponseWriter, r *http.Request, localFileName string) {
+	serveFileOrEmptyContent := func(w http.ResponseWriter, r *http.Request, localFileName, contentTypeName string) {
```
lafriks, line 75: contentTypeName not really what is passed there as file name is being passed to this param

nitram509: thanks for the feedback.
well, the variable name might not be named optimal.
According to the docs, `http.ServeContent(..., name, ...)` just using the name to determine the content type.

Any proposal on how to improve the name, @lafriks ?

lafriks: I would probably call it `fileName`

6543 (approved):

PR #2518, opened 2023-10-03
Fix incorrect yaml syntax for `ref` in docs

categulario:
`when` contents for the `ref` option were written as a mapping, but it must be a list.

### DIFF
--- a/docs/docs/20-usage/20-pipeline-syntax.md
+++ b/docs/docs/20-usage/20-pipeline-syntax.md
@@ -348,8 +348,8 @@ This allows you to filter, for example, tags that must start with **v**:
 
 ```yaml
 when:
-  event: tag
-  ref: refs/tags/v*
+  - event: tag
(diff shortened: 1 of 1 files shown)


categulario:
you're right... in any case, the rest of the examples use explicit lists even if they consist of only one mapping. Maybe for consistency it is worth making this change?

categulario:
if that's a yaml-specific thing then the woodpecker docs might not be the right place to document it. Having a uniform style across examples helps reduce the cognitive load. Adding options (like the fact that you can use single-element lists or a mapping) doesn't seem necessary or beneficial. People who know the yaml spec will surely use it.

in any case, do you have a link to the yaml spec that says this?

categulario:
Ohhh, I see. Then it is worth documenting. In which places is it implemented? is it only for the `when` directive or does it apply everywhere a list is expected?

categulario:
:bow:

PR #2527, opened 2023-10-05
Refactor pipeline parsing and forge refreshing

qwerty287:
- refactor pipeline parsing
    - do not parse the pipeline multiple times to perform filter checks, do this once and perform checks on the result directly
    - code deduplication
- refactor forge token refreshing
    - move refreshing to a helper func to reduce code

### DIFF
--- a/server/pipeline/create.go
+++ b/server/pipeline/create.go
@@ -87,8 +87,8 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline 
 		}
 		configs = append(configs, config)
 	}
-        // link pipeline to persisted configs
-	if err := persistPipelineConfigs(_store, configs, pipeline.ID); err != nil {
+	// link pipeline to persisted configs
+	if err := linkPipelineConfigs(_store, configs, pipeline.ID); err != nil {
 		msg := fmt.Sprintf("failed to find or persist pipeline config for %s", repo.FullName)
 		log.Error().Err(err).Msg(msg)
 		return nil, fmt.Errorf(msg)
--- a/server/pipeline/restart.go
+++ b/server/pipeline/restart.go
@@ -88,7 +88,7 @@ func Restart(ctx context.Context, store store.Store, lastPipeline *model.Pipelin
 		}
 		return newPipeline, nil
 	}
-	if err := persistPipelineConfigs(store, configs, newPipeline.ID); err != nil {
+	if err := linkPipelineConfigs(store, configs, newPipeline.ID); err != nil {
 		msg := fmt.Sprintf("failure to persist pipeline config for %s.", repo.FullName)
 		log.Error().Err(err).Msg(msg)
 		return nil, fmt.Errorf(msg)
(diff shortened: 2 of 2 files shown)


server/forge/refresh.go
```diff
@@ -0,0 +1,38 @@
+package forge
```
anbraten, line 1: missing license header

anbraten (commented):
Thanks for the work. Always wanted to unclutter this part as well 😅. Looks good in general, but haven't tested it yet.

server/pipeline/topic.go
```diff
@@ -33,6 +33,7 @@ func publishToTopic(c context.Context, pipeline *model.Pipeline, repo *model.Rep
 		},
 	}
 	pipelineCopy := *pipeline
+	pipelineCopy.Workflows = nil // never send workflow lists
```
anbraten, line 36: From testing I think this is needed. If we need pipelines to be approved we do not get the workflow list after accepting by the update.

anbraten: AFAIK this function is sending an event to the UI with pipeline updates. If we approve a pipeline the workflows get parsed and were previously send via this event as part of the pipeline. Now we do not get the workflow list anymore until we refresh the full page doing a fetch for the pipeline containing the added workflows list.

```suggestion
```

server/pipeline/create.go
```diff
@@ -37,109 +36,65 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline
...
-
-		updatePipelineStatus(ctx, pipeline, repo, repoUser)
-
-		return pipeline, nil
-	}
-
-	pipeline, pipelineItems, err := createPipelineItems(ctx, _store, pipeline, repoUser, repo, forgeYamlConfigs, nil)
-	if err != nil {
-		msg := fmt.Sprintf("failure to createPipelineItems for %s", repo.FullName)
+	if err := persistPipelineConfigs(_store, configs, pipeline.ID); err != nil {
```
anbraten, line 90: Was a bit confused about `findOrPersistPipelineConfig` and `persistPipelineConfigs` first. The second is basically the linkage of an existing config (in this case prev created one) to a pipeline, right? Do you maybe know better names for them to prevent confusion?

server/pipeline/create.go
```diff
@@ -37,109 +36,65 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline
...
 	if err != nil {
-		msg := fmt.Sprintf("failure to save pipeline for %s", repo.FullName)
-		log.Error().Err(err).Msg(msg)
-		return nil, fmt.Errorf(msg)
+		msg := fmt.Errorf("failure to save pipeline for %s", repo.FullName)
+		log.Error().Err(err).Msg(msg.Error())
+		return nil, msg
 	}
 
+	pipeline = setPipelineStepsOnPipeline(pipeline, pipelineItems)
```
anbraten: Should we move this line below `setGatedState(repo, pipeline)` so those two adjusments are kinda bundeled?

anbraten:
Tested most cases I could imagine and seems to work now. I removed a property from the UI event as it was removed from the server code. Hope you are okay with it being added in here.

server/pipeline/create.go
anbraten: Maybe sth like:

```suggestion
        // link pipeline to persisted configs
	if err := persistPipelineConfigs(_store, configs, pipeline.ID); err != nil {
```

anbraten: That would make the code way more self-explanatory as well 😅

anbraten (approved):

Issue #2505, opened 2023-09-27
Node Selector assigns wrong architecture for builds


pat-s:
### Component

agent

### Describe the bug

Recently, I am seeing builds being assigned to the wrong nodes on our cluster. 
Inspecting the pods, I can see that the `platform` env, which comes from the matrix definition, is set to `amd64`. However, the nodeSelector is set to `arm64` instead.

```
Environment:
  platform:                        linux/amd64 

[...]

Node-Selectors:              kubernetes.io/arch=arm64 
```

The WF definition is as follows

```
platform: ${platform}

matrix:
  include:
    - DISTRIBUTION: ubuntu
      RELEASE: focal
      OS_RELEASE: 20.04
      platform: linux/amd64
    - DISTRIBUTION: ubuntu
      RELEASE: jammy
      OS_RELEASE: 22.04
      platform: linux/amd64
    - DISTRIBUTION: redhat/ubi8-minimal
      RELEASE: latest
      platform: linux/amd64
      OS_RELEASE: rhel8
    - DISTRIBUTION: redhat/ubi9-minimal
      RELEASE: latest
      platform: linux/amd64
      OS_RELEASE: rhel9
```

Could it be that https://github.com/woodpecker-ci/woodpecker/pull/2048 changed the logic in a way that `nodeSelector` is always set to `CI_SYSTEM_PLATFORM` and does not honor `platform` env anymore?

https://github.com/woodpecker-ci/woodpecker/blob/0eacbe8a33823b75e1ecdd738f6c0cad00bd7d68/pipeline/backend/kubernetes/pod.go#L112

Or maybe one of the changes within the recent months caused this behavior?
Tagging @6543 and @zc-devs as you both did modifications in this area lately.

I am not having enough experience in GO to make an educated judgement right now - all I can see is that `CI_SYSTEM_PLATFORM` is not honoring `platform` env and therefore the `nodeSelector` is wrong in the end :)

### System Info

```shell
`next-f8e91f00aa`
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #2531, opened 2023-10-05
Update documentation WRT to recent `$platform` changes

pat-s:
I am unsure if the following `when:` syntax is also deprecated and should be updated:

```yml
when:
  - platform: linux/amd64
```

@qwerty287 Do you know more here?

### DIFF
--- a/docs/docs/20-usage/30-matrix-workflows.md
+++ b/docs/docs/20-usage/30-matrix-workflows.md
@@ -120,7 +120,8 @@ matrix:
     - linux/amd64
     - linux/arm64
 
-platform: ${platform}
+labels:
+  platform: ${platform}
 
 steps:
   test:
@@ -136,3 +137,7 @@ steps:
     when:
       platform: linux/arm*
 ```
+
+:::note
+If you want to control the architecture of a runner on Kubernetes, see [the nodeSelector documentation of the Kubernetes backend](./40-kubernetes.md#nodeSelector).
+:::
--- a/docs/docs/30-administration/22-backends/40-kubernetes.md
+++ b/docs/docs/30-administration/22-backends/40-kubernetes.md
@@ -80,9 +80,31 @@ See the [kubernetes documentation](https://kubernetes.io/docs/concepts/security/
 
 ### nodeSelector
 
-Specify the label which is used to select the node where the job should be executed. Labels defined here will be appended to a list already containing "kubernetes.io/arch".
-By default the pod will use "kubernetes.io/arch" inferred from top-level "platform" setting which is deducted from the agents' environment variable CI_SYSTEM_PLATFORM. To overwrite this, you need to specify this label in the nodeSelector section.
-See the [kubernetes documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) for more information on using nodeSelector.
+Specify the label which is used to select the node where the job should be executed.
+Labels defined here will be appended to a list already containing "kubernetes.io/arch".
+By default the pod will use "kubernetes.io/arch" inferred from top-level "platform" setting which is deducted from the agents' environment variable CI_SYSTEM_PLATFORM.
+To overwrite this, you need to specify this label in the `nodeSelector` section of the `backend_options`.
+See the [kubernetes documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) for more information on about the `nodeSelector` setting.
+
+A practical example is when you are running a matrix-build and you want to delegate specific elements of the matrix to run on a certain architecture.
+In this case, define an arbitrary key in the matrix section of the respective matrix element:
+
+```yml
+matrix:
+  include:
+    - NAME: runner1
+      ARCH: arm64
+```
+
+And then overwrit the `nodeSelector` in the `backend_options` section of the step(s):
+
+```yml
+[...]
+    backend_options:
+      kubernetes:
+        nodeSelector:
+          kubernetes.io/arch: "${ARCH}"
+```
 
 ### tolerations
 


pat-s:
Maybe @zc-devs?

docs/docs/30-administration/22-backends/40-kubernetes.md
```diff
@@ -80,9 +80,32 @@ See the [kubernetes documentation](https://kubernetes.io/docs/concepts/security/
...
 ### nodeSelector
 
-Specify the label which is used to select the node where the job should be executed. Labels defined here will be appended to a list already containing "kubernetes.io/arch".
-By default the pod will use "kubernetes.io/arch" inferred from top-level "platform" setting which is deducted from the agents' environment variable CI_SYSTEM_PLATFORM. To overwrite this, you need to specify this label in the nodeSelector section.
-See the [kubernetes documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) for more information on using nodeSelector.
+Specifies the label which is used to select the node on which the job will be executed.
+
+Labels defined here will be appended to a list which already contains `"kubernetes.io/arch"`.
+`"kubernetes.io/arch"` is inferred from the agents' environment variable `CI_SYSTEM_PLATFORM`.
+Without a manual overwrite, builds will be randomly assigned to the runners and inherit their respective architectures.
```
zc-devs, line 87: This and below should be more concrete and placed in [Codeberg-CI/examples](https://codeberg.org/Codeberg-CI/examples) or in the separate `examples` directory here, I think.

docs/docs/20-usage/30-matrix-workflows.md
```diff
@@ -136,3 +137,7 @@ steps:
     when:
       platform: linux/arm*
 ```
+
+:::note
+If you want to control the architecture of a pipeline on a Kubernetes runner, see [the nodeSelector documentation of the Kubernetes backend](../30-administration/22-backends/40-kubernetes.md#nodeSelector).
```
zc-devs, line 142: This is not related directly to the matrix builds.

Also you can control architecture of pipeline by labels, which is more canonical.

docs/docs/30-administration/22-backends/40-kubernetes.md
```diff
@@ -80,9 +80,32 @@ See the [kubernetes documentation](https://kubernetes.io/docs/concepts/security/
...
+
+To overwrite this, one needs to set the label in the `nodeSelector` section of the `backend_options`.
+A practical example for this is when running a matrix-build and delegating specific elements of the matrix to run on a specific architecture.
+In this case, one must define an arbitrary key in the matrix section of the respective matrix element:
```
zc-devs, line 91: This is one approach, but there is another: pipelines distribution by labels. The last should be preferable, I think.

zc-devs (commented):
Wait a little bit, I'll explain more precisely later.

docs/docs/30-administration/22-backends/40-kubernetes.md
pat-s: What do you mean by "more concrete"? I think having a short code example directly here is fine, https://codeberg.org/Codeberg-CI/examples is more for full pipelines.

docs/docs/20-usage/30-matrix-workflows.md
pat-s: I put it there as this is common use case if you're running on k8s and working with large matrices. And having a link to the respective docs explaining "how to" seems quite useful to me?

docs/docs/30-administration/22-backends/40-kubernetes.md
pat-s: Happy to change it if there's a better way. Curious to see the "labels" approach as I failed to get it working.

zc-devs:
_I've not used matrix and architecture builds. All that I'll say below is based on [documentation](https://woodpecker-ci.org/docs/next/usage/pipeline-syntax#labels), [experiments](https://github.com/woodpecker-ci/woodpecker/issues/2505#issuecomment-1746794776), common sense and my expectations :)_

---

First of all, let's work out what is [the Matrix workflows](https://woodpecker-ci.org/docs/next/usage/matrix-workflows) and how they work. Consider next example:
```yaml
matrix:
  OS:
    - linux
    - windows
  ARCH:
    - x86
    - arm

steps:
  build:
    image: alpine
    commands:
      - echo "This is a build for ${OS} operating system and ${ARCH} processor architecture"
```

Matrix - is just set of various variables combinations. And workflow is just a template. Example becomes 4 workflows after interpolation:
```yaml
steps:
  build:
    image: alpine
    commands:
      - echo "This is a build for linux operating system and x86 processor architecture"
```
```yaml
steps:
  build:
    image: alpine
    commands:
      - echo "This is a build for linux operating system and arm processor architecture"
```
```yaml
steps:
  build:
    image: alpine
    commands:
      - echo "This is a build for windows operating system and x86 processor architecture"
```
```yaml
steps:
  build:
    image: alpine
    commands:
      - echo "This is a build for windows operating system and arm processor architecture"
```
These workflows are scheduled to run.

---

Next, let's look at [labels](https://woodpecker-ci.org/docs/next/usage/pipeline-syntax#labels). I can hardly add anything, it is clear enough. Note that each Agent have `platform=agent-os/agent-arch` label.

If we look at next paragraph, we'll see example of pipeline distribution by labels.

---

Assume, we have 2 Agents: one on `linux/arm` and one on `windows/amd64`. And next workflow definition:
```yaml
matrix:
  PLATFORM:
    - linux/arm
    - windows/amd64

steps:
  build:
    image: golang:${PLATFORM}
    commands:
      - go build
```

It interpolates into 2 workflows:
```yaml
steps:
  build:
    image: golang:linux/arm
    commands:
      - go build
```
```yaml
steps:
  build:
    image: golang:windows/amd64
    commands:
      - go build
```

They would be run randomly on Agents, because we didn't set any label.

If you add:
```yaml
labels:
  location: mordor
```
then they won't run at all, because there is no Agent in Mordor (with label `location=mordor`) :)

Let's see what if you add standard `platform` label:
```yaml
matrix:
  PLATFORM:
    - linux/arm
    - windows/amd64

labels:
  platform: ${PLATFORM}

steps:
  build:
    image: golang:${PLATFORM}
    commands:
      - go build
```

Then we'll have 2 pipelines:
```yaml
labels:
  platform: linux/arm

steps:
  build:
    image: golang:linux/arm
    commands:
      - go build
```
```yaml
labels:
  platform: windows/amd64

steps:
  build:
    image: golang:windows/amd64
    commands:
      - go build
```

They will be scheduled each one on dedicated Agent:




That is how it works on backends, other than Kubernetes. This is _labels approach_. It should work in Kubernetes too:




This looks more canonical to me (from Woodpecker perspective) and should be used in general.

---

What's about `nodeSelector`?

I think `nodeSelector` in Kubernetes is similar to `label` in Woodpecker, but exists in Kube only. It contains platform by default like labels.

So, pipelines distribution/scheduling by platfrom should work out-the box via labels. We can also use labels and node selector to schedule _workflows_ (entirely) more precisely, by location, for example. In that case labels approach is prefferable, probably, too. Althogh, in Kube backend `taints` may be better than `nodeSelector`. Arguable...

What doess node selector allow to do, that lablels don't? This is scheduling by _workload_ - by step. Consider next example:
```
steps:
  prepare:
    image: alpine
    commands:
      - tar -xf huge_archive.tar
    backend_options:
      kubernetes:
        nodeSelector:
          storage: "nvme" // we need fast storage to extract archive quickly
  build:
    image: golang
    commands:
      - go build
    backend_options:
      kubernetes:
        nodeSelector:
          cpu: "256-core" // we need fast cpu to build quickly
  build:
    image: aws-cli
    commands:
      - aws s3 upload // don't remember exact command
    backend_options:
      kubernetes:
        nodeSelector:
          network: "40Gbit" // we need fast uplink to upload quickly
```

Taints should be better, but you understand general idea. That is _specific use case_ from my comment.

---

Hope, this help in some way.

It's time for scheduling via node selector.

```
matrix:
  PLATFORM:
    - mac/m1
    - windows/amd64

steps:
  build:
    image: golang:${PLATFORM}
    commands:
      - go build
    backend_options:
      kubernetes:
        nodeSelector:
          kubernetes.io/arch: ${PLATFORM}
```

Interpolates into:
```
steps:
  build:
    image: golang:mac/m1
    commands:
      - go build
    backend_options:
      kubernetes:
        nodeSelector:
          kubernetes.io/arch: mac/m1
```
```
steps:
  build:
    image: golang:windows/amd64
    commands:
      - go build
    backend_options:
      kubernetes:
        nodeSelector:
          kubernetes.io/arch: windows/amd64
```



Note, that labels are not used (explicitly) at all. Server and Agents run on completely different platfroms than build nodes.

pat-s:
@zc-devs Oh wow, that must have taken you an hour or more 😅 I am a bit overwhelmed almost by that large response. Highly appreciated though!

Overall, I think most of it was already clear to me 😅 I was more focusing on the exact notation with `labels` only - instead of combining it with the `backend_options`.
I am referring to this example of yours

```yml
matrix:
  PLATFORM:
    - linux/arm
    - windows/amd64

labels:
  platform: ${PLATFORM}

steps:
  build:
    image: golang:${PLATFORM}
    commands:
      - go build
```

Which sadly does not work when writing it as (which is what I need in my matrix)

```yml
### NOT WORKING
matrix:
  include:
    - name: A
      PLATFORM: linux/arm
    - name: B
      PLATFORM: windows/amd64

labels:
  platform: ${PLATFORM}

steps:
  build:
    image: golang:${PLATFORM}
    commands:
      - go build
```

as it will result in 

```
Execution Error:
failed to parse pipeline: Cannot unmarshal '<nil>' of type <nil> into a string value
```

This might be a general issue of the `include` property. In general though I agree that writing it that way would be the "way to go" (as beforehand I was doing it exactly like this but using only `platform: ${PLATFORM}` instead of `labels.platform: ${PLATFORM}`).

Interestingly, the above works when `${PLATFORM}` is sent to the `backend_options` and the top-level `labels:` is not used. Which is why I used the notation in this way for this PR. So maybe it's a backend issue in this specific case of using `include` within `matrix`.

And summing @qwerty287 here again due to #2480 - shouldn't the `labels.platform: ${PLATFORM}` notation work in general and is this potentially a bug? Maybe because the vars are handed over in a map instead of a string in the `include:` example?

docs/docs/30-administration/22-backends/40-kubernetes.md
zc-devs: Now you have only two small pieces of your pipeline, but make statements like `builds will be randomly assigned` and `one must define`. This is not so obvious, not always true and you didn't provide full config when it's true. Each part is described well enough in its own page. If you want to show some use case when they are combined, you should show full pipeline. Of course, it should be as much as possible short, but full. Good examples are [1](https://woodpecker-ci.org/docs/next/usage/matrix-workflows#example-matrix-pipeline-using-multiple-platforms) and the last example from [comment](https://github.com/woodpecker-ci/woodpecker/pull/2531#issuecomment-1757460570).

Though, placing in Codeberg-CI/examples or in the separate examples directory might be overkill. Depends on pipeline length and specificity.

pat-s: > This is not so obvious, not always true and you didn't provide full config when it's true.

Hmm. It's the default which is inferred by "Without a manual overwrite,". When is it not true? Builds will be assigned to runners based on resource needs and what the scheduler thinks it's best - no? So each build could go to a different arch - it might not be explicitly random but from a user perspective it is.

> If you want to show some use case when they are combined, you should show full pipeline

Hm, I don't agree here. I think this is a style question.

pat-s:
@zc-devs The current state here is valid code and works (and works around the issue reported when not using `backend_options`). Hence I'd propose to merge this and update it in another PR (if motivation exists). 
You are also welcome to do it, possibly using full examples etc. What do you think?
It's definitely better than before as the current documentation is not resulting in a working YAML.

zc-devs:
I think that it _was_ an issue. Also I wanted to see more rounded sentences as it is one of the possibilities. And from documentation it should be clear what exactly matters in that case (are there something between matrix and backend options like labels? should they be there?)

> merge this and update it in another PR

Why not? There is no blocking from me :)

> You are also welcome to do it

May not me? :) Although I would glad to see information from my comment in documentation (with the schemas drew in a proper way).

zc-devs:
> wanna join the maintainers team?

Thank you, but may be some time later.

PR #2537, opened 2023-10-07
Fix gitlab hooks and simplify config extension

### LINKED ISSUE #2534: Missing root path in GitLab webhook
### Component

server

### Describe the bug

When enabling a new GitLab repository, the webhook is set in GitLab with the API address `https://example.com/api/hook`. However, the server has the `WOODPECKER_ROOT_PATH=/ci`, and it should be `https://example.com/ci/api/hook` in my case.

Changing it manually in the GitLab configuration fails because the GitLab webhook token gets erased and I don't know where to get it from, so I cannot use the webhook.

I suspect the problem is here: https://github.com/woodpecker-ci/woodpecker/blob/931af16d14e94fbf6fbb94a167251fc2d6bedded/server/forge/gitlab/gitlab.go#L450C57-L450C57

### System Info

```shell
I'm using woodpecker-server 1.0.2 in NixOS:

{"source":"https://github.com/woodpecker-ci/woodpecker","version":"1.0.2"}
```

### Additional context

_No response_

### Validations

qwerty287:
- closes #2534 
- remove `IsConfigured` func from config extension. If `server.Config.Services.ConfigService != nil` it is always configured


server/pipeline/restart.go
```diff
@@ -51,8 +51,8 @@ func Restart(ctx context.Context, store store.Store, lastPipeline *model.Pipelin
 		pipelineFiles = append(pipelineFiles, &forge_types.FileMeta{Data: y.Data, Name: y.Name})
 	}
 
-	// If config extension is active we should refetch the config in case something changed
-	if server.Config.Services.ConfigService != nil && server.Config.Services.ConfigService.IsConfigured() {
+	// If the config extension is active we should refetch the config in case something changed
+	if server.Config.Services.ConfigService != nil {
```
anbraten, line 55: Wouldn't this also use the config service even if the url is empty?

anbraten (approved):

PR #1889, opened 2023-06-27
Add pull-request implementation for bitbucket forge 

mzampetakis:

PR #2539, opened 2023-10-07
Fix PR event trigger and list for bitbucket repos

mzampetakis:
Without this woodpecker ci could not trigger a CI when a new PR is created within the repo.


mzampetakis:
The registered events before the change:
Screenshot 2023-10-07 at 14 08 59

and after the fix:
Screenshot 2023-10-07 at 14 26 34

Now CI jobs can be triggered when a new PR is creted:
Screenshot 2023-10-07 at 14 27 10

anbraten (approved):

anbraten:
Does the bitbucket backend already support PR listing as well? In that case we could list those in the UI (similar to branches).

mzampetakis:
> Does the bitbucket backend already support PR listing as well? In that case we could list those in the UI (similar to branches).

Yes, we could! As we have [this](https://github.com/woodpecker-ci/woodpecker/pull/1889) in place we could enable them. Should I (try to) add it to this PR?

anbraten (approved):

Issue #2545, opened 2023-10-08
Support AppArmor annotations and securityContext in kubernetes backend


dominic-p:
### Clear and concise description of the problem

I'm trying to run a container image build step in my pipeline that uses [buildah](https://quay.io/repository/buildah/stable?tab=info). I'm running into a lot of permission errors like this:

```
time="2023-10-07T19:59:26Z" level=error msg="writing \"0 0 4294967295\\n\" to /proc/17/uid_map: write /proc/17/uid_map: operation not permitted"
time="2023-10-07T19:59:26Z" level=error msg="(Unable to determine exit status)"
```

After some experimentation with podman and buildah containers on my cluster, I have determined that I need to configure a few different settings at the Pod level for this to work:

- `metadata.annotations`: I need to be able to set `container.apparmor.security.beta.kubernetes.io/<container_name>` to use a custom apparmor profile for my buildah container. I would like to use the `WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS` env variable, but since the container names are randomly generated I'm not sure how to make this work.
- `spec.securityContext`: The only way I have found to make buildah work is to run it as non-root. This means that I need to set `runAsUser`, `runAsGroup`, and `fsGroup` on the Pod level `securityContext` (not the container level).

### Suggested solution

Adding support for `securityContext` seems like it should be fairly straightforward. We could just follow that same approach that is currently used for `resources`, `nodeSelector`, etc. [documented here](https://woodpecker-ci.org/docs/next/administration/backends/kubernetes#resources).

The AppArmor profile is trickier since the container name needs to be included in the annotation key. I'm imagining a configuration like `backend_options.kubernetes.appArmorProfile` that could take all of the possible values for the annotation [documented here](https://kubernetes.io/docs/tutorials/security/apparmor/). Then, the pod generation logic would just look for this configuration and add the appropriate annotation if it's found.

### Alternative

I could disable AppArmor entirely, but that's not really advisable for a production environment. I could also try to find a way to get the buildah container running as root (it should be possible), but again, that's not ideal from a security perspective.

### Additional context

This is somewhat of a follow up from #2510. I'm trying to get Woodpecker to run on my kubeadm/CRI-O based cluster running on Debian which I believe makes me some kind of leprechaun/unicorn/lochness monster. There definitely aren't many of us out there.

### Validations

[...]

PR #2550, opened 2023-10-08
Simple security context options (Kubernetes)

zc-devs:
Part of #2545

Adds security context options:` privileged`, `runAsUser`, `runAsGroup`, `runAsNonRoot`, `readOnlyRootFilesystem`, `allowPrivilegeEscalation`.

---

Pipeline:
```
steps:
  test:
    image: alpine
    commands:
      - echo Hello from $(whoami)
    backend_options:
      kubernetes:
        securityContext:
          runAsUser: 999
```

Log:
```
{"level":"trace","Security context":{"runAsUser":999},"time":"2023-10-08T15:14:39Z","caller":"/src/pipeline/backend/kubernetes/pod.go:200","message":"Security context that will be used for containers"}
```

Pod:
```
spec:
  containers:
    - name: wp-01hc7qqcf5q5gqd86yjytb37q3-0-step-0
      image: alpine
      securityContext:
        privileged: false
        runAsUser: 999
```

Step output:
```
+ echo Hello from $(whoami)
Hello from
whoami: unknown uid 999
```


zc-devs:
@dominic-p, could you test it and maybe write some docs?

dominic-p:
Thanks for working on this! Yes, I'd be happy to test and write up some docs. Two questions:

1. Could you point me in the right direction for testing? I'm not a go developer, so I'm not sure what the right approach would be. I'm currently running Woodpecker via the Helm chart. If there's an image published for this PR, I should be able to reference it, but it looks like the CI hasn't run yet.

2. I didn't see `fsGroup` listed in the `securityContext` struct definitions. I believe that will be needed in addition to `runAsGroup` to make sure the mounted volume gets appropriate permissions.

dominic-p:
@qwerty287 thanks for doing that. I was able to test the PR. Here's what I'm seeing:

1. The `securityContext` is being set at the container level (i.e. `spec.containers[0].securityContext`). That may be useful for some circumstances, but I believe a more common approach (and more useful in general) is to set the `securityContext` at the Pod level (i.e. `spec.securityContext`). Maybe both could be supported? The YAML could look like this:

```yml
    backend_options:
      kubernetes:
        securityContext:
          runAsNonRoot: true
          runAsUser: 1000
          runAsGroup: 1000
          fsGroup: 1000
        containerSecurityContext:
          runAsUser: 1000
```

That said, my vote would be to just support setting the Pod level `securityContext` since these Pods will typically (always?) have just one container.

2. We're still missing the `fsGroup` setting which makes sense because I don't believe that's supported at the container level. But, if we switch to the Pod level `fsGroup` should also be added as a supported option.

pat-s:
I wonder how much we want to bake into the `backend_options` or actually port these options to the helm chart?
Most of the options in these PR are usually set for all pods of an instance and shouldn't be necessary to define on the step level?

zc-devs:
@dominic-p, you're right [1](https://github.com/kubernetes/api/blob/9a776fe3a720323e4f706b3ebb462b3dd661634f/core/v1/types.go#L6850), [2](https://github.com/kubernetes/api/blob/9a776fe3a720323e4f706b3ebb462b3dd661634f/core/v1/types.go#L3865). I did it yesterday, but faced issue with stucking pipeline. I'm figuring it out now.
 
I made one security context:
```
type SecurityContext struct {
	// Pod Security Context
	RunAsUser          *int64  `yaml:"runAsUser,omitempty"`
	RunAsGroup         *int64  `yaml:"runAsGroup,omitempty"`
	RunAsNonRoot       *bool   `yaml:"runAsNonRoot,omitempty"`
	SupplementalGroups []int64 `yaml:"supplementalGroups,omitempty"`
	FSGroup            *int64  `yaml:"fsGroup,omitempty"`
	// Container Security Context
	Privileged               *bool `yaml:"privileged,omitempty"`
	ReadOnlyRootFilesystem   *bool `yaml:"readOnlyRootFilesystem,omitempty"`
	AllowPrivilegeEscalation *bool `yaml:"allowPrivilegeEscalation,omitempty"`
}
```

@pat-s, why should it be ported to the helm chart?

> Most of the options are usually set for all pods of an instance and shouldn't be necessary to define on the step level?

Do you mean, I should add `WOODPECKER_BACKEND_K8S_POD_SECURITY_*` options? I agree. Then step level options could be treated as exclusion: for example, we set `runAsNonRoot=true` instance-wide, but for some step/pod/container we make exclusion in step config.

> I wonder how much we want to bake into the backend_options?

¯\_(ツ)_/¯ Probably, all Kube spec 🤣 Seriously, at least I plan to add [seccomp](https://kubernetes.io/docs/tutorials/security/seccomp/) and [AppArmor](https://kubernetes.io/docs/tutorials/security/apparmor/). @dominic-p, what of simple SC options (above) do you need and what we can get rid of (except `privileged`)?

zc-devs:
Added `supplementalGroups` and `fsGroup`.

Pipeline:
```
steps:
  test:
    image: alpine
    commands:
      - echo Hello from $(whoami)
    backend_options:
      kubernetes:
        securityContext:
          runAsUser: 999
          readOnlyRootFilesystem: true
```

Pod:
```
kind: Pod
spec:
  containers:
    - name: wp-01hca2t4zh0r9xsqpbze5ryqww-0-step-0
      image: alpine
      securityContext:
        privileged: false
        readOnlyRootFilesystem: true
  securityContext:
    runAsUser: 999
```

dominic-p:
@zc-devs for my use case I just need `runAsNonRoot`, `runAsUser`, `runAsGroup`, and `fsGroup` (although `runAsUser` may imply `runAsNonRoot` I'm not sure on that).

I would hesitate to combine Pod and container level config in the same object just because as a user I wouldn't be sure which one I was setting. That said, it's probably mostly a matter of style/preference in a single container Pod.

As for global options in the helm chart, I agree with @zc-devs. If I have multiple steps running in different containers, it's very unlikely that they are all going to need to run as the same UID/GID, for instance. I would still need to configure/override some of the settings at the `step` level.

zc-devs:
* Deleted unnecessary options
* Added pipeline schema
* Made `runAsNonRoot` agent-wide
* Don't set `securityContext` if all fields aren't set

Test with `WOODPECKER_BACKEND_K8S_SECCTX_NONROOT: 'true'`

1. Pipeline:
```
skip_clone: true
steps:
  test:
    image: alpine
    commands:
      - echo Hello from $(whoami)
```
Error:
```
Error: container has runAsNonRoot and image will run as root (container: wp-01hcd9bzzxydxr5fes0fhq97rz-0-step-0)
```

2. Pipeline:
```
skip_clone: true
steps:
  test:
    image: alpine
    commands:
      - echo Hello from $(whoami)
    backend_options:
      kubernetes:
        securityContext:
          runAsUser: 999
          runAsGroup: 999
```
Pod:
```
kind: Pod
spec:
  containers:
    - name: wp-01hcd7tjy79ctkkx3g3hg8hae2-0-step-0
      image: alpine
  securityContext:
    runAsUser: 999
    runAsGroup: 999
    runAsNonRoot: true
```

3. Pipeline:
```
skip_clone: true
steps:
  test:
    image: alpine
    commands:
      - echo Hello from $(whoami)
    backend_options:
      kubernetes:
        securityContext:
          runAsUser: 999
          runAsGroup: 999
          privileged: true
```
Pod:
```
kind: Pod
spec:
  containers:
    - name: wp-01hcd83q7be5ymh89k5accn3k6-0-step-0
      image: alpine
      securityContext:
        privileged: true
  securityContext:
    runAsUser: 999
    runAsGroup: 999
    runAsNonRoot: true
```

zc-devs:
> I would hesitate to combine Pod and container level config in the same object just because as a user I wouldn't be sure which one I was setting. That said, it's probably mostly a matter of style/preference in a single container Pod.

It's abstraction here which doesn't have to match 1:1 with Kubernetes. Also as a step is a pod with only one container now, there is no reason to make things more complex.

> have these options set on the instance level

>  If I have multiple steps running in different containers, it's very unlikely that they are all going to need to run as the same UID/GID

Agree, it doesn't make sense to have _user_ and _group_ options agent-wide. Consider this, I added only `runAsNonRoot` in the agent configuration.

>  admin might want to control if users can override security specific settings on the step level

Then [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission) will help, I think.

dominic-p:
@zc-devs that all makes sense to me. If we run the CI again, I can smoke test the revisions on my cluster.

Once everyone's settled on the semantics for the config I can submit a PR with some docs.

zc-devs:
@pat-s, could you add `build_pr_images` label here?

pat-s:
@zc-devs Your PR images are finally ready!

zc-devs:
Thanks.
@dominic-p, could you test it (`woodpeckerci/woodpecker-server:pull_2735`, `woodpeckerci/woodpecker-agent:pull_2735`)?

dominic-p:
Sorry for the radio silence here. I've been stuck working on a pretty consuming project the last couple weeks. That's wrapping up now, so I should be able to test this in the next day or two.

dominic-p:
Ok, I was finally able to test this tonight. It looks good to me.

I'm still not able to run my particular pipeline as we're still missing AppArmor functionality, but the security context setting seem to be working as expected.

Thanks so much for your work on this. Did you want me to take a stab at some documentation on this for the website?

zc-devs:
> Did you want me to take a stab at some documentation on this for the website?

Yeah, that would be great!

zc-devs:
@qwerty287, could you review again?

PR #2554, opened 2023-10-09
Refactor/simplify pubsub

qwerty287:
- remove interface, we can use the `struct` directly
- remove support for different topics (we don't use it)
- remove unused methods and returns
- remove specific LICENSE file (is this fine? I changed so many lines that there shouldn't be that much from the original library)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/go-ap/httpsig v0.0.0-20221203064646-3647b4d88fdf
 	github.com/go-sql-driver/mysql v1.7.1
 	github.com/golang-jwt/jwt/v5 v5.0.0
-	github.com/google/go-github/v55 v55.0.0
+	github.com/google/go-github/v56 v56.0.0
 	github.com/google/tink/go v1.7.0
 	github.com/google/uuid v1.3.1
 	github.com/gorilla/securecookie v1.1.1
@@ -67,12 +67,10 @@ require (
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
 	github.com/KyleBanks/depth v1.2.1 // indirect
 	github.com/Microsoft/go-winio v0.6.1 // indirect
-	github.com/ProtonMail/go-crypto v0.0.0-20230217124315-7d5c6f04bbb8 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/bytedance/sonic v1.9.1 // indirect
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/chenzhuoyu/base64x v0.0.0-20221115062448-fe3a3abad311 // indirect
-	github.com/cloudflare/circl v1.3.3 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/davidmz/go-pageant v1.0.2 // indirect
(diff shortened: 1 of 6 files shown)


server/pubsub/pub.go
```diff
@@ -19,71 +19,53 @@ import (
...
 
-type publisher struct {
+// Receiver receives published messages.
+type Receiver func(Message)
+
+type Publisher struct {
 	sync.Mutex
 
-	topics map[string]*topic
+	done chan struct{}
```
anbraten, line 40: What is this channel used for?

server/pubsub/pub.go
```diff
@@ -19,71 +19,53 @@ import (
...
+	subs map[*Receiver]struct{}
```
anbraten, line 41: Is using an empty struct a common way for this type of map? I saw it done with a `bool` a few times.

anbraten (commented):
In general a nice cleanup. Could be that we need to add the error returns & context back if we want to use redis etc for HA

anbraten (approved):

anbraten:
> Just would like to make this sure again: Is it fine that I removed the old `LICENSE` file?

To be safe I would just leave the file for now.

server/pubsub/pub.go
```diff
@@ -19,71 +19,48 @@ import (
...
 
-type publisher struct {
+// Receiver receives published messages.
+type Receiver func(Message)
+
+type Publisher struct {
 	sync.Mutex
 
-	topics map[string]*topic
+	subs map[*Receiver]bool
```
lafriks, line 40: Why does it needs to be bool? I don't see where false value would be used
If map value is not used than `struct{}` is better option

server/pubsub/pub.go
```diff
@@ -19,71 +19,48 @@ import (
...
-	}
-	t.publish(message)
-	return nil
 }
 
-func (p *publisher) Subscribe(c context.Context, dest string, receiver Receiver) error {
+func (p *Publisher) Subscribe(c context.Context, receiver Receiver) {
 	p.Lock()
-	t, ok := p.topics[dest]
+	p.subs[&receiver] = true
```
lafriks, line 60:
```suggestion
	p.subs[&receiver] = struct{}{}
```

server/pubsub/pub.go
```diff
@@ -19,71 +19,48 @@ import (
...
+	subs map[*Receiver]bool
 }
 
 // New creates an in-memory publisher.
-func New() Publisher {
-	return &publisher{
-		topics: make(map[string]*topic),
+func New() *Publisher {
+	return &Publisher{
+		subs: make(map[*Receiver]bool),
```
lafriks, line 46:
```suggestion
		subs: make(map[*Receiver]struct{}),
```

server/pubsub/pub.go
lafriks: struct is faster and uses less memory than bool: https://itnext.io/set-in-go-map-bool-and-map-struct-performance-comparison-5315b4b107b

lafriks: is still valid also in 1.21:
```
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
                           │  bench.txt  │
                           │   sec/op    │
SetWithBoolValueWrite        3.154 ± 10%
SetWithStructValueWrite      3.075 ± 10%
SetWithInterfaceValueWrite   5.760 ±  5%

                           │  bench.txt   │
                           │     B/op     │
SetWithBoolValueWrite        811.4Mi ± 0%
SetWithStructValueWrite      737.6Mi ± 0%
SetWithInterfaceValueWrite   1.800Gi ± 0%

                           │  bench.txt  │
                           │  allocs/op  │
SetWithBoolValueWrite        380.2k ± 0%
SetWithStructValueWrite      379.8k ± 0%
SetWithInterfaceValueWrite   379.6k ± 0%
```

anbraten: Thanks for sharing ❤️

6543 (changes requested):

server/pubsub/pubsub.go
```diff
@@ -1,56 +0,0 @@
...
-
-	// Data is the actual data in the entry.
-	Data []byte `json:"data"`
-
-	// Labels represents the key-value pairs the entry is labeled with.
-	Labels map[string]string `json:"labels,omitempty"`
-}
-
-// Receiver receives published messages.
-type Receiver func(Message)
```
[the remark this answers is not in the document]
6543: https://docs.nats.io/nats-concepts/core-nats/pubsub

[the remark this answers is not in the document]
anbraten: Just add the correct interface and functions when you add the alternative pubsub service. It will probably also need a context and should return an error, but its way easier to do that when we are actually adding it. We don't need to guess about that code now!

PR #2555, opened 2023-10-09
Move skip-ci back into central place

anbraten:
The skip logic should be in a central place and not in the compiler as it also needs to work independent of the compiler.

reverts #2216

solves #2313

### DIFF
--- a/pipeline/frontend/yaml/constraint/constraint.go
+++ b/pipeline/frontend/yaml/constraint/constraint.go
@@ -19,19 +19,15 @@ import (
 	"fmt"
 	"maps"
 	"path"
-	"regexp"
 	"strings"
 
 	"github.com/antonmedv/expr"
 	"github.com/bmatcuk/doublestar/v4"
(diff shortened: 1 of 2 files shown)


anbraten:
> How does this solve #2313?

Yeah not directly, but it allows to use `[skip ci]` again without even having a config file present.

> In the past, I was able to work around this issue by adding [skip ci] to all commits to branches without a CI config. This is now also broken, see https://github.com/woodpecker-ci/woodpecker/pull/2216.

xoxys:
Still not happy with this approach and a bit strange this got merged within hours while others were blocked form months... However at leadt one way back to skip ci...

6543:
I considder this a hotfix!!!

6543:
a.k.a. workaround!

anbraten:
> I considder this a hotfix!!!

It should stay in the current place. The code is independent of the config-compiler and should work with every compiler we implement in the future.

PR #2559, opened 2023-10-09
Bump to mysql 8

### LINKED ISSUE #2558: Bump mysql version to 8 in tests
### Discussed in https://github.com/woodpecker-ci/woodpecker/discussions/2392

<div type='discussions-op-text'>

<sup>Originally posted by **pat-s** September  8, 2023</sup>
Because of https://github.com/woodpecker-ci/woodpecker/pull/2232.

There will most likely never be a arm64 image for mysql 5.x and this DB type/version might anyhow not be the most recommended one anyhow?

</div>

pat-s:
fix #2558


.woodpecker/test.yml
```diff
@@ -136,7 +136,7 @@ services:
     when: *when
 
   service-mysql:
-    image: mysql:5.6.27
```
anbraten, line 139: Should we maybe even switch it to `mariadb`?

pat-s: What would be the advantage? They have arm builds for older versions? 

I am not an expert in mysql/mariadb but given that MySQL 5 is EOL quite soon...

anbraten: Actually it doesn't matter too much. Some users / distros etc seem to move to mariadb as of license / philosophy reasons.

PR #2562, opened 2023-10-10
Fix fetch request content type

anbraten:


anbraten:
> Should this also be applied if the request has no body?

In that case its not needed to set the content type.

PR #2571, opened 2023-10-11
Add plugin "NixOS Remote Builder" to plugin index

JohnWalkerx:
I created a small Woodpecker plugin to build nix flake paths on a remote NixOS system.
So this adds it to the plugin index because this plugin could be interesting for other users.


JohnWalkerx:
> 
> Deployment of preview was successful: https://woodpecker-ci-woodpecker-pr-2571.surge.sh

@qwerty287 Is this a bug in the preview that I can't see the added plugin in the generated site?

PR #2576, opened 2023-10-13
Run tests on `main` branch

qwerty287:
We currently don't have coverage information on codecov, the tests must run for `main` pushes so we can publish them afterwards


.woodpecker/test.yml
```diff
@@ -64,9 +65,12 @@ steps:
   lint-editorconfig:
     image: mstruebing/editorconfig-checker:2.7.2
     group: test
+    when:
+      - event: [pull_request, tag, deployment]
```
anbraten, line 69: I think we can remove `deployment` from this file completely as we don't use it.

anbraten (approved):

PR #2577, opened 2023-10-13
Update to docker 23

### LINKED ISSUE #1807: Bump docker dep to v23
e.g. revert #1792

pat-s:
fix #1807


pat-s:
Yeah but not all distros are on 24 yet -> compatibility. If you remember we excluded Debian anyhow but Suse only upgraded to 23 for now.

pat-s:
I don't fully know. The last time it was caused by an API incompatibility as the API changed between 1.41 (docker 20) and 1.42 (docker 23). Docker 24 has 1.43 and it might be the same issue.

It seems older (go) libs in WP can handle newer host docker versions but not the other way round. And as we don't want to build multiple images...

PR #2581, opened 2023-10-14
Add plugin "Gitea Create Pull Request" to plugin index

JohnWalkerx:
I wrote a plugin to create a new pull request on a Gitea/Forgejo instance.
So I wanna add it to the plugin index because I think this is very useful for automated workflows.

### DIFF
--- a/docs/plugins/woodpecker-plugins/plugins.json
+++ b/docs/plugins/woodpecker-plugins/plugins.json
@@ -95,6 +95,11 @@
       "verified": true
     },
     {
+      "name": "Gitea Create Pull Request",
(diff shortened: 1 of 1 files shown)


pat-s:
@JohnWalkerx How about moving it into the woodpecker org on codeberg while maintaining permissions? Might be easier for users to discover it and report something etc...

JohnWalkerx:
> @JohnWalkerx How about moving it into the woodpecker org on codeberg while maintaining permissions? Might be easier for users to discover it and report something etc...

Yeah, why not.
I think you are talking about the https://codeberg.org/woodpecker-plugins org?

PR #2585, opened 2023-10-14
Fix netrc injection -> allow injection into untrusted containers

### LINKED ISSUE #2583: Injecting netrc credentials into non-trusted containers does not work
### Component

agent

### Describe the bug

By default netrc credentials (i.e. git credentials) are not injected into builds unless

- the repo is set to trusted (which can only be done by admins)
- "Only inject netrc credentials into trusted containers" is unchecked

However, the latter does not work as executing a "git push" back to the checked out repo does not work when **the option is unchecked**.
It only works if the repo is set to "trusted" - which cannot be enabled by a normal user.

Showcasing this in an example repo is hard as normal users cannot open the settings of a repo.
I've tested this multiple times in an example repo toggling different options on and off and running a simple `git push` at the end (after a dummy modification).

Also I think the current docs are wrong

> Cloning pipeline step may need git credentials. They are injected via netrc. By default, they're only injected if this option is enabled, the repo is trusted ([see above](#trusted)) or the image is a trusted clone image. If you uncheck the option, git credentials will be injected into any container in clone step.

I think it should read as:

"By default, they're only injected if this option is enabled **and** the repo is trusted - or if the "inject" option is unchecked."

Also "git credentials will be injected into any container in clone step" -> should probably be "into any container **in addition** to the 'clone' step". 

### System Info

```shell
`next-8629a418f8`
```

### Additional context

_No response_

### Validations

pat-s:
fix #2583 

And improve settings documentation.


pat-s:
Tested that the fixes work - credentials are now injected as expected. The change in line 256 was the important one.

anbraten:
I am fine with fixing it first

anbraten:
However AFAIK this is a feature not a fix, right? It was on purpose that only the clone image would receive netrc credentials when used in the steps.

pipeline/frontend/yaml/compiler/compiler.go
```diff
@@ -187,7 +187,7 @@ func (c *Compiler) Compile(conf *yaml_types.Workflow) (*backend_types.Config, er
 			step := c.createProcess(name, container, backend_types.StepTypeClone)
 
 			// only inject netrc if it's a trusted repo or a trusted plugin
-			if !c.netrcOnlyTrusted || c.trustedPipeline || (container.IsPlugin() && container.IsTrustedCloneImage()) {
+			if !c.netrcOnlyTrusted || c.trustedPipeline || (c.netrcOnlyTrusted && container.IsPlugin() && container.IsTrustedCloneImage()) {
```
anbraten, line 190: This change is not needed as it should be true:
- if netrc passing is allowed for not trusted images
- if repo is trusted
- if its a plugin and using a trusted clone image (regardless of the netrc option)

pipeline/frontend/yaml/compiler/compiler.go
```diff
@@ -253,7 +253,7 @@ func (c *Compiler) Compile(conf *yaml_types.Workflow) (*backend_types.Config, er
 		step := c.createProcess(name, container, stepType)
 
 		// inject netrc if it's a trusted repo or a trusted clone-plugin
-		if c.trustedPipeline || (container.IsPlugin() && container.IsTrustedCloneImage()) {
+		if !c.netrcOnlyTrusted || c.trustedPipeline || (c.netrcOnlyTrusted && container.IsPlugin() && container.IsTrustedCloneImage()) {
```
anbraten, line 256: Thinking about it I would prefer following cases:
- if trusted repo (needs admin access)
- if plugin and trusted clone image (clone image should not steal /leak the credentials)

New case:
- if `netrcOnlyTrusted` is active only allow plugins as the admin could trust them, so we might need a list of `passNetrcToImages` instead

pat-s:
> However AFAIK this is a feature not a fix, right? I think it was on purpose that only the clone image would receive netrc credentials when used in the steps.

It sounds like nobody really knows 😅 To me, reading the option description, it sounds like this is a "bugfix" as I would have expected that unchecking this option injects netrc into **all** containers - as the option did not say that it's limited to the clone step.
Yes, the [docs](https://woodpecker-ci.org/docs/next/usage/project-settings#only-inject-netrc-credentials-into-trusted-containers) give a hint about that but that can be overlooked easily.

WRT to security: I fully understand the hesitation and why it should not be injected into all containers by default. Yet I wonder if there's really a different in limiting it to the clone step only or not - *if* there is malicious motivation, the step does not really matter. Or am I overlooking something? If this is the case, then the PR can be merged. 

> instance with multiple users (company, multiple friends, community like codeberg) => not everyone was made an admin somehow (probably they do not 100% trust everybody), so to prevent miss-configuration it could be best to still be pretty strict about options like passing netrc

This is my use case (at least in a specific env). Not everybody is admin (as this is not good per se and not everyone knows what's possible etc.). But everybody should have the option to decide for their own repos whether they want to inject netrc - at least that's my view.

I see the point when an admin wants to prevent this entirely. So maybe an additional admin option is needed which allows/forbids changing the repo setting in the first place?

anbraten:
Although it was document in a bad way / the option implied sth else, no one was currently able to use it in steps therefore we should carefully thing how we provide this power to users as it would be breaking if we remove it later on again.

Regarding the security aspect I currently think it would be best to replace the current option with a comma separated list of images. As netrc credentials are somehow a secret as well this would match the logic we use for the secret image list. The credentials should therefore only be passed to plugin of that list. If empty they should not be passed to anything else then the official plugin.

pat-s:
I am not sure because AFAICS #2601 only tackles the plugin case but doesn't allow arbitrary images to use the netrc creds (which should also be an option for repo admins IMO). Besides the one to give netrc to plugins. But I didn't had time to dive deep into this topic in the last days.

I don't think we can finish this for 2.0, so let's at least move it out of this milestone.

xoxys:
@woodpecker-ci/maintainers Can we fix conflicts and merge this fix as is? Right now, the option `Only inject netrc credentials into trusted containers` is just broken, as unchecking it doesn't work.

If we decided to drop this feature completely, fine for me, but right now the checkbox exist in the latest version of wp and as long as it exists it should at least work. I cannot believe that we have not fixed this issue since more than a year just because we might or might not want to drop it (which also doesn't happen till now).

xoxys:
@6543 this is definitely not an enhancement. This checkbox exists today, and unchecking the checkbox has no effect. That's what I would call a bug.

nemunaire:
We agreed with that: the default behavior should be to not share credentials with containers except clone. The fact there is a dedicated option is to let the user choose if it requires a *less secure* alternative in all conscience.
For example, this can [help to `git push`](https://github.com/woodpecker-ci/woodpecker/issues/2583).

Although I migrated to WP a long time ago, I'm still required to run Drone for repositories making `git push`...

pat-s:
While I understand all discussion around this feature in general, as @xoxys said: the checkbox exists currently and is broken. This PR only fixes that.

Everything else (whether to restrict the option to admins-only or other ideas) should be done in a separate discussion and PR then IMO.

I am gonna resolve conflicts and then leave it to somebody else to possibly approve or deny this.

anbraten:
@pat-s It actually works fine (just checked the code and function again). You can use a [custom clone plugin](https://woodpecker-ci.org/docs/usage/workflow-syntax#clone) and set this option to still allow the netrc injection. It was not intended to be used for normal steps as this PR is doing it. Therefore this PR is rather a new feature and should probably start migrating to sth like #2601. We could also change the description of the current option to sth like `Only inject netrc credentials into trusted clone step` for the beginning.

pat-s:
I understand (now) better. Yes, I think updating the wording here is very important to avoid these misconceptions.

I think also the description should be much more verbose here to explain the setting. Currently it is just repeating the title of the setting, which doesn't help at all.

pat-s:
Closing as due to the latest comment by @anbraten.

xoxys:
Well if that is really the intended way then the description is completely misleading. So what was the intention to just write such a description then back in the days where this setting was introduced?

PR #2590, opened 2023-10-15
Refactor UI dark/bright mode

qwerty287:
- collapse pipeline configs by default, closes https://github.com/woodpecker-ci/woodpecker/discussions/2557
- refactor ui themes: instead of just differentiating between "is dark" or "is not dark", add a third "auto" option following the browser settings (closes https://github.com/woodpecker-ci/woodpecker/discussions/2204) and put everything into an enum. also move the option from the navbar to user settings.

### DIFF
--- a/web/src/compositions/useTheme.ts
+++ b/web/src/compositions/useTheme.ts
@@ -25,7 +25,7 @@ updateTheme();
 
 export function useTheme() {
   return {
-    theme,
+    theme: resolvedTheme,
     storeTheme,
   };
 }


web/src/components/layout/header/Navbar.vue
```diff
@@ -24,13 +24,6 @@
     </div>
     <!-- Right Icons Box -->
     <div class="flex ml-auto -m-1.5 items-center space-x-2">
-      <!-- Dark Mode Toggle -->
-      <IconButton
-        :icon="darkMode ? 'dark' : 'light'"
-        :title="$t(darkMode ? 'color_scheme_dark' : 'color_scheme_light')"
-        class="navbar-icon"
-        @click="darkMode = !darkMode"
-      />
```
anbraten, line 33: What do you think about adding an `A` for automatic mode? Its sometimes pretty nice to quickly switch the mode especially as chromium is still having this bug that it does not reflect the OS setting 🙈

web/src/compositions/useTheme.ts
```diff
@@ -0,0 +1,61 @@
+import { computed, ref, watch } from 'vue';
+
+export enum Theme {
+  Auto = 'auto',
+  Light = 'light',
+  Dark = 'dark',
+}
+
+const LS_THEME = 'woodpecker:theme';
+const activeTheme = ref(Theme.Auto);
```
anbraten, line 10: Vue-use now has such a thing as well: https://vueuse.org/core/useColorMode/

web/src/compositions/useTheme.ts
```diff
@@ -30,30 +18,20 @@ watch(activeTheme, (theme) => {
...
-    setTheme(Theme.Auto);
-  } else {
-    setTheme(isActive);
-  }
+function setTheme(theme: BasicColorSchema) {
+  store.value = theme;
 }
 
-load();
+setTheme(store.value);
```
anbraten, line 25: Should not be needed

web/src/compositions/useTheme.ts
```diff
@@ -1,23 +1,11 @@
...
-const activeTheme = ref(Theme.Auto);
-
-function resolveAuto(theme: Theme) {
-  if (theme === Theme.Auto) {
-    return window.matchMedia('(prefers-color-scheme: dark)').matches ? Theme.Dark : Theme.Light;
-  }
-  return theme;
-}
+const { system, store } = useColorMode();
+const resolvedTheme = computed(() => (store.value === 'auto' ? system.value : store.value));
```
anbraten, line 5: Seems not to be needed from the code (docs seem to be misleading a bit): https://github.com/vueuse/vueuse/blob/main/packages/core/useColorMode/index.ts#L142

web/src/compositions/useFavicon.ts
```diff
@@ -1,10 +1,10 @@
 import { computed, ref, watch } from 'vue';
 
 import useConfig from '~/compositions/useConfig';
-import { useDarkMode } from '~/compositions/useDarkMode';
+import { useTheme } from '~/compositions/useTheme';
 import { PipelineStatus } from '~/lib/api/types';
 
-const darkMode = computed(() => (useDarkMode().darkMode.value ? 'dark' : 'light'));
+const darkMode = computed(() => useTheme().resolvedTheme.value);
```
anbraten, line 7:
```suggestion
const { resolvedTheme } = useTheme();
const darkMode = computed(() => resolvedTheme.value);
```

web/src/compositions/useTheme.ts
```diff
@@ -0,0 +1,38 @@
...
+  return {
+    resolvedTheme,
+    theme: computed({
+      get() {
+        return storeTheme.value;
+      },
+      set(theme: BasicColorSchema) {
+        storeTheme.value = theme;
+      },
+    }),
```
anbraten, line 36: Might be a good idea to switch both as you probably normally only need the `resolvedTheme`:

- `resolvedTheme` => `theme`
- `theme` => `storedTheme`

```suggestion
    resolvedTheme,
    theme: storeTheme,
```

web/src/components/user/UserGeneralTab.vue
```diff
@@ -3,6 +3,16 @@
     <InputField :label="$t('user.settings.general.language')">
       <SelectField v-model="selectedLocale" :options="localeOptions" />
     </InputField>
+    <InputField :label="$t('user.settings.general.theme.theme')">
```
anbraten, line 6: Just as I notice it here again, should we start to stop deeply nesting translations to have it easier to re-use them. So in this case simply `theme.theme`

anbraten (approved):

anbraten (approved):

Issue #336, opened 2021-09-21
Extend Gated Feature with an Allow/Block List


6543:
extend feature added (back) in #51

---
ref: https://codeberg.org/Codeberg/Community/issues/428#issuecomment-195258
ref: #193

xoxys:
Do you already have some ideas in mind on how this should be implemented? I would like to work on this, as at least for me, it gets annoying very fast :see_no_evil: 

A repo setting to configure a list of authors that can be checked in https://github.com/woodpecker-ci/woodpecker/blob/main/server/pipeline/gated.go#L20

6543:
well there are ideas like:
 - allow listing: if a user has X accepted pipelines it gets accepted automatically
 - allow listing: if the pipeline config already did run allow it automatically
 - deny listing: if pipeline target branch X make it gated ...

so basically there are rules and you could have a allow- or a deny- block-list

xoxys:
Would it make sense to integrate [open policy agent](https://github.com/open-policy-agent/opa) for stuff like this? We could allow admins to write the own policies. We could also make use of it in e.g. extensions as proposed in https://github.com/woodpecker-ci/woodpecker/pull/915

anbraten:
[opa](https://github.com/open-policy-agent/opa) is a great tool, but for WP it feels to me it would add to much complexity.

PR #2597, opened 2023-10-16
Update renovate schedule & use central config repo

pat-s:
To reduce the `renovate` noise and relax their influence on PRs a bit, I suggest the following updates:

- Run renovate only on weekends (this should create the PRs initially)
- Automerge daily (once automerge is in place -> currently testing in helm repo). This should ideally automerge the PRs from the weekend early morning on each day. Assuming there are more than one opened on the weekend, only one can be merged each day due to required branch rebasing. Ofc automerge will only happen if tests pass.
- Use a central config repo for easier maintenance of a shared config

Central config repo: https://github.com/woodpecker-ci/renovate-config/blob/main/default.json

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,11 +1,6 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "config:recommended",
-    ":maintainLockFilesWeekly",
-    "schedule:automergeDaily",
-    "schedule:weekends"
-  ],
+  "extends": ["github>woodpecker-ci/renovate-config"],
   "prConcurrentLimit": 5,
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


pat-s:
The plan is to use the `renovate-approve` bot, which is already in place in the helm repo. https://github.com/apps/renovate-approve

pat-s:
That's still an issue yes and addressed in https://github.com/woodpecker-ci/woodpecker/issues/336.

Until then it might only work for plugin repos for which no workflow approval is in place.

pat-s:
But also besides the automerge idea, I think having renovate PR popup once per week is also "enough"? The good thing about that is that updates of the same manager are grouped usually and we don't need to deal with so many single PRs every few days.

xoxys (approved):
Don't know if it makes sense to enable it on this repo under these circumstances.  LGTM otherwise.

pat-s:
Maybe it's best we add a preset config repo which hosts some presets for all repos and extend the individual configs in the repos.

Then things like `automergeDaily` would only appear there and not be explicitly present in a repo config. And we don't need so many updates in all repos.

xoxys:
Yes, if a broader use is planned a preset repo is a better approach :+1:

PR #2232, opened 2023-08-17
make it possible to run pipeline on arm64

6543:
test pipeline to run on arm64

```
Architecture:           aarch64
  CPU op-mode(s):       32-bit, 64-bit
  Byte Order:           Little Endian
CPU(s):                 8
  On-line CPU(s) list:  0-7
Vendor ID:              ARM
  Model name:           Neoverse-N1
    Model:              1
    Thread(s) per core: 1
    Core(s) per socket: 8
    Socket(s):          1
    Stepping:           r3p1
    BogoMIPS:           50.00
    Flags:              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
```

## TODOs

- [x] web: [works](https://ci.woodpecker-ci.org/repos/3780/pipeline/7363/46)
- [ ] test
  - [x] lint-editorconfig: ~~[fail](https://ci.woodpecker-ci.org/repos/3780/pipeline/7361/27)~~ -> works after image was updated on dockerhub
  - [x] codecov: fail ... will be removed
  - [ ] mysql: [failed](https://ci.woodpecker-ci.org/repos/3780/pipeline/7365/42) -> service-mysql does not have arm64 image
- [x] securityscan:  [works](https://ci.woodpecker-ci.org/repos/3780/pipeline/7363/46)
- [x] docs: [works](https://ci.woodpecker-ci.org/repos/3780/pipeline/7695/28)
- [ ] docker
  - [ ] cross-compile-server: [fail](https://ci.woodpecker-ci.org/repos/3780/pipeline/7366/20) -> https://github.com/techknowlogick/xgo/issues/213
  - [x] build-agent/cli: [fail](https://ci.woodpecker-ci.org/repos/3780/pipeline/7366/11) -> unrelated -> fixed by #2240
- [ ] binaries
  - [ ] cross-compile-server: [fail](https://ci.woodpecker-ci.org/repos/3780/pipeline/7366/10) -> https://github.com/techknowlogick/xgo/issues/213

PR #2605, opened 2023-10-18
Allow running tests on arm64 runners

pat-s:
builds upon https://github.com/woodpecker-ci/woodpecker/pull/2232

NB: the change `pats22/xgo:go-1.21.3` is only as long as https://github.com/techknowlogick/xgo/pull/224 is not yet merged and an arm image is available upstream.


6543:
Nice ... i still want to wait until xgo has the patches merged :)

pat-s:
Sure. Let's see how upstream goes. But a PoC was needed anyhow :)

pat-s:
@6543 even though CI is green, compile-server-preview fails but the failure is not reflected by the ci

.woodpecker/binaries.yml
```diff
@@ -72,7 +72,6 @@ steps:
     commands:
       - ls -la dist/*.*
       - cat dist/checksums.txt
```
6543, line 74:
```suggestion
      - cat dist/checksums.txt

```

pat-s:
@6543 Promising new image, all tests worked so far. Test failure in `binaries` unrelated.

pat-s:
@6543 If you want to test with an arm runner again, it might be worth it. Unclear how long it will take until tech really has the time to test this upstream.

6543:
ok - let me deploy the autoscaler and then update this pull to let it run.

might take some time

6543:
xgo works :tada:  - nfpm fails :D

pat-s:
> nfpm fails :D

That issue looks unrelated AFAICS - it's not a compile issue but a missing file.

The issue is `dist/server/linux/amd64/woodpecker-server"` vs `dist/server/linux_amd64/woodpecker-server"` in the output locations.

I couldn't find a run where `build-deb-rpm` succeeded lately, so it might be a general issue unrelated to this PR.

pat-s:
Given that the `xgo` compiler seems working now, I think we should give it a go, i.e. deploy arm runners. There's a lot to gain/save and doing so might also help tech adding upstream support. What do you think?

pat-s:
Updated images to 1.22.1. 

Feedback would be great ;)

6543:
it did run on x86 right now ... let me change the autoscaler to spawn arm ... not jet but soon ish

6543:
what we could do is to explizite lable workflows to use x86 and then cange it to arm in here ...

pat-s:
> what we could do is to explizite lable workflows to use x86 and then cange it to arm in here ...

Certainly good for testing. But a bit of effort to do. Though I am meanwhile pretty sure the arm build fine.

pat-s:
@6543 A bit hard to make some progress here. I am slowly loosing motivation a bit as I can't do anything on my side anymore besides asking/pinging for external action. both here and in `xgo`.

If this is not important enough, we can also close it and just stay on `amd64` (?).

6543:
the only thing blocking it is https://github.com/techknowlogick/xgo/pull/224

so we can just make the autoscaler more inteligent and deploy arm and only if needed by workflow (label selector) use x64

pat-s:
> the only thing blocking it is https://github.com/techknowlogick/xgo/pull/224

is not blocking, there is a functional image built from this PR under my account `pats22` which is also referenced in this PR and tested to work in both ways.

6543:
well in this case we can just use that meanwhile ... just add a codecoment (todo: switch to main once X is merged...)

pat-s:
> well in this case we can just use that meanwhile ... just add a codecoment (todo: switch to main once X is merged...)

This is what I am trying to express since the last five comments or so 😅

xoxys (commented):
Some IDEs can create a tasks tree from `TODO:` commends.

xoxys (approved):

PR #2617, opened 2023-10-21
doc: improve prometheus docs

Janik-Haag:
Adding docs about how to read the bearer token from a file


docs/docs/30-administration/90-prometheus.md
```diff
@@ -30,6 +30,20 @@ scrape_configs:
        - targets: ['woodpecker.domain.com']
 ```
 
+You can also read the token from a file like so:
```
xoxys, line 33:
```suggestion
As an alternative, the token can also be read from a file:
```

docs/docs/30-administration/90-prometheus.md
```diff
@@ -1,6 +1,6 @@
 # Prometheus
 
-Woodpecker is compatible with Prometheus and exposes a `/metrics` endpoint. Please note that access to the metrics endpoint is restricted and requires an authorization token with administrative privileges.
+Woodpecker is compatible with Prometheus and exposes a `/metrics` endpoint if you set the `WOODPECKER_PROMETHEUS_AUTH_TOKEN` environment variable. Please note that access to the metrics endpoint is restricted and requires an authorization token with administrative privileges.
```
xoxys, line 3:
```suggestion
Woodpecker is compatible with Prometheus and exposes a `/metrics` endpoint if the environment variable `WOODPECKER_PROMETHEUS_AUTH_TOKEN` is set. Please note that access to the metrics endpoint is restricted and requires an authorization token with administrative privileges.
```

xoxys (changes requested):
Thanks! Only small adjustments in the wording, the rest looks good.

Janik-Haag:
Should be good now?

xoxys (approved):

PR #2626, opened 2023-10-23
Do not run docker prepare steps

qwerty287:
if nothing else will run

from #2625

### DIFF
--- a/.woodpecker/docker.yml
+++ b/.woodpecker/docker.yml
@@ -55,6 +55,16 @@ steps:
     pull: true
     commands:
       - go mod vendor
+    when:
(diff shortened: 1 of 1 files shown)


pat-s:
I wonder if it makes sense to add a label (in the UI) to steps which are skipped entirely, i.e. only have a `clone` step.

pat-s (approved):

anbraten:
> I wonder if it makes sense to add a label (in the UI) to steps which are skipped entirely, i.e. only have a clone step.

At some point we want workflows to not clone at all if the workflow contains no other step

PR #2635, opened 2023-10-23
Remove SSH backend

qwerty287:
The SSH backend is, similar to Gogs and Coding for forges, completely unmaintained and seems unused (it is likely broken but we didn't get any reports).
Instead, you should directly run the agent on the SSH machine with the `local` backend.


6543 (approved):
yes it has bugs as it was once similar to local backend witch had been fixed meanwhile ... and differged quite a bit

PR #2636, opened 2023-10-24
local backend ignore errors in commands inbetween

### LINKED ISSUE #2612: local backend dont work with "cmd" shell
if you have this config:
```
steps:
  # a demo step that say hello and show current work path
  hello:
    image: cmd.exe
    commands:
      - echo hello world
      - echo %CD%
```

for a **local** backend on **windows** you will get:
image

6543:
for normal posix shells we have to add the `-e ` option ... but as there are more shells out there we have to handle this edgecases on base per base case.

create a switch case statement in woodpecker should be fine as there is only a finite number of shells, used in production.

also close  #2612

---


pipeline/backend/common/script.go
```diff
@@ -37,10 +37,3 @@ func GenerateContainerConf(commands []string) (env map[string]string, entry, cmd
 
 	return env, entry, cmd
 }
-
-func GenerateScript(commands []string) string {
```
6543, line 41: not needed anymore ... we dont have an ssh backend now ;)

Issue #1764, opened 2023-05-17
Rogue log line when running `woodpecker-cli`


ambroisie:
### Component

cli

### Describe the bug

The latest update (v0.15.9) has updated the dependencies, and it looks like this might have changed a default value in the logging infrastructure.

When simply running `woodpecker-cli`, one can read the following message:

```
9:08PM ??? github.com/woodpecker-ci/woodpecker/cli/common/zerologger.go:20 > LogLevel = trace
```

### System Info

```shell
N/A, this is purely a CLI thing.
```

### Additional context

This was noticed when bumping the version in nixpkgs: NixOS/nixpkgs#232073.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

ambroisie:
The issue is still present in v0.15.10.

PR #2638, opened 2023-10-24
Do not print log level on CLI

qwerty287:
if it's not in debug log level

closes https://github.com/woodpecker-ci/woodpecker/issues/1764

### DIFF
--- a/cmd/common/logger.go
+++ b/cmd/common/logger.go
@@ -97,6 +97,6 @@ func SetupGlobalLogger(c *cli.Context, printLvl bool) {
 	}
 
 	if printLvl {
-		log.Debug().Msgf("LogLevel = %s", zerolog.GlobalLevel().String())
+		log.Info().Msgf("LogLevel = %s", zerolog.GlobalLevel().String())
 	}
 }


anbraten (approved):

xoxys:
This will also remove the line from the server and agent logs, where I find it quite helpful to always print the configured log level on startup. Can we just disable it for the cli?

cmd/cli/app.go
```diff
@@ -41,6 +41,7 @@ func newApp() *cli.App {
 	app.Usage = "command line utility"
 	app.EnableBashCompletion = true
 	app.Flags = common.GlobalFlags
+	app.Before = common.SetupGlobalLogger
```
xoxys, line 44: Does this work? `app.Before` accepts `(c *cli.Context, printLvl bool)` functions?

xoxys: I was just wondering because `app.Before` requires a type `BeforeFunc` and `common.SetupGlobalLogger` does not match the function type https://pkg.go.dev/github.com/urfave/cli#BeforeFunc

anbraten: its v2: https://pkg.go.dev/github.com/urfave/cli/v2#BeforeFunc

xoxys: Ah of course its v2 sorry for the wrong link. Still stragen :sweat_smile:

xoxys (approved):

xoxys:
@qwerty287 Looks like this is not working as expected. The log level is no longer shown on server startup.

PR #2639, opened 2023-10-24
Remove deprecated API paths

qwerty287:
and add some missing swagger docs


anbraten:
Removing /hook probably needs a migration for old webhooks to work

anbraten:
We could think about adding a repair button to the admin settings

xoxys:
Something like this? Or do you mean a button for each repo on this page?

image

6543 (approved):

PR #2642, opened 2023-10-24
Add "Repair all" button

qwerty287:
to the admin repo list

ref #2639

### DIFF
--- a/server/api/repo.go
+++ b/server/api/repo.go
@@ -401,70 +401,15 @@ func DeleteRepo(c *gin.Context) {
 //	@Summary	Repair a repository
 //	@Router		/repos/{repo_id}/repair [post]
 //	@Produce	plain
-//	@Success	200
+//	@Success	204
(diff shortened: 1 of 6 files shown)


anbraten (approved):

server/api/repo.go
```diff
@@ -575,3 +520,91 @@ func GetAllRepos(c *gin.Context) {
...
+		}
+	}
+
+	c.Status(http.StatusNoContent)
+}
+
+func repairRepo(c *gin.Context, repo *model.Repo, withPerms bool) {
+	forge := server.Config.Services.Forge
+	_store := store.FromContext(c)
+	user := session.User(c)
```
6543, line 554: Uh the woodpecker admin could be a normal user on the forge ... this case will break things.

For single repair we should alter the user as we do now (witch is used to sync repo stuff) but for itterating over all of them we should use the orig one ...

PR #2648, opened 2023-10-25
Add woodpecker-lint tool

qwerty287:


xoxys:
It's not a tool that was developed. The tool is https://pre-commit.com/ and you just need to place a [single yaml file](https://git.schmidl.dev/schtobia/woodpecker-lint/src/branch/master/.pre-commit-hooks.yaml) in a repo to get it working. See https://pre-commit.com/#creating-new-hooks

xoxys (approved):
If you still prefer to keep it external, works for me as well.

pat-s:
@xoxys I am also using precommit for many projects and was already thinking about proposing it's use here as well. I think it could be helpful in many cases.

PR #2649, opened 2023-10-25
Pin alpine versions in Dockerfile

pat-s:
And versionize via renovate. For more control about versions.


xoxys:
Not sure if pinning package versions in the Dockerfiles is really what we want to do. I can't remember a single case in the last years when I ever had the use case to use a specific package version in a container. Do you have one?

What we should do is pinning the `FROM` image digests of the containers, as this often indicates package updates for the default packages of a base image that can also include security updates of the os.

pat-s:
>  I can't remember a single case in the last years when I ever had the use case to use a specific package version in a container. Do you have one?

It was/is a lint rec/warning from `hadolint` (which is only used locally in pre-commit and not in CI).
I don't have a use case and it's probably a bit over-picky but I thought if renovate does the update job for us...

xoxys:
You still dont know it anyway as all preinstalled packages of the base image could have been updated by a digest change 🙂

xoxys (approved):

pat-s:
You know it at least for the ones pinned 😄 they are usually not of big importance...

But hey, it's a best practice and if the linter says so. But I still think it's better than digest pinning - if you think about it, you have libs and versions. With digests for rolling tags, it's way harder to infer the changes.

PR #2652, opened 2023-10-26
Add some tests

qwerty287:
and some cleanups


server/store/datastore/agent_test.go
```diff
@@ -29,28 +30,62 @@ func TestAgentFindByToken(t *testing.T) {
 		Name:  "test",
 		Token: "secret-token",
 	}
-	if err := store.AgentCreate(agent); err != nil {
-		t.Errorf("Unexpected error: insert agent: %s", err)
-		return
-	}
+	err := store.AgentCreate(agent)
+	assert.Nil(t, err)
```
6543, line 34: assert.NoError

server/store/datastore/server_config.go
```diff
@@ -43,7 +43,9 @@ func (s storage) ServerConfigSet(key, value string) error {
 		return err
 	}
 
-	_, err = s.engine.Where("key = ?", config.Key).AllCols().Update(config)
+	_, err = s.engine.Cols("value").Update(config, &model.ServerConfig{
```
6543, line 46: we should not change the coding style here ...

it'S first explite filter ... then opt cols to update ... then the changed object

6543: Yes in this case it **IS** an xorm bug

We can merge your sol. but witj a todo connent to an open upstream bug report

6543 (approved):

Issue #1513, opened 2022-12-30
[Summary] Kubernetes backend support


6543:
basic support (#9) was added with #552

# Current state:

- [x] Basic pipeline support
- [x] Proper service support
- [x] Default resource limits: 512MB, 1 core
- [x] Integrate it to drone-server
- [x] RWX volume support with `storageclasses`
- [x] Deploy it and run builds for 10 repos with it for 2 weeks
- [x] support resource limits for pods (#1767)
- [x] make services work
  - [x] support service ports (#2656)

maltegrosse:
@6543 great work, can I test it somehow?
Is it possible to include any resources into it, e.g. nvidia gpus ?

6543:
hmm @maltegrosse at the moment I would say, best help is to test and point out it's limitations.

passthrough hardware axeleration like gpus, I never thought of. if it's about help via $, we have an openCollective account

maltegrosse:
@6543 is there a different behavior regarding cpu/mem resources and other resources available on the node?  
https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/

6543:
I guess gpu's are just not jet taken into account - but that's an interesting usecase

Dan6erbond:
@maltegrosse as long as the cluster has device plugins providing resources such as GPUs, it's pretty much the same as defining a resource limit for CPU/memory.

maltegrosse:
@6543 I seeing great progress in k8s backend support.
I think I wanna give it a try and update from 0.x to the latest stable version. But now I just saw that 2.0 is in the pipeline.
Would you recommend me to wait for the new 2.x major release? if yes, is there any ETA ?

pat-s:
WRT to k8s specifically: I'd say it's useable for production needs, at least I do so across many instances and a few dozen repos with some complex configs.

maltegrosse:
sounds nice, thanks for the feedback @pat-s . the only point which confuses me are the resource limits. Seems like every step requires the resources definitions - or can I simple add it to any agent globally by using normal k8s syntax? (so it will be applied to any job) - see https://woodpecker-ci.org/docs/next/administration/backends/kubernetes#resources
The reason is that i dont trust my users that they can assume / predict proper resource usage :)

Addionally, are there any breaking changes regarding my db (postgres) ? (currently using 0.15.6)
 - I couldnt find anything at https://woodpecker-ci.org/docs/next/migrations

[...]

Issue #1238, opened 2022-10-07
Missing required Service.spec.ports in Kubernetes


mmoya:
### Component

agent

### Describe the bug

 `Service.spec.ports` is required and not being passed.

Same environment and pipeline of #1232. This time using agent version `next-3130a1c5`. The error is:

```
default/woodpecker-woodpecker-agent-56df456494-b62fg[woodpecker-agent]: {
  "level": "error",
  "error": "rpc error: code = Unknown desc = Proc finished with exitcode 1, Service \"wp-0-3996753301819885534-services-0\" is invalid: spec.ports: Required value",
  "time": "2022-10-07T21:27:22Z",
  "message": "grpc error: wait(): code: Unknown: rpc error: code = Unknown desc = Proc finished with exitcode 1, Service \"wp-0-3996753301819885534-services-0\" is invalid: spec.ports: Required value"
}
```

I see that [`kubernetes.Service`](https://github.com/woodpecker-ci/woodpecker/blob/master/pipeline/backend/kubernetes/service.go#L12) has a `ports` parameter but all calls are done with an empty array as its value.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-3130a1c5"}
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

6543:
might could solve the open todo of #552

anbraten:
So you are trying to run a pipeline with this context (no ports defined)?:

```yml
services:
  docker:
    image: docker:22.06-rc-dind

pipeline:
  build:
    image: docker:22.06-rc-dind
    commands:
      - printenv | sort
```

I will try to debug it against a cluster again.

mmoya:
> So you are trying to run a pipeline with this context (no ports defined)?:

Correct, no ports defined.

zc-devs:
> no ports defined

> all calls are done with an empty array as its value

[...]

PR #2656, opened 2023-10-27
Add ports into pipeline backend step model

### LINKED ISSUE #2655: reopen - Missing required Service.spec.ports in Kubernetes
_Originally posted by @J-Ha in https://github.com/woodpecker-ci/woodpecker/issues/1238#issuecomment-1776908019_

this problem still exists.
this problem has nothing todo with the helm-chart.
it is more about the kubernetes backend and the service that should be created by a job

the service of the job awaits a port:
https://github.com/woodpecker-ci/woodpecker/blob/next-release/1.0.3/pipeline/backend/kubernetes/service.go#L26

but there is no option to pass a service port into this by the `.woodpecker.yml` 
and the job terminates on creation with the error of:
`is invalid: spec.ports: Required value`

there is something within the schema.json defined, nothing in the documentation. But even when setting a port in the array im getting the same error.

https://github.com/woodpecker-ci/woodpecker/blob/v1.0.3/pipeline/schema/schema.json#L580

zc-devs:
Closes #2655.

[Pipeline](https://woodpecker-ci.org/docs/next/usage/services#complete-pipeline-example):
```yaml
services:
  database:
    image: mysql
    environment:
      - MYSQL_DATABASE=test
      - MYSQL_ROOT_PASSWORD=example
    ports:
      - 3306

steps:
  get-version:
    image: ubuntu
    commands:
      - ( apt update && apt dist-upgrade -y && apt install -y mysql-client 2>&1 )> /dev/null
      - sleep 60s # need to wait for mysql-server init
      - echo 'SHOW VARIABLES LIKE "version"' | mysql -uroot -hdatabase test -pexample
```

Service:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: wp-01hdq6gbkw1mn6k1655fs3rntf-0-services-0
  namespace: woodpecker-runtime
  ...
  selfLink: >-
    /api/v1/namespaces/woodpecker-runtime/services/wp-01hdq6gbkw1mn6k1655fs3rntf-0-services-0
status:
  loadBalancer: {}
spec:
  ports:
    - protocol: TCP
      port: 3306
      targetPort: 3306
  selector:
    step: database
  clusterIP: 10.43.180.120
  clusterIPs:
    - 10.43.180.120
  type: ClusterIP
  sessionAffinity: None
  ipFamilies:
    - IPv4
  ipFamilyPolicy: SingleStack
  internalTrafficPolicy: Cluster
```


J-Ha:
this solves also #1238

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -154,9 +154,7 @@ func (e *kube) SetupWorkflow(ctx context.Context, conf *types.Config, taskUUID s
 					return err
 				}
 				log.Trace().Str("pod-name", stepName).Msgf("Creating service: %s", step.Name)
-				// TODO: support ports setting
-				// svc, err := Service(e.config.Namespace, step.Name, stepName, step.Ports)
-				svc, err := Service(e.config.Namespace, step.Name, stepName, []string{})
+				svc, err := Service(e.config.Namespace, step.Name, step.Alias, step.Ports)
```
xoxys, line 157: Was it intended to use `step.Alias` instead of `stepName` while `stepName` is now only used in the trace log?

zc-devs: > stepName is now only used in the trace log

No, it becomes `metadata.name` field - `wp-01hdq6gbkw1mn6k1655fs3rntf-0-services-0` in the example.

> Was it intended to use step.Alias instead of stepName

Yes, it becomes `spec.selector.step` field - `database` in the example.

pipeline/backend/types/step.go
```diff
@@ -48,6 +48,7 @@ type Step struct {
 	NetworkMode    string            `json:"network_mode,omitempty"`
 	IpcMode        string            `json:"ipc_mode,omitempty"`
 	Sysctls        map[string]string `json:"sysctls,omitempty"`
+	Ports          []int32           `json:"ports,omitempty"`
```
6543, line 51:
```suggestion
	Ports          []uint16           `json:"ports,omitempty"`
```

ports are uint16 if we want to be specific :)

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -165,6 +165,11 @@ func (c *Compiler) createProcess(name string, container *yaml_types.Container, s
 		cpuSet = c.reslimit.CPUSet
 	}
 
+	var ports []int32
+	for _, port := range container.Ports {
+		ports = append(ports, int32(port))
```
6543, line 170:
```suggestion
		ports = append(ports, uint16(port))
```

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -165,6 +165,11 @@ func (c *Compiler) createProcess(name string, container *yaml_types.Container, s
 		cpuSet = c.reslimit.CPUSet
 	}
 
+	var ports []int32
```
6543, line 168:
```suggestion
	var ports []uint16
```

pipeline/backend/kubernetes/service_test.go
```diff
@@ -54,7 +54,7 @@ func TestService(t *testing.T) {
 	  }
 	}`
 
-	s, _ := Service("foo", "bar", "baz", []string{"1", "2", "3"})
+	s, _ := Service("foo", "bar", "baz", []int32{1, 2, 3})
```
6543, line 57:
```suggestion
	s, _ := Service("foo", "bar", "baz", []uint16{1, 2, 3})
```

pipeline/backend/kubernetes/service.go
```diff
@@ -15,24 +15,17 @@
...
-	"fmt"
-	"strconv"
-
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-func Service(namespace, name, podName string, ports []string) (*v1.Service, error) {
+func Service(namespace, name, podName string, ports []int32) (*v1.Service, error) {
```
6543, line 23:
```suggestion
func Service(namespace, name, podName string, ports []uint16) (*v1.Service, error) {
```

6543 (approved):

6543:
@zc-devs could you report if we can check the "services" part of  #1513 now ?

J-Ha:
hej @zc-devs i found an issue with this fix.

when you create a job pipeline it results into a wrongly cofigured service or service step pod (depends on the point of view)

for example:

the service step pod gets generated with the following label:

```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2023-11-02T21:31:42Z"
  labels:
    step: wp-01he8z5bt3gt6t6bx0n838j486-0-services-0
  name: wp-01he8z5bt3gt6t6bx0n838j486-0-services-0
```

and the service gehts a not matiching selector:
```

apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2023-11-02T21:25:25Z"
  name: wp-01he8yteq8ge6pa3e8z3g4fnkj-0-services-0
  namespace: woodpecker
  resourceVersion: "32674961"
  uid: 571fa3d4-19a1-4fa5-a29a-dc8b3a96e372
spec:
  clusterIP: 100.70.144.123
  clusterIPs:
  - 100.70.144.123
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - port: 2375
    protocol: TCP
    targetPort: 2375
  selector:
    step: docker

```

PR #2674, opened 2023-10-30
Fix secret image filter regex

### LINKED ISSUE #2641: Secret image filter with port invalid
### Component

server, web-ui

### Describe the bug

I'm running a local registry on a custom port (:8443) . So when I'm using a local image in my pipelines I reference it by specifying the following image:
```
steps:
  image:
    image: myregistry.example.com:8443/myimage:mytag
```

Now I tried to add a global secret and restrict it to an image from my registry. So I entered the following image name in the webui: `myregistry.example.com:8443/myimage:mytag`

This results in the error message 
`Error updating global secret. Invalid Secret Image: image 'myregistry.example.com:8443/myimage:mytag' do not match regexp '^([\w\d\-_\.\/]*[\w\d\-_]+)+(:[\w\d\-_]+)?$'`

I searched the code and found the following RegEx
https://github.com/woodpecker-ci/woodpecker/blob/a1a35eb7be62309aad1c5ecf35746a9df636568f/server/model/secret.go#L117-L122
This RegEx does not allow the port in the url part.

A RegEx like this should fix the problem
```
var validDockerImageString = regexp.MustCompile(
	`^([\w\d\-_\.\/]*` + // optional url prefix
        `(:\d*/)?` + // optional port prefix
	`[\w\d\-_]+` + // image name
	`)+` +
	`(:[\w\d\-_]+)?$`, // optional image tag
)
```

### System Info

```shell
V1.0.2
```

### Additional context

_No response_

### Validations

HamburgerJungeJr:
closes #2641 

This PR should fix the issue described in #2641. 

While working on this issue I noticed that the regex would accept an image starting with `/` like `/mysql`
Therefore I fixed this too and added some more tests which should cover most combinations of image names and also most combinations of missing parts of the image name.


anbraten (approved):

6543 (approved):

6543:
I think we should backport that one :)

6543:
backport ->  #2686

PR #2180, opened 2023-08-09
fail hard on deprecated pipeline keys

6543:
so we can drop it in next minor version complete

PR #2677, opened 2023-10-31
Revert breaking pipeline changes

qwerty287:
Revert #2180 and #2480 so we can release v2.0 without breaking changes in pipeline config.

After merging #2476 we should apply these changes to a new v2.

After merging this, we should be ready for the 2.0 release.

### DIFF
--- a/cli/deploy/deploy.go
+++ b/cli/deploy/deploy.go
@@ -75,7 +75,7 @@ func deploy(c *cli.Context) error {
 	status := c.String("status")
 
 	pipelineArg := c.Args().Get(1)
-	var number int
+	var number int64
 	if pipelineArg == "last" {
 		// Fetch the pipeline number from the last pipeline
 		pipelines, berr := client.PipelineList(repoID)
@@ -100,7 +100,7 @@ func deploy(c *cli.Context) error {
 			return fmt.Errorf("Cannot deploy failure pipeline")
 		}
 	} else {
-		number, err = strconv.Atoi(pipelineArg)
+		number, err = strconv.ParseInt(pipelineArg, 10, 64)
 		if err != nil {
 			return err
 		}
--- a/cli/log/log_purge.go
+++ b/cli/log/log_purge.go
@@ -42,7 +42,7 @@ func logPurge(c *cli.Context) (err error) {
 	if err != nil {
 		return err
 	}
-	number, err := strconv.Atoi(c.Args().Get(1))
+	number, err := strconv.ParseInt(c.Args().Get(1), 10, 64)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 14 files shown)


pat-s:
> After merging https://github.com/woodpecker-ci/woodpecker/pull/2476 we should apply these changes to a new v2.

You mean v3?

---

I would also be ok with waiting a few days if you think #2476 can make it in and we're good to go then. I can't judge its status...

6543 (changes requested):
I'm against it ... as it was deprecated last release already ...

anbraten:
> Yes, and it's still deprecated (olds are used only as fallback). However, I'd make this only breaking in a new pipeline config version after we merged https://github.com/woodpecker-ci/woodpecker/pull/2476 .

I am also for leaving the fallback in for now. Biggest issue as mentioned before is that server admins are not enforced to tell their users to update their configs and users might not read migration notes. That's why I would like to get #1572 in to show warnings to the user if the pipeline contains deprecated parts.

anbraten (commented):
We need to undo the changes in the yaml schema as well

pipeline/frontend/yaml/types/workflow.go
```diff
@@ -36,6 +36,13 @@ type (
 		Cache    base.StringOrSlice `yaml:"cache,omitempty"`
 		Networks WorkflowNetworks   `yaml:"networks,omitempty"`
 		Volumes  WorkflowVolumes    `yaml:"volumes,omitempty"`
+
+		// Deprecated
+		PlatformDontUseIt string `yaml:"platform,omitempty"` // TODO: remove after v1.2.x version
+		// Deprecated
+		BranchesDontUseIt *constraint.List `yaml:"branches,omitempty"` // TODO: remove after v1.1.x version
+		// Deprecated
+		PipelineDontUseIt ContainerList `yaml:"pipeline,omitempty"` // TODO: remove after v1.1.x version
```
anbraten, line 45:
```suggestion
		// Deprecated
		PlatformDontUseIt string `yaml:"platform,omitempty"` // TODO: remove in next major version
		// Deprecated
		BranchesDontUseIt *constraint.List `yaml:"branches,omitempty"` // TODO: remove in next major version
		// Deprecated
		PipelineDontUseIt ContainerList `yaml:"pipeline,omitempty"` // TODO: remove in next major version
```

anbraten:
> @anbraten the schema was never updated so it should still work

Ahh great 😅

anbraten (approved):

6543 (approved):
well I still want to get rid of it

anbraten:
> well I still want to get rid of it

Sure, next release to do so will come 😉

PR #2680, opened 2023-10-31
Do not sanitzie secrets with 3 or less chars

6543:
as this secrets have to low entropy they can not be valid secrets and e.g. make log only unredable

just add a secret with value `a` to a repo an run a pipeline ...

---


anbraten (approved):

anbraten:
Do we need `security` (it was and is kinda secure before and now) and `breaking` (what should the user change / be careful of) labels?

6543:
well it breaks with how it did behave in the past ... so we need that label at least - if we should add security ... well I'll go with your decision :)

PR #2681, opened 2023-10-31
Stop steps after they are done

anbraten:
We haven't stopped steps immediateley after they were done. Therefore in case of the docker & kubernetes backends the containers of finished steps existed til the whole workflow was done.


pipeline/backend/types/engine.go
```diff
@@ -44,6 +44,9 @@ type Engine interface {
 	// TailStep the workflow step logs.
 	TailStep(ctx context.Context, step *Step, taskUUID string) (io.ReadCloser, error)
 
+	// StopStep stop the workflow step.
+	StopStep(ctx context.Context, step *Step, taskUUID string) error
```
6543, line 48: we should call it what it is: DestroyStep ?

6543: or more CleanStep

anbraten: Wanted to be consistent with `StartStep` which is also a `create` somehow.

6543:
if a container is already deleted ... does this lead to errors if DestoryWorkflow is exec ?

anbraten:
> DestoryWorkflow

We have `isErrContainerNotFoundOrNotRunning` checks in the `DestroyWorkflow` part, so that should be fine.

pipeline/backend/types/engine.go
anbraten: Renamed it to `DestroyStep` as its also `DestroyWorkflow`

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -325,6 +325,29 @@ func (e *kube) TailStep(ctx context.Context, step *types.Step, taskUUID string)
...
+
+	if err := e.client.CoreV1().Pods(e.config.Namespace).Delete(ctx, podName, deleteOpts); err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	return nil
+}
+
 // Destroy the pipeline environment.
 func (e *kube) DestroyWorkflow(_ context.Context, conf *types.Config, taskUUID string) error {
```
6543, line 352: could we remove the log in 376  as it is default now

anbraten: done

6543 (approved):

strk:
Will this be backported to 1.0 branch too ?

PR #2686, opened 2023-10-31
Fix secret image filter regex (#2674)

6543:
Backport #2674


lafriks:
Looks like deps need to be updated also to fix security checks

6543:
> Looks like deps need to be updated also to fix security checks

well yes but it's not required to merge this pull

lafriks (approved):

PR #2688, opened 2023-11-01
Let the backend engine report the current platform

6543:
if you run woodpecker-agent on windows and connect it to an docker daemon, there could be two different platforms possible, as you can switch from linux to windows mode and visa versa

---


6543:
this will also make it possible to improve the situation for kubernetes ... as there you can have different platforms too

pipeline/backend/types/engine.go
```diff
@@ -47,3 +47,9 @@ type Engine interface {
 	// DestroyWorkflow the workflow environment.
 	DestroyWorkflow(ctx context.Context, conf *Config, taskUUID string) error
 }
+
+// EngineInfo represents the reported information of a loaded engine
+type EngineInfo struct {
+	Platform string
+	Backend  string
```
anbraten, line 54: Why not stick and directly use `backend.Name()` seems kinda duplicated.

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -120,12 +122,17 @@ func (e *kube) Load(context.Context) error {
...
 
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	e.client = kubeClient
 
-	return nil
+	// TODO: use info resp of kubeClient to define platform var
```
anbraten, line 130: Wont be possible as Kubernetes is a cluster and therefore it could have multiple architectures. We can only stick to the architecture of the agents host.

pipeline/backend/docker/docker.go
```diff
@@ -344,3 +349,15 @@ func isErrContainerNotFoundOrNotRunning(err error) bool {
 	// Error: No such container: ...
 	return err != nil && (strings.Contains(err.Error(), "No such container") || strings.Contains(err.Error(), "is not running"))
 }
+
+// normalizeArchType converts what docker info reports als arch
+// and convert it into the runtime.GOARCH format
+// TODO: find out if we we need to convert other arch types too
```
anbraten, line 355:
```suggestion
// normalizeArchType converts the arch type reported by docker info into
// the runtime.GOARCH format
// TODO: find out if we we need to convert other arch types too
```

pipeline/backend/kubernetes/kubernetes.go
6543: Well if it's `linux/amd64`,`linux/riscv64`

We could report `linux/multi` or `linux/{amd64,riscv64}` ...

6543: And if we only have one type we do report as with docker?

anbraten: Would be a tough one to match such labels. Maybe adding it as regex could do the trick. Okay, but that's really a future PR I guess

pipeline/backend/types/engine.go
anbraten: I would just remove it from `EngineInfo` and use `backend.Name()` in the places as before.

pipeline/backend/kubernetes/kubernetes.go
6543: I'll create a new issue and link the todo to it ... it is worth it

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -120,12 +122,16 @@ func (e *kube) Load(context.Context) error {
...
```
6543, line 130:
```suggestion
	// TODO(2693): use info resp of kubeClient to define platform var
```

anbraten (approved):

PR #2691, opened 2023-11-01
Generate go-client from swagger

### LINKED ISSUE #3053: Generate sdk from swagger

as of https://github.com/woodpecker-ci/woodpecker/pull/2691

~~blocked by https://github.com/swaggo/swag/issues/386~~ -> https://github.com/woodpecker-ci/woodpecker/pull/3897

### Suggested solution

auto-generate client from swagger docs so we don't have to do it manually

### Alternative

_No response_

### Additional context

_No response_

### Validations

anbraten:
close #3053
close #3652

Creates a new go api client automatically out of the swagger definition


runephilosof-karnovgroup:
Won't this type int or 'latest' be a problem?
https://github.com/woodpecker-ci/woodpecker/blob/main/server/api/pipeline.go#L126

Should it be changed to something like?
```go
//	@Param		number		path	int|string		true	"the number of the pipeline, OR 'latest'"
```

runephilosof-karnovgroup:
The current client does takes `int` instead of `int64` as arguments for repo_id and pipeline number etc.
It should take `int64`, how will this generated client know to take int64 arguments?

woodpecker-go/stream.go
```diff
@@ -0,0 +1,22 @@
...
+	"context"
+	"fmt"
+
+	"github.com/r3labs/sse/v2"
+)
+
+func (c *Client) StreamPipelineStepLog(ctx context.Context, repoID, pipelineNumber, stepID int64) error {
+	logs := make(chan *sse.Event)
+
+	client := sse.NewClient(fmt.Sprintf("%s/api/stream/logs/%d/%d/%d", c.uri, repoID, pipelineNumber, stepID))
```
6543, line 13: how is the logs chan exposed ?

PR #2692, opened 2023-11-01
Use full qualifyer for images

xoxys:
Supersedes https://github.com/woodpecker-ci/woodpecker/pull/1659

### DIFF
--- a/.woodpecker/binaries.yml
+++ b/.woodpecker/binaries.yml
@@ -2,6 +2,10 @@ depends_on:
   - test
   - web
 
+# TODO: upload build artifacts for pushes to ${CI_REPO_DEFAULT_BRANCH}
+when:
+  event: tag
+
 variables:
   - &golang_image 'docker.io/golang:1.21.3'
   - &node_image 'docker.io/node:21-alpine'
--- a/.woodpecker/binaries.yml
+++ b/.woodpecker/binaries.yml
@@ -2,14 +2,10 @@ depends_on:
   - test
   - web
 
-# TODO: upload build artifacts for pushes to ${CI_REPO_DEFAULT_BRANCH}
-when:
-  event: tag
-
 variables:
-  - &golang_image 'golang:1.21.3'
-  - &node_image 'node:21-alpine'
-  - &xgo_image 'techknowlogick/xgo:go-1.21.3'
+  - &golang_image 'docker.io/golang:1.21.3'
+  - &node_image 'docker.io/node:21-alpine'
+  - &xgo_image 'docker.io/techknowlogick/xgo:go-1.21.3'
   - &xgo_version 'go-1.21.2'
 
 steps:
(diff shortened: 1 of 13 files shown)


shared/constant/constant.go
```diff
@@ -16,10 +16,10 @@ package constant
 
 // PrivilegedPlugins can be changed by 'WOODPECKER_ESCALATE' at runtime
 var PrivilegedPlugins = []string{
-	"plugins/docker",
-	"plugins/gcr",
-	"plugins/ecr",
-	"woodpeckerci/plugin-docker-buildx",
+	"docker.io/plugins/docker",
```
xoxys, line 19: Unsure about this change, this would be a breaking change as auto-escalate doesn't work anymore for image names without the prefix.

shared/constant/constant.go
```diff
@@ -33,7 +33,7 @@ var DefaultConfigOrder = [...]string{
 
 const (
 	// DefaultCloneImage can be changed by 'WOODPECKER_DEFAULT_CLONE_IMAGE' at runtime
-	DefaultCloneImage = "docker.io/woodpeckerci/plugin-git:2.1.1"
+	DefaultCloneImage = "quay.io/woodpeckerci/plugin-git:2.1.1"
```
anbraten, line 36: Why should we change this one? I would prefer to stick with the old one.

.woodpecker/binaries.yml
```diff
@@ -7,9 +7,9 @@ when:
   event: tag
 
 variables:
-  - &golang_image 'golang:1.21.3'
-  - &node_image 'node:21-alpine'
-  - &xgo_image 'techknowlogick/xgo:go-1.21.3'
+  - &golang_image 'docker.io/golang:1.21.3'
```
anbraten, line 10: What's the reason to add `docker.io`?

anbraten (changes requested):

shared/constant/constant.go
xoxys: I've just picked up the outdated PR and fixed conflicts. If you want to stay with docker.io by default fine for me.

.woodpecker/binaries.yml
xoxys: The container engine that is executing the pipeline could have other default registries, make it explicit what registry to use for images. See https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name

shared/constant/constant.go
anbraten: Would like to stay with it for now. To me using `docker.io` or `quay.io` makes no difference both are some "random" companies hosting images for free.

shared/constant/constant.go
xoxys: Reverted.

6543 (approved):

anbraten (approved):

PR #2696, opened 2023-11-01
Fix typo in golang image name

6543:


anbraten (approved):

6543:
Haha i thought the title would be funny enouth be be commited into history ...

... well you could also rename it to a serious title 😆

Issue #2657, opened 2023-10-27
Properly link commit hash in run UI


pat-s:
### Component

web-ui

### Describe the bug

From https://codeberg.org/Codeberg-CI/feedback/issues/91

A commit hash link in CRON runs only links to `user/repo` instead of `user/repo/commit/xxx`

![image](https://github.com/woodpecker-ci/woodpecker/assets/8430564/1ae24e61-f566-4567-88b2-f6105d60c60a)

### System Info

```shell
next-7b3eeec65c # Oct 17
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

anbraten:
Should be related to https://github.com/anbraten/woodpecker/blob/ccdab26deb7f4052e0914b80978bc2a5f1c5a574/server/cron/cron.go#L142-L143

PR #2698, opened 2023-11-01
Some UI fixes

anbraten:
- fix first log line being dropped if channel was opened
- link org and repo on repo logs view
- fix decline pipeline status does not update


pat-s:
Maybe #2657 could be included here while at it?

server/pipeline/start.go
```diff
@@ -33,30 +32,13 @@ func start(ctx context.Context, store store.Store, activePipeline *model.Pipelin
 		log.Error().Err(err).Msg("Failed to cancel previous pipelines")
 	}
 
-	if err := store.WorkflowsCreate(activePipeline.Workflows); err != nil {
```
6543, line 36: ah nice that could solve the issue to not be able to restart declined pipelines :)

anbraten: removed it for now as it wasn't running when approving pipelines as we were missing the already existing workflows then and ran into a unique constraint error.

6543 (approved):

PR #2706, opened 2023-11-02
Switch to go vanity urls

6543:
close https://github.com/woodpecker-ci/infrastructure/issues/4

switch from `github.com/woodpecker-ci/woodpecker/...` to `go.woodpecker-ci.org/woodpecker/...`


anbraten (changes requested):
Would block this until we get the needed PRs for the release merged.

6543:
I'd like to get that int v2.0.0 as backports will be easyer this way :)

pat-s:
Couldn't this also go into 2.1.0? If @6543 resolves all issues quickly, 2.0.0 is also fine, but I don't think we should necessarily wait for this?

6543:
I'll resolve conflicts :)

anbraten (commented):
Had a look at all changes and those seem fine. However some files are missing.

anbraten (approved):

anbraten:
Pleas upgrade depending packages like autoscaler accordingly.

PR #2708, opened 2023-11-02
Remove `lint` from `test.yml`

pat-s:
pre-commit.ci is also running `golangci-lint run` but almost instant while the step in question takes almost 7 mins 🤯 

Example run: https://results.pre-commit.ci/run/github/179344069/1698940277.6X0mpSLoRbSqB0fkONEFaQ


pat-s:
@anbraten Similar to #2762. I think in this case pre-commit heavily profits from the dep cache WRT to speed. Are you ok with removing this workflow step? No plugin involved here, just a makefile call.

pat-s:
Let's stick with the current approach.

PR #2725, opened 2023-11-03
Add deprecation warnings

### LINKED ISSUE #2720: Add deprecation warning to linter for `pipeline`
We reverted some changes in #2677 which should be finally deleted in the next major release. Until then we want to show the users still using the old syntax a deprecation warning.

anbraten:
closes #2720
- allow to lint multiple files (related to #2685)
- add file name to linter errors


anbraten:
> This won't work. See https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/yaml/parse.go#L44

Good catch. Any idea how I could check it otherwise? Parsing the raw yaml string and checking for the key to exists?

> Also, would it be possible to return multiple errors there so we can lint all three fields?

Definitely, that's why I introduced multierr support. Just forgot about it again

PR #2732, opened 2023-11-04
Update docusaurus to v3

qwerty287:
@pat-s 

This would be my variant…

I'd prefer it as it does not changes the MD files.

### DIFF
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -1,8 +1,9 @@
-const codeThemes = require('prism-react-renderer').themes;
-const path = require('path');
+import { themes } from 'prism-react-renderer';
+import type {Config} from '@docusaurus/types';
+import type * as Preset from '@docusaurus/preset-classic';
+import * as path from 'path';
 
-/** @type {import('@docusaurus/types').Config} */
-module.exports = {
+const config: Config = {
   title: 'Woodpecker CI',
   tagline: 'Woodpecker is a simple CI engine with great extensibility.',
   url: 'https://woodpecker-ci.org',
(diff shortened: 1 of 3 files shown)


pat-s:
Looks way cleaner! Let's see if it builds and how it looks.

One thing I think we should change nevertheless are the "format" outputs in `cli.md` which are not formatted as code and contain c/p ascii characters.

pat-s:
CI fails at the same error I encountered locally

> message: 'Hook useColorMode is called outside the <ColorModeProvider>. 

this seems (for now) unrelated to the plugins?

> I agree, but isn't this document auto-generated? [main/cmd/cli/docs.go#L26](https://github.com/woodpecker-ci/woodpecker/blob/main/cmd/cli/docs.go?rgh-link-date=2023-11-04T09%3A08%3A03Z#L26)

Ah, didn't know this. Makes sense though.
Then we need another approach (maybe some post-render hook) but as long as it's not blocking the general upgrade...

pat-s:
Yeah, but I don't see how `useColorMode` and the `/api` relate. Until now, I thought of this being a general issue with the `useColorMode` definition in docusaurus v3 as described in the link of the error. But TBH I don't have enough react knowledge to really troubleshoot this.

pat-s:
`redocusaurus` might take some time as the maintainer said he has no time. Anyhow, the error with the colorMode has been mentioned there before the 3.0 update and I wonder if it's actually related to it - or if we can fix it ourselves and maybe no issues exists with `redocusaurus` at all.

anbraten:
I assume redocusaurus is using a component / context-hook to access the current color-mode, but that seems to be removed / renamed from the error.

docs/docusaurus.config.js
```diff
@@ -139,6 +139,7 @@ module.exports = {
       prism: {
```
anbraten, line 139: Could you change it to the .ts file format

anbraten: https://docusaurus.io/docs/typescript-support#typing-config

anbraten: thanks

anbraten (approved):
deployment & changes lgtm

PR #2736, opened 2023-11-04
Refactor auth

anbraten:


server/router/router.go
```diff
@@ -57,13 +57,17 @@ func Load(noRouteHandler http.HandlerFunc, middleware ...gin.HandlerFunc) http.H
 	{
 		base.GET("/web-config.js", web.Config)
 
-		base.GET("/logout", api.GetLogout)
```
6543, line 60: breaking api?

server/router/router.go
```diff
@@ -57,13 +57,17 @@ func Load(noRouteHandler http.HandlerFunc, middleware ...gin.HandlerFunc) http.H
...
-		base.GET("/login", api.HandleLogin)
-		auth := base.Group("/authorize")
+		auth := base.Group("/auth")
 		{
-			auth.GET("", api.HandleAuth)
-			auth.POST("", api.HandleAuth)
-			auth.POST("/token", api.GetLoginToken)
+			auth.GET("/login", api.Auth)
+			auth.GET("/logout", api.GetLogout)
```
lafriks, line 63: Why this needs to be GET? From security perspective this is bad practice because you can redirect user to this endpoint to force his logout would be better to use endpoints:

GET /auth -> get token
DELETE /auth -> logout

lafriks: Oh, it's for redirects. Than it should be called GET /authorize and logout can be together with GET /token - to get token and DELETE /token - logout (call it as api and then push home route)

lafriks: If needed /logout can be the frontend route that can be redirected externally that would ask the user if he really wants to logout

anbraten:
will not adjust the auth for

PR #2750, opened 2023-11-05
pre-commit: test hadolint-docker for CI

pat-s:


pat-s:
To make this work in CI, we would need to run `pre-commit` in WP with a dedicated image. AFACIS no maintained images are available and only adding one for hadolint is not adding up.

So let's stick with the local hook.

PR #2751, opened 2023-11-05
Support go plugins for forges and agent backends

qwerty287:
As of #2520 

Support to load new forges and agent backends at runtime using go's plugin system. (https://pkg.go.dev/plugin)

I also added a simple example addon (a new forge which just prints log statements), it should be removed later of course, but you can see an example.


anbraten:
Looks pretty nice already. 
Two questions / ideas that came up to me:
- How can we add a plugin to the docker image? Would we create a docker image using the "normal" image as base and simply add the plugin binary?
- Can we combine this with the requested compiler, secrets, registry addons somehow? Problem would be as those should be selectable by users they would need to be executed in a rather separated environment like a step by the agent and therefore we might need to consider sth like <https://github.com/hashicorp/go-plugin>

docs/docs/30-administration/75-addons/20-creating-addons.md
```diff
@@ -0,0 +1,82 @@
...
+
+```sh
+go build -buildmode plugin
+```
+
+The output file is your addon which is now ready to use.
+
+## Restrictions
+
+Addons must directly directly depend on Woodpecker's core (`go.woodpecker-ci.org/woodpecker/woodpecker`).
```
mzampetakis, line 40: Nitpick: Duplicate `directly`.

mzampetakis:
This looks great! 

It seems that any loaded forge through this add-on system will override the server's configuration for any forge. This could be documented at someplace (e.g. `docs/docs/30-administration/75-addons/00-addons.md`) so that the administrators will have a more clear view about server's bevahiour.

mzampetakis:
Is there any particular reason for using `hashicorp/go-plugin` instead of the built in one? 
I believe that it might be more convenient to stick with the go's built in one in terms of upgradability.
What do you think?

anbraten:
"users" would be all users (including non-instance-admins) in my idea. The idea I had some time ago was that the server asks an agent to start a service (exp container) (similar to running a pipeline somewhere). This service would then provide an http endpoint. This endpoint could for example provide the config-service endpoint, so it could be used as an compiler. So as soon as the server needs the config it would check that the service is running and forwards the http request through the grpc connection via the agent to the service. But all of that was just a first idea.

mzampetakis:
Is this completed or blocked by something? I could run some tests with a dummy addon forge to check if everything is working as expected. Please let me know if I could provide any assistance.

anbraten:
> Your suggestion won't work. hashicorp/go-plugin does not support RPC over remote networks, only locally.

That was just a reference. I think we need to create our own http endpoints definitions if we want to go that path. I would use an http api similar to what we have for the config-service already as grpc wouldn't be resistent to possible changes.

anbraten:
> So this would be totally different from this approach. Can I revert it back to go plugins?

Sure, I would be fine adding the go-plugin approach for now as experimental and would not document it, so we could drop / replace it at any point.

mzampetakis:
> > So this would be totally different from this approach. Can I revert it back to go plugins?
> 
> Sure, I would be fine adding the go-plugin approach for now as experimental and would not document it, so we could drop / replace it at any point.

I am fond of moving towards the native go plugin approach as wel. 

Being not documented/experimental means that it could be dropped at some point of time right? So, any implemented addons that will use this approach to integrate with woodpecker will be dropped as well right?

If that's the case I would recommend to be documented but noted as experimental (alongside the possibility of dropping this in the future) in order to be crystal clear but also be able to collect any feedback from implemented addons.

cmd/server/flags.go
```diff
@@ -256,6 +256,10 @@ var flags = append([]cli.Flag{
 		Name:    "enable-swagger",
 		Value:   true,
 	},
+	&cli.StringSliceFlag{
+		EnvVars: []string{"WOODPECKER_ADDONS"},
+		Name:    "addons",
```
mzampetakis, line 261: Maybe the `usage` could be added here.

docs/docs/30-administration/75-addons/00-overview.md
```diff
@@ -0,0 +1,42 @@
...
+
+```diff
+# docker-compose.yml
+version: '3'
+
+services:
+  woodpecker-server:
+    [...]
+    environment:
++     - WOODPECKER_PLUGIN=/path/to/your/addon/file.so
```
mzampetakis, line 33: I think it's `WOODPECKER_ADDONS` here:
```suggestion
+     - WOODPECKER_ADDONS=/path/to/your/addon/file.so
```

docs/docs/30-administration/75-addons/00-overview.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+You may need to [mount the addon file as volume](https://docs.docker.com/storage/volumes/#create-and-manage-volumes) to access it from inside the Docker container.
+
+You can list multiple addons, Woodpecker will automatically determine their type. If you specify multiple addons with the same type, only the first one will be used.
```
mzampetakis, line 38: Maybe it should state that any forge addon will override other setup forges.

cmd/agent/agent.go
```diff
@@ -247,6 +251,25 @@ func run(c *cli.Context) error {
 	return nil
 }
 
+func getEngine(backendCtx context.Context, engineName string, addons []string) (types.Engine, error) {
+	addonEngine, err := addon.Load[types.Engine](addons, addonTypes.TypeEngine)
+	if err != nil {
+		log.Error().Err(err).Msg("cannot load addon")
```
anbraten, line 257:
```suggestion
		log.Error().Err(err).Msg("cannot load backend addon")
```

docs/docs/30-administration/75-addons/20-creating-addons.md
```diff
@@ -0,0 +1,88 @@
...
+   2. An error. If this error is not `nil`, Woodpecker exits.
+
+Directly import Woodpecker's Go package (`go.woodpecker-ci.org/woodpecker/woodpecker`) and use the interfaces and types defined there.
+
+### Return types
+
+| Addon type | Return type                                                                  |
+| ---------- | ---------------------------------------------------------------------------- |
+| `Forge`    | `"go.woodpecker-ci.org/woodpecker/woodpecker/server/forge".Forge`            |
+| `Engine`   | `"go.woodpecker-ci.org/woodpecker/woodpecker/pipeline/backend/types".Engine` |
```
anbraten, line 28:
```suggestion
| `Forge`    | `"go.woodpecker-ci.org/woodpecker/woodpecker/v2/server/forge".Forge`            |
| `Engine`   | `"go.woodpecker-ci.org/woodpecker/woodpecker/v2/pipeline/backend/types".Engine` |
```

docs/docs/30-administration/75-addons/20-creating-addons.md
```diff
@@ -0,0 +1,88 @@
...
+
+In addition to this, addons are only supported on Linux, FreeBSD and macOS.
+
+:::info
+It is recommended to at least support the latest released version of Woodpecker.
+:::
+
+### Compile for different versions
+
+As long as there were no changes to Woodpecker's interfaces or they are backwards-compatible, you can easily compile the addon for multiple version by changing the version of `go.woodpecker-ci.org/woodpecker/woodpecker` using `go get` before compiling.
```
anbraten, line 54:
```suggestion
As long as there were no changes to Woodpecker's interfaces or they are backwards-compatible, you can easily compile the addon for multiple version by changing the version of `go.woodpecker-ci.org/woodpecker/woodpecker/v2` using `go get` before compiling.
```

docs/docs/30-administration/75-addons/20-creating-addons.md
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+  "context"
+  "net/http"
+
+  "github.com/rs/zerolog"
+  "go.woodpecker-ci.org/woodpecker/woodpecker/server/forge"
+  forge_types "go.woodpecker-ci.org/woodpecker/woodpecker/server/forge/types"
+  "go.woodpecker-ci.org/woodpecker/woodpecker/server/model"
+  addon_types "go.woodpecker-ci.org/woodpecker/woodpecker/shared/addon/types"
```
anbraten, line 73:
```suggestion
  "go.woodpecker-ci.org/woodpecker/woodpecker/v2/server/forge"
  forge_types "go.woodpecker-ci.org/woodpecker/woodpecker/v2/server/forge/types"
  "go.woodpecker-ci.org/woodpecker/woodpecker/v2/server/model"
  addon_types "go.woodpecker-ci.org/woodpecker/woodpecker/v2/shared/addon/types"
```

shared/addon/addon.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+			pluginCache[file] = p
+		}
+
+		typeLookup, err := pluginCache[file].Lookup("Type")
+		if err != nil {
+			return nil, err
+		}
+		if addonType, is := typeLookup.(*types.Type); !is {
+			return nil, errors.New("addon type has incorrect type")
```
anbraten, line 38:
```suggestion
			return nil, errors.New("addon type is incorrect")
```

shared/addon/addon.go
```diff
@@ -0,0 +1,66 @@
...
+		} else if *addonType != t {
+			continue
+		}
+
+		mainLookup, err := pluginCache[file].Lookup("Addon")
+		if err != nil {
+			return nil, err
+		}
+		fmt.Println(reflect.TypeOf(mainLookup))
```
anbraten, line 47:
```suggestion
```

shared/addon/addon.go
```diff
@@ -0,0 +1,66 @@
...
+		main, is := mainLookup.(func(zerolog.Logger, []string) (T, error))
+		if !is {
+			return nil, errors.New("addon main has incorrect type")
```
anbraten, line 50:
```suggestion
			return nil, errors.New("addon main function has incorrect type")
```

docs/docs/30-administration/75-addons/00-overview.md
```diff
@@ -0,0 +1,44 @@
...
+    [...]
+    environment:
++     - WOODPECKER_ADDONS=/path/to/your/addon/file.so
+```
+
+You may need to [mount the addon file as volume](https://docs.docker.com/storage/volumes/#create-and-manage-volumes) to access it from inside the Docker container.
+
+You can list multiple addons, Woodpecker will automatically determine their type. If you specify multiple addons with the same type, only the first one will be used.
+
+Using an addon always overwrites Woodpecker's internal setup. This means, that a forge addon will be used if specified, no matter what's configured for the forges natively supported by Woodpecker.
```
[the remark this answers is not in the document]
anbraten: Yep, just wanted to mention that.

docs/docs/30-administration/75-addons/20-creating-addons.md
```diff
@@ -0,0 +1,88 @@
...
+### Return types
+
+| Addon type | Return type                                                                     |
+| ---------- | ------------------------------------------------------------------------------- |
+| `Forge`    | `"go.woodpecker-ci.org/woodpecker/woodpecker/v2/server/forge".Forge`            |
+| `Engine`   | `"go.woodpecker-ci.org/woodpecker/woodpecker/v2/pipeline/backend/types".Engine` |
+
+## Compiling
+
+After you wrote your addon code, compile your addon:
```
mzampetakis, line 32:
```suggestion
After you write your addon code, compile your addon:
```

docs/docs/30-administration/75-addons/20-creating-addons.md
```diff
@@ -0,0 +1,88 @@
...
+
+In addition to this, addons are only supported on Linux, FreeBSD and macOS.
+
+:::info
+It is recommended to at least support the latest released version of Woodpecker.
+:::
+
+### Compile for different versions
+
+As long as there were no changes to Woodpecker's interfaces or they are backwards-compatible, you can easily compile the addon for multiple version by changing the version of `go.woodpecker-ci.org/woodpecker/woodpecker/v2` using `go get` before compiling.
```
mzampetakis, line 54:
```suggestion
As long as there are no changes to Woodpecker's interfaces or they are backwards-compatible, you can easily compile the addon for multiple version by changing the version of `go.woodpecker-ci.org/woodpecker/woodpecker/v2` using `go get` before compiling.
```

cmd/agent/flags.go
```diff
@@ -97,4 +97,8 @@ var flags = []cli.Flag{
 		Usage:   "backend to run pipelines on",
 		Value:   "auto-detect",
 	},
+	&cli.StringSliceFlag{
+		EnvVars: []string{"WOODPECKER_ADDONS"},
+		Name:    "addons",
```
6543, line 102:
```suggestion
		Name:    "addons",
		Usage: "list of addon files",
```

mzampetakis (approved):
Looks great! ♥️

docs/docs/30-administration/75-addons/20-creating-addons.md
```diff
@@ -0,0 +1,88 @@
...
+   2. An error. If this error is not `nil`, Woodpecker exits.
+
+Directly import Woodpecker's Go package (`go.woodpecker-ci.org/woodpecker/woodpecker/v2`) and use the interfaces and types defined there.
+
+### Return types
+
+| Addon type | Return type                                                                     |
+| ---------- | ------------------------------------------------------------------------------- |
+| `Forge`    | `"go.woodpecker-ci.org/woodpecker/woodpecker/v2/server/forge".Forge`            |
+| `Engine`   | `"go.woodpecker-ci.org/woodpecker/woodpecker/v2/pipeline/backend/types".Engine` |
```
6543, line 28:
```suggestion
| `Backend`   | `"go.woodpecker-ci.org/woodpecker/woodpecker/v2/pipeline/backend/types".Backend` |
```

6543 (approved):

PR #2752, opened 2023-11-05
Add docs about Gitea on same host and update docker-compose example

qwerty287:
Closes #1660

Taken out the part of it that's still up-to-date.
Everything else in this PR is outdated as we don't primarily use quay.io

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -16,7 +16,7 @@
   "dependencies": {
     "@docusaurus/core": "^3.0.0",
     "@docusaurus/preset-classic": "^3.0.0",
-    "@easyops-cn/docusaurus-search-local": "^0.36.0",
+    "@easyops-cn/docusaurus-search-local": "^0.37.1",
     "@mdx-js/react": "^3.0.0",
     "@svgr/webpack": "^8.0.0",
     "clsx": "^2.0.0",
@@ -54,12 +54,6 @@
     "overrides": {
       "trim": "^0.0.3",
       "got": "^11.8.5"
-    }
-  },
-  "overrides": {
-    "@easyops-cn/docusaurus-search-local": {
-      "@docusaurus/core": "^3.0.0",
-      "@docusaurus/theme-common": "^3.0.0"
     }
   }
 }
--- a/docs/pnpm-lock.yaml
+++ b/docs/pnpm-lock.yaml
@@ -14,13 +14,13 @@ importers:
     dependencies:
       '@docusaurus/core':
         specifier: ^3.0.0
-        version: 3.0.0(@docusaurus/types@3.0.0)(react-dom@18.2.0)(react@18.2.0)(typescript@5.2.2)
+        version: 3.0.0(@docusaurus/types@3.0.0)(debug@4.3.4)(react-dom@18.2.0)(react@18.2.0)(typescript@5.2.2)
       '@docusaurus/preset-classic':
         specifier: ^3.0.0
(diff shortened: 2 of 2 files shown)


docker-compose.example.yml
```diff
@@ -2,12 +2,11 @@ version: '3'
 
 services:
   woodpecker-server:
-    build:
-      context: .
-      dockerfile: Dockerfile
     image: woodpeckerci/woodpecker-server:local
```
anbraten, line 5: That's `local`? Shouldn't it be `latest`?

```suggestion
    image: woodpeckerci/woodpecker-server:latest
```

docs/docs/30-administration/11-forges/30-gitea.md
```diff
@@ -20,6 +20,24 @@ services:
     [...]
 ```
 
+## Gitea on the same host
+
+If you have Gitea also running on the same host within a container, make sure the agent does have access to it.
+The agent try to clone by the clone URL Gitea reports via API. So you might want to specify a docker network for the agent too:
```
anbraten, line 26:
```suggestion
The agent tries to clone by the clone URL Gitea reports via API. So you might want to specify a docker network for the agent too:
```

docs/docs/30-administration/11-forges/30-gitea.md
```diff
@@ -20,6 +20,24 @@ services:
...
+# docker-compose.yml
+version: '3'
+
+services:
+  [...]
+  woodpecker-agent:
+    [...]
+    environment:
+      - [...]
++     - WOODPECKER_BACKEND_DOCKER_NETWORK=gitea
```
anbraten, line 38: Needs to be `woodpecker` or we should adjust the sample compose file to use `gitea` as network.

docs/docs/30-administration/11-forges/30-gitea.md
```diff
@@ -20,6 +20,24 @@ services:
     [...]
 ```
 
+## Gitea on the same host
+
+If you have Gitea also running on the same host within a container, make sure the agent does have access to it.
+The agent tries to clone by the clone URL Gitea reports via API. So you might want to specify a docker network for the agent too:
```
pat-s, line 26:
```suggestion
The agent tries to clone using the URL which Gitea reports through its API. For simplified connectivity, it helps to add the woodpecker agent to the same docker network as Gitea is in:
```

docs/docs/30-administration/11-forges/30-gitea.md
pat-s: The most important thing is to highlight that both should be in the same network. And if not, that the communication should go via the docker0 gateway (usually `172.17.0.1`). Maybe it's worth pointing this out in more detail instead of just mentioning the hard-coded docket network value?

docker-compose.example.yml
```diff
@@ -2,12 +2,11 @@ version: '3'
 
 services:
   woodpecker-server:
-    build:
-      context: .
-      dockerfile: Dockerfile
-    image: woodpeckerci/woodpecker-server:local
+    image: woodpeckerci/woodpecker-server:latest
```
pat-s, line 5: Do we want to suggest latest here? (Sorry, I know somewhat off-topic but I just saw it here). I think it should be possible to also let this being updated dynamically via renovate without manual work by ourselve.

pat-s (approved):

docker-compose.example.yml
pat-s: Yes but that's exactly the point: you will also never know which version/commit you're actually on. If we suggest such examples, users will use it like this and never change it. Hence I suggest we always put the most recent version into the example.

PR #2754, opened 2023-11-05
UI enhancements/fixes

qwerty287:
- disallow collapsing pipeline configs if it's only one config
- do not show name if only one config

### DIFF
--- a/web/src/views/repo/pipeline/PipelineWrapper.vue
+++ b/web/src/views/repo/pipeline/PipelineWrapper.vue
@@ -39,7 +39,6 @@
               @click="cancelPipeline"
             />
             <Button
-              v-if="pipeline.status !== 'error'"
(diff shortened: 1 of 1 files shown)


xoxys:
> do not show name if only one config

Is this open for discussions? I don't like this approach and find it pretty confusing to not display the pipeline name.

xoxys:
Ah, It's just in cases where only a single wp pipeline config file exists?

Example: Multiple pipeline configs exist, all run on main branch merges, only one pipeline runs on PRs. What happens in this case? Because that's my main issue. I don't like that the output looks different just because only one pipeline out of all existing ones is executed due to defined conditions.

xoxys:
🤦‍♀️ sorry misread it. Then all good 😅

xoxys (approved):

Issue #895, opened 2022-04-29
Improve status messages for protected pipelines


qwerty287:
### Clear and concise description of the problem

Currently, the status sent back to the forge looks like this:

|state|status|message|
|---|---|---|
|pipeline waits for approval|pending|`Pipeline is pending`|
|pipeline is declined|pending|`Pipeline is pending`|

### Suggested solution

It should look like:

|state|status|message|
|---|---|---|
|pipeline waits for approval|pending|`Pipeline waits for approval`|
|pipeline is declined|failed|`Pipeline was declined`|

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

xoxys:
> pipeline is declined 	failed 	Pipeline was declined

Isn't there a `canceled` state? I would use this for declined pipelines instead of `failed`.

PR #2757, opened 2023-11-06
Store workflows/steps for blocked pipeline

qwerty287:
This stores workflows and steps to DB even if it is not yet approved and thus blocked.

I'm not really happy with this, because even though it is stored, it must parse the pipeline again and set back the original UUID. If you have any ideas how to fix/improve this just comment.

In addition, this allows to view step list and side panel for approved pipelines, https://github.com/woodpecker-ci/woodpecker/pull/2345 is partially not longer necessary.

Closes https://github.com/woodpecker-ci/woodpecker/issues/895


pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -32,7 +32,7 @@ import (
 
 func (c *Compiler) createProcess(container *yaml_types.Container, stepType backend_types.StepType) (*backend_types.Step, error) {
 	var (
-		uuid = ulid.Make()
+		uuid string
```
6543, line 36: The uuid is **always** required!! We should not make it optional.

Overwrite it with an external/existing uuid is ok through

cli/exec/exec.go
```diff
@@ -207,6 +207,7 @@ func execWithAxis(c *cli.Context, file, repoPath string, axis matrix.Axis) error
 		compiler.WithMetadata(metadata),
 		compiler.WithSecret(secrets...),
 		compiler.WithEnviron(pipelineEnv),
+		compiler.WithUUID(true),
```
6543, line 210: I'm against addin this option witch you always need ...

6543 (changes requested):
See comment

pipeline/frontend/yaml/compiler/convert.go
6543: Yes but if we add that option. It add the posibility to create a bug wich result in strange behaviour in completly unrelated parts. I rather have it created and stored even if not jet jused, as to have a posibility to create a bug in the future, because some refactorung did not had this in mind.

Anyway i think i should also add a integrity check to the backend data ... i open a new issue

6543 (approved):

PR #2762, opened 2023-11-06
Add `prettier` pre-commit hook

pat-s:
~~We don't need to run it twice and pre-commit execution is almost instant while the WP plugin takes ~ 30s.~~

So people can catch prettier issues before committing.


anbraten:
We need to make sure why WP is slower (if it really is) as if the plugin-image is present both should just clone the repo and run prettier. In addition it might be kinda weird if we start replacing WP with pre-commit.ci, not really useful to show WP users how to do things.

pat-s:
precommit.ci takes ~7s for prettier execution (compare https://results.pre-commit.ci/run/github/179344069/1699281343.TWyETzQATxO-wl6WiT8rMA (40s) with https://results.pre-commit.ci/run/github/179344069/1699280456.0pM_T8CpT4CFWso9fC8d-w (33s)).

On top there is caching from precommit.ci on the exec env.

WP: 4s image DL + 21s exec
Precommit: cached exec env (-> effectively instant readiness) + 7s exec

The VMs WP uses might not be as performant as the precommit ones (switching to arm might speed things up). 

> In addition it might be kinda weird if we start replacing WP with pre-commit.ci, not really useful to show WP users how to do things.

Hmm. I am not a friend of "we need to use our stuff even though it's not so good". It's an offer but if there are better alternatives (doesn't matter in what regards), why stick blindly to the self-baked ones?
I am aware that precommit.ci is only available for GH and that alternatives should exist and are great to have. And we should definitely use them on CB for `woodpecker-plugins` etc. But why not save time and resources here? The WP main repo should not necessarily be the demo repo for everything, IMO.

(btw the same applies to #2708, which is just different in the sense that no plugin exists for it).

anbraten:
WP when not waiting on unrelated tasks: https://ci.woodpecker-ci.org/repos/3780/pipeline/9569/3

anbraten:
> Hmm. I am not a friend of "we need to use our stuff even though it's not so good". It's an offer but if there are better alternatives (doesn't matter in what regards), why stick blindly to the self-baked ones?

I guess the main point is you wont improve it if you don't use it. So comparing the two on a fair basis now shows us that we are ~2 seconds slower on clone and wp takes ~6s to check using prettier with a total of ~10s. Isn't that a starting point to rather improve the workflow structure. In addition pre-commit.ci is using `ghcr.io/pre-commit-ci/runner-image` as a prepared image which we could do as well if that helps.

pat-s:
> I guess the main point is you wont improve it if you don't use it. So comparing the two on a fair basis now shows us that we are ~2 seconds slower on clone and wp takes ~6s to check using prettier with a total of ~10s. Isn't that a starting point to rather improve the workflow structure. In addition pre-commit.ci is using ghcr.io/pre-commit-ci/runner-image as a prepared image which we could do as well if that helps.

Fair enough. I didn't see the waiting time for other tasks.
The checking time then is almost equal, and it's good to have a provider-agnostic solution.

The caching part might be hard to beat though anyway.

I still think it's good to add it as a hook, so prettier issues are caught before committing (for people which use it).

pat-s:
@anbraten good to add as a hook?

pat-s:
ping @anbraten See recent changes :)

.pre-commit-config.yaml
```diff
@@ -28,6 +28,10 @@ repos:
     rev: v2.12.1-beta
     hooks:
       - id: hadolint
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v3.0.3
```
6543, line 32:
```suggestion
    rev: v3.1.0
```

6543:
there is a change that produce different format between version `v3.0.3` and `v3.1.0` (witch we use now)

6543 (approved):

PR #2764, opened 2023-11-06
adds semver related environment vars

OCram85:
- adds additional environment variables in pipelines to get more inf a semver tag is used.


pat-s:
There was a discussion in the "develop" channel about this a few days ago.
My opinion was that its ok if we don't add another library for this. Given that we already use `semver` I am actually in favor of this.

I am not sure if I would want to use a plugin for this? And yes, one can of course do manual string manipulation of the git tag etc. but why not let it be automatically done by WP itself if the event is a tag? I don't see many downsides. The code change is small and env vars are cheap.

Another argument was that users coming from drone are used to these variables as they are present there by default: https://docs.drone.io/pipeline/environment/reference/drone-semver/.

pat-s:
@OCram85 Seems more people are against it than in favor.

OCram85:
That is absolutely fine with me. 😄 
Just didn't want to create another plugin to setup an *advanced* environment variables - If it's just for the semver ones.
So to solve my special use case I could:
- contribute to `plugins-gitea-release`  to auto detect it the new release should be marked as prerelease if the there was a semver string containing a prerelease.
- or just use another step which sets an `.env` file prior using the `plugins-gitea-release `.

anbraten:
Don't know about the gitea release plugin, but maybe I can add it to the woodpecker-plugin go  utils.

OCram85:
> Don't know about the gitea release plugin, but maybe I can add it to the woodpecker-plugin go utils.

@anbraten: Which repository do you mean? - Is it this one: https://codeberg.org/woodpecker-plugins/go-plugin ?

anbraten:
> Which repository do you mean? - Is it this one: https://codeberg.org/woodpecker-plugins/go-plugin ?

Yes I think that's the one. I am not sure if it is used atm. My idea would be to update it and use it in most of our plugins as source for the env-data and there we could then add the helper.

OCram85:
So I guess we can close this pr for now. - Because there are better ways to add and use additional environment variables than implementing them to the core codebase?

anbraten:
> So I guess we can close this pr for now. - Because there are better ways to add and use additional environment variables than implementing them to the core codebase?

I guess for now, yes. Sorry for the work, but we are happy for your interest in improving WP ❤️

PR #2765, opened 2023-11-06
Replace linter icons to match theme

xoxys:
@qwerty287 @anbraten What do you think about this approach? This way the icons fit more into the existing design. I've also moved the icon in the tab to the end as this looks IMO a bit better on mobile view, but open for discussions.

image
 

Alternative idea for errors:

Or just use one symbol and only change the color for errors/warnings?


xoxys:
PR is not finished and polished yet, want to get some feedback first.

xoxys:
Switched to one icon for now.

xoxys:
image
image

xoxys:
@anbraten What do you think?

anbraten:
I liked the triangle for warnings. The icons inside the circle are kinda bulky for the tabs. Was just trying if we might replace the circle icons in general. Icons need to be vertical centered I guess. For desktop I would place them in front again.

image

xoxys:
Ok will adjust it. Do you prefer to use different icons for warning and error? Or can we use the triangle in red and yellow?

xoxys:
I've changed it back to the triangle. The order of text and icon in the tab is now handled individually for mobile and desktop view.

 

image

@anbraten just let me know if you want to use two different icons.

web/src/components/atomic/Icon.vue
```diff
@@ -22,6 +22,8 @@
   <i-carbon-in-progress v-else-if="name === 'status-running' || name === 'status-started'" class="h-6 w-6" />
   <i-bi-slash-circle-fill v-else-if="name === 'status-skipped'" class="h-6 w-6" />
   <i-bi-check-circle-fill v-else-if="name === 'status-success'" class="h-6 w-6" />
+  <i-bi-exclamation-triangle-fill v-else-if="name === 'attention'" class="h-5 w-5" />
+  <i-bi-exclamation-triangle v-else-if="name === 'warning'" class="h-5 w-5" />
```
anbraten, line 26: What's the difference between those two icons?

anbraten (approved):

web/src/components/atomic/Icon.vue
xoxys: One is filled one is not.

The filled one is used for the linter errors page/tab, the other one replaces the emoji used in the warning hint.

xoxys: This one.

After:
image

Before:
image

I don't see a good way to use the same icon for both cases, but let me know if you would like to change something and I can do so.

lafriks:
Contrast for red color for dark theme is not very good, red should be lighter

xoxys:
That's the same color that is also used for all other red icons... And at least accessibility tools don't complain, so contrast can't be that bad, even if it might not be perfect. Feel free to improve the contrast of icons in a dedicated PR.

PR #2774, opened 2023-11-07
Fix release pipeline for 1.x.x

6543:


pat-s:
Needs a force merge bc we won't update 1.x to docker 24 and hence the sec scan will never pass. 
Same for pre-commit

PR #2777, opened 2023-11-08
Add `privileged` schema definition

pat-s:
I noticed that setting `privileged` in the pipeline config is triggering a warning due to a missing schema definition (even though its a valid and documented one).

image


pipeline/frontend/yaml/linter/schema/schema.json
```diff
@@ -212,6 +212,12 @@
         "image": {
           "$ref": "#/definitions/step_image"
         },
+        "privileged": {
+          "description": "Run the step in privileged mode. Read more: https://woodpecker-ci.org/docs/usage/pipeline-syntax#privileged",
+          "type": "boolean",
```
zc-devs, line 217: Here should be reference only.
Put your definition on line around 431, between `step_image` and `step_pull`.
Look at `step_image`, for example.

pipeline/frontend/yaml/linter/schema/schema.json
```diff
@@ -561,6 +567,12 @@
...
```
zc-devs, line 572: Here should be reference only.

zc-devs (approved):

pipeline/frontend/yaml/linter/schema/schema.json
```diff
@@ -561,6 +569,12 @@
...
+          "default": false,
```
6543, line 575:
```suggestion
```

6543 (approved):

PR #2778, opened 2023-11-08
Fix error color contrast in dark theme

lafriks:
* Also fixes small selected radio button inner circle to be centered correctly

Before: https://colourcontrast.cc/2d313d/9e3737

After: https://colourcontrast.cc/2d313d/ff6666

Before:

attels

attels

After:

attels

attels


anbraten (commented):
Looks a bit like piggy red 😅

anbraten (approved):

xoxys:
Looks like I have to switch to the light version again... I don't like it and it looks weird now :shrug:

xoxys:
We have introduced CSS custom attributes for exactly such cases, to simplify theming, I don't understand why we merge such changes that quickly. All colors were desaturated to the same level for the dark mode to create a unique look. Now a single one way brighter than all others and is during away your eyes.

lafriks:
The problem was that the red color was way too dark in the dark theme that it was unreadable

xoxys:
There would have been other ways instead... Also, that the error button is now the only button that has a black/dark text while all others have white/light text recreates another inconsistency that was removed already. However, I'm going to use a custom theme now.

lafriks:
I understand that colors are subjective but they must have accessibility contrast, dark theme really needs review colors, I agree this is not completely ideal but previously red color was completely unreadable for red elements, especially text but introducing multiple red shades for different elements would be even worse. Other option of course could be use previous red colors but than do not use red color for any text colors in red anywhere this could be other solution

PR #2786, opened 2023-11-09
Revert to previous red colors but do not use red color for text

lafriks:
Revert red colors back introduced by #2778 

But do not use red color for text on dark background anywhere

example:
attels


xoxys:
Thanks for looking into it again! I really like this approach. What do you think about using a template similar to the [warning](https://github.com/woodpecker-ci/woodpecker/blob/main/web/src/components/atomic/Warning.vue) one?

xoxys:
I like the idea to re-use the look of the Vue notifications, and we can also adjust the `warning` template to this looks, but I think creating an `error` template might make sense.

lafriks:
@xoxys done

web/src/components/layout/Container.vue
```diff
@@ -1,11 +1,12 @@
 <template>
-  <div class="w-full p-4 mx-auto" :class="{ 'max-w-5xl': !fullWidth, 'md:px-0': fullWidth }">
+  <div class="w-full p-4 mx-auto" :class="{ 'max-w-5xl': !fullWidth && !fillWidth, 'md:px-0': fullWidth }">
```
xoxys, line 2: Do we really want to use the full page width here? We don't need it and all other tabs like config, warnings, etc also dont have it.

xoxys (commented):
Besides that, looks good to me.

web/src/components/layout/Container.vue
lafriks: I have changed it only here as otherwise it will look
attels

vs 

attels

so when you click between steps it's jumping around that does not look consistent and jumpy

lafriks: after this change log view and error message has same size and when you click between steps it is same size at least in width

xoxys (approved):

PR #2789, opened 2023-11-09
Bump github.com/docker/docker v20.10.24 -> v24.0.7

6543:
address https://github.com/advisories/GHSA-jq35-85cj-fj4p

make sure the docker deamon, the agent connect to, is updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/caddyserver/certmagic v0.17.2
 	github.com/docker/cli v20.10.24+incompatible
 	github.com/docker/distribution v2.8.2+incompatible
-	github.com/docker/docker v20.10.24+incompatible
+	github.com/docker/docker v24.0.7+incompatible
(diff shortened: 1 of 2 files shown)


lafriks:
Docker Engine 20.10 will be EOL soon (if I remember correctly 10th December) so I think we should definitely update this but probably not for 1.x release as this could break some instances 🤷🏻‍♂️

lafriks:
I always prefer security over compatibility and as this would be released as v1.1.0 not v1.0.x I would say 👍🏻 for this 😉

pat-s:
This has been discussed among maintainers in the matrix channel and until now there was no clear conclusion on how to proceed. There are good reasons for both approaches, but the tl;dr is: it's more than "just" a security patch and would be breaking due to API incompatibility with older OSes and hence not really a clean minor/patch release.

6543:
well @anbraten did his final veto ... so v1.x.x is now unmaintained as of now

PR #2794, opened 2023-11-09
Kubernetes refactor

zc-devs:
Kubernetes backend refactoring and tests


pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -45,7 +44,10 @@ const (
 	EngineName = "kubernetes"
 )
 
-var noContext = context.Background()
+var (
+	noContext            = context.Background()
```
6543, line 48: I'd like to get rid of **noContext** !!!

pipeline/backend/kubernetes/volume.go
```diff
@@ -59,3 +62,41 @@ func PersistentVolumeClaim(namespace, name, storageClass, size string, storageRw
 
 	return pvc, nil
 }
+
+func VolumeName(name string) (string, error) {
```
6543, line 66: why export internal package functions?

pipeline/backend/kubernetes/kubernetes.go
zc-devs: Agree. Check if did it right.

go.mod
```diff
@@ -107,6 +107,7 @@ require (
 	github.com/imdario/mergo v0.3.12 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/kinbiko/jsonassert v1.1.1
```
6543, line 110: unrelated?

zc-devs: I added [Pod tests](https://github.com/woodpecker-ci/woodpecker/pull/2794/commits/4b77add7a276c5d6beac87d043ddbe5baff8a13d#diff-ceff1c2127ca61c02154522c696ceafe06aa75b46deb1b9e12903867bec25e8f) and it worked couple of times then.

Yesterday I fixed review notes and ran tests again. But Pod tests failed. It was because of `env` array: elements there appear in random order.

I found a [library](https://pkg.go.dev/github.com/kinbiko/jsonassert) to address this issue. [5413406](https://github.com/woodpecker-ci/woodpecker/pull/2794/commits/541340652336a6639e07701b6811e30e0c5ce9e3) dedicated to that. I included that library, use it and mark `env` array as `unordered`.

Also, I ran `go mod tidy` - that's why this library is in `go.mod` and some others are gone.

So, looks related to me, but maybe I miss something. Should I delete something?

zc-devs: You're right. I cleared `go.mod`, reran `go mod tidy` and dependency placed in the first `require` block. Thank you for suggestion.

zc-devs:
I've just resolved conflicts. @6543, could you review again?

6543 (approved):

PR #2796, opened 2023-11-10
Use labels for platform in example form Matrix workflows

miry:
I used example from https://woodpecker-ci.org/docs/usage/matrix-workflows
Then I see warning message from Woodpecker:

image


anbraten (approved):

pat-s (changes requested):
The PR changes the 1.0 docs which still use the old notation AFAIK? The change is already in the next documentation.
I think the user just didn't switch to the next documentation but is running next.

anbraten:
> The PR changes the 1.0 docs which still use the old notation AFAIK? The change is already in the next documentation.
I think the user just didn't switch to the next documentation but is running next.

Good catch

PR #2799, opened 2023-11-10
Update docs npm deps non-major

pat-s:
To fix security scan issues of `axios`. (renovate would have only scheduled this tomorrow or on Sunday). I force-triggered the creation.

Usually renovate would also create the PR but it was somehow stuck between automerge and missing branch tests (which is also why #2798 is good to have).


pat-s:
Ah we need a major bump :/

lafriks:
looks like related `Broken lockfile: no entry for '/wait-on/7.1.0(debug@4.3.4)' in pnpm-lock.yaml`

pat-s:
Yes, but we need to wait for https://github.com/jeffbski/wait-on/issues/143 anyhow. It should then resolve itself.

6543 (approved):

pat-s:
Why merge? The issue I links above is still open and actually required?

6543:
the usual bump worked - only the sec fix did not went in, I think that can be done in a dedicated pull

PR #2808, opened 2023-11-11
Fix repo owner filter

### LINKED ISSUE #2784: WOODPECKER_REPO_OWNERS is not working
### Component

server

### Describe the bug

I am looking for a way to limit which orgs can be activated in Woodpecker. According to the description in the documentation, `WOODPECKER_REPO_OWNERS` should do this, but it doesn't work. Even if `WOODPECKER_REPO_OWNERS=myorg` is set, I can still enable repos of `myorg` and `otherorg` in Woodpecker.

Looking at the code, it looks like the variable is no longer used:

```
❯ grep -ri "OwnersWhitelist"
server/model/settings.go:	OwnersWhitelist map[string]bool // Owners whitelist
server/router/middleware/config.go:		OwnersWhitelist: sliceToMap2(c.StringSlice("repo-owners")),
```

Based on the Git history, it looks like the code was removed in https://github.com/woodpecker-ci/woodpecker/pull/1658

```
❯ git log -SOwnersWhitelist
commit 0970f35df5034633a40b8d9f1a2ab901963aebee
Author: qwerty287 <80460567+qwerty287@users.noreply.github.com>
Date:   Tue Mar 21 23:01:59 2023 +0100

    Do not store inactive repos (#1658)
    
    Do not sync repos with forge if the repo is not necessary in DB.
    
    In the DB, only repos that were active once or repos that are currently
    active are stored. When trying to enable new repos, the repos list is
    fetched from the forge instead and displayed directly. In addition to
    this, the forge func `Perm` was removed and is now merged with `Repo`.
    
    Solves a TODO on RepoBatch.
    
    ---------
    
    Co-authored-by: Anbraten <anton@ju60.de>
```

Was it removed by accident or is there now another way to limit organizations in Woodpecker and `WOODPECKER_REPO_OWNERS` is just a leftover?

\cc @qwerty287 

### System Info

```shell
next
```

### Additional context

_No response_

### Validations

qwerty287:
and move to server config instead of middleware

cc @xoxys 

closes #2784

### DIFF
--- a/server/api/login.go
+++ b/server/api/login.go
@@ -75,7 +75,7 @@ func HandleAuth(c *gin.Context) {
 			return
 		}
 
-		// if self-registration is enabled for whitelisted organizations we need to
+		// if self-registration is enabled for allowed organizations we need to
 		// check the user's organization membership.
 		if server.Config.Permissions.Orgs.IsConfigured {
 			teams, terr := _forge.Teams(c, tmpuser)
@@ -125,7 +125,7 @@ func HandleAuth(c *gin.Context) {
 	u.Login = tmpuser.Login
 	u.Admin = u.Admin || server.Config.Permissions.Admins.IsAdmin(tmpuser)
 
-	// if self-registration is enabled for whitelisted organizations we need to
+	// if self-registration is enabled for allowed organizations we need to
 	// check the user's organization membership.
 	if server.Config.Permissions.Orgs.IsConfigured {
 		teams, terr := _forge.Teams(c, u)


server/plugins/permissions/util.go
```diff
@@ -0,0 +1,13 @@
+package permissions
+
+// sliceToMap is a helper function to convert a string slice to a map.
+func sliceToMap(s []string) map[string]bool {
```
6543, line 4: could you move and make it similar to
https://github.com/woodpecker-ci/woodpecker/blob/905bb9585371b3e86393d994f31983d936e24a91/shared/utils/slices.go#L53

?

xoxys (commented):
As we touch it anyway can we call it allowlist instead of whitelist?

xoxys (approved):
Thanks for the fix. I like the new permissions struct, makes the code more readable.

pat-s:
A partly unrelated thought: could we apply a similar approach to solve https://github.com/woodpecker-ci/woodpecker/issues/336?

Issue #1155, opened 2022-09-01
Allow to download pipline metadata via UI to replay with cli


anbraten:
I would be nice if I could download the metadata object via the UI as something like json to then use it to locally test with it (maybe adjusting some parts of it) via `cli exec`.

6543:
pipeline/frontend/metadata.go create the info from a pipeline (stored in db etc...)  and pipeline/frontend/metadata/* contains al types and is ready to be de-/serialized

PR #2812, opened 2023-11-12
Rename `link` to `url`

qwerty287:
As of https://woodpecker-ci.org/docs/next/usage/terminiology#conventions

This is breaking (changes api fields), so I'd like to get it into 2.0. Is this fine as we released an RC already?

### DIFF
--- a/.woodpecker/docs.yml
+++ b/.woodpecker/docs.yml
@@ -76,7 +76,7 @@ steps:
       - git clone --depth 1 --single-branch git@github.com:woodpecker-ci/woodpecker-ci.github.io.git /repo
       # update latest and next version
       - if [ "$CI_PIPELINE_EVENT" == "tag" ] ; then jq '.latest = ${CI_COMMIT_TAG}' /repo/version.json > /repo/version.json.tmp && mv /repo/version.json.tmp /repo/version.json ; fi
-      - if [ "$CI_PIPELINE_EVENT" == "push" ] ; then jq "next-" '.next = ${CI_COMMIT_SHA:0:10}' /repo/version.json > /repo/version.json.tmp && mv /repo/version.json.tmp /repo/version.json ; fi
+      - if [ "$CI_PIPELINE_EVENT" == "push" ] ; then jq '.next = "next-${CI_COMMIT_SHA:0:10}"' /repo/version.json > /repo/version.json.tmp && mv /repo/version.json.tmp /repo/version.json ; fi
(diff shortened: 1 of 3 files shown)


pipeline/frontend/metadata/types.go
```diff
@@ -17,101 +17,101 @@ package metadata
...
-		Secrets     []Secret `json:"secrets,omitempty"`
-		Branch      string   `json:"default_branch,omitempty"`
-		Trusted     bool     `json:"trusted,omitempty"`
+		ID          int64
+		Name        string
+		Owner       string
+		RemoteID    string
+		ForgeURL    string
+		CloneURL    string
+		CloneSSHURL string
```
6543, line 38: are you sure this do translate to `clone_url_ssh` ?

pipeline/frontend/metadata/types.go
```diff
@@ -17,101 +17,101 @@ package metadata
...
-		ChangedFiles      []string `json:"changed_files,omitempty"`
-		PullRequestLabels []string `json:"labels,omitempty"`
+		Sha               string
+		Ref               string
+		Refspec           string
+		Branch            string
+		Message           string
+		Author            Author
+		ChangedFiles      []string
+		PullRequestLabels []string
```
6543, line 71: does nor convert to labels

6543 (changes requested):

6543:
why did you remove all explicite tags for json serialisation ?!?

6543:
well they are used as we encode it send it over grps to the agent and decode it.

**for now** we don't use it elsewhere so it would not break anything beside agents -> you would have to +1 the grpc version

but I'd like to have this metadata downloadable to reproduce and run pipeline environments via local cli exec

6543:
> I don't think so. Can you point me to the `json.Marshal`/`json.Unmarshal` statements? If there are any, we should remove it. Why should the agents need the metadata? They only need the environment.


https://github.com/woodpecker-ci/woodpecker/blob/342b25826c93f76ec72ecf7d0d5b6e1e05485f50/server/grpc/server.go#L87
https://github.com/woodpecker-ci/woodpecker/blob/342b25826c93f76ec72ecf7d0d5b6e1e05485f50/pipeline/backend/types/config.go#L18
https://github.com/woodpecker-ci/woodpecker/blob/342b25826c93f76ec72ecf7d0d5b6e1e05485f50/pipeline/backend/types/step.go#L28

well yes in this case we dont have metadata anymore

anyway I want to use it at #1155  in the future

PR #2818, opened 2023-11-13
mark v1.0.x as unmaintained

6543:


pat-s (approved):
Nice feature, though we should wait until 2.0 is present as otherwise the logic doesn't make sense

image

6543:
well next is currently the only save option until be branch v2.0.0 out witch should happen relay soon

PR #2820, opened 2023-11-14
Revert to singular CLI args

qwerty287:
revert parts of https://github.com/woodpecker-ci/woodpecker/pull/2213


xoxys:
https://github.com/woodpecker-ci/woodpecker/blob/main/docs/docs/91-migrations.md?plain=1#L9 needs to be reverted as well

cli/secret/secret_set.go
```diff
@@ -46,11 +46,11 @@ var secretUpdateCmd = &cli.Command{
...
 		},
 		&cli.StringSliceFlag{
-			Name:  "events",
+			Name:  "event",
 			Usage: "secret limited to these events",
 		},
 		&cli.StringSliceFlag{
-			Name:  "images",
+			Name:  "image",
 			Usage: "secret limited to these images",
```
xoxys, line 54: Not related to this PR but the `Usage` seems to be a copy/paste one and is wrong.

xoxys: Ah no... My bad

6543 (approved):

anbraten:
Was there an issue with the renamings?

6543:
it was discussted at maintainers channel via chat

tldr: this means `--evens cron` `--events tags` instead of `--event cron` `--event tags`

we can always have alias if that is relay a problem

PR #2826, opened 2023-11-15
Signal to clients the hook and event routes where removed

### LINKED ISSUE #2824: 2.0.0: moving webhook endpoint behind `/api` prefix should probably be considered a breaking change
### Component

server

### Describe the bug

Hi, apologies if this has already been reported (I searched but couldn't find any issues), but I've recently updated my installation to 2.0.0-rc.0, and I was confused as to why webhooks didn't seem to work anymore. 

I went over the changelog for the RC, and found a reference to https://github.com/woodpecker-ci/woodpecker/pull/2212

This change breaks all webhooks created prior to the upgrade to 2.0.0-rc.0, as those still point at `${HOST}/hook`. And by leaving no route behind for the old hooks endpoint, it appears as though incoming the hook succeeds, as Woodpecker responds with a statuscode `200` (the body being the landing page).

It would've been helpful if:

1. The change had been marked as a breaking change
2. There had been some route left behind that responds with some kind of error statuscode (perhaps `410 Gone`?) so some kind of alerting would go off in whatever called the hook. Bonus points if this logs some warning in the woodpecker server logs!

I did eventually find https://github.com/woodpecker-ci/woodpecker/blob/main/docs/docs/91-migrations.md, which does describe what could be done. Still, it would have helped me personally if the changelog would have mentioned this more prominently.

### System Info

```shell
{
    "version": "2.0.0-rc.0"
}
```

### Additional context

_No response_

### Validations

6543:
Signal to clients the hook and event routes where removed at #2639

close #2824

### DIFF
--- a/server/router/api.go
+++ b/server/router/api.go
@@ -15,6 +15,8 @@
 package router
 
 import (
+	"net/http"
+
 	"github.com/gin-gonic/gin"
 	"github.com/rs/zerolog"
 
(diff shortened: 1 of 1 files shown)


6543:
lint error is unrelated!

Issue #2827, opened 2023-11-15
Server crashes and after a restart SQL lite db is locked


aozomaro:
### Component

server

### Describe the bug

During the execution of a pipeline, often times the server crashes. After crash, when the server is restarted the server is in a very slow state, often times reporting that "db is locked"

When the server crashes the following info is outputted to the logs:

goroutine 33930 [running]:
net/http.(http2responseWriter).write(0xc0017c6780?, 0xc00111fea8?, {0x0?, 0x12634a0?, 0x1362a80?}, {0x141f1bf?, 0xc000074d00?})
    /usr/local/go/src/net/http/h2_bundle.go:6797 +0x13f
net/http.(http2responseWriter).WriteString(0xc00111fee0?, {0x141f1bf?, 0xc000074d00?})
    /usr/local/go/src/net/http/h2_bundle.go:6790 +0x28
io.WriteString({0x7f428c48e2b8, 0xc000074d00}, {0x141f1bf, 0x18})
    /usr/local/go/src/io/io.go:316 +0x58
github.com/gin-gonic/gin.(*responseWriter).WriteString(0xc00154a000, {0x141f1bf, 0x18})
    /woodpecker/src/github.com/woodpecker-ci/woodpecker/vendor/github.com/gin-gonic/gin/response_writer.go:90 +0x73
io.WriteString({0x7f428c48e278, 0xc00154a000}, {0x141f1bf, 0x18})
    /usr/local/go/src/io/io.go:316 +0x58
go.woodpecker-ci.org/woodpecker/server/api.LogStreamSSE.func2()
    /woodpecker/src/github.com/woodpecker-ci/woodpecker/server/api/stream.go:224 +0x134
created by go.woodpecker-ci.org/woodpecker/server/api.LogStreamSSE in goroutine 33971
    /woodpecker/src/github.com/woodpecker-ci/woodpecker/server/api/stream.go:208 +0x7a5

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.0.0-rc.0"}
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #2830, opened 2023-11-16
Add option disable cron

6543:


6543:
... i'd like to have it as option too :)

aozomaro:
fyi, disabling the cron jobs proved to not be necessary to solve #2827. what really solved the problem was two things:
1) Using non sql lite db 
2) Adding a reverse proxy in front of woodpecker-server (this was the big one that really got me up and running)

pat-s:
@6543 What would be the motivation to disable cron? Just asking, as the code becomes more complex by these additions and I don't really see what the advantage is as cron is one of the core features of a CI. Security?

aozomaro:
> @6543 What would be the motivation to disable cron? Just asking, as the code becomes more complex by these additions and I don't really see what the advantage is as cron is one of the core features of a CI. Security?

Original problem was database being locked and receiving locking errors frequently which caused nothing to work (agents unable to report that they’ve closed the stream and as a result the server would panic when it sent a message after the stream had already closed)

I ended up solving the problem by:

* switching to Postgres (partially worked, server took longer to crash but crashed when the job queue had 4 pull requests running by checks)
* Adding a reverse proxy (this made the problem vanish)
* 
that being said we probably don’t need this anymore and just need to update docs/add disclaimers around how to run this in production mode

anbraten (changes requested):
Per comment

pat-s:
@aozomaro Thanks for the summary.

I'm a bit confused now: @anbraten Does your comment imply to close this PR? @6543 seems to view this PR as a feature despite not being needed anymore for the linked issue.

(No opinion from my side, just wanted to moderate this one initially to move forward)

anbraten:
Yes, I would close this PR as I don't see a serious reason to disable cron.

Issue #2748, opened 2023-11-05
Fix version check


qwerty287:
### Component

web-ui

### Describe the bug

@anbraten 

- [x] This does not work on light theme (fix: #2765):
    ![](https://github.com/woodpecker-ci/woodpecker/assets/80460567/df78e937-a2e9-44cd-b824-82b99872a0f0)
    Looks like the SVG has a hardcoded backend color which if `fff`
- [x] And if you're on next if there's an update it links to a not existing release page. (see https://ci.woodpecker-ci.org/admin#info), e.g. https://github.com/woodpecker-ci/woodpecker/releases/tag/next-fe48928 #2871 
- [ ] if our CI does not update the docs because of `path` filters the version.json file is not updated.
- [ ] if only `docs` wf runs the version.json is updated even though there's no new docker image
- [x] RC versions are never respected #2871 
    ![](https://github.com/woodpecker-ci/woodpecker/assets/80460567/dbc33bc8-20e4-43c2-bb8b-377e07debc40)

### System Info

```shell
after #2722
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

xoxys:
Please add an option to disable the check. It creatse noise in cases where deployments and updates are automated anyway.

xoxys:
I've added the icon color fix to https://github.com/woodpecker-ci/woodpecker/pull/2765

zc-devs:
> option to disable the check

On the other hand, as it's implemented on client side, we can "disable" it via Content Security Policy.

xoxys:
[...]

PR #2871, opened 2023-11-25
Fix version check partially

qwerty287:
ref #2748 

- fix link to releases
- fix jq syntax
- support rc versions (separate json field)


web/src/compositions/useVersion.ts
```diff
@@ -56,27 +59,31 @@ export function useVersion() {
...
   onMounted(async () => {
     const versionInfo = await fetchVersion();
 
-    let needsUpdate = false;
+    let latest;
     if (versionInfo) {
       if (usesNext) {
-        needsUpdate = versionInfo.next !== current;
+        latest = versionInfo.next;
+      } else if (current.includes('rc')) {
```
anbraten, line 74: I am checking using '-' or using the semver package most of the time as you could also release '1.2.3-beta.0'.

Maybe using the semver package is in general a nice idea as we could check that the upstream version is newer.

.woodpecker/docs.yml
```diff
@@ -75,7 +75,8 @@ steps:
       - git config --global user.name "woodpecker-bot"
       - git clone --depth 1 --single-branch git@github.com:woodpecker-ci/woodpecker-ci.github.io.git /repo
       # update latest and next version
-      - if [ "$CI_PIPELINE_EVENT" == "tag" ] ; then jq '.latest = ${CI_COMMIT_TAG}' /repo/version.json > /repo/version.json.tmp && mv /repo/version.json.tmp /repo/version.json ; fi
+      - if [[ "$CI_PIPELINE_EVENT" == "tag" && "${CI_COMMIT_TAG}" != *"rc"* ]] ; then jq '.latest = "${CI_COMMIT_TAG}"' /repo/version.json > /repo/version.json.tmp && mv /repo/version.json.tmp /repo/version.json ; fi
+      - if [[ "$CI_PIPELINE_EVENT" == "tag" ]] ; then jq '.rc = "${CI_COMMIT_TAG}"' /repo/version.json > /repo/version.json.tmp && mv /repo/version.json.tmp /repo/version.json ; fi
```
anbraten, line 79: So the rc channel will always be the latest release including rc and stable ones, right?

lafriks (approved):

web/src/compositions/useVersion.ts
anbraten: So you would ignore the check / case that the user is running a newer version than the upstream config is "aware" of for now?

.woodpecker/docs.yml
anbraten: Just checked again and the added `"` should solve the current ci issue as well:

```
+ if [ "$CI_PIPELINE_EVENT" == "push" ] ; then jq '.next = "next-0fc428aa8e"' /repo/version.json > /repo/version.json.tmp && mv /repo/version.json.tmp /repo/version.json ; fi
+ rsync -r --exclude .git --exclude CNAME --exclude index.yaml --exclude README.md --exclude version.json --delete docs/build/ /repo
jq: error: Invalid numeric literal at EOF at line 1, column 4 (while parsing '.0.0') at <top-level>, line 1:
.latest = v2.0.0            
jq: error: syntax error, unexpected LITERAL, expecting $end (Unix shell quoting issues?) at <top-level>, line 1:
.latest = v2.0.0            
jq: 2 compile errors
```

web/src/compositions/useVersion.ts
anbraten: fine

anbraten (approved):

Issue #2672, opened 2023-10-29
CI_PIPELINE_NUMBER is set as 0 


praveen-livspace:
### Component

server, agent

### Describe the bug

Whenever the pipelines are triggered using github PR or push events, the step pods have CI_PIPELINE_NUMBER set as "0". 

However on restarting the build, the env variable is correctly set.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-5b3bba726d"}
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

PR #2876, opened 2023-11-26
Fix pipeline-related environment

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/2672

pipeline model must be persisted first to have some fields like `Number` and `ID`

### DIFF
--- a/docs/docs/10-intro.md
+++ b/docs/docs/10-intro.md
@@ -11,8 +11,7 @@ If you are already using containers in your daily workflow, you'll for sure love
 - Pipeline steps can be named as you like
 - Run any command in the commands section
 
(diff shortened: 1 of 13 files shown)


server/pipeline/create.go
```diff
@@ -77,15 +84,6 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline
 		return nil, ErrFiltered
 	}
 
-	setGatedState(repo, pipeline)
-
-	err = _store.CreatePipeline(pipeline)
```
6543, line 82: did you look into if pipeline did got altered after creation - if so we have to update the db record here

Issue #2912, opened 2023-12-06
CLI doesn't show any user info when listing users


hrachovm:
### Component

cli

### Describe the bug

* login as `woodpecker` admin user (via gitea) :heavy_check_mark: 
* see and use API Access Token via swagger :heavy_check_mark: 
* connection of `cli` to server using the `woodpecker` user token works :heavy_check_mark: 
* getting list of users shows one single empty line :x: 
  * used command `docker compose run cli user ls`
* adding new user using cli successful :heavy_check_mark: 
  * used command `docker compose run cli user add joe`
    * output `Successfully added user joe`
  * creation of user verfied in swagger
  * user is able successfully loign into `woodpecker` (via gitea)
* getting list of users shows **two** empty lines now :x: 
  * used command `docker compose run cli user ls`
* deleting new user using cli successful :heavy_check_mark:
  * used command `docker compose run cli user rm joe`
    * output `Successfully removed user joe`
* getting list of users shows one empty line again :x: 
  * used command `docker compose run cli user ls`

My assumption is that communication with the API works fine, and the issue is only in the rendering output part.
I have tired to pass `--pretty` and `--log-level trace` arguments but without any effect.

### System Info

```shell
services:
  svc:
    image: woodpeckerci/woodpecker-server:v2.0.0
    environment:
    - WOODPECKER_ROOT_PATH=/woodpecker
    - WOODPECKER_ADMIN=woodpecker
    ...not relevant...
    networks:
      lan:

  cli:
    image: woodpeckerci/woodpecker-cli:v2.0.0
    env_file:
    - $DATA/cli.secrets
    environment:
    - WOODPECKER_SERVER=http://svc:8000/woodpecker/
    networks:
      lan:
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

PR #2914, opened 2023-12-07
Fix flags added multiple times

qwerty287:
Closes https://github.com/woodpecker-ci/woodpecker/issues/2912

Global flags should not be added to the commands/subscommands, only globally.


6543:
Well i remember that global commands did not propagate to subcommands ... but that is fixed in v3 - should we switch to it?

xoxys:
Do we really need this? I mean yes it might be convenient to be able to use `cli-bin subcommand --global-flag` but IMO it's also ok to be strict and enforce users to use the flags in the appropriate order.

Nevertheless, we might want to migrate to v3 anyway, if we are ok to switch to beta-releases this might be an option as well.

PR #2920, opened 2023-12-07
Fix schema validation with array syntax for clone and services

lafriks:
Currently it gives warnings even when pipeline syntax is supported and compiles/runs just fine


xoxys (commented):
LGTM in general.

Wasn't this removed/deprecated intentionally? I thought there were discussions somewhere in the past to not support both list and object-style syntax anymore, but I might be wrong

pipeline/frontend/yaml/linter/schema/.woodpecker/test-new-syntax.yml
```diff
@@ -0,0 +1,25 @@
...
+clone:
+  - name: git
+    image: woodpeckerci/plugin-git
+    settings:
+      partial: true
+  - name: testdata
+    image: woodpeckerci/plugin-git
+    settings:
+      remote: https://gitserver/owner/testdata.git
+      path: testdata
```
anbraten, line 12: Wasn't aware that you could use this for multiple clone plugins as well

pipeline/frontend/yaml/linter/schema/schema_test.go
```diff
@@ -101,6 +101,11 @@ func TestSchema(t *testing.T) {
 			testFile: ".woodpecker/test-broken.yml",
 			fail:     true,
 		},
+		{
+			name:     "New syntax",
+			testFile: ".woodpecker/test-new-syntax.yml",
+			fail:     false,
+		},
```
anbraten, line 108: I would suggest to call the file sth like `test-array-steps.yml` or `test-array-syntax.yml`.

anbraten (approved):

pipeline/frontend/yaml/linter/schema/schema_test.go
```diff
@@ -101,6 +101,11 @@ func TestSchema(t *testing.T) {
 			testFile: ".woodpecker/test-broken.yml",
 			fail:     true,
 		},
+		{
+			name:     "New syntax",
```
anbraten, line 105:
```suggestion
			name:     "Array syntax",
```

xoxys (approved):

pipeline/frontend/yaml/linter/schema/.woodpecker/test-new-syntax.yml
lafriks: yes we are using this quite much, especially for test data, integration tests and sometimes in place of submodules and they also have some drawbacks

Issue #2002, opened 2023-07-15
failed to parse pipeline: yaml: line 4: mapping values are not allowed in this context


XDream8:
Error:
![20230715_09h12m46s_grim](https://github.com/woodpecker-ci/woodpecker/assets/62709801/864c2eb7-4313-4cdd-b43f-91bb0e5452c4)

Pipeline:
```yaml
pipeline:
  test:
    group: test
    image: rust:${TAG}
    environment:
      - CARGO_TERM_COLOR=always
      - CARGO_BUILD_TARGET=${TARGET}
    commands:
      - rustup default stable
      - rustup target add ${TARGET}
      - rustup component add clippy
      - cargo test
      - cargo clippy

  build:
    group: build
    image: rust:${TAG}
    when:
      event: tag
    environment:
      - CARGO_TERM_COLOR=always
      - CARGO_BUILD_TARGET=${TARGET}
    commands:
      - rustup default stable
      - rustup target add ${TARGET}
      - cargo build --profile optimized
      - tar -zcvf file-rs-${TARGET}.tar.gz target/${TARGET}/optimized/file-rs

  publish:
    group: publish
    image: woodpeckerci/plugin-gitea-release
    when:
      event: tag
    settings:
      base_url: https://codeberg.org
      files:
        - "file-rs-${TARGET}.tar.gz"
      file-exists: overwrite
      api_key:
        from_secret: CODEBERG_ACCESS_TOKEN
      target: main
      title: file-rs
      skip_verify: true

matrix:
  include:
    - TARGET: x86_64-unknown-linux-musl
      TAG: alpine
    - TARGET: x86_64-unknown-linux-gnu
      TAG: latest
    - TARGET: aarch64-unknown-linux-musl
      TAG: alpine
    - TARGET: aarch64-unknown-linux-gnu
      TAG: latest
    - TARGET: x86_64-pc-windows-gnu
      TAG: latest
```

I followed the [documentation for matrix builds](https://woodpecker-ci.org/docs/usage/matrix-builds) but in the end this error came up.

When I use normal matrix combinations it works fine but when I try to use specific matrix combinations this happens.

6543:
@XDream8 what server vetsion?

XDream8:
it is codeberg ci so i suppose it is the latest version. i fixed it by using full image name `image: ${IMAGE}`

xoxys:
[...]

PR #2928, opened 2023-12-11
Fix skipped pipelines model (v2)

xoxys:
Alternative approach to https://github.com/woodpecker-ci/woodpecker/pull/2923

This approach is using the existing mechanism to update pipeline definitions used for [gated repos](https://github.com/woodpecker-ci/woodpecker/blob/fb7619a5c842fe7e4b93fb8f86c849c9049084a3/server/pipeline/approve.go#L53-L60).

To avoid executing it twice in case of gated repos, I've moved the logic after the `model.StatusBlocked` check. Calling `setPipelineStepsOnPipeline` was removed, as this is done by `createPipelineItems` already.

Of course this approach is also not perfect as multiple calls to `parsePipeline` also have a price, but IMO avoids some race conditions and extra handling required for adding + deleting pipelines from the DB.


lafriks (approved):

lafriks:
I think this is better approach than creating and deleting records in db

xoxys:
Thanks for your feedback. As I'm not sure about possible side effects I would like for a second review.

xoxys:
> Yes, this will probably cause some side effects.
> 
> 1. Pipelines are parsed twice. (Not sure if there's a way to go around this)

True but the exact code is executed twice on every single ci run for gated repos.


> 2. Empty pipeline env vars can cause yaml parsing issues like https://github.com/woodpecker-ci/woodpecker/issues/2002

Same here, if that is an issue the same issue exist for all gated repo jobs or am I missing something?

xoxys:
Ok so we should better take the other approach and delete empty pipelines from DB again?

xoxys:
Fully agree 🤣 If you would like to look into the entire issue on your own thats ok for me as well. We dont need to use one of my approaches.

xoxys:
Sure, fine for me as well.

Issue #2931, opened 2023-12-11
When trying to set several ports in k8s, they must be named


allanger:
### Component

agent

### Describe the bug

When you use kubernetes backend and need to have several ports exposed from a service, the pipeline is getting canceled (that is actually also not very obvious, because there is no error in UI, you can see it only when checking logs from the agent directly)

The error is caused by a requirement to have ports names if there are more than one. So this configuration won't work: 
```yaml
services:
  docker:
    image: docker:dind
    commands: dockerd -H tcp://0.0.0.0:2375 --tls=false
    privileged: true
    ports:
      - 2375
      - 16443
```

Error: 
```
{"level":"error","error":"rpc error: code = Unknown desc = Step finished with exit code 1, Service \"wp-01hhdbk671qq8d5hewnmp98f3e-0-services-0\" is invalid: [spec.ports[0].name: Required value, spec.ports[1].name: Required value]","time":"2023-12-11T21:12:46Z","message":"grpc error: wait(): code: Unknown: rpc error: code = Unknown desc = Step finished with exit code 1, Service \"wp-01hhdbk671qq8d5hewnmp98f3e-0-services-0\" is invalid: [spec.ports[0].name: Required value, spec.ports[1].name: Required value]"}
```

I think that the easiest solution would be to give random names to ports, since they are used by woodpecker only, so k8s users should actually not care about them. Then the pipeline syntax won't have to be updated, and it won't be a breaking change. If this solution sounds reasonable, I can implement it myself.

UPD: Or put a port number to the port name too, I think it's even easier, we just need to make sure it's a string

### System Info

```shell
I'm using Woodpecker 2.0.0 installed by the official helm chart.
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

PR #2933, opened 2023-12-12
feat(k8s): Add a port name to service definition

allanger:
It should cover this issue: https://github.com/woodpecker-ci/woodpecker/issues/2931

To sum up, when several ports need to be specified, they must be named

### DIFF
--- a/pipeline/backend/kubernetes/service.go
+++ b/pipeline/backend/kubernetes/service.go
@@ -16,6 +16,7 @@ package kubernetes
 
 import (
 	"fmt"
+
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
--- a/pipeline/backend/kubernetes/service.go
+++ b/pipeline/backend/kubernetes/service.go
@@ -15,6 +15,8 @@
 package kubernetes
 
 import (
+	"fmt"
+
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -24,6 +26,7 @@ func Service(namespace, name string, ports []uint16) (*v1.Service, error) {
 	var svcPorts []v1.ServicePort
 	for _, port := range ports {
 		svcPorts = append(svcPorts, v1.ServicePort{
+			Name:       fmt.Sprintf("port-%d", port),
 			Port:       int32(port),
 			TargetPort: intstr.IntOrString{IntVal: int32(port)},
 		})
--- a/pipeline/backend/kubernetes/service.go
+++ b/pipeline/backend/kubernetes/service.go
@@ -16,7 +16,6 @@ package kubernetes
 
 import (
 	"fmt"
-
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
--- a/pipeline/backend/kubernetes/service_test.go
+++ b/pipeline/backend/kubernetes/service_test.go
@@ -32,17 +32,17 @@ func TestService(t *testing.T) {
 	  "spec": {
 	    "ports": [
 	      {
-				  "name": "port-1",
+	        "name": "port-1",
 	        "port": 1,
 	        "targetPort": 1
 	      },
 	      {
-				  "name": "port-2",
+	        "name": "port-2",
 	        "port": 2,
 	        "targetPort": 2
 	      },
 	      {
-				  "name": "port-3",
+	        "name": "port-3",
 	        "port": 3,
 	        "targetPort": 3
 	      }
--- a/pipeline/backend/kubernetes/service_test.go
+++ b/pipeline/backend/kubernetes/service_test.go
@@ -32,14 +32,17 @@ func TestService(t *testing.T) {
 	  "spec": {
 	    "ports": [
 	      {
+				  "name": "port-1",
 	        "port": 1,
 	        "targetPort": 1
 	      },
 	      {
+				  "name": "port-2",
 	        "port": 2,
 	        "targetPort": 2
 	      },
 	      {
+				  "name": "port-3",
 	        "port": 3,
 	        "targetPort": 3
 	      }


pipeline/backend/kubernetes/service.go
```diff
@@ -24,6 +26,7 @@ func Service(namespace, name string, ports []uint16) (*v1.Service, error) {
 	var svcPorts []v1.ServicePort
 	for _, port := range ports {
 		svcPorts = append(svcPorts, v1.ServicePort{
+			Name:       strconv.FormatUint(uint64(port), 10),
```
lafriks, line 29: maybe we could name them `port-X` - `fmt.Sprintf("port-%d", port)`?

allanger: I think we could, but `strconv` seems a way faster, if my benchmark is correct 

```go
package main

import (
        "fmt"
        "time"
        "strconv"
)

func printf(num uint16) time.Duration {
        start := time.Now()
        _ = fmt.Sprintf("port-%d", num)
        return time.Since(start)
}

func convert(num uint16) time.Duration {
        start := time.Now()
        _ = strconv.FormatUint(uint64(num), 10)
        return time.Since(start)
}
func main(){
        var num uint16 = 1000

        fmt.Printf("%v - %v\n\n", convert(num).Nanoseconds(), printf(num).Nanoseconds())
        fmt.Printf("%v - %v\n\n", convert(num).Nanoseconds(), printf(num).Nanoseconds())
        fmt.Printf("%v - %v\n\n", convert(num).Nanoseconds(), printf(num).Nanoseconds())
        fmt.Printf("%v - %v\n\n", convert(num).Nanoseconds(), printf(num).Nanoseconds())
        fmt.Printf("%v - %v\n\n", convert(num).Nanoseconds(), printf(num).Nanoseconds())
        fmt.Printf("%v - %v\n\n", convert(num).Nanoseconds(), printf(num).Nanoseconds())
        fmt.Printf("%v - %v\n\n", convert(num).Nanoseconds(), printf(num).Nanoseconds())
}
```

```bash 
go run main.go

438 - 12970

157 - 327

96 - 174

92 - 161

63 - 141

64 - 152

46 - 167

```

Ofc it's just nanoseconds, but anyway, why making it slower

lafriks: Yes, of course specific convert function will be faster than generic format but this was more about readability just in case you need to debug it in k8s cluster etc so that's it's easy to identify that it's port name

allanger: Not sure how it's going to help, because in code it's under the `Name` element, and if you get a service from k8s, it's going to look like 
```yaml
apiVersion: v1
kind: Service
metadata:
  name: wp-01hhf980hdzs25ewwe2am6g3ad-0-services-0
  namespace: woodpecker-ci
spec:
  clusterIP: 10.152.183.105
  clusterIPs:
  - 10.152.183.105
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: "2375"
    port: 2375
    protocol: TCP
    targetPort: 2375
  - name: "16443"
    port: 16443
    protocol: TCP
    targetPort: 16443
  selector:
    step: wp-01hhf980hdzs25ewwe2am6g3ad-0-services-0
  sessionAffinity: None
  type: ClusterIP
```

So it's like pretty obvious that it's name and nothing else.

But if you'd like to have it with the `port` prefix anyway, I don't mind changing, just don't think it's making it more readable.

allanger: I think, nevermind, I've just updated to `sptrinf`

lafriks: At current point it probably does not matter but I have not tested how it does behave in SRV record nslookup etc when name is just numeric just in case we would use it at some point

allanger:
Would it be possible to have it in the next release?

pipeline/backend/kubernetes/service_test.go
```diff
@@ -32,14 +32,17 @@ func TestService(t *testing.T) {
 	  "spec": {
 	    "ports": [
 	      {
+				  "name": "port-1",
```
pat-s, line 35: Could you please align the indentation with the other items? I.e. using tab+whitespace in the same combination.

pipeline/backend/kubernetes/service.go
```diff
@@ -15,6 +15,8 @@
 package kubernetes
 
 import (
+	"fmt"
+
```
pat-s, line 19: Unnecessary newline

pipeline/backend/kubernetes/service_test.go
allanger: It should be fixed

pipeline/backend/kubernetes/service.go
allanger: It should be gone

allanger: I guess `gofumpt` wants this

```bash
$ cat pipeline/backend/kubernetes/service.go 
...
import (
        "fmt"
        v1 "k8s.io/api/core/v1"
        metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
        "k8s.io/apimachinery/pkg/util/intstr"
)
...

$ make lint
Running golangci-lint
golangci-lint run --timeout 15m
pipeline/backend/kubernetes/service.go:18: File is not `gofumpt`-ed with `-extra` (gofumpt)
        "fmt"
make: *** [Makefile:139: lint] Error 1

$ make format 
$ git diff
 import (
        "fmt"
+
        v1 "k8s.io/api/core/v1"
        metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
        "k8s.io/apimachinery/pkg/util/intstr"
```

pipeline/backend/kubernetes/service.go
```diff
@@ -15,6 +15,7 @@
 package kubernetes
 
 import (
+	"fmt"
```
lafriks, line 18:
```suggestion
	"fmt"

```

allanger: https://github.com/woodpecker-ci/woodpecker/pull/2933#discussion_r1428841966

lafriks: It must have newline between import groups std, local and external

allanger: It should look better now

pipeline/backend/kubernetes/service.go
allanger: https://github.com/woodpecker-ci/woodpecker/pull/2933#discussion_r1429210373

6543 (approved):

PR #2934, opened 2023-12-12
init CI_COMMIT_TAG if commit ref is a tag

gapanyc:
When triggering a deployment event on an existing pipeline, there is no way to get the tag used to trigger the parent pipeline, even if this parent was a tag event. 

In our company CI/CD current setup with drone, we use the tag event to trigger a kaniko image build step, using the git tag as an image tag, and the deployment/promotion step to effectively deploy this image using the tag reference in our cluster. This is the only point blocking us to completely switch to woodpecker and get rid of drone...

What's done:
- changed the metadata environ() method to populate CI_COMMIT_TAG env var if commit ref starts with /refs/tags (like it's done in drone), independently of event type EventTag.

Please let me know if I'm wrong, I will happily contribute in this nice project.

### DIFF
--- a/pipeline/frontend/metadata/environment.go
+++ b/pipeline/frontend/metadata/environment.go
@@ -119,7 +119,7 @@ func (m *Metadata) Environ() map[string]string {
 		// TODO Deprecated, remove in 3.x
 		"CI_COMMIT_URL": m.Curr.ForgeURL,
 	}
-	if m.Curr.Event == EventTag {
+	if strings.HasPrefix(m.Curr.Commit.Ref, "refs/tags/") {
 		params["CI_COMMIT_TAG"] = strings.TrimPrefix(m.Curr.Commit.Ref, "refs/tags/")
 	}
 	if m.Curr.Event == EventPull {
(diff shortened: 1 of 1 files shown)


pipeline/frontend/metadata/environment.go
```diff
@@ -119,7 +119,7 @@ func (m *Metadata) Environ() map[string]string {
 		// TODO Deprecated, remove in 3.x
 		"CI_COMMIT_URL": m.Curr.ForgeURL,
 	}
-	if m.Curr.Event == EventTag {
+	if strings.HasPrefix(m.Curr.Commit.Ref, "refs/tags/") {
```
lafriks, line 122: It would be better to keep both:
```suggestion
	if m.Curr.Event == EventTag || strings.HasPrefix(m.Curr.Commit.Ref, "refs/tags/") {
```

gapanyc: Thanks for the review. Did that at first, then tried to keep it simple, maybe a bit too much ;-)

lafriks: technically I don't think there could be a case when tag event would not have ref/tags/ prefix but better is to keep both just in case

lafriks (approved):

PR #2935, opened 2023-12-13
Export changed files via builtin environment variables

### LINKED ISSUE #853: Add changed files env var

There isn't a way to get the changed files of a commit/PR, which would be useful for static code checks so that they don't have to run on all files, only changed ones.

### Suggested solution

Add an environment variable that contains changed files, i.e. the list under the `Changed files` tab. To make it machine-readable, a json-style string could used. For example, [this build](https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/2300) would be:
```json
[".woodpecker/docs.yml"]
```

And an example with more files from a PR, https://ci.woodpecker-ci.org/woodpecker-ci/woodpecker/build/2303:

```json
["web/src/components/repo/build/BuildLog.vue","web/package.json","web/yarn.lock","web/src/components/repo/build/BuildProcList.vue","web/src/compositions/useBuildProc.ts","web/src/views/repo/build/Build.vue","web/src/views/repo/build/BuildWrapper.vue"]
```

### Alternative

I don't know if that's possible, but you could extract the build URL and then request the files from the api (if there's one).

### Additional context

I'm working on a plugin which checks files and I don't want this plugin to check all files of a repo, only changed ones.

### Validations

6543:
add **`CI_PIPELINE_FILES`** to builtin env vars

close  #853

---


anbraten (changes requested):
I wouldn't do this as the string can easily be extremely long. Getting the changed files can if needed be done using git already.

6543:
I'm for adding it.

first:
1. long env variables dont create any issue
2. on bigger projects this is a question of efficience (if checked via git in pipeline, it would e.g. add 9min just to checkout the whole code to get the changed files)
3. we already have the info in metadata and database anyway

6543:
if long env var are a concern they posibly can only make issues on windows, I can add a check for this case if desired

anbraten (commented):
Still feels wrong to possibly have those super long env vars, but as it seems to be supported its fine to me.

lafriks:
I would rather see this implemented differently than env variable, maybe having a token available for some special api call to the woodpecker server from the pipeline that would allow it to retrieve such data?

6543:
So basically: https://github.com/woodpecker-ci/woodpecker/discussions/2291#discussioncomment-7806713 ?

6543:
->  #2954

confusedsushi:
For the user it is so convenient to just have that information. Of course he could get this information via git, or maybe later via some sort of API call. But shouldn't you make it for the user as easy as possible? You already have that information, why not forward it to the user? If the user need to bother with some git command, some HTTP API call or some other external tool he will just make mistakes.

xoxys:
I'm also not sure about the current way of implementation. Why not use the approach from the linked issue and let e.g. plugin-git create a file that contains all changed files, one per line? The current JSON string can't be easily parsed by bash/shell.

6543:
well first, we do shallow clone by default.
so recover the information the server already has is just **realy** inefficient.

just to point out some numbers: big repos will rise clone time from ~3min to 12min just to get that info

6543:
and json can be parsed by [`jq`](https://repology.org/project/jq/versions) or `yq` witch are widely available. powershell (windows) support it natively.

xoxys:
Ah yeah that make sense. Missed this point.

6543:
> I think this fixes #853?

yes it does :) - updated pull description

xoxys (approved):
Code LGTM.

PR #2936, opened 2023-12-13
RegisterAgent with hostname

runephilosof-karnovgroup:
Transfer the agent hostname to the server's model.agent.Name field.

Previously the autoscaler had to create the agent with the server and get a unique token to pass to the agent at boot up. This allows to get the right name for the agent while using the shared master agent token at agent boot up.
Both ways are still supported.


runephilosof-karnovgroup:
I have no secure way to transfer a unique token to the instances I create in my autoscaling setup on Google Cloud.
My autoscaler then has no way to identify which woodpecker agents have which agent ids.
So this would solve that problem by saving the mapping in the woodpecker server.

runephilosof-karnovgroup:
It only updates the hostname if it is already empty.

server/grpc/rpc_test.go
```diff
@@ -0,0 +1,108 @@
...
+				LastContact: 0,
+				Platform:    "platform",
+				Backend:     "backend",
+				Capacity:    2,
+				Version:     "version",
+				NoSchedule:  false,
+			}
+
+			store.On("AgentFind", int64(1337)).Once().Return(storeAgent, nil)
+			store.On("AgentUpdate", &updatedAgent).Once().Return(nil)
```
runephilosof-karnovgroup, line 51: I wanted to do something like
```suggestion
			var updatedAgent *model.Agent

			store.On("AgentFind", int64(1337)).Once().Return(storeAgent, nil)
			store.On("AgentUpdate", mock.AnythingOfType("*model.Agent")).Once().Return(func(agent *model.Agent) {
				updatedAgent = agent
			})
```
And then in the end of the test, check 
```
			assert.EqualValues(t, "hostname", updatedAgent.Name)
			assert.EqualValues(t, "platform", updatedAgent.Platform)
			assert.EqualValues(t, "backend", updatedAgent.Backend)
			assert.EqualValues(t, "version", updatedAgent.Version)
```
But I could not get that to work.
It says 
```
panic: assert: arguments: Error(0) failed because object wasn't correct type: 0x906f80
```

runephilosof-karnovgroup: It is not the `mock.AnythingOfType("*model.Agent")`. I get the same error if I use `mock.Anything`.
Also, for some reason, the same error if I use
```go
			var updatedAgent model.Agent

			store.On("AgentFind", int64(1337)).Once().Return(storeAgent, nil)
			store.On("AgentUpdate", mock.Anything).Once().Return(func(agent model.Agent) {
				updatedAgent = agent
			})
```

6543:
PS: thanks for adding tests :heart:

server/grpc/rpc_test.go
runephilosof-karnovgroup: I am not going to use more effort on improving the tests

runephilosof-karnovgroup:
Do I need to do anything extra to get this merged?

6543 (approved):

PR #2939, opened 2023-12-13
Allow to restart pipelines that has warnings

### LINKED ISSUE #2753: Restarting pipelines with errors is returning 500
If I try to restart a pipeline with errors I am getting a 500

### LINKED ISSUE #2921: Failure to create pipeline using API (plugin-trigger) if definition has lint warnings
### Component

server

### Describe the bug

When using plugin-trigger to trigger builds for other pipeline they will fail with error:

```
2023-12-07T10:59:20Z ERR failure to createPipelineItems for ***/*** | error=[linter] Invalid type. Expected: object, given: array; [linter] Invalid type. Expected: object, given: array
2023-12-07T10:59:20Z ERR error=Error #01: failure to createPipelineItems for ***/***
 ip=10.255.0.2 latency=34.869784 method=POST path=/api/repos/1243/pipelines/1221 status=500 user-agent=Go-http-client/1.1
 ```
 
Pipeline itself will be created but won't be and can not be started because of this as it does not contain any steps or definition

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.0.0"}
```

### Additional context

_No response_

### Validations

lafriks:
Fixes #2753, fixes  #2921

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -42,6 +42,7 @@
       "runtimeArgs": ["start"],
       "cwd": "${workspaceFolder}/web",
       "resolveSourceMapLocations": ["${workspaceFolder}/web/**", "!**/node_modules/**"],
+      "envFile": "${workspaceFolder}/.env",
       "skipFiles": ["<node_internals>/**"]
     }
   ]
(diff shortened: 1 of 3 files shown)


lafriks:
> Lgtm, but untested. I assume you did?

yes, I did test it

lafriks:
> And I assume `createPipelineItems` is not used anywhere else except in approval (or all other usages are just like their for approval)?

It's used only in approve and restart but returned error was handled differently. In approve it was handled correctly but in restart it failed even it should not have as error was just pipeline parsing warning - so I moved that logic into createPipelineItems so that no such errors could happen in future this function usage

PR #2943, opened 2023-12-13
Remove separate root path config

qwerty287:
You must add the path to `WOODPECKER_HOST` anyways, so it doesn't make a difference.

This *could* be marked as breaking but since it won't change something for systems that are configured to work, I don't think that's necessary.

### DIFF
--- a/docs/docs/91-migrations.md
+++ b/docs/docs/91-migrations.md
@@ -4,7 +4,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 
 ## `next`
 
-- Remove `WOODPECKER_ROOT_PATH` and `WOODPECKER_ROOT_URL` config variables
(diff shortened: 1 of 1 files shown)


docs/docs/91-migrations.md
```diff
@@ -4,7 +4,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 
 ## `next`
 
-No breaking changes yet
+- Remove `WOODPECKER_ROOT_PATH` and `WOODPECKER_ROOT_URL` config variables
```
anbraten, line 7:
```suggestion
- Removed `WOODPECKER_ROOT_PATH` and `WOODPECKER_ROOT_URL` config variables. Use `WOODPECKER_HOST` with a path instead
```

anbraten (approved):

Issue #2937, opened 2023-12-13
When a pipeline creation is failed on agent side, steps are marcked as canceled without error messages


allanger:
### Component

agent

### Describe the bug

NOTE: I've noticed it only on the k8s backend since I'm not using others. 

I had this issue with the pipelines: https://github.com/woodpecker-ci/woodpecker/issues/2931

But when I was trying to trigger a workflow, all the steps are just marked as canceled: you can see it here: https://ci.badhouseplants.net/repos/2/pipeline/10

To understand what the problem is, I had to go to logs of the agent which should have executed this job and find the reason there. 

It would be a way nicer to have the error in the UI.

### System Info

```shell
The problem was spotted on the Woodpecker 2.0.0, installed with the official helm chart
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

allanger:
I think I could try taking care of it

PR #2945, opened 2023-12-14
Display runners errors in UI

allanger:
It should resolve this issue: #2937 

I've only tested on k8s backend, and here you can see what it looks like in UI: https://ci.badhouseplants.net/repos/2/pipeline/55 

I'm not sure about change here: `server/pipeline/cancel.go`. So I still want to figure out what I've actually done, but I guess maybe @qwerty287, could you please look at the result and say what you think already?

### DIFF
--- a/server/grpc/rpc.go
+++ b/server/grpc/rpc.go
@@ -274,10 +274,13 @@ func (s *RPC) Done(c context.Context, id string, state rpc.State) error {
 	}
 	s.completeChildrenIfParentCompleted(workflow)
 
+			if currentPipeline, err = pipeline.UpdateToStatusKilled(s.store, *currentPipeline, errors.New(state.Error)); err != nil {
+				logger.Error().Err(err).Msgf("pipeline.UpdateStatusToDone: cannot update workflow final state")
+			}
 	if !model.IsThereRunningStage(currentPipeline.Workflows) {
-		if currentPipeline, err = pipeline.UpdateStatusToDone(s.store, *currentPipeline, model.PipelineStatus(currentPipeline.Workflows), workflow.Stopped); err != nil {
-			logger.Error().Err(err).Msgf("pipeline.UpdateStatusToDone: cannot update workflow final state")
-		}
+			if currentPipeline, err = pipeline.UpdateStatusToDone(s.store, *currentPipeline, model.PipelineStatus(currentPipeline.Workflows), workflow.Stopped); err != nil {
+				logger.Error().Err(err).Msgf("pipeline.UpdateStatusToDone: cannot update workflow final state")
+			}
 	}
 
 	s.updateForgeStatus(c, repo, currentPipeline, workflow)
--- a/server/pipeline/cancel.go
+++ b/server/pipeline/cancel.go
@@ -82,7 +82,7 @@ func Cancel(ctx context.Context, store store.Store, repo *model.Repo, user *mode
 		}
 	}
 
-	killedPipeline, err := UpdateToStatusKilled(store, *pipeline)
+	killedPipeline, err := UpdateToStatusKilled(store, *pipeline, err)
 	if err != nil {
 		log.Error().Err(err).Msgf("UpdateToStatusKilled: %v", pipeline)
 		return err
--- a/server/pipeline/pipelineStatus.go
+++ b/server/pipeline/pipelineStatus.go
@@ -56,7 +56,8 @@ func UpdateToStatusError(store model.UpdatePipelineStore, pipeline model.Pipelin
 	return &pipeline, store.UpdatePipeline(&pipeline)
 }
 
-func UpdateToStatusKilled(store model.UpdatePipelineStore, pipeline model.Pipeline) (*model.Pipeline, error) {
+func UpdateToStatusKilled(store model.UpdatePipelineStore, pipeline model.Pipeline, err error) (*model.Pipeline, error) {
+	pipeline.Errors = errors.GetPipelineErrors(err)
 	pipeline.Status = model.StatusKilled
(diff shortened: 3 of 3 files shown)


server/grpc/rpc.go
```diff
@@ -274,10 +274,13 @@ func (s *RPC) Done(c context.Context, id string, state rpc.State) error {
 	}
 	s.completeChildrenIfParentCompleted(workflow)
 
+			if currentPipeline, err = pipeline.UpdateToStatusKilled(s.store, *currentPipeline, errors.New(state.Error)); err != nil {
+				logger.Error().Err(err).Msgf("pipeline.UpdateStatusToDone: cannot update workflow final state")
+			}
```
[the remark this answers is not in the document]
allanger: Missing if, I've removed it while trying to understand if UI job status in UI not getting updated without reloading a page  is frontend related or not, but then never had it back. I'll fix

server/pipeline/cancel.go
```diff
@@ -82,7 +82,7 @@ func Cancel(ctx context.Context, store store.Store, repo *model.Repo, user *mode
 		}
 	}
 
-	killedPipeline, err := UpdateToStatusKilled(store, *pipeline)
+	killedPipeline, err := UpdateToStatusKilled(store, *pipeline, err)
```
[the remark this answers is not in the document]
allanger: Yep, thanks, I'll fix. Was just making compiler compile again

server/pipeline/pipelineStatus.go
```diff
@@ -56,7 +56,8 @@ func UpdateToStatusError(store model.UpdatePipelineStore, pipeline model.Pipelin
 	return &pipeline, store.UpdatePipeline(&pipeline)
 }
 
-func UpdateToStatusKilled(store model.UpdatePipelineStore, pipeline model.Pipeline) (*model.Pipeline, error) {
+func UpdateToStatusKilled(store model.UpdatePipelineStore, pipeline model.Pipeline, err error) (*model.Pipeline, error) {
+	pipeline.Errors = errors.GetPipelineErrors(err)
```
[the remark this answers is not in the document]
allanger: it should, there is a test for that case, isn't it?
```go
func TestGetPipelineErrors(t *testing.T) {
	t.Parallel()

	tests := []struct {
		title    string
		err      error
		expected []*pipeline_errors.PipelineError
	}{
		{
			title:    "nil error",
			err:      nil,
			expected: nil,
		},
                ...
	}

	for _, test := range tests {
		assert.Equalf(t, pipeline_errors.GetPipelineErrors(test.err), test.expected, test.title)
	}
}


```

server/pipeline/cancel.go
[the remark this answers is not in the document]
allanger: It should be fixed

server/grpc/rpc.go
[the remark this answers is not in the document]
allanger: It should be fixed

allanger:
Some pipelines as example of how it's working now:

- Green one: https://ci.badhouseplants.net/repos/3/pipeline/12
- Job is failed: https://ci.badhouseplants.net/repos/3/pipeline/11
- Scheduling is failed: https://ci.badhouseplants.net/repos/2/pipeline/61

They all should be public, if they're not, let me know

server/grpc/rpc.go
```diff
@@ -273,10 +273,15 @@ func (s *RPC) Done(c context.Context, id string, state rpc.State) error {
 		return err
 	}
 	s.completeChildrenIfParentCompleted(workflow)
-
 	if !model.IsThereRunningStage(currentPipeline.Workflows) {
-		if currentPipeline, err = pipeline.UpdateStatusToDone(s.store, *currentPipeline, model.PipelineStatus(currentPipeline.Workflows), workflow.Stopped); err != nil {
-			logger.Error().Err(err).Msgf("pipeline.UpdateStatusToDone: cannot update workflow final state")
+		if len(state.Error) > 0 {
```
anbraten, line 277: Shouldn't the error be set on the workflow level?

allanger:
> @allanger please check linter https://ci.woodpecker-ci.org/repos/3780/pipeline/10530/25
I hope it's fixed now

server/grpc/rpc.go
allanger: Maybe it should. But for example, when I have a syntax error in of workflows, the whole pipeline is failed. 

Though when one of workflows in a pipeline is killed with an error, the error is not displayed, I'm not sure why. Example: https://ci.badhouseplants.net/repos/2/pipeline/67

And is this kind of errors already supported for workflows? Cause it didn't seem so to me, but I'm still trying to understand how everything is implemented, so I most probably might be wrong

allanger: Alright, I'll move it there then

allanger: Hm, it seems that the error is already there. Just not displayed in UI at all

allanger: I'm running the latest upstream image here 
```bash
$ curl -Q -i ${WOODPECKER_SERVER}/api/repos/2/pipelines/74 -H "Authorization: Bearer ${WOODPECKER_TOKEN}" | jq '.workflows.[] | select(.pid==2) | .error'
"Service \"wp-01hhrxhkcw3zp5j3m38f8epb8v-0-services-0\" is invalid: [spec.ports[0].name: Required value, spec.ports[1].name: Required value]"
```

allanger: I've got a feeling that to get it done right, the pipeline page should be redesigned. Because one can't really interact with workflows now. What do you think about that?

PR #2953, opened 2023-12-15
Draft: Agent pipeline callback api

6543:
Current STATE: this pull is more of an idea just wrote down ... anybody can start fresh ... still till no issue exist, I'll rather have this open

## Reason

If we have an propper api that can be gated by on demand random generated tokens.
Each Step can have a back hannel to woodpecker that can be propper serilaized, gate keepted, monitored and secured.
We efen could allow plugins to show addiginal UI elements in the server this way as long as we make it sainitizable e.g. no XSS issues etc...

That way we could for example let one step change/set settings for the next plugin etc... and still let the plugin decide it it wana accept that ...

in general we need more reasoning how such api should look like, what state is stored where.

with in mind:
 - not to create side channels that can be atacked
 - be secure
 - be reproducable with same input...
 - backend transparent

## TODOs:
- [ ] create agent callback server
- [ ] allow access token as auth method for server api to change the pipeline specific resources
- [ ] ...

### DIFF
--- a/pipeline/backend/local/local.go
+++ b/pipeline/backend/local/local.go
@@ -40,6 +40,7 @@ type workflowState struct {
 	homeDir         string
 	workspaceDir    string
 	pluginGitBinary string
+	accessToken     string
 }
 
 type local struct {
@@ -48,6 +49,7 @@ type local struct {
 	output          io.ReadCloser
 	pluginGitBinary string
 	os, arch        string
+	callbackServer  string
 }
 
 // New returns a new local Backend.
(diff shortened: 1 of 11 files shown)


pipeline/backend/local/local.go
```diff
@@ -126,6 +135,9 @@ func (e *local) StartStep(ctx context.Context, step *types.Step, taskUUID string
 		}
 	}
 
+	env = append(env, "CI_PIPELINE_CALLBACK_URL="+e.callbackServer)
```
lafriks, line 138: Maybe would be better to call it `CI_PIPELINE_API_URL`?

pat-s:
@6543 This has many conflicts now, are you planning to continue here?

6543:
I still think if we dont want to parse all stuff through stdout ... we need some callback functionality

PR #2956, opened 2023-12-16
Update README badges

qwerty287:
- use repo id
- remove tickgit (the service is broken)
- use dynamic pre-commit.ci badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@
 </p>
 <br/>
 <p align="center">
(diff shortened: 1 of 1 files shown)


anbraten (approved):

xoxys:
Is it planned to remove the ability to use the repo name on urls? Personally I would prefer to us the name because its readable for humans. That we use ids internally makes sense but for urls thats it creates alreay some ux issues e.g. if a user wants go to bookmark an url just from reading the url you never know the repo.

PR #2959, opened 2023-12-17
Fix broken gated repos

xoxys:
Fixes a bug introduced in https://github.com/woodpecker-ci/woodpecker/pull/2923. I'll also try to add a test case.


xoxys:
My attempts to test the `Create` method were not successful.... Looks like a lot of mocking is required, and I'm not able to put it together properly...

lafriks (approved):

PR #2960, opened 2023-12-17
Enable more go linters [ready to split]

xoxys:
Supersedes: https://github.com/woodpecker-ci/woodpecker/pull/977


xoxys:
I'm a bit confused about how pre-commit is handling `golangci-lint`. While `make lint` still returns a lot of errors (PR not ready yet), pre-commit seems to be happy. Any idea @pat-s?

xoxys:
Some remaining issues where I would like to get some input:

1) Why is `!= 2` correct here? Would like to document what `2` means and why it is the intended value.
```
server/forge/github/github.go:493:22: mnd: Magic number: 2, in <condition> detected (gomnd)
		if len(matches) != 2 {
		                   ^
```

2) Why 8 bytes? Would like to document why it is the intended value.
```
pipeline/multipart/reader.go:55:21: mnd: Magic number: 8, in <argument> detected (gomnd)
	out, _ := buf.Peek(8)
	                   ^
```

3) Why is `< 2` correct here? Would like to document what `2` means and why it is the intended value.
```
pipeline/backend/docker/convert.go:125:19: mnd: Magic number: 2, in <condition> detected (gomnd)
		if len(parts) < 2 {
		                ^
```

```
pipeline/backend/docker/convert.go:152:19: mnd: Magic number: 2, in <condition> detected (gomnd)
		if len(parts) < 2 {
		                ^
```

4) If the goal is to get seconds from the given duration, would it be better to use `time.Duration(cf.timeout.Seconds())`?
```
server/forge/configFetcher.go:64:30: Multiplication of durations: `time.Second * cf.timeout` (durationcheck)
		files, err = cf.fetch(ctx, time.Second*cf.timeout, strings.TrimSpace(cf.configPath))
		                           ^
```

5) Why is `<= 3` correct here? Would like to document what `3` means and why it is the intended value.
```
pipeline/shared/replace_secrets.go:23:18: mnd: Magic number: 3, in <condition> detected (gomnd)
		if len(old) <= 3 {
		               ^
```

6) Is it acceptable to not exec the defer function in this case, or do we need to refactor this part?
```
cmd/server/server.go:154:4: exitAfterDefer: log.Fatal will exit, and `defer func(){...}(...)` will not run (gocritic)
			log.Fatal().Err(err).Msg("failed to create web engine")
```

shared/httputil/httputil.go
```diff
@@ -40,14 +40,17 @@ func IsHTTPS(r *http.Request) bool {
 
 // SetCookie writes the cookie value.
 func SetCookie(w http.ResponseWriter, r *http.Request, name, value string) {
+	// the cookie value (token) is responsible for expiration
+	cookieMaxAgeSeconds := 2147483647
```
xoxys, line 44: Why was `2147483647` seconds chosen here?

xoxys: Ah, I guess we want to "never expire" here and `2147483647` is the 32-bit integer limit.

cli/internal/util.go
```diff
@@ -108,7 +108,7 @@ func ParseKeyPair(p []string) map[string]string {
 	params := map[string]string{}
 	for _, i := range p {
 		parts := strings.SplitN(i, "=", 2)
-		if len(parts) != 2 {
+		if len(parts) != 2 { //nolint:gomnd
```
lafriks, line 111: all strings.SplitN(..., 2) should be refactored to use strings.Cut

xoxys: Done

agent/runner.go
```diff
@@ -29,6 +29,7 @@ import (
 	"go.woodpecker-ci.org/woodpecker/v2/pipeline"
 	backend "go.woodpecker-ci.org/woodpecker/v2/pipeline/backend/types"
 	"go.woodpecker-ci.org/woodpecker/v2/pipeline/rpc"
+	"go.woodpecker-ci.org/woodpecker/v2/server/model"
```
6543, line 32: agent should not import server

int looks like  model.ExitCodeKilled should be moved into `pipeline/*` package somewhere

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -222,7 +216,7 @@ func (e *kube) WaitStep(ctx context.Context, step *types.Step, taskUUID string)
 	finished := make(chan bool)
 
 	podUpdated := func(old, new any) {
-		pod := new.(*v1.Pod)
+		pod, _ := new.(*v1.Pod)
```
6543, line 219: well in this case we still need error handling

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -282,7 +276,7 @@ func (e *kube) TailStep(ctx context.Context, step *types.Step, taskUUID string)
 	up := make(chan bool)
 
 	podUpdated := func(old, new any) {
-		pod := new.(*v1.Pod)
+		pod, _ := new.(*v1.Pod)
```
6543, line 279: same

pipeline/backend/local/local.go
```diff
@@ -257,7 +257,9 @@ func (e *local) getState(taskUUID string) (*workflowState, error) {
 	if !ok {
 		return nil, ErrWorkflowStateNotFound
 	}
-	return state.(*workflowState), nil
+
+	workflowState, _ := state.(*workflowState)
```
6543, line 261: if we ignore the ok here we should add a comment that we always only expect this to work

pipeline/frontend/yaml/constraint/constraint.go
```diff
@@ -195,7 +195,8 @@ func (c *Constraint) Match(m metadata.Metadata, global bool, env map[string]stri
 		if err != nil {
 			return false, err
 		}
-		match = match && result.(bool)
+		bresult, _ := result.(bool)
```
6543, line 198: can result be something beside bool?

pipeline/stepBuilder.go
```diff
@@ -26,17 +26,15 @@ import (
 
 	backend_types "go.woodpecker-ci.org/woodpecker/v2/pipeline/backend/types"
 	"go.woodpecker-ci.org/woodpecker/v2/pipeline/errors"
-	pipeline_errors "go.woodpecker-ci.org/woodpecker/v2/pipeline/errors"
```
6543, line 29: could we use pipeline_errors as there is the std package `errors` witch can be used here in the future?

pipeline/tracer.go
```diff
@@ -17,6 +17,8 @@ package pipeline
 import (
 	"strconv"
 	"time"
+
+	"go.woodpecker-ci.org/woodpecker/v2/server/model"
```
6543, line 21: pipeline should have it's own enums if i remember correctly

in any case we dont want a circular dependency

pipeline/backend/backend.go
```diff
@@ -29,11 +29,11 @@ var (
 	backends       []types.Backend
 )
 
-func Init(ctx context.Context) {
+func Init() {
```
6543, line 32: can we move this change into it's own pull and merge it first?

xoxys:
Thanks! Will take care.

agent/runner.go
xoxys: Done

pipeline/backend/backend.go
xoxys: I've reverted this change. Not even sure why it ended in this PR, as at least the linter doesn't complain.

pipeline/tracer.go
xoxys: Makes sense, but how should this case be handled? Copy the status consts from the servers model package to the pipeline package?

pipeline/stepBuilder.go
xoxys: Done.

pipeline/backend/kubernetes/kubernetes.go
xoxys: Done

pipeline/backend/kubernetes/kubernetes.go
xoxys: Done

pipeline/stepBuilder.go
```diff
@@ -25,18 +25,16 @@ import (
...
-
 	"go.woodpecker-ci.org/woodpecker/v2/pipeline/frontend"
 	"go.woodpecker-ci.org/woodpecker/v2/pipeline/frontend/metadata"
 	"go.woodpecker-ci.org/woodpecker/v2/pipeline/frontend/yaml"
 	"go.woodpecker-ci.org/woodpecker/v2/pipeline/frontend/yaml/compiler"
 	"go.woodpecker-ci.org/woodpecker/v2/pipeline/frontend/yaml/linter"
 	"go.woodpecker-ci.org/woodpecker/v2/pipeline/frontend/yaml/matrix"
+	yaml_types "go.woodpecker-ci.org/woodpecker/v2/pipeline/frontend/yaml/types"
 	"go.woodpecker-ci.org/woodpecker/v2/server"
+	forge_types "go.woodpecker-ci.org/woodpecker/v2/server/forge/types"
```
6543, line 37: can we add a todo that this import should not be allowed and so must be refactored - it is there because last refactors where already to big to fit it in and this got forgotten at some point

pipeline/tracer.go
6543: make a copy and put it into pipeline/const.go migt be the best option atm.
later we could use that type in server code instead

pipeline/stepBuilder.go
xoxys: Done

pipeline/tracer.go
xoxys: Done

pipeline/backend/local/local.go
xoxys: Just added error handling.

pipeline/frontend/yaml/constraint/constraint.go
xoxys: Dont know I just added error handling for this assertion as well.

xoxys:
@6543 Could you take another look and maybe also give me some hints on https://github.com/woodpecker-ci/woodpecker/pull/2960#issuecomment-1860028847?

xoxys:
Going to close this one... We might need to split this into smaller chunks to get it merged faster. Otherwise, the conflict/rebase loop is just annoying.

6543:
do you split or are the changes now lost?!?

xoxys:
Not yet. Branch still exist.

6543:
in that case I'll reopen so it dont get deleted and the pull remains as TODO

xoxys:
I've added dedicated PRs for all missing linters. Closing again.

PR #2962, opened 2023-12-18
Update slogan & logo

6543:
- Update woodpecker slogan to repo too
  we are not just a simple fork anymore 🎉
- use https://github.com/woodpecker-ci/design/blob/main/logo/non-square.svg


README.md
```diff
@@ -46,13 +46,13 @@
 </p>
 <br/>
 
-Woodpecker is a community fork of the Drone CI system.
+Woodpecker is a simple CI engine with great extensibility.
```
anbraten, line 49: We could maybe even leave out the simple as users could think its not powerful.

6543: well I want to have the simple in it - as we try to be simple - without cuting down on being powerfull

6543: what about "... simple jet powerful ..." ?

6543, line 49:
```suggestion
Woodpecker is a simple jet powerful CI/CD engine with great extensibility.
```
?

pat-s: I get your idea with "simple" though it's quite a subjective word and people might understand it negatively (which is what @anbraten might have referred to).
(Same goes for "powerful" in the end.)

Maybe "golang-based" or "efficient" might also transmit the message? In the end people care about

- speed
- resource-efficiency
- extensibility/flexibility

and having these somehow reflected in this sentence would be great 🙂

anbraten: In terms of UX it could also be intuitive?

xoxys: It's just a single sentence... I would go with "Woodpecker is a simple yet powerful CI/CD engine with great extensibility." as a good compromise. IMO we tend to overcomplicate simple problems sometimes :see_no_evil:

xoxys:
If we change the slogan in the readme to something else, this change should be reflected in all other places as well, see https://github.com/search?q=repo%3Awoodpecker-ci%2Fwoodpecker+%22Woodpecker+is+a+simple+CI+engine+with+great+extensibility%22&type=code

lafriks (approved):

xoxys (approved):

Issue #2483, opened 2023-09-20
New javascript/css files are not loaded because index.html is cached


runephilosof-karnovgroup:
[indexHTML](https://github.com/woodpecker-ci/woodpecker/blob/6d86ec7fe77e2949e6351bd1f9b009e8c1afb237/server/web/web.go#L143-L144) should not be cached with a long expiry.
`index.html` references which version of the javascript and css files should be loaded.
So when index.html is cached with a long max-age the browser will not notice when it is time to load new javascript/css.

https://github.com/woodpecker-ci/woodpecker/blob/6d86ec7fe77e2949e6351bd1f9b009e8c1afb237/server/web/web.go#L176

Users need to reload the page to get the new js/css files. And even worse, if they do that while on /user for instance, the browser will only cache the new js/css files when you load woodpecker on /user next time. If you go to /repos or /, your browser will still have cached the old index file for those endpoints...

runephilosof-karnovgroup:
Recently experienced problems with outdated javascript code again in https://github.com/woodpecker-ci/woodpecker/issues/2821#issuecomment-1821058712

runephilosof-karnovgroup:
If Cache-Control is set to max-age=0, then etag will be used.
But this is only part of the problem.
Since a lot of the pages changes are within the spa, then the index.html won't get loaded, just new json responses.
So the server should probably set a header for the json responses with its version, to let the frontend detect whether it is outdated.
Or the spa should poll /version every minute or so.

runephilosof-karnovgroup:
But `index.html` is not reloaded, because `max-age=31536000`.
So the end user won't get the new version unless they manually click reload in the browser (it just happened to me again, when upgrading to v2.0.0).

It is fairly easy to reproduce.
1. Install `next` on the server.
2. Visit in a browser.
3. Change server to `v2.0.0`.
4. Click around in the browser
5. Open a new window, go to your woodpecker installation

Actual:
Version still shows `next` in the top left

Expected:
Version shows `v2.0.0` in the top left

runephilosof-karnovgroup:
`max-age=0`, or a low number.
This will affect how many requests the server receives for assets (those requests will have `If-None-Match: <etag>`, so probably easily processed, but they will still have to be processed).
Assets which have hashes, timestamps or similar in the request, can be cached a long time.

[...]

PR #2972, opened 2023-12-19
Server: Do not cache index.html

runephilosof-karnovgroup:
Or at least only cache it a short while. Set caching to 1 minute. Should not increase server load much, since most page changes are with the javascript spa app and does not ask for index.html.

This will ensure that the right version of the javascript file is loaded.

Fixes https://github.com/woodpecker-ci/woodpecker/issues/2483


lafriks:
Replaced by #2975

PR #2975, opened 2023-12-19
Fix static file caching

### LINKED ISSUE #2483: New javascript/css files are not loaded because index.html is cached

lafriks:
Replaces #2972

Fixes #2483

Removed etag header as etag is used incorrectly, it should be based on content not startup time and we don't handle it from request headers anyway.


anbraten (approved):
not tested, but lgtm

PR #2977, opened 2023-12-20
Add pagination component

anbraten:
I created this one some time ago for testing, but it might be worth considering to use the `Load more` button in general.


anbraten:
When adding it I would do it for all lists. Started and tested this some time ago and just tried again for the secrets list, but actually noticed the pagination works quite fine as long as a page-size is set explicitly, so no interest (time) in working on this from my side atm. Just wanted to push my tests.

xoxys:
We still need a `Load more` component, e.g. for the pipelines list.

xoxys:
But this one is in a fork repo, Ill pick it up in a new branch...

PR #2978, opened 2023-12-20
feat: add pipeline name even if there is only one

Tchoupinax:
Hey!

I noticed that when there is only one triggered pipeline, the name of the pipeline is not displayed. However, the information is valuable.
This MR tries to display it when there is only one pipeline


Tchoupinax:
With several pipelines
image

With one pipeline
image

zc-devs:
Is [this](https://github.com/woodpecker-ci/woodpecker/pull/2978#issuecomment-1865161496) behavior prior PR? If yes, could you provide screenshots after applying your changes?

Perhaps, you mentioned [workflow](https://woodpecker-ci.org/docs/usage/terminiology).

6543 (changes requested):
this was a UX dessision to not show it in single workflow mode ... as in 90% of cases it will just show "woodpecker"

6543:
through @Tchoupinax thanks for submitting a pull request :)

An idea to address your problem, what about a mouse hover over wicht will show a info text with the name in it?
through i dont know where that area should be :sweat_smile:

Tchoupinax:
@zc-devs Thank for the architecture schema, nice to understand!

> this was a UX dessision to not show it in single workflow mode ... as in 90% of cases it will just show "woodpecker"
Yes I see the problem... but basically is it not an invitation to correctly name your yaml file?

The "mouse over" solution is not ideal but can work if I really want to know which workflow is running

xoxys:
I can't find the original PR again, which has changed this behavior. What happens if a user has 5 workflows, but on PR's 4 are filtered out by a condition and only one workflow remains? Will this also not show the name of the single remaining workflow?

6543:
Good point, that should probably show the name. It does not atm

6543:
A fix should be simple, instead of count workflows, count reported config files

xoxys:
Or just always show the name and call it `Default` or similar? Personally, I don't like UI layout changes on the exact same view. The pipeline step list should have the same layout all the time.

lafriks:
Maybe hide only if the name is `woodpecker`?

Tchoupinax:
> Maybe hide only if the name is `woodpecker`?

I love the idea! So great!

lafriks:
> I don't think that's a good idea as you can change the default file name and you wouldn't have any benefit from showing it. I like @6543 's approach to just count workflow configs instead of workflows.

The problem is that you can have 3 pipelines with different conditions with the same step names and by opening the workflow there might be only one and you won't be able to tell which is executed as name is hidden

xoxys:
> The problem is that you can have 3 pipelines with different conditions with the same step names and by opening the workflow there might be only one and you won't be able to tell which is executed as the name is hidden.

The approach was to count config __files__ not workflows and I still think that's the best approach.

6543:
@Tchoupinax could you fix the conflicts and update acording to the discussion

anbraten:
I've applied the requested changes, but haven't tested it yet.

PR #2981, opened 2023-12-20
gitlab: support nested repos

6543:
close  #2311

PR #2994, opened 2023-12-22
gitlab cast ForgeRemoteID to string

6543:
regression of #2981

### DIFF
--- a/server/forge/gitlab/gitlab.go
+++ b/server/forge/gitlab/gitlab.go
@@ -234,7 +234,7 @@ func (g *GitLab) getProject(ctx context.Context, client *gitlab.Client, mr *mode
 		err  error
 	)
 	if mr != nil && mr.ForgeRemoteID.IsValid() {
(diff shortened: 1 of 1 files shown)


anbraten:
I also tried it and I think the problem is that it actually needs to be an integer instead. Will open a PR tomorrow

server/forge/gitlab/gitlab.go
```diff
@@ -234,7 +234,7 @@ func (g *GitLab) getProject(ctx context.Context, client *gitlab.Client, mr *mode
 		err  error
 	)
 	if mr != nil && mr.ForgeRemoteID.IsValid() {
-		repo, _, err = client.Projects.GetProject(mr.ForgeRemoteID, nil, gitlab.WithContext(ctx))
+		repo, _, err = client.Projects.GetProject(string(mr.ForgeRemoteID), nil, gitlab.WithContext(ctx))
```
lafriks, line 237: If it's integer you can't really do this

anbraten:
#3000

PR #3003, opened 2023-12-23
Hide queue visualization if nothing to show

qwerty287:


anbraten:
Why do you want to hide free agents capacity? Could you add some screenshots / more description?

web/src/components/admin/settings/queue/AdminQueueStats.vue
```diff
@@ -15,10 +15,10 @@
               {{ stats.completed_count }}
             </h4>
           </div>
-          <div class="pb-4 lg:pb-6">
+          <div v-if="total > 0" class="pb-4 lg:pb-6">
```
anbraten, line 18: Total IMO is the capacity your system with all agents can handle. So if there are 0 tasks it would be 100% green / free. If you have no agents and no tasks it should probably have some backgroundish color.

lafriks:
But free is also count of free capacity to run workflows not agent count

anbraten (approved):

PR #3007, opened 2023-12-23
fix and lint swagger file

6543:


cmd/server/woodpecker_docs_gen.go
```diff
@@ -38,8 +41,23 @@ func main() {
...
+	m := make(map[string]interface{})
+	if err := json.Unmarshal([]byte(jsonIn), &m); err != nil {
+		panic(err)
+	}
+	delete(m, "host")
+	raw, err := json.Marshal(m)
 	if err != nil {
 		panic(err)
 	}
+	return string(pretty.Pretty(raw))
```
6543, line 62: not sure if we have to format it nicely?

6543:
Well yes but we lib we use to gen dont detect errors as you can see.

6543:
https://github.com/woodpecker-ci/woodpecker/blob/9d9bcbf363bbd82b88481108f1bc0cf305de6899/.woodpecker/test.yml#L59-L66

hmm looks like i can move the check a bit :)

6543:
@qwerty287 updated :)

6543:
@qwerty287 updated :)

6543:
Why it's just generated json?

You should not manually edit it anyway.

6543:
Well we additionally could do that, but it did not detect the errors in the first place so it can not be used for linting 😆 

PS: can you push the codeformat onto this branch?

6543:
@qwerty287 this format should go into precommit ... do you know how to do that?

PR #3008, opened 2023-12-24
step depends_on ignore empty strings

### LINKED ISSUE #3055: depends_on: [] doesn't enable DAG mode
### Component

server

### Describe the bug

`depends_on: []` like specified in https://github.com/woodpecker-ci/woodpecker/blob/c1a1f7c10b3c0840bad9c0b25fcd7f5ff8b0173b/.woodpecker/securityscan.yml is useless, and doesn't enable DAG mode as documented on https://woodpecker-ci.org/docs/usage/workflow-syntax#depends_on.

This is because DAG is only enabled for non-empty lists: https://github.com/woodpecker-ci/woodpecker/blob/c1a1f7c10b3c0840bad9c0b25fcd7f5ff8b0173b/pipeline/frontend/yaml/compiler/dag.go#L44-L51

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.1.1"}
```

### Additional context

_No response_

### Validations

6543:
Just extend the test and ignore empty string dependencies

close #3055
close  #3047


pipeline/frontend/yaml/compiler/dag_test.go
```diff
@@ -59,4 +59,74 @@ func TestConvertDAGToStages(t *testing.T) {
 	}
 	_, err = convertDAGToStages(steps, "")
 	assert.ErrorIs(t, err, &ErrStepMissingDependency{})
+
+	steps = map[string]*dagCompilerStep{
+		"echo env": {
+			position:  0,
+			name:      "echo env",
+			group:     "",
+			dependsOn: []string{""},
```
anbraten, line 68: @6543 Why do you want to support such a wrong configuration? Shouldn't we treat it like an error if someone writes: `depends_on: [""]` instead of `depends_on: []` Or is it because of this super weirdish way go-yaml is converting the slice?

6543: I just want to let the server follow our idea: if run and on strange edgecases make sure it still works

the linter of course could point out a warning, but if you do `depends_on: [""]` as **a naive user** i expect it to not depend on something ... since the string is empty right :laughing:

anbraten: I would just throw an error instead of supporting such configs

6543: Well I'm for supporting users

anbraten: Yes, but not bad habits and wrong configs. Using an empty string is just wrong IMO.

6543: well we now even have a good use-case, and a user intuitively try to use it that way already

anbraten: I still think it supports wrong configs. The mentioned issue was a bug solved by #3097

6543:
#3097 got merged

PR #3010, opened 2023-12-24
Use dag in ci config

anbraten:


.woodpecker/securityscan.yml
```diff
@@ -14,13 +14,13 @@ variables:
 
 steps:
   check backend:
-    group: check
+    depends_on: []
```
lafriks, line 17: Why empty depends_on?

anbraten: it depends on nothing, but you need to define at least one `depends_on` per config file to enable dag mode.

lafriks: We should probably add/implement an option to pipeline `parallel: true` or something like that to set pipeline to do this without needing to add empty dependencies for steps

anbraten: I prefer this way as it is simpleand works the same way in drone. It was also discussed like this in the original issue.

lafriks (approved):

Issue #738, opened 2022-01-31
Add spell checker to pipeline


6543:
as title

pat-s:
Added in #3018

PR #3018, opened 2023-12-25
Add spellcheck config

lukashass:
Part of #738 

```
pnpx cspell lint --gitignore '{**,.*}/{*,.*}'
```


anbraten:
CI job works now: https://ci.woodpecker-ci.org/repos/3780/pipeline/11011/33
Just "some" typos left.

pat-s:
@lukashass Do you want to clean up here? :)

lukashass:
Yes, its just a lot :see_no_evil:

.cspell.json
```diff
@@ -0,0 +1,39 @@
...
+    "xormigrate",
+    "zerolog",
+    "Msgf"
+  ],
+  "ignorePaths": [
+    "**/node_modules/**/*",
+    "pnpm-lock.yaml",
+    ".cspell.json",
+    "web/src/assets/locales/**/*",
+    "server/store/datastore/migration/testfiles/sqlite.db",
```
6543, line 28:
```suggestion
    "server/store/datastore/migration/**/*",
```

anbraten: Why should the code / other files in migrations not be checked? I am pretty sure this was intentional. CC @lukashass

anbraten:
> Yes, its just a lot 🙈

We could split it by ignoring paths first and for example just correct the `agent/` folder in this PR. This way others could correct code as well and we could simply open new PRs removing the excluded area with all translation fixes.

lukashass:
> We could split it by ignoring paths first and for example just correct the `agent/` folder in this PR. This way others could correct code as well and we could simply open new PRs removing the excluded area with all corresponding translation fixes.

Most of the work to be done is adding words to the allow list, so this wouldn't work very well. I'll have some time next week to continue with this.

lukashass:
I decided to do this in steps after all. So `web/` is done, others are started, but ignored for now.

.cspell.json
```diff
@@ -0,0 +1,113 @@
...
+    ".cspell.json",
+    ".golangci.yaml",
+    ".vscode/extensions.json",
+    "web/src/assets/locales/**/*",
+    "web/components.d.ts",
+    "server/store/datastore/migration/**/*",
+    "go.mod",
+    "go.sum",
+    "*.excalidraw",
+    "CHANGELOG.md",
```
anbraten, line 98: Should probably be also be checked in the future. It will fail in the release PR so we will know if we need to rename a PR.

anbraten (approved):

web/src/components/atomic/Icon.vue
```diff
@@ -7,7 +8,7 @@
   <i-mdi-gesture-tap v-else-if="name === 'manual-pipeline'" class="h-6 w-6" />
   <i-mdi-tag-outline v-else-if="name === 'tag'" class="h-6 w-6" />
   <i-clarity-deploy-line v-else-if="name === 'deployment'" class="h-6 w-6" />
-  <i-mdisource-commit v-else-if="name === 'commit'" class="h-6 w-6" />
+  <i-mdi-source-commit v-else-if="name === 'commit'" class="h-6 w-6" />
```
6543, line 11: was this a bug? does that icon exist?

.gitpod.yml
```diff
@@ -105,6 +105,7 @@ ports:
 
 vscode:
   extensions:
+    # cSpell:disable
```
6543, line 108: ?

.cspell.json
lukashass: Checking commit messages should probably be done before merging PRs then. Will move it down to the TODOs

web/src/components/atomic/Icon.vue
lukashass: I think both work, with the `-` even the inline preview of my VSCode plugin works, but not tested in the app.

.gitpod.yml
lukashass: Extensions have mostly usernames, not really worth spell checking IMO

6543: No you have two comments here, should ir not be one?

lukashass: Ah the second comment enables cSpell for any lines that might be added below

.cspell.json
```diff
@@ -0,0 +1,113 @@
...
+  ],
+  "ignorePaths": [
+    "**/node_modules/**/*",
+    "pnpm-lock.yaml",
+    ".gitignore",
+    ".git/**/*",
+    ".cspell.json",
+    ".golangci.yaml",
+    ".vscode/extensions.json",
+    "web/src/assets/locales/**/*",
```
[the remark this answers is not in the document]
lukashass: I'm not sure if this is possible with the glob library cSpell is using: https://cspell.org/docs/globs/

> Glob patterns used to exclude files and directories from being checked. ! are only partially supported due to an issue with the Glob library[1](https://github.com/isaacs/node-glob/issues/409)

.gitpod.yml
6543: Diff looked straange

6543 (approved):

6543:
Thanks :)

PR #3037, opened 2023-12-27
make backend step dag generation deterministic

6543:
the the generation for backend steps if a dag is used deterministic.

this also fix where the test randomly fail like in:
- https://ci.woodpecker-ci.org/repos/3780/pipeline/11057/30
- https://ci.woodpecker-ci.org/repos/3780/pipeline/11076/25


6543:
PS: I'll push 100% pass once test failed

lafriks:
Would it be possible to keep order defined in the pipeline definition?

6543:
@lafriks yes - done

lafriks (approved):

PR #3039, opened 2023-12-27
trim v on version check

### LINKED ISSUE #3034: Update notice showing for v2.1.0 while already on v2.1.0
After switching to the v2.1.0 docker image for woodpecker, the admin interface says to update to the current version.

image

This is from the docker image: `woodpeckerci/woodpecker-server:v2.1.0`

6543:
close #3034

### DIFF
--- a/web/src/compositions/useVersion.ts
+++ b/web/src/compositions/useVersion.ts
@@ -72,9 +72,9 @@ export function useVersion() {
       if (usesNext) {
         latest = versionInfo.next;
       } else if (current.includes('rc')) {
-        latest = versionInfo.rc;
+        latest = versionInfo.rc.replace(/^v/, '');
       } else {
-        latest = versionInfo.latest;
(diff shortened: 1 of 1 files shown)


6543:
a simple hotfix to release a v2.1.1

web/src/compositions/useVersion.ts
```diff
@@ -72,9 +72,9 @@ export function useVersion() {
       if (usesNext) {
         latest = versionInfo.next;
       } else if (current.includes('rc')) {
-        latest = versionInfo.rc;
```
anbraten, line 75: @6543 Or should we directly do this in the ci config file updating the version.json file.

6543: I want to release a v2.1.1 anyway ... :)

and later i suggest to use a version parse lib

6543: #3041

xoxys (approved):

PR #3042, opened 2023-12-27
use semver for version comparsion

### LINKED ISSUE #3041: use semversion parsing on version check
we currently just compare strings, but some (e.g. _me_) run custom versions like e.g. "v2.1.0-win10wsl".

by using a semversion parsing lib we get rid of issues with trailing v and still can compare if current <= latest.

6543:
close  #3041
close #3043


6543:
@qwerty287 pnpm told me that the dependency is already in the project it was just an dependency-dependency till now

lafriks (approved):

lafriks:
JS library management etc is not the best but reimplementing this also probably is not the best way to go...

web/src/compositions/useVersion.ts
```diff
@@ -72,17 +73,17 @@ export function useVersion() {
...
+        latest = versionInfo.latest;
       }
     }
 
     version.value = {
       latest,
       current,
       currentShort: usesNext ? 'next' : current,
-      needsUpdate: latest !== undefined && latest !== current,
+      needsUpdate: semver.gt(latest, semver.coerce(current)),
```
6543, line 86:
```suggestion
      needsUpdate: latest !== undefined && semver.gt(latest, semver.coerce(current)),
```

web/src/compositions/useVersion.ts
```diff
@@ -72,17 +73,17 @@ export function useVersion() {
...
+        latest = versionInfo.latest;
       }
     }
 
     version.value = {
       latest,
       current,
       currentShort: usesNext ? 'next' : current,
-      needsUpdate: latest !== undefined && latest !== current,
+      needsUpdate: latest !== undefined && semver.gt(latest, semver.coerce(current)),
```
6543, line 86:
```suggestion
      needsUpdate: latest !== undefined && latest !== null && semver.gt(latest, semver.coerce(current)),
```

6543:
this only adds `8397` bytes :tada:

PR #3043, opened 2023-12-27
Fix version updating

anbraten:
proper fix for #3039


6543:
does the link generated to the release still work?

6543:
Thats why i'm for  #3042

It also fixes other edgecases too

lafriks (approved):

6543 (changes requested):
Break release link

Issue #2780, opened 2023-11-09
Deprecations tracker


anbraten:
This issue contains parts that are deprecated and the date when they will finally be removed within the next major release. 

Deprecated parts have to be announced in the release notes and if related to the pipeline config via the linter / etc. After that they have to stay for at least (TBD months).

# Active deprecations

## https://github.com/woodpecker-ci/woodpecker/pull/2181

remove in x.2024

## #2180

remove in x.2024

## #2180

remove in x.2024

## #2181

remove in x.2024

## Removed

- https://github.com/woodpecker-ci/woodpecker/pull/2048
- https://github.com/woodpecker-ci/woodpecker/pull/1644
- https://github.com/woodpecker-ci/woodpecker/pull/1752
- https://github.com/woodpecker-ci/woodpecker/pull/2090
- https://github.com/woodpecker-ci/woodpecker/pull/2635

zc-devs:
Seems, in `no PR yet` category only `update Docker to v27` left.

PR #3052, opened 2023-12-28
Revert "Add workflow version"

qwerty287:
This reverts commit 1bc4415075488d1bdfb496a5954505459b252032.

Workflow YAML versions are not longer necessary as we have deprecations now. See https://github.com/woodpecker-ci/woodpecker/issues/2780

This isn't breaking as it wasn't documented yet and didn't has any usecase.


pipeline/frontend/yaml/linter/schema/schema.json
```diff
@@ -493,7 +374,7 @@
       }
     },
     "event_enum": {
-      "enum": ["push", "pull_request", "pull_request_closed", "tag", "deployment", "cron", "manual"]
```
anbraten, line 496: Should stay

pipeline/frontend/yaml/linter/schema/schema.json
```diff
@@ -14,118 +14,42 @@
...
+      "items": { "type": "string" }
     },
     "runs_on": {
       "type": "array",
       "minLength": 1,
-      "items": {
-        "type": "string"
-      }
-    },
-    "version": {
```
anbraten, line 65: We might leave it in here as it will throw an error otherwise if someone added the version property.

anbraten (approved):

PR #3060, opened 2023-12-29
Ignore missing dependencies

qwerty287:
To fix issues with steps/workflows filtered by `when`.


anbraten:
Shouldn't the depending step bein reauired to have the same filter as well in which case it would not be an issue?

lafriks:
I still think this could be useful, maybe not in our pipeline but in general otherwise there might be cases where you could have to duplicate steps just to match conditions. Skipped steps imho should be treated as completed in DAG resolution

PR #3061, opened 2023-12-29
Use CI vars for plugin

qwerty287:
we don't have to set them manually


anbraten:
Let's wait for the next release of the plugin, so we can directly add the pull_request_closed event as well

anbraten (approved):

Issue #1389, opened 2022-11-02
Document how to create an own plugin


6543:
- mention how settings are de-serialized
- mention build-in-env-vars
- mention best practics
- mention multiarch build, as official immages try to always have linux/amd64 & linux/arm64 images, and so other plugins should try to do so to

6543:
tldr of how to do plugin is: all settings (arguments) are environment variables start with prefix PLUGIN_
most plugins use the build-in-env-vars for default values ...
and as plugin you define the container entrypoint

simplest example I know of:
https://github.com/vividboarder/drone-webdav

smainz:
As I need to write some plugins, I am going to put together a small template bash plugin. What I plan is:

- Provide an example `Dockerfile` (alpine based)
- Provide a simple script `plugin.sh` which
  - reads settings 
  - reads env vars
  - uses some built in env vars
  - doses some simple action
  - sets exit code
- Template for a Readme
- Build instructions / build script for the image
- CI config using `woodpeckerci/plugin-docker-buildx` to build the images
- Readme on how to use it

I could need some help:

1. Input on the structure / content of the Readme template. ATM I am going to structure it like this:  https://github.com/smainz/woodpecker-git-setup/blob/main/Readme.md
2. Should the plugin prefer DRONE_* or CI_* env vars?
3. What are the 'best practices' mentioned above
4. What more would you like to see?

Question 2. somehow requires some vision on where Woodpecker will go and if we strive to be compatible with drone for the forseable future on the plugin level.

anbraten:
> 1. Input on the structure / content of the Readme template. ATM I am going to structure it like this: https://github.com/smainz/woodpecker-git-setup/blob/main/Readme.md

That structure is nice. To be able to list projects in the [official Woodpecker plugins list](https://woodpecker-ci.org/plugins) its in addition quite / even more important to provide a `docs.md` file. 

> 2. Should the plugin prefer DRONE_* or CI_* env vars?

If the plugin is mainly for Woopecker it should go with `CI_` variables as they are more feature-rich. They are not as vendor specific as well.

> 3. What are the 'best practices' mentioned above

I guess that's a list that can be extended from time to time we experience any pitfalls. Currently I can think of things like supporting both `CI_` and `DRONE_` in case you want to use the plugin with `DRONE_` as well. We could for example also provide a https://github.com/drone-plugins/drone-plugin-lib version for Woodpecker and link it here.

> 4. What more would you like to see?

[...]

Issue #1033, opened 2022-07-18
Add new page "plugins" into Development section


6543:
Add a new page (https://woodpecker-ci.org/docs/development/plugins) where:
- **all assumptions** are written down
- how things are handled for the docker backend
- ...

anbraten:
I think this is already explaining most points of it: https://woodpecker-ci.org/docs/usage/plugins/sample-plugin

PR #3062, opened 2023-12-29
Extend create plugin docs

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/1389
closes https://github.com/woodpecker-ci/woodpecker/issues/1033


lafriks:
It would also be nice to add a go plugin sample and it's dockerfile

pat-s:
We should really find a central place for plugin documentation (both for dev and users). I still find it unfortunate that plugins are split between CB and GH. 

Btw there is also https://codeberg.org/woodpecker-plugins/go-plugin.

pat-s:
> But docs are completely here, right?

I guess this would be the best.

We have a recent use case which also exactly tackles this topic: https://codeberg.org/woodpecker-plugins/node-pm/issues/21

Any effort on improving the situation is great 👍

docs/docs/20-usage/51-plugins/20-creating-plugins.md
```diff
@@ -0,0 +1,121 @@
+# Creating plugins
+
+Creating a new plugin is simple: Build a Docker container which uses your plugin logic as entrypoint.
```
pat-s, line 3:
```suggestion
Creating a new plugin is simple: Build a Docker container which uses your plugin logic as the ENTRYPOINT.
```

docs/docs/20-usage/51-plugins/20-creating-plugins.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Settings
+
+To allow users to configure the behavior of your plugin, you should use settings.
```
pat-s, line 7:
```suggestion
To allow users to configure the behavior of your plugin, you should use `settings:`.
```

docs/docs/20-usage/51-plugins/20-creating-plugins.md
```diff
@@ -0,0 +1,121 @@
...
+
+These are passed to your plugin as uppercase env vars with `PLUGIN_` prefix.
```
pat-s, line 9:
```suggestion
These are passed to your plugin as uppercase env vars with a `PLUGIN_` prefix.
```

docs/docs/20-usage/51-plugins/20-creating-plugins.md
```diff
@@ -0,0 +1,121 @@
...
+Using a setting like `url` results in `PLUGIN_URL` as env var.
```
pat-s, line 10:
```suggestion
Using a setting like `url` results in an env var named `PLUGIN_URL`.
```

docs/docs/20-usage/51-plugins/20-creating-plugins.md
```diff
@@ -0,0 +1,121 @@
...
+        list:
+          - 2
+          - 3
+```
+
+Values like this are converted to JSON and then passed to your plugin. In the example above, the env value would be `{"abc": "2", "list": [ "2", "3" ]}`.
+
+### Secrets
+
+Secrets should be passed as settings too. Therefore, users should use [`from_secret`](../40-secrets.md#use-secrets-in-settings).
```
pat-s, line 45:
```suggestion
Secrets should be passed as settings, too. Therefore, users should use [`from_secret`](../40-secrets.md#use-secrets-in-settings).
```

docs/docs/20-usage/51-plugins/20-creating-plugins.md
```diff
@@ -0,0 +1,121 @@
...
+  -d ${PLUGIN_BODY} \
+  ${PLUGIN_URL}
+```
+
+### Package it
+
+Create a Dockerfile that adds your shell script to the image, and configures the image to execute your shell script as the main entrypoint.
+
+```dockerfile
+FROM alpine
```
pat-s, line 88:
```suggestion
```dockerfile
# please pin the version, e.g. alpine:3.19
FROM alpine
```

We should encourage version pinning already here

docs/docs/20-usage/51-plugins/20-creating-plugins.md
```diff
@@ -0,0 +1,121 @@
...
+  -e PLUGIN_BODY="hello world" \
+  foo/webhook
+```
+
+## Best practices
+
+- Build your plugin for different architectures to allow many users to use them.
+  At least, you should support `amd64` and `arm64`.
+- Provide binaries for users using the `local` backend.
+  These should also be built for different OS/architectures.
```
pat-s, line 117: If we list this, we should add more context how this can be done.

pat-s (commented):
Two additional points worth mentioning:

- Plugins on CB are also published to the CB package registry
- `renovate` should be used to keep dependencies up-to-date
- Encourage the use of https://github.com/woodpecker-ci/plugin-ready-release-go (?)

6543:
i agree the docs should not be for our official plugins but rather for everybody who want to make there own.
and the programming practices might vary

pat-s:
Programming practices is one thing but if there are no general "rules" how to format/maintain/publish/release plugins, everybody will do something else and in the end it's just a mess for users (and devs).
That's what linters are for, that's what other implicit "contributing" rules are for.

*At least* in repos which live within the GH and CB woodpecker org we should have such rules. In a central place. How devs then comply with these rules is a different thing. I am actually aiming to draft something for a long time already (content TBD) but I don't think being "too open" or "too lax" is really helping in the long-term. Yes, it's FOSS, yes individual contributors with (many) own opinions, but if this project wants to grow in a smooth way, there is no way around such rules in the mid and long-term.

> This should be added to the individual plugin docs.

I disagree. This is a perfect point showcasing what I wanted to say above: sure, it doesn't hurt anyone if a plugin is published there or if not. But if there is no rule, the ratio of plugins available there will be maybe like 70/30 (available/not available) and users have to look up every time the specific state for this plugin. This is already the case with the availability of plugins in CB or GH. There was no rule, nobody knows why the current state is like it is and users are also partly confused.
This is just one example, the same applies for 

- semver
- conventional commits
- updates
- naming
- README content
- etc.

I hope you understand what I want to say 😄 I don't want to make things too complicated and surely enjoyable for everyone. And many things are great. But there are (quite some) points where everyone is just 🤷🤷🤷 and in the end this will cause issues/lot of unnecessary typing/unprofessional look of the project.

> i agree the docs should not be for our official plugins but rather for everybody who want to make there own.

And I also disagree with this with the arguments from above.

No hard feelings, just opinions ❤️

docs/docs/20-usage/51-plugins/20-creating-plugins.md
```diff
@@ -0,0 +1,122 @@
...
+    image: foo/plugin
+    settings:
+      complex:
+        abc: 2
+        list:
+          - 2
+          - 3
+```
+
+Values like this are converted to JSON and then passed to your plugin. In the example above, the env value would be `{"abc": "2", "list": [ "2", "3" ]}`.
```
6543, line 41:
```suggestion
Values like this are converted to JSON and then passed to your plugin. In the example above, the environment variable `PLUGIN_COMPLEX` would contain `{"abc": "2", "list": [ "2", "3" ]}`.
```

6543 (approved):

6543:
@pat-s @qwerty287 as a owner I think this pull is ready.

And so I'm going it merge it _now_ :)

for the guidelines, I open a new discussion

6543:
- >  #3085

PR #3069, opened 2023-12-29
Cleanups + prefer .yaml

qwerty287:
- docs cleanups
    - move to `ini` for config env vars instead of `docker-compose`
- rename files from `*.yml` -> `*.yaml` (see https://yaml.org/faq.html)
    - except gitpod because their docs don't mention `.yaml` ending
    - also renamed in docs

### DIFF
--- a/pipeline/frontend/yaml/linter/schema/schema_test.go
+++ b/pipeline/frontend/yaml/linter/schema/schema_test.go
@@ -34,76 +34,76 @@ func TestSchema(t *testing.T) {
 	}{
 		{
 			name:     "Branches",
-			testFile: ".woodpecker/test-branches.yml",
+			testFile: ".woodpecker/test-branches.yaml",
 		},
 		{
 			name:     "Branches Array",
-			testFile: ".woodpecker/test-branches-array.yml",
+			testFile: ".woodpecker/test-branches-array.yaml",
 		},
 		{
 			name:     "Branches exclude & include",
-			testFile: ".woodpecker/test-branches-exclude-include.yml",
+			testFile: ".woodpecker/test-branches-exclude-include.yaml",
 		},
 		{
 			name:     "Clone",
-			testFile: ".woodpecker/test-clone.yml",
+			testFile: ".woodpecker/test-clone.yaml",
 		},
 		{
 			name:     "Clone skip",
-			testFile: ".woodpecker/test-clone-skip.yml",
+			testFile: ".woodpecker/test-clone-skip.yaml",
 		},
 		{
 			name:     "Matrix",
-			testFile: ".woodpecker/test-matrix.yml",
+			testFile: ".woodpecker/test-matrix.yaml",
 		},
 		{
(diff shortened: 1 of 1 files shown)


docs/docs/20-usage/20-workflow-syntax.md
```diff
@@ -726,58 +726,58 @@ Example conditional execution by branch:
        channel: dev
 ```
 
-> The step now triggers on main, but also if the target branch of a pull request is `main`. Add an event condition to limit it further to pushes on main only.
+The step now triggers on main, but also if the target branch of a pull request is `main`. Add an event condition to limit it further to pushes on main only.
```
xoxys, line 729:
```suggestion
The step now triggers on `main`, but also if the target branch of a pull request is `main`. Add an event condition to limit it further to pushes on main only.
```

docs/docs/30-administration/00-deployment/10-docker-compose.md
```diff
@@ -130,7 +130,7 @@ Image variants:
 - The `latest` image is the latest stable release
 - The `vX.X.X` images are stable releases
 - The `vX.X` images are based on the current release branch (e.g. `release/v1.0`) and can be used to get bugfixes asap
-- The `next` images are based on the current `main` branch and should not be used for production environments
```
xoxys, line 133: I would keep it. `main` may break at any time and should be used with caution.

docs/docs/30-administration/10-server-config.md
```diff
@@ -12,63 +12,36 @@ Registration is closed by default (`WOODPECKER_OPEN=false`). If registration is
...
-+     - WOODPECKER_ADMIN=johnsmith,janedoe
+```ini
+WOODPECKER_OPEN=false
+WOODPECKER_ADMIN=johnsmith,janedoe
 ```
 
-### To only allow registration of users, who are members of approved organizations
+### Only allow registration of users, who are members of approved organizations
 
-```diff title="docker-compose.yml"
```
xoxys, line 47: Don't know if it is really helpful to remove all the compose examples.

docs/docs/30-administration/22-backends/20-local.md
```diff
@@ -98,35 +62,10 @@ steps:
 If no commands are provided, we treat them as plugins in the usual manner.
 In the context of the local backend, plugins are simply executable binaries, which can be located using their name if they are listed in `$PATH`, or through an absolute path.
 
-### Using labels to filter tasks
```
xoxys, line 101: Why? Not valid anymore?

xoxys (approved):
Few smaller findings, LGTM otherwise.

6543:
linter to enforce it ?!?

xoxys:
A simple `find ` command that searches for files with yml ending and fail if found?

xoxys:
There is a better way by just using pre-commit https://stackoverflow.com/a/74701364

PR #3081, opened 2023-12-30
Make k8s backend configuration docs in the same format as others

micash545:


pat-s (changes requested):
IMO this should be other way round, i.e. the k8s format should be applied to the others. We should not use markdown headers for config descriptions.

Yes I know, headers create link which can be used as direct references but they also bloat the always-visible sidebar in the docs - and this looks quite strange (having only capitalized config values there).

zc-devs:
> they also bloat the always-visible sidebar in the docs

This. One option is move it at the end of the page, like in [Agent](https://woodpecker-ci.org/docs/administration/agent-config) and [Server](https://woodpecker-ci.org/docs/administration/server-config) docs. JFYI: [WOODPECKER_ADDONS](https://woodpecker-ci.org/docs/administration/agent-config#woodpecker_addons) doesn't look as others.

> headers create link which can be used as direct references

Another option is set them as H4 and hide headers >H3 from menu. Something like [that](https://docusaurus.io/docs/next/markdown-features/toc#example-subsubsection-1-a-i).

---

[This](https://github.com/woodpecker-ci/woodpecker/assets/127358482/c68faa5d-0491-4d3b-83fd-1efe478cde26) should be fixed as well as [this](https://github.com/woodpecker-ci/woodpecker/assets/127358482/3a9c2304-7c89-480f-97dd-793c8439b36d).

pat-s:
Moving them to the end and reducing the header level sounds good to me too.

Should there be a new PR? I don't think we should urge @micash545 to do all of the mentioned improvements on every page.

pat-s:
Thanks @qwerty287 

> Another option is set them as H4 and hide headers >H3 from menu. Something like [that](https://docusaurus.io/docs/next/markdown-features/toc#example-subsubsection-1-a-i).

Could we also get this in as suggested by @zc-devs? If we do so, `options` would be alone in the ToC which also looks a bit weird. But likely still better than having all individual options as a single entry.

I would also vote for moving the list of options to the end to avoid long scrolling at the top of the page.

pat-s (approved):

Issue #3071, opened 2023-12-29
pull request close envent not selectable on edit secret page


6543:
### Component

web-ui

### Describe the bug

you can not have secrets for the close pull event atm

### System Info

```shell
v2.1.1
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

PR #3084, opened 2023-12-30
Allow PR secrets to be used on close

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/3071

1. If a secret can be used on PRs, it can also be used on PR close.
2. If no events are set, disallow access to secret. This was different before, secrets without any event set were allowed for all events.
3. Compare strings instead of patterns.


server/model/secret.go
```diff
@@ -105,11 +104,11 @@ func (s Secret) IsRepository() bool {
 
 // Match returns true if an image and event match the restricted list.
 func (s *Secret) Match(event WebhookEvent) bool {
-	if len(s.Events) == 0 {
-		return true
+	if event == EventPullClosed {
+		event = EventPull
 	}
-	for _, pattern := range s.Events {
-		if match, _ := filepath.Match(string(pattern), string(event)); match {
```
6543, line 112: why was there a filepath match in the first place?!?

6543: never mind it's just a droneCI legacy nonesence

https://github.com/woodpecker-ci/woodpecker/blame/f1e2f3dcad3c5eff9255f58cf8eb0430cd80880f/server/model/secret.go#L70-L73

server/model/secret.go
```diff
@@ -105,11 +104,11 @@ func (s Secret) IsRepository() bool {
 
 // Match returns true if an image and event match the restricted list.
```
6543, line 105:
```suggestion
// Match returns true if an image and event match the restricted list.
// Note that EventPullClosed are treated as EventPull.
```

6543 (approved):

6543:
it was so in the past, it was meant as "optional filter" and if you have no events set, it was for all.

we did change the UI from a simple textfield to checkboxes those it's a bit confusing as UI<->backend handle things different now.

so the only thing for "filter them all" is the idea of missusing the events filter for "disable feature", I think we can&should:
- add an enable/disable secrets option
- then show all checked if filter is is empty and if you mannualy deselect all then automatically make secrete disabled (via UI) and handle in backend acordingly

that way we dont have to break anything and can indicate this clearly

6543:
moved to #3094 so we can merge this pull now (if @qwerty287 comment lgtm or other maintainers did lgtm).

PR #3086, opened 2023-12-30
Blocked: Fix alterTableSecretsAddUserCol to handle NULL cases of secret_repo_id

6543:
see https://discord.com/channels/838698813463724034/838698813463724037/1190775891282907238

"... SInce migrating to v2 I get this error "fatal could not migrate datastore | error=migration alter-table-add-secrets-user-id failed: Error 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'COLUMN secret_repo_id SET NOT NULL' at line 1 " ..."


6543:
blocked: need response of **bunkerman**@Discord

6543:
did not fix the issue ...

more debugging needed

PR #3089, opened 2023-12-30
make sure cross-compile-server does not run with other compile steps

6543:
I noticed it as it crashed the agent on the main server at some point as we did use almost all swap and had 100% ram usage

in total we used 16G of ram


6543:
no as it does not consume that much ram, we only compile for a subset of targets there

PR #3092, opened 2023-12-31
Move `cmd/common` to `shared`

qwerty287:
`cmd` should only contain the entrypoints of the binaries.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -141,6 +141,6 @@ issues:
       linters:
         - forbidigo
     # allow some setup functions to use log.Fatal()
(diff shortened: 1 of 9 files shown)


6543:
well `cmd` should also contain all the config & interaction to build those cli tools

like agrs, options (flags) ...

and the logger handling mostly handles those args ... so i think it should stay within cmd

6543:
does the text in https://woodpecker-ci.org/docs/next/development/architecture#main-package-hierarchy not reflect this ?!?

6543 (approved):
hmm thats a point ... I still thin we should make "shared" as lean as possible

6543:
such "shared" places **tend to** collect more and more stuff witch should be seperated

PR #2019, opened 2023-07-20
docs: continuous deployment

lonix1:
Continuous deployment page, for newbies like me.

Thanks to @techknowlogick who advised me on [discord](https://discord.com/channels/838698813463724034/838698813463724037/1131383762962370671).

PR #3098, opened 2024-01-01
Add continous deployment cookbook

qwerty287:
Supersedes https://github.com/woodpecker-ci/woodpecker/pull/2019


lafriks:
Probably should mention also about manual step to run deploy

pat-s:
This sounds great overall. I wonder though if the "simple blog" format is the best format for this content. Over time, the posts will move down the page due to their publish age. 

What about a format similar to https://docusaurus.io/showcase? With some favorites pinned at the top?

pat-s:
> I'm not sure what is wrong with having a page in the docs with ci info. To "keep it simple".

I'm not sure if I fully understand you here. Can you rephrase?

@qwerty287 What is your opinion on the "gallery style" instead of the blog style?

pat-s:
> I'm fine with it, but I don't think I'd allow pinning to top. But in general yes.

Is "pinning on top" referring to adding a navbar item?
In this case I'd say: a navbar exists to navigate, i.e. an additional item which links to the (integrated) cookbook should be fine? The current preview also lists it and was added by you @qwerty287, right? Or am I misunderstanding something here?

I also think it shouldn't belong to "Docs" but be a separate page in a gallery style, listed in the navbar.

pat-s:
> No I mean pin to top in the gallery.
I.e. have some items always on top.

Ah. Hmm, no strong opinion about this. I would probably just sort alphabetically. Or by date. If there's a search function, it doesn't matter much.

But the gallery format is still missing - at least in this PR. I think I am partly confused that we talk about this format but the current preview shows the blog style format 😅

zc-devs:
So many words (in PR), but no mention of using dedicated CD tool like AgroCD. Or what category does it fall?

All points (in PR) are ether too general or implementation details. Does AgroCD use webhooks? Can it poll? Can you call some API from the last WP step?

> I think it's too generic to be really useful for docs. Our docs should be only for woodpecker-related stuff, and this document can be applied to any ci/cd system. Our docs shouldn't be an "all-in-one" solution, they should document our software.

I'm with this. There are already docs that either wrong or for a specific case, but have label `must`.

Should I write _in docs_ how I deploy my Woodpecker in Kubernetes? How I terminate TLS? How I deploy remote Agent using Traefik and K3s? (No)

I like the blog format. Cases above and PR are incredibly fit that format.

> Over time, the posts will move down the page due to their publish age.

That is cool. Because you may use Woodpecker 0.15 in your showcase, which may not be relevant today.

> I mean a "CD page" alongside the other pages

At least because it is not CD tool. The name points to it and should be in the docs (not checked, but saw / read somewhere).

However, I glad to see something about `Deploy` button in the docs (sorry, if there is)
[Screenshot 2024-01-07](https://github.com/woodpecker-ci/woodpecker/assets/127358482/fd4b72f3-ed7b-42aa-84ed-b7d52b3d987c)

6543 (approved):
This is good enougth. Every thing else is bike shading.

If there are errors/things to update, just open a new pull.

Lets publish it.

PR #3125, opened 2024-01-07
Remove steps for publishing release branches

qwerty287:
from #3069


.woodpecker/docker.yml
```diff
@@ -3,7 +3,6 @@ when:
   - event: push
     branch:
       - ${CI_REPO_DEFAULT_BRANCH}
-      - release/*
```
6543, line 6: why remove release branches - this will prohibit releasing from them patch versions

6543 (changes requested):

.woodpecker/docker.yml
6543: thats a point

6543 (approved):

6543 (approved):

Issue #3093, opened 2023-12-31
Error when adding two equal workflow configs


qwerty287:
### Component

server

### Describe the bug

Add two YAML files with exactly the same content (copy-paste). (I know that this isn't a common usecase ;) )

```
11:39AM ERR server/pipeline/create.go:114 > failed to find or persist pipeline config for qwerty287/woodpecker error="UNIQUE constraint failed: pipeline_config.config_id, pipeline_config.pipeline_id"
11:39AM ERR server/router/middleware/logger.go:57 > error="Error #01: failed to find or persist pipeline config for qwerty287/woodpecker\n" ip=::1 latency=343.384457 method=POST path=/api/hook status=500 user-agent=Go-http-client/1.1
```

### System Info

```shell
latest next
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

PR #3128, opened 2024-01-07
Make config linking non-unique

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/3093

### DIFF
--- a/server/model/config.go
+++ b/server/model/config.go
@@ -35,6 +35,6 @@ type Config struct {
 
 // PipelineConfig is the n:n relation between Pipeline and Config
 type PipelineConfig struct {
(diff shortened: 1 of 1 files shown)


anbraten (approved):

6543:
we should not do that ... this is a hotfix at most!!

@qwerty287 could you check why the dublication did ocured in the first place.

or at lesat add a dedicated pipelineConfigID and retrive it based on that.

this table is just for n-m relation mapping and such should never have a dublicate!!!

6543:
I found the root cause

so i created a fix and add tests for that #3215

PR #3123, opened 2024-01-05
Kubernetes AppArmor and seccomp

zc-devs:
Closes #2545

seccomp
https://kubernetes.io/docs/tutorials/security/seccomp/
https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/135-seccomp/README.md

AppArmor
https://kubernetes.io/docs/tutorials/security/apparmor/
https://github.com/kubernetes/enhancements/blob/fddcbb9cbf3df39ded03bad71228265ac6e5215f/keps/sig-node/24-apparmor/README.md
Went ahead and implemented API from KEP-24 above.

PR #3135, opened 2024-01-08
Use UUID as podName and cleanup arguments for Kubernetes backend

6543:
to much args are just horrible to maintain. And we already have it nice structured stored as step.


pipeline/backend/kubernetes/service.go
```diff
@@ -56,24 +67,12 @@ func serviceName(step *types.Step) (string, error) {
...
-		StepLabel: podName,
-	}
-
-	svc, err := mkService(engine.config.Namespace, name, step.Ports, selector)
+	svc, err := mkService(step, engine.config.Namespace)
 	if err != nil {
 		return nil, err
 	}
 
+	log.Trace().Str("name", svc.Name).Interface("selector", svc.Spec.Selector).Interface("ports", svc.Spec.Ports).Msg("Creating service")
```
anbraten, line 75:
```suggestion
	log.Trace().Str("name", svc.Name).Interface("selector", svc.Spec.Selector).Interface("ports", svc.Spec.Ports).Msg("creating service")
```

pipeline/backend/kubernetes/utils_test.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func TestDNSName(t *testing.T) {
+	name, err := dnsName("wp_01he8bebctabr3kgk0qj36d2me_0_services_0")
+	assert.NoError(t, err)
+	assert.Equal(t, "wp-01he8bebctabr3kgk0qj36d2me-0-services-0", name)
+
+	name, err = dnsName("wp-01he8bebctabr3kgk0qj36d2me-0\\services-0")
+	assert.NoError(t, err)
+	assert.Equal(t, "wp-01he8bebctabr3kgk0qj36d2me-0\\services-0", name)
```
anbraten, line 30: Do we want those `\\` to be in the dns name? (Not that I expect the uuids to contain such chars 😅 )

6543: no idea but the current regex do allow it ...

6543: https://github.com/woodpecker-ci/woodpecker/blob/a63135363b513ba4bb13f0a721f4b7998cf3f85d/pipeline/backend/kubernetes/utils.go#L30

was added at #1661

@NeilHanlon might have more insight?

NeilHanlon: looking at the change, it appears, though I was the last to touch that line, I didn't write the regex. If it's incorrect it was incorrect before I touched it... promise! :P

6543: Haha no problem thanks for the fast headsup :)

It would be interesting why exactly this regex and not something else.

I guess i can just simplify it :)

NeilHanlon: in fact, i think that this is intended not to match `\\` but is just escaping the `\`, so it should be matching `\` followed by anything.. it's the 'followed by anything' bit that is problematic--and that should probably be `\\[a-z0-9]` without the `.`

6543: done

pipeline/backend/kubernetes/pod.go
```diff
@@ -378,12 +374,10 @@ func startPod(ctx context.Context, engine *kube, step *types.Step) (*v1.Pod, err
...
 	}
-
-	pod, err := mkPod(engine.config.Namespace, podName, step.Image, step.WorkingDir, engine.goos, step.BackendOptions.Kubernetes.ServiceAccountName,
-		step.Pull, step.Privileged,
-		step.Commands, step.Volumes, engine.config.ImagePullSecretNames,
-		engine.config.PodLabels, engine.config.PodAnnotations, step.Environment, step.BackendOptions.Kubernetes.NodeSelector,
-		step.ExtraHosts, step.BackendOptions.Kubernetes.Tolerations, step.BackendOptions.Kubernetes.Resources, step.BackendOptions.Kubernetes.SecurityContext, engine.config.SecurityContext)
+	pod, err := mkPod(step, &step.BackendOptions.Kubernetes, engine.config.Namespace, podName, engine.goos,
+		engine.config.ImagePullSecretNames,
+		engine.config.PodLabels, engine.config.PodAnnotations,
```
zc-devs, line 379: This comment, probably, is not relevant to this PR, but we should make this config immutable. Seems, we should make and pass a copy.
I've got an issue because of that. In #3123 I have the code below:
```go
	if securityContext != nil {
		key, value := apparmorAnnotation(name, securityContext.ApparmorProfile)
		if key != nil && value != nil {
			meta.Annotations[*key] = *value
		}
	}
```
This adds annotation in Agent's configuration. When I run pipelines without security context, then this annotation still exist and pod fails.
I can make a copy of map, but think, that making whole config copy is better.

pipeline/backend/kubernetes/pod.go
```diff
@@ -378,12 +374,10 @@ func startPod(ctx context.Context, engine *kube, step *types.Step) (*v1.Pod, err
...
 		return nil, err
 	}
-
-	pod, err := mkPod(engine.config.Namespace, podName, step.Image, step.WorkingDir, engine.goos, step.BackendOptions.Kubernetes.ServiceAccountName,
-		step.Pull, step.Privileged,
-		step.Commands, step.Volumes, engine.config.ImagePullSecretNames,
-		engine.config.PodLabels, engine.config.PodAnnotations, step.Environment, step.BackendOptions.Kubernetes.NodeSelector,
-		step.ExtraHosts, step.BackendOptions.Kubernetes.Tolerations, step.BackendOptions.Kubernetes.Resources, step.BackendOptions.Kubernetes.SecurityContext, engine.config.SecurityContext)
+	pod, err := mkPod(step, &step.BackendOptions.Kubernetes, engine.config.Namespace, podName, engine.goos,
+		engine.config.ImagePullSecretNames,
```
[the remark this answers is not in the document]
6543: same thought ... but kept it for an other day

pipeline/backend/kubernetes/pod.go
6543: golang has no concept of native immutable maps :/

pipeline/backend/kubernetes/pod.go
```diff
@@ -378,12 +374,10 @@ func startPod(ctx context.Context, engine *kube, step *types.Step) (*v1.Pod, err
 	if err != nil {
 		return nil, err
 	}
-
-	pod, err := mkPod(engine.config.Namespace, podName, step.Image, step.WorkingDir, engine.goos, step.BackendOptions.Kubernetes.ServiceAccountName,
-		step.Pull, step.Privileged,
-		step.Commands, step.Volumes, engine.config.ImagePullSecretNames,
-		engine.config.PodLabels, engine.config.PodAnnotations, step.Environment, step.BackendOptions.Kubernetes.NodeSelector,
-		step.ExtraHosts, step.BackendOptions.Kubernetes.Tolerations, step.BackendOptions.Kubernetes.Resources, step.BackendOptions.Kubernetes.SecurityContext, engine.config.SecurityContext)
+	pod, err := mkPod(step, &step.BackendOptions.Kubernetes, engine.config.Namespace, podName, engine.goos,
```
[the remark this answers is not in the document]
anbraten: True. We should probably use them from step as well.

pipeline/backend/kubernetes/pod.go
6543: well there are third-partys ...

https://github.com/benbjohnson/immutable
https://github.com/fishy/go-immutable
...
but if they work for us, i dont know. as we have to de-/serialize from/to json inbetween

6543: moved discussion into https://github.com/woodpecker-ci/woodpecker/discussions/3165

pipeline/backend/kubernetes/pod.go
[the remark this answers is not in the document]
6543: well just did it

zc-devs (approved):

anbraten (approved):

Issue #3109, opened 2024-01-03
Allow steps/services to have same name


qwerty287:
See https://github.com/woodpecker-ci/woodpecker/pull/3079

Services can't have names that steps have, also adding another step `clone` does not work.

I think we can also disallow this but then we should add a linter for it.

6543:
No woodpecker should just handle it!

We use uuid's for all steps internaly so equal names should not have any impact

6543:
(So i considder it a bug)

6543:
found it finally:  #3143

tested with docker backend and config:
```yml
steps:
    wait:
        image: "bash"
        commands:
            - "echo type: step"
            - "sleep 10s"
            - "echo hihi step"
            - "sleep 30s"

services:
    wait:
        image: "bash"
        commands:
            - "echo type: service"
            - "sleep 10s"
            - "echo hihi service"
            - "sleep 30s"
```

PR #3136, opened 2024-01-08
Use UUID instead of step name where possible

6543:
things I found while looking at  #3109

### DIFF
--- a/pipeline/backend/local/local.go
+++ b/pipeline/backend/local/local.go
@@ -203,7 +203,7 @@ func (e *local) WaitStep(_ context.Context, step *types.Step, taskUUID string) (
 
 	cmd, ok := state.stepCMDs[step.UUID]
 	if !ok {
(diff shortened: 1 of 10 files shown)


pipeline/backend/local/local.go
```diff
@@ -201,9 +201,9 @@ func (e *local) WaitStep(_ context.Context, step *types.Step, taskUUID string) (
 		return nil, err
 	}
 
-	cmd, ok := state.stepCMDs[step.Name]
+	cmd, ok := state.stepCMDs[step.UUID]
 	if !ok {
-		return nil, fmt.Errorf("step cmd %s not found", step.Name)
+		return nil, fmt.Errorf("step cmd %s not found", step.UUID)
```
anbraten, line 206:
```suggestion
		return nil, fmt.Errorf("step cmd for %s not found", step.UUID)
```

anbraten (approved):

PR #3137, opened 2024-01-08
Add step name as label to docker containers

6543:
and add a test


anbraten (approved):

anbraten:
@6543 Could you do the same for k8s?

6543:
no k8s is handled differently ... beside there is already a similar thing:
https://github.com/woodpecker-ci/woodpecker/blob/549d095ff35560b2bed6c6cc4c59fd9ef496f637/pipeline/backend/kubernetes/pod.go#L83

PR #3139, opened 2024-01-08
Remove duplicate assignment of `CI_STEP_NAME`

6543:
https://github.com/woodpecker-ci/woodpecker/blob/cd0c89ced9d87cb92f1b3fc1288ec66a200deecb/pipeline/frontend/metadata/environment.go#L86

will set it


anbraten (approved):
not tested, but hopefully correct. only case I could image why such a thing would be needed is for checking when conditions which does not seem to be the case here

6543:
the metadata package should be the single source of truth and is also called before we do eval ... else it's a bug anyway ;)

PR #3140, opened 2024-01-09
`cli exec`: let override existing environment values but print a warning

6543:


cli/exec/exec.go
```diff
@@ -121,13 +122,13 @@ func execWithAxis(c *cli.Context, file, repoPath string, axis matrix.Axis) error
...
+	pipelineEnv := make(map[string]string)
 	for _, env := range c.StringSlice("env") {
 		envs := strings.SplitN(env, "=", 2)
-		droneEnv[envs[0]] = envs[1]
-		if _, exists := environ[envs[0]]; exists {
-			// don't override existing values
-			continue
+		pipelineEnv[envs[0]] = envs[1]
+		if oldVar, exists := environ[envs[0]]; exists {
+			// let override existing values but print a warning
```
anbraten, line 130:
```suggestion
			// override existing values but print a warning
```

dominic-p:
Would this PR allow us to override things like the `$HOME` environment variable in the kubernetes backend using the [standard pipeline config](https://woodpecker-ci.org/docs/usage/environment)?

This came up in #3123 so I thought it would be worth double checking.

6543:
no this is just for the local execution via cli tool

dominic-p:
Thanks for the clarification. I'll open a separate issue about env override.

pat-s (approved):

PR #3141, opened 2024-01-09
Use step type to detect services in Kubernetes backend

6543:
and use the correct name for tail log

### DIFF
--- a/pipeline/backend/kubernetes/kubernetes.go
+++ b/pipeline/backend/kubernetes/kubernetes.go
@@ -173,8 +173,8 @@ func (e *kube) SetupWorkflow(ctx context.Context, conf *types.Config, taskUUID s
 
 	extraHosts := []types.HostAlias{}
 	for _, stage := range conf.Stages {
-		if stage.Alias == "services" {
-			for _, step := range stage.Steps {
+		for _, step := range stage.Steps {
+			if step.Type == types.StepTypeService {
 				svc, err := startService(ctx, e, step)
 				if err != nil {
 					return err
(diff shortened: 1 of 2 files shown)


pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -327,13 +332,14 @@ func (e *kube) TailStep(ctx context.Context, step *types.Step, taskUUID string)
...
 	return rc, nil
-
-	// rc := io.NopCloser(bytes.NewReader(e.logs.Bytes()))
-	// e.logs.Reset()
-	// return rc, nil
 }
 
 func (e *kube) DestroyStep(_ context.Context, step *types.Step, taskUUID string) error {
+	if step.Type == types.StepTypeService {
+		// this one should be stopped by DestroyWorkflow so we can ignore it
```
anbraten, line 339:
```suggestion
		// a service should be stopped by DestroyWorkflow so we can ignore it
```

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -196,6 +196,11 @@ func (e *kube) SetupWorkflow(ctx context.Context, conf *types.Config, taskUUID s
 
 // Start the pipeline step.
 func (e *kube) StartStep(ctx context.Context, step *types.Step, taskUUID string) error {
+	if step.Type == types.StepTypeService {
+		// this one should be started by SetupWorkflow so we can ignore it
```
anbraten, line 200:
```suggestion
		// a service should be started by SetupWorkflow so we can ignore it
```

anbraten (approved):

PR #3148, opened 2024-01-09
Make service icon rotate

6543:
make #3078 work as enum was wrong.

also add animation for color blind people

### DIFF
--- a/web/src/style.css
+++ b/web/src/style.css
@@ -150,3 +150,16 @@ body,
 .code-box-inline {
   @apply bg-wp-code-200 rounded-md text-wp-code-text-100;
 }
+
+/* TODO(2194): https://tailwindcss.com/docs/animation */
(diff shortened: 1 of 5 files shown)


6543:
image

web/src/style.css
```diff
@@ -150,3 +150,16 @@ body,
 .code-box-inline {
   @apply bg-wp-code-200 rounded-md text-wp-code-text-100;
 }
+
+/* TODO(2194): https://tailwindcss.com/docs/animation */
+@keyframes rotate-kf {
```
6543, line 155: if you have a better way just tell me :)

@anbraten

6543:
moved rotation too  #3149 and made this a bugfix only

6543:
skip-changelog -> was not released

PR #3149, opened 2024-01-09
Make service icon rotate

6543:
add animation for color blind people
[video](https://cloud.obermui.de/s/adpxqpMw6dEgZcC/download/recording_woodpecker-pull-3148.mkv)

### DIFF
--- a/web/src/style.css
+++ b/web/src/style.css
@@ -150,3 +150,16 @@ body,
 .code-box-inline {
   @apply bg-wp-code-200 rounded-md text-wp-code-text-100;
 }
+
+/* TODO(2194): https://tailwindcss.com/docs/animation */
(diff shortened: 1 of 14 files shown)


web/src/style.css
```diff
@@ -150,3 +150,16 @@ body,
 .code-box-inline {
   @apply bg-wp-code-200 rounded-md text-wp-code-text-100;
 }
+
+/* TODO(2194): https://tailwindcss.com/docs/animation */
+@keyframes rotate-kf {
```
6543, line 155: if you have a better way just tell me :)

@anbraten

6543:
Should not be importand for a service ... if it fails the pipeline will still be successfull in the end

PR #3156, opened 2024-01-09
More docker backend test remove more undocumented

6543:
remove Sysctls and IpcMode


6543:
sysctls: https://docs.openshift.com/container-platform/4.6/nodes/containers/nodes-containers-sysctls.html#nodes-containers-sysctls-about_nodes-containers-using

so we should not need them at all, as i dont expect a CI step in any means that it should change the agents host kernel setup

IpcMode: how the docker deamon communicates with the running container.
this should also not be changed.

we have this options, as original drone did just copy the **internal** docker container struct for there step configuration ...

if we do at some point have docker specific needs we should do so as we do with kubernetis now

6543:
beside we still have the rule: if it's not documented, its free to remove/change at any time :)

6543:
see https://github.com/woodpecker-ci/woodpecker/commit/0bb62be3034694f839682118eefade37f5203cea 

that was where is was embeded ... before the internal struct of the docker project was simpy used.

I want to remove more in the future :)

PR #3159, opened 2024-01-10
Enable `gocritic` and don't ignore globally

qwerty287:
Use `nolint` directives instead.

From #2960

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -161,28 +161,7 @@ linters:
     - unparam
     - wastedassign
     - whitespace
+    - gocritic
 
 run:
   timeout: 15m
-
-issues:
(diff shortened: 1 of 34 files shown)


xoxys:
Instead of adding nolint 100 times I would keep global exclude rules for files that are expect to be excluded.

xoxys:
I'm not sure if this has unwanted side effects. If, e.g. the host can't be parsed or is not set, the server should be terminated instead of just returning the error, no?

xoxys:
Ah I guess this is covered by the entrypoint now and we just need one `Fatal` linting exception per package?

6543 (approved):

PR #3162, opened 2024-01-10
Remove panic recovering

qwerty287:
from #3161


6543:
thanks - well lets have a look if it still causes bugs if we do remove it :)

6543 (approved):

PR #3169, opened 2024-01-11
Enable golangci linter goconst

xoxys:
Split out from https://github.com/woodpecker-ci/woodpecker/pull/2960

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -164,6 +164,13 @@ linters:
     - gocritic
     - nolintlint
     - stylecheck
+    - goconst
+
+issues:
+  exclude-rules:
+    - path: '_test.go'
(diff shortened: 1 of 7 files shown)


6543 (approved):

lafriks:
Not sure I like this linter 🤣

xoxys:
No strong opinion, feel free to close unmerged.

lafriks:
It's too strict on values as can be seen already in a lot of places need to add nolint comment, personally I don't like it because of it

xoxys:
Wel... There are only a few places where excludes were added... However.

xoxys:
Please make a decision first if the linter should be activated or not before I spend more time on it.

PR #3174, opened 2024-01-11
Move docker resource limit settings from server to agent

6543:
so you can set it per agent and not per server

related to  #3156 to cleanup the step definition


pipeline/backend/docker/flags.go
```diff
@@ -56,4 +56,37 @@ var Flags = []cli.Flag{
...
+	},
+	&cli.Int64Flag{
+		EnvVars: []string{"WOODPECKER_BACKEND_DOCKER_LIMIT_MEM", "WOODPECKER_LIMIT_MEM"},
+		Name:    "backend-docker-limit-mem",
+		Usage:   "maximum memory allowed in bytes",
+	},
+	&cli.Int64Flag{
+		EnvVars: []string{"WOODPECKER_BACKEND_DOCKER_LIMIT_SHM_SIZE", "WOODPECKER_LIMIT_SHM_SIZE"},
+		Name:    "backend-docker-limit-shm-size",
+		Usage:   "docker compose /dev/shm allowed in bytes",
```
6543, line 75:
```suggestion
		Usage:   "docker /dev/shm allowed in bytes",
```

lafriks:
Maybe server can still provide defaults?

6543:
no I'm against this ...

we e.g. set the tmp path for the local backend on the server side

lafriks:
I have no strong opinion about this but that would avoid it being breaking change

anbraten:
I think it the default fits well into the agent settings, specific settings however should probably be still settable via the pipeline config. We just need to move it into the backend config section.

6543:
well make it non breaking for now then ... but add a TODO comment so we can do so on next majour bump

also I add a docker specific config like with kubernetis for limiting per step

pipeline/backend/docker/flags.go
```diff
@@ -56,4 +56,37 @@ var Flags = []cli.Flag{
 		Name:    "backend-docker-volumes",
 		Usage:   "backend docker volumes (comma separated)",
 	},
+	//
+	// resource limit parameters
+	//
+	&cli.IntFlag{
+		Sources: cli.EnvVars("WOODPECKER_BACKEND_DOCKER_LIMIT_MEM_SWAP", "WOODPECKER_LIMIT_MEM_SWAP"),
+		Name:    "backend-docker-limit-mem-swap",
+		Usage:   "maximum swappable memory allowed in bytes",
```
6543, line 65:
```suggestion
		Usage:   "maximum memory used for swap in bytes",
```

docs/docs/91-migrations.md
```diff
@@ -4,6 +4,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 
 ## `next`
 
+- Move docker resource limit settings into agent configuration
```
6543, line 7:
```suggestion
- Move docker resource limit settings from server into agent configuration
```

pat-s (approved):
Fits better into the agent than server but I agree, it should be settable in the step config like for k8s, as every container usually has a different requirement.

Given it's breaking due to the move, it fits well into 3.0.

6543:
yes have it alser per stes should be an backend option like done for kubes settings. this will be a new issue to address ...

PR #3179, opened 2024-01-12
Deprecate HTTP config service

qwerty287:
As we have an addon system now, we can migrate http config services to them and remove support for http config services in the next major (so deprecate for now).

If you're currently using the http config service, you need to write the server yourself, there's only an example provided by us.
Then you can also write the addon I think. This is much faster than a HTTP request, simplifies typings and thus reduces errors etc. and does not require you to run an additional server.

Of course, we would need to update https://github.com/woodpecker-ci/example-config-service repository for this, I'm happy to open a PR for this.

### DIFF
--- a/cmd/server/setup.go
+++ b/cmd/server/setup.go
@@ -330,6 +330,7 @@ func setupConfigService(c *cli.Context) (config.Extension, error) {
 	}
 
 	if endpoint := c.String("config-service-endpoint"); endpoint != "" {
+		log.Warn().Msg("the HTTP config service is deprecated, use addons instead")
 		return config.NewHTTP(endpoint, server.Config.Services.SignaturePrivateKey), nil
 	}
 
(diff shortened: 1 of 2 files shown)


xoxys:
I tend to disagree and Im not sure yet if the addon system is really the way to go. The approach chosen by drone to extend parts by an external http service (https://docs.drone.io/extensions/overview/) is IMO way more flexible and easier do adopt by end users:

- extensions can be written in any language
- extensions are unrelated to the server version
- ...

xoxys:
Per repo doesnt really scale in my opinion. Just wanted to add my experience with drone in the last year and the http extension system was quiet handy.

6543:
I would also like to have it as core component.

6543:
As long as we dont have a good way for plugins to:
1. Have core components work on windows too
2. Deliver them in a "cloud native way"

I'm strictly against relaying on them

6543:
Plugins currently overload stuff instead of extending things. It"s great for thirdparty but not for our own components.

PR #3184, opened 2024-01-12
Document core development ideas

6543:


docs/docs/92-development/01-getting-started.md
```diff
@@ -1,5 +1,12 @@
 # Getting started
 
+## Core ideas
+
+- A configuration should never be [turing complete](https://en.wikipedia.org/wiki/Turing_completeness) (We have agents to exec things 🙂).
```
6543, line 5:
```suggestion
- A (e.g. pipeline) configuration should never be [turing complete](https://en.wikipedia.org/wiki/Turing_completeness) (We have agents to exec things 🙂).
```

?

6543:
looks good enouth to  me ... so i dont have to write it again and again - instead i can link to the docs now ;)

PR #3206, opened 2024-01-15
Add to agent exponential backoff for fetching new tasks

6543:


agent/runner.go
```diff
@@ -59,10 +98,10 @@ func (r *Runner) Run(runnerCtx context.Context) error {
 	// get the next workflow from the queue
 	work, err := r.client.Next(runnerCtx, r.filter)
```
anbraten, line 99: The server is blocking `Next` request until it has a job, so there should be no need to backoff here.

agent/runner.go
```diff
@@ -47,10 +59,37 @@ func NewRunner(workEngine rpc.Peer, f rpc.Filter, h string, state *State, backen
...
 		backend:  backend,
+		backOff: backOff{
+			pollInitialInterval: backOffInit,
+			pollMaxInterval:     backOffMax,
+		},
 	}
 }
 
 func (r *Runner) Run(runnerCtx context.Context) error {
+	retry := backoff.NewExponentialBackOff()
```
anbraten, line 70: Why do you want to backoff exponential instead of linear here?

anbraten (changes requested):

agent/runner.go
6543: Because if you fetched a task last time its more likely that there are some next to come

Else in the worst case wait 10s

6543:
With this in the worstcase we have ~1.7Mb per day

anbraten:
> With this in the worstcase we have ~1.7Mb per day

Did you enabled debug logs?

6543:
no ... that would be no bug then

image

6543:
:point_up: the mentioned bug should already be fixed ... so that's why i labled it enhancement not bug

6543:
and no the server should and do not block the agent on a next() call.

so just enable debug and then disable the agent on the server ... you can see that the agent trys to fetch in a loop nonstop.
why should we waist cpu resources by query if we can have a backoff

agent/rpc/client_grpc.go
```diff
@@ -34,16 +34,28 @@ import (
...
+	pollInitialInterval time.Duration
 }
 
 // NewGrpcClient returns a new grpc Client.
 func NewGrpcClient(conn *grpc.ClientConn) rpc.Peer {
 	client := new(client)
 	client.client = proto.NewWoodpeckerClient(conn)
 	client.conn = conn
+
+	client.pollInitialInterval = backOffInit
```
6543, line 56: this prepairs to make it config later

Issue #3210, opened 2024-01-17
Slow fetch of repos info after login when using Bitbucket Cloud forge


healdropper:
### Component

server

### Describe the bug

The current implementation of Bitbucket Forge makes an excessive number of API requests to fetch all repositories and the current user permissions associated with them. It performs an extra API call per every repository found.

This significantly impacts users, particularly those with access to numerous repositories, especially when the Bitbucket API experiences high latency. It can take minutes after logging in or pressing the 'Reload Repositories' button.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"dev"}
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

PR #3211, opened 2024-01-17
Retrieve all user repo perms with a single API call

healdropper:
This pull request addresses the issue https://github.com/woodpecker-ci/woodpecker/issues/3210.

Ideally, the Bitbucket API should include repository permissions when utilizing the 'get all repositories' endpoint. However, as it currently does not provide this information, a viable solution is to fetch all permissions for every repository and then employ a dictionary to associate each repository with its respective permissions.

Without implementing this fix, logging in becomes problematic for users with access to a substantial number of repositories (300+), as the process takes over 2 minutes to complete.

### DIFF
--- a/server/forge/bitbucket/bitbucket.go
+++ b/server/forge/bitbucket/bitbucket.go
@@ -203,6 +203,16 @@ func (c *config) Repos(ctx context.Context, u *model.User) ([]*model.Repo, error
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


anbraten (approved):
Code lgtm, haven't tested it

healdropper:
I will fix the failing test asap

6543 (approved):

Issue #2653, opened 2023-10-26
Allow loading more log lines


runephilosof-karnovgroup:
### Clear and concise description of the problem

Allow loading more log lines.

### Suggested solution

It seems the log lines are all sent from the backend to the browser, but just not rendered, because that would take too much memory in some cases.

1. scroll what content is rendered as the page is scrolled

Might result in a messy scrollbar movement, unless the real size of the document was calculated initially.

### Alternative

put a button to load all log lines (if that crashes the user's browser, the user can then just download the logs instead).

or increase the maxLineCount, I was able to load 3500 lines without problems (it seems the browser goes from 5 MB memory usage to 10 MB).

### Additional context

As mentioned in the code comment
https://github.com/woodpecker-ci/woodpecker/blob/3607886be2b20bc2df1d1b2adff1f41b9d313dc5/web/src/components/repo/pipeline/PipelineLog.vue#L161

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

6543:
note https://github.com/woodpecker-ci/woodpecker/pull/6250 adds a workaround

PR #3212, opened 2024-01-17
Load more pipeline log lines (500 => 5000)

runephilosof-karnovgroup:
I was able to load 3500 lines without problems (it seems the browser goes from 5 MB memory usage to 10 MB).

This way, we will hit the boundary less often, lowering the need to get lazy loading #2653 fixed


6543:
well the underlaying issue still remains ... I consider this a workaround and we should revert it asap lazyloading is implemented.

if we are going to merge this or on ... I leave it to the more frontend-related maintainers ;)

PR #3221, opened 2024-01-19
Switch to `slog`

qwerty287:
Use stdlib instead of zerolog.

slog misses two features from zerolog:
- `trace` level
- native error logging (I has to write a wrapper)

### DIFF
--- a/.woodpecker/docker.yaml
+++ b/.woodpecker/docker.yaml
@@ -1,11 +1,3 @@
-when:
-  - event: [pull_request, tag]
-  - event: push
-    branch:
-      - ${CI_REPO_DEFAULT_BRANCH}
-      - release/*
-      - renovate/*
-
 variables:
   - &golang_image 'docker.io/golang:1.22.0'
   - &node_image 'docker.io/node:21-alpine'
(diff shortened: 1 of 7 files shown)


6543:
@qwerty287 does slog have structured logging ?

zc-devs:
Why? What is the problem with Zerolog? What benefits does Slog bring?

zc-devs:
First off, I'm not strictly against this. If majority is for it, then OK.

> stdlib is better than an external library

Yes, if it provides the same functionality. But there are already drawbacks. I'm afraid, that we will want some feature, that Zerolog provides, but Slog does not. Then we have to find a way again.

Also there is aspect of laziness (my :) and application stability. We use library (yes, but battle-tested and standard de-facto, IMO), we know it works great. Who knows, what pitfalls does bring other implementation? And all risks are here when there are no pros for Slog and cons/bugs/something-important of Zerolog, except Slog is in core.
Example. Suppose, we have app in Spring. Can we rework it on core Java? Yes, but for what?

Besides, I prefer Zerolog's "stream" syntax over Slog:
> log.Error().Err(err).Str("step-uuid", w.stepUUID).Msg("fail to write pipeline log to peer")
> slog.Error("fail to write pipeline log to peer", logger.Error(err), slog.String("step-uuid", w.stepUUID))

It is just my opinion. We need more voices.

anbraten:
> Also there is aspect of laziness (my :) and application stability. 

That's a good point. So my vote can be rather treated as not against it, but also not willing to adjust it in the first place 😅

lafriks:
I'm not strictly against it but I'm somehow sceptical :)

PR #3222, opened 2024-01-19
Lint for event filter and deprecate `exclude`

qwerty287:
Closes https://github.com/woodpecker-ci/woodpecker/discussions/2174

- return bad habit error if no event filter is set
- If this is applied, it's useless to allow `exclude`s on events. Therefore, deprecate it together with `include`s which should be replaced by `base.StringOrSlice` later.


6543 (changes requested):
let me have a look into the discussion first ...

... as i told - discussions are a bad "representation" i normaly dont get either a notice nore the current consensus from it

pipeline/errors/error.go
```diff
@@ -14,6 +14,7 @@ const (
 	PipelineErrorTypeDeprecation PipelineErrorType = "deprecation" // using some deprecated feature
 	PipelineErrorTypeCompiler    PipelineErrorType = "compiler"    // some error with the config semantics
 	PipelineErrorTypeGeneric     PipelineErrorType = "generic"     // some generic error
+	PipelineErrorTypeBadHabit    PipelineErrorType = "bad_habit"   // some generic error
```
anbraten, line 17:
```suggestion
	PipelineErrorTypeBadHabit    PipelineErrorType = "bad_habit"   // some bad-habit error
```

docs/docs/91-migrations.md
```diff
@@ -7,6 +7,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 - Deprecated `steps.[name].group` in favor of `steps.[name].depends_on` (see [workflow syntax](./20-usage/20-workflow-syntax.md#depends_on) to learn how to set dependencies)
 - Removed `WOODPECKER_ROOT_PATH` and `WOODPECKER_ROOT_URL` config variables. Use `WOODPECKER_HOST` with a path instead
 - Pipelines without a config file will now be skipped instead of failing
+- Deprecated `includes` and `excludes` support from event filter
```
anbraten, line 10: We might directly remove it from the available docs as well, so no one will start using it.

anbraten (approved):

anbraten:
@6543 Did you had the chance to read the discussion?

6543:
Not realy ... and i dont get why we should remove exclude?

6543:
Thanks github ... i had a detailed text. I recap it in a tldr now: 

When filter currently has AND, OR and NAND, to build all combinations.

removing exclude is removing NAND

There is a workaround: use eval; but we should not encurage long eval statements if the simple filters just work fine

6543 (changes requested):
PS: mentioned dicsuccion does not argue at all why exclude justd be removed!

6543:
Ah ok ... well i do use it ... looks like i have to migrate on repo.

Only for events is fine but the codecomment do not indicate that

docs/docs/91-migrations.md
6543, line 10:
```suggestion
- Deprecated `includes` and `excludes` support from **event** filter
```

pipeline/frontend/yaml/linter/linter.go
```diff
@@ -263,10 +263,86 @@ func (l *Linter) lintDeprecations(config *WorkflowConfig) (err error) {
...
 
+	for i, c := range parsed.When.Constraints {
+		if len(c.Event.Exclude) != 0 {
+			err = multierr.Append(err, &errors.PipelineError{
+				Type:    errors.PipelineErrorTypeDeprecation,
+				Message: "Please only use allow lists for events",
+				Data: errors.DeprecationErrorData{
+					File:  config.File,
+					Field: fmt.Sprintf("when[%d].event", i),
+					Docs:  "https://woodpecker-ci.org/docs/next/usage/workflow-syntax#event-1",
```
6543, line 274:
```suggestion
					Docs:  "https://woodpecker-ci.org/docs/usage/workflow-syntax#event-1",
```

pipeline/frontend/yaml/linter/linter.go
```diff
@@ -263,10 +263,86 @@ func (l *Linter) lintDeprecations(config *WorkflowConfig) (err error) {
...
+	for _, step := range parsed.Steps.ContainerList {
+		for i, c := range step.When.Constraints {
+			if len(c.Event.Exclude) != 0 {
+				err = multierr.Append(err, &errors.PipelineError{
+					Type:    errors.PipelineErrorTypeDeprecation,
+					Message: "Please only use allow lists for events",
+					Data: errors.DeprecationErrorData{
+						File:  config.File,
+						Field: fmt.Sprintf("steps.%s.when[%d].event", step.Name, i),
+						Docs:  "https://woodpecker-ci.org/docs/next/usage/workflow-syntax#event",
```
6543, line 290:
```suggestion
						Docs:  "https://woodpecker-ci.org/docs/usage/workflow-syntax#event",
```

pipeline/frontend/yaml/linter/linter_test.go
```diff
@@ -172,3 +181,41 @@ func TestLintErrors(t *testing.T) {
...
+		from string
+		want string
+	}{
+		{
+			from: "steps: { build: { image: golang } }",
+			want: "Please set an event filter",
+		},
+		{
+			from: "when: [{branch: xyz}, {event: push}]\nsteps: { build: { image: golang } }",
+			want: "Please set an event filter",
```
6543, line 196: why does it suggest an event filter if global does already have it?

6543:
fixed the "code comment" issue ...
... I only found one issue left - but it's a minor thing and if it's hard to fix it now we can let it open for later

pipeline/frontend/yaml/linter/linter_test.go
6543: Well i guess, user having questions abour that message or not will finaly tell us

6543 (approved):

PR #764, opened 2022-02-08
Added support for release event (Github)

LamaAni:
As a github developer (both public and Enterprise) I would like to be able to catch and run CI on the release event, allowing to add artifacts for release via the github GUI. 

Reasons:
1. It is easy for all users (management/lower level devs) to release, revert, get artifacts etc from release page.
2. It is a part of the github UI.

Changes:
1. Added global webhook "release"
2. Added support for parsing webhook "release" to the github vendor.
   * No release commit. We are using the /refs/tags/... Ref.
   * No branch is associated.

PR #3226, opened 2024-01-20
Add release event

qwerty287:
Supersedes #764 

Bitbucket does not support release webhooks.

### DIFF
--- a/web/src/compositions/useTheme.ts
+++ b/web/src/compositions/useTheme.ts
@@ -1,7 +1,11 @@
 import { useColorMode } from '@vueuse/core';
 import { watch } from 'vue';
 
-const { store: storeTheme, state: resolvedTheme } = useColorMode({
+const {
+  store: storeTheme,
+  state: resolvedTheme,
+  system: systemTheme,
+} = useColorMode({
   storageKey: 'woodpecker:theme',
 });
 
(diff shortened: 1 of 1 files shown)


docs/docs/20-usage/50-environment.md
```diff
@@ -77,6 +77,7 @@ This is the reference list of all environment variables available to your pipeli
 | `CI_COMMIT_AUTHOR`               | commit author username                                                                                             |
 | `CI_COMMIT_AUTHOR_EMAIL`         | commit author email address                                                                                        |
 | `CI_COMMIT_AUTHOR_AVATAR`        | commit author avatar                                                                                               |
+| `CI_COMMIT_PRERELEASE`           | release is a pre-release (empty if event is not `release`)                                                         |
```
anbraten, line 80: I like that approach. Could we use sth similar for draft releases? Currently it will only handle release creation, right? Do you have any ideas how we could add update / delete in the long term if needed?

server/forge/gitea/parse.go
```diff
@@ -118,3 +121,20 @@ func parsePullRequestHook(payload io.Reader) (*model.Repo, *model.Pipeline, erro
 	pipeline = pipelineFromPullRequest(pr)
 	return repo, pipeline, err
 }
+
```
lafriks, line 124:
```suggestion
```

pat-s (approved):
Would be OK to get this into 2.3.0 - CI is passing and I assume @qwerty287 also tested this locally in a dev env.

pat-s (changes requested):
Test failures

pat-s (approved):

PR #3227, opened 2024-01-20
Parse backend options in backend

qwerty287:
Currently, backend options are parsed in the yaml parser.
This has some issues:
- backend specific code should be in the backend folders
- it is not possible to add backend options for backends added via addons

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,10 +14,10 @@ repos:
     hooks:
       - id: golangci-lint
   - repo: https://github.com/igorshubovych/markdownlint-cli
-    rev: v0.38.0
+    rev: v0.39.0
     hooks:
       - id: markdownlint
-        exclude: '^docs/versioned_docs/.*$'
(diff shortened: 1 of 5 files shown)


zc-devs (commented):
I like the idea in general 👍 

> added via plugins

Did you mean `addons`?

zc-devs (approved):

pipeline/backend/kubernetes/backend_options.go
```diff
@@ -0,0 +1,77 @@
+package kubernetes
+
+import (
+	"github.com/mitchellh/mapstructure"
```
6543, line 4: without looking into the lib - why was it used ?

anbraten (approved):
Apart from the discussions regarding the data format passed to the backend, code lgtm

6543:
Yes that concerns me too ... but i dont have time to look into it so, lgtm

anbraten:
Just to clarify. lgtm. Just wanted to ping as 6543 requested changes.

PR #3251, opened 2024-01-22
Reduce default privileged plugins

qwerty287:
We should not allow privileged to plugins we don't control ourselves.

Also, `plugins/docker` and `plugins/ecr` should be replaced by `docker-buildx`.

Does somebody know whether our buildx plugin has support for GCR?


xoxys:
This is a breaking change.

xoxys:
We should consider to not escate anything by default. Users should be aware of contsiners running in privilrged mode and should only enable it it required and knowing the impact.

PR #3252, opened 2024-01-22
chore: make services ephemeral on k8s

eliasscosta:
# Context

When we run `services` and a storage `rwx` is set to `false`, sometimes pipelines get stuck when the steps run in another host. 

```
  Warning  FailedMount     	19s	kubelet              	Unable to attach or mount volumes: unmounted volumes=[wp-01hmhhfewr9cwqsj2q4v440rwa-0-default], unattached volumes=[aws-iam-token wp-01hmhhfewr9cwqsj2q4v440rwa-0-default 
```

# How 
I propose to treat them as ephemeral pods without PVC and just serve the service as is. 
Removing the volume when the pod is a `services` and when `rwx` is `false`. To prevent unnecessary mount volumes.

### DIFF
--- a/pipeline/backend/kubernetes/pod.go
+++ b/pipeline/backend/kubernetes/pod.go
@@ -107,6 +107,10 @@ func podSpec(step *types.Step, config *config) (v1.PodSpec, error) {
 		Tolerations:        tolerations(step.BackendOptions.Kubernetes.Tolerations),
 		SecurityContext:    podSecurityContext(step.BackendOptions.Kubernetes.SecurityContext, config.SecurityContext),
 	}
+
+	if step.Type == types.StepTypeService {
+		return spec, nil
+	}
 	spec.Volumes, err = volumes(step.Volumes)
 	if err != nil {
 		return spec, err
@@ -146,6 +150,9 @@ func podContainer(step *types.Step, podName, goos string) (v1.Container, error) 
 		return container, err
 	}
 
+	if step.Type == types.StepTypeService {
+		return container, nil
+	}
 	container.VolumeMounts, err = volumeMounts(step.Volumes)
 	if err != nil {
 		return container, err
--- a/pipeline/backend/kubernetes/pod.go
+++ b/pipeline/backend/kubernetes/pod.go
@@ -108,7 +108,7 @@ func podSpec(step *types.Step, config *config) (v1.PodSpec, error) {
 		SecurityContext:    podSecurityContext(step.BackendOptions.Kubernetes.SecurityContext, config.SecurityContext),
 	}
 
-	if (step.Type == types.StepTypeService) && (config.StorageRwx == false) {
+	if (step.Type == types.StepTypeService) && (!config.StorageRwx) {
 		return spec, nil
 	}
 	spec.Volumes, err = volumes(step.Volumes)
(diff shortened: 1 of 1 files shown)


pipeline/backend/kubernetes/pod.go
```diff
@@ -146,6 +150,9 @@ func podContainer(step *types.Step, podName, goos string) (v1.Container, error)
 		return container, err
 	}
 
+	if (step.Type == types.StepTypeService) && (!storageRwx) {
+		return container, nil
```
zc-devs, line 154: `container` returns on line 161.
Gathering volume mounts (lines 156-159) should be wrapped in that `if`.

pipeline/backend/kubernetes/pod.go
```diff
@@ -107,6 +107,10 @@ func podSpec(step *types.Step, config *config) (v1.PodSpec, error) {
 		Tolerations:        tolerations(step.BackendOptions.Kubernetes.Tolerations),
 		SecurityContext:    podSecurityContext(step.BackendOptions.Kubernetes.SecurityContext, config.SecurityContext),
 	}
+
+	if (step.Type == types.StepTypeService) && (!config.StorageRwx) {
+		return spec, nil
```
zc-devs, line 112: The same as with volume mounts.

Why I'm disallowed to mount my volume Read-Only?

eliasscosta: Maybe I'm not clear. My main idea is to create services without PVC when the storage is read-only. 
Why? This is because the next step runs in another node making it impossible to mount.

zc-devs: > My main idea is to create services without PVC when the storage is read-only

I got it and asked:

> Why I'm disallowed to mount my volume in Read-Only?

[Volumes](https://woodpecker-ci.org/docs/administration/backends/kubernetes#volumes)
Why I disallowed from mounting [a secret (or configmap) with a file ca.crt of a self-hosted registry](https://github.com/woodpecker-ci/helm/pull/47) in a service, for example.

---

> the next step runs in another node making it impossible to mount

That is the cause. Seems should be fixed. But not by disallowing Read-Only mode in service.

Consider the example:
```yaml
skip_clone: true
services:
  server:
    image: alpine
    commands:
      - echo Hello from service
    ports:
      - 1234
steps:
  client:
    image: alpine
    commands:
      - echo Hello from step
  two:
    image: alpine
    commands:
      - echo Hello from step
  three:
    image: alpine
    commands:
      - echo Hello from step
```

What if all of them will be scheduled on separate node?
Why Read-Only mode matters? Why only in service?

Maybe [NodeSelector](https://woodpecker-ci.org/docs/administration/backends/kubernetes#nodeselector) will serve as a workaround.

zc-devs:
Seems, related https://github.com/woodpecker-ci/woodpecker/discussions/3068

pipeline/backend/kubernetes/pod.go
eliasscosta: I got. I try to consider the mount when is declared  in the step.

zc-devs: Then we should do some affinity like K3s do with their Local Storage. Haven't dug deep, though.

PR #3268, opened 2024-01-24
Rework addons (use rpc)

qwerty287:
Migrate addons to https://github.com/hashicorp/go-plugin.

Closes https://github.com/woodpecker-ci/woodpecker/discussions/3214

This will likely take a lot of time to complete, please don't review yet.

### DIFF
--- a/docs/docs/30-administration/11-forges/100-addon.md
+++ b/docs/docs/30-administration/11-forges/100-addon.md
@@ -20,8 +20,6 @@ WOODPECKER_ADDON_FORGE=/path/to/your/addon/forge/file
 
 In case you run Woodpecker as container, you probably want to mount the addon binary to `/opt/addons/`.
 
-Using an addon forge always overwrites Woodpecker's internal forge configuration.
-
 ### Bug reports
 
 If you experience bugs, please check which component has the issue. If it's the addon, **do not raise an issue in the main repository**, but rather use the separate addon repositories. To check which component is responsible for the bug, look at the logs. Logs from addons are marked with a special field `addon` containing their addon file name.


anbraten:
One thing I noticed when thinking about the config api again ist that we should probably decided how we allow admins & maybe users to use plugins. For forges I think this approach makes sense. The previous building seemed to be already causing some issues, but using pure http seems a bit off for sth deeply intergreated as forges are currently.

For sth like config, secrets & registries it would be nice to allow repo admins to use their own plugins. We could allow grpc for that as well, but I actually think the http approach similar to the config-api as I started it months ago in #915 could be better suited and would be easier for developers to implement?

pipeline/errors/error.go
```diff
@@ -35,12 +19,8 @@ type DeprecationErrorData struct {
 	Docs  string `json:"docs"`
 }
 
-func (e *PipelineError) Error() string {
-	return fmt.Sprintf("[%s] %s", e.Type, e.Message)
-}
-
-func (e *PipelineError) GetLinterData() *LinterErrorData {
-	if e.Type != PipelineErrorTypeLinter {
+func GetLinterData(e *types.PipelineError) *LinterErrorData {
```
mzampetakis, line 22: This func always returns nil. Something does not seem correct eve thought it wan't introduced from this PR.

.github/renovate.json
```diff
@@ -4,7 +4,7 @@
   "packageRules": [
     {
       "matchManagers": ["docker-compose"],
-      "matchFileNames": ["docker-compose.gitpod.yml"],
+      "matchFileNames": ["docker-compose.gitpod.yaml"],
```
anbraten, line 7: Should we move unrelated docs changes like this one to a separate PR?

anbraten: thanks

docs/docs/30-administration/11-forges/100-addon.md
```diff
@@ -0,0 +1,70 @@
...
+To use an addon forge, download the correct addon version. Then, you can add the following to your configuration:
+
+```ini
+WOODPECKER_ADDON_FORGE=/path/to/your/addon/forge/file
+```
+
+In case you run Woodpecker as container, you probably want to mount the addon binary to `/opt/addons/`.
+
+Using an addon forge always overwrites Woodpecker's internal forge configuration.
+
```
anbraten, line 24:
```suggestion
```

PR #3270, opened 2024-01-24
Make agent usable for external backends

qwerty287:
Currently, you can extend agents with custom backends using go plugins (addons).

With https://github.com/woodpecker-ci/woodpecker/pull/3268, this is migrated to RPC.

I don't think it's a good idea to use hashicorp go-plugins in the agent because they us RPC, but the agent uses gRPC too. So it's kinda duplicated.

Thus, I had the idea in this PR: Allow to import the agent as external package and run it with custom backends:

```go
package main

import (
	"go.woodpecker-ci.org/woodpecker/v2/cmd/agent/core"
)

func main() {
	core.RunAgent(<your backend>)
}
```

This takes care of everything else and allows you to easily develop your own backend.

This replaces the addon approach for backends.


anbraten:
> Instead of compiling all backends (currently docker, local, kubernetes) into one agent and allowing to select at runtime using env var, only compile one backend which is then automatically executed.

I am not sure about that point. Sure it would be a bit smaller (afaik it isn't that large anyways), but the maintenance overhead for us (building, releasing, adjusting them, ...) and for instance admins (downloading the correct agent, upps I got the wrong agent ...) could be expected to increase quite a bit. 

> I don't think it's a good idea to use hashicorp go-plugins in the agent because they us RPC, but the agent uses gRPC too. So it's kinda duplicated.

However both connections use totally different settings, scopes etc. I think we could allow custom agents to reuse the official agent code as a lib they can hook into, but would leave the standard agent as it is.

anbraten:
> So you say it would also be fine to connect the backend via RPC to agent?

That could also work, but I was mainly referring to the idea that the agent could also be use a lib. Where someone could add their own backend as well.

Apart from that are there currently any attempts to add a custom backend?

anbraten (approved):

PR #3271, opened 2024-01-24
Improve secret available checks

### LINKED ISSUE #2800: Linter: show warning if a secret is not populated to the current build due to filtering

Upstream issue: https://codeberg.org/Codeberg-CI/feedback/issues/40

### Suggested solution

Now that the UI linter is parsing the config, it could also check for secrets which are not inserted into the build because

- they are limited to certain images (still would actually create a permanent note then, not sure...)
- they are limited to certain event triggers which are not matched for the respective build

### Alternative

_No response_

### Additional context

_No response_

### Validations

anbraten:
closes #2800

- return more specific errors for missing secrets
  - secret not found
  - secret not allowed to be used by event
  - secret not allowed to be used by image
  - secret is not a plugin
- use some error handling for `secrets` and `settings -> from_secret` options


pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -102,6 +103,15 @@ func (c *Compiler) createProcess(container *yaml_types.Container, stepType backe
 		}
 	}
 
+	for _, requested := range container.Secrets.Secrets {
+		secret, ok := c.secrets[strings.ToLower(requested.Source)]
+		if ok && secret.Available(container) {
+			environment[strings.ToUpper(requested.Target)] = secret.Value
+		} else {
+			return nil, fmt.Errorf("secret %q not found or not allowed to be used", requested.Source)
```
anbraten, line 111: If a secret is requested as part of the `secrets` list property of a step we check here.

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -89,6 +89,7 @@ func (c *Compiler) createProcess(container *yaml_types.Container, stepType backe
 		workingdir = c.stepWorkdir(container)
 	}
 
+	// TODO: why only if not detached?
 	if !detached {
 		pluginSecrets := secretMap{}
 		for name, secret := range c.secrets {
```
anbraten, line 95: If a secret is requested through settings we do not check here right now?!

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -86,20 +86,38 @@ func (c *Compiler) createProcess(container *yaml_types.Container, stepType backe
...
+
+		err := secret.Available(container)
+		if err != nil {
+			return "", err
+		}
+
+		return secret.Value, nil
+	}
+
+	// TODO: why don't we pass secrets to detached steps?
```
anbraten, line 107: Why don't we allow `settings` options for detached steps / services?

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -86,20 +86,38 @@ func (c *Compiler) createProcess(container *yaml_types.Container, stepType backe
...
 	if !detached || len(container.Commands) != 0 {
-		workingdir = c.stepWorkdir(container)
+		workingDir = c.stepWorkingDir(container)
 	}
 
+	getSecretValue := func(name string) (string, error) {
+		name = strings.ToLower(name)
+		secret, ok := c.secrets[name]
+		if !ok {
+			return "", fmt.Errorf("secret %q not found or not available for this event", name)
```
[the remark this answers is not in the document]
anbraten: nice catch. should work now

PR #3299, opened 2024-01-30
Fix Gitpod: Gitea auth token creation

HerHde:
The Gitea CLI `generate-access-token` has updated the OAuth scopes and [posts some warning to the STDOUT when creating a token](https://github.com/go-gitea/gitea/issues/28758):

```
.../setting/security.go:168:loadSecurityFrom() [W] Enabling Query API Auth tokens is not recommended. DISABLE_QUERY_AUTH_TOKEN will default to true in gitea 1.23 and will be removed in gitea 1.24.
0adftokenblablubb
```

When logging in to the Gitpod Gitea instance from Woodpecker, I get an error:

Authorization failed: Client ID not registered

This PR fixes the problems leading you to a dialogue authorizing Woodpecker after logging in to Gitea.


anbraten (approved):
not tested code lgtm

PR #3300, opened 2024-01-30
Revert "Add release event"

anbraten:
Reverts woodpecker-ci/woodpecker#3226


anbraten:
> I prefer the first option. You can still filter out pre-releases and drafts with whens.

I would be fine with that as well. So should we add a note to the docs and maybe the `CI_RELEAE_TYPE` directly?

PR #3314, opened 2024-02-01
Allow editing all environment variables in pipeline popups

### LINKED ISSUE #3292: UI/UX: Apply Non-Empty Environment Variables

When triggering a manual pipeline run in the web UI, I often forget to apply env vars with the `+ Add variable` button, as I interpret it as adding _another_ variable.

image

### Suggested solution

- Apply all variables with non-empty variable names
- Add a new input line once the variable name is not empty
- Keep the variables editable
- Remove the `+ Add variable` button

Example from GitLab:

image

### Alternative

_No response_

### Additional context

Might also be useful if my issue #3291 gets implemented

### Validations

lukashass:
Closes #3292 

[Screencast from 02.02.2024 00:15:21.webm](https://github.com/woodpecker-ci/woodpecker/assets/32853499/a52aa9bb-4c12-4e49-a631-5a96663789dd)


pat-s:
How will this look like for non-empty vars in the config? IIUC #3292 proposes to show/list existing env vars in the config and then allow adding additional ones.

Also, if so, the existing ones from the config should be highlighted somehow so they can be distinguished from the additional ones.

lukashass:
I basically tried to rebuild the functionality of Gitlab because I think that was a good suggestion.

@HerHde is this what you imagined?

HerHde:
> I understood #3292 differently that the env var in the field where you enter it was not added with the `Add variable ` button and thus not added to the pipeline. The issue requests to also respect this env var.

This is what I meant

> I basically tried to rebuild the functionality of Gitlab because I think that was a good suggestion.
> 
> @HerHde is this what you imagined?

Perfect!

Issue #3330, opened 2024-02-05
(Kubernetes backend) terminated node causes runtime error when handling step exit code


fernandrone:
### Component

agent

### Describe the bug

We were running Woodpecker v2.1.1 on Kubernetes backend at a multi-node cluster on AWS.

We've got a few `panic: runtime error` logs in our agent like this:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x16f7594]

goroutine 99699 [running]:
[go.woodpecker-ci.org/woodpecker/v2/pipeline/backend/kubernetes.(*kube).WaitStep(0xc00050a140](http://go.woodpecker-ci.org/woodpecker/v2/pipeline/backend/kubernetes.%28*kube%29.WaitStep%280xc00050a140), {0x1e32748, 0xc0005048c0}, 0xc00116c500?, {0xc0024529b0, 0x5})
    /src/pipeline/backend/kubernetes/kubernetes.go:251 +0x594
[go.woodpecker-ci.org/woodpecker/v2/pipeline.(*Runtime).exec(0xc001d80b80](http://go.woodpecker-ci.org/woodpecker/v2/pipeline.%28*Runtime%29.exec%280xc001d80b80), 0xc00116c500)
    /src/pipeline/pipeline.go:269 +0x196
[go.woodpecker-ci.org/woodpecker/v2/pipeline.(*Runtime).execAll.func1()](http://go.woodpecker-ci.org/woodpecker/v2/pipeline.%28*Runtime%29.execAll.func1%28%29)
    /src/pipeline/pipeline.go:206 +0x1ba
[golang.org/x/sync/errgroup.(*Group).Go.func1()](http://golang.org/x/sync/errgroup.%28*Group%29.Go.func1%28%29)
    /src/vendor/[golang.org/x/sync/errgroup/errgroup.go:75](http://golang.org/x/sync/errgroup/errgroup.go:75) +0x56
created by [golang.org/x/sync/errgroup.(*Group).Go](http://golang.org/x/sync/errgroup.%28*Group%29.Go) in goroutine 41
    /src/vendor/[golang.org/x/sync/errgroup/errgroup.go:72](http://golang.org/x/sync/errgroup/errgroup.go:72) +0x96
```

[...]

PR #3331, opened 2024-02-05
fix: agent panic when node is terminated during step execution

fernandrone:
Fixes https://github.com/woodpecker-ci/woodpecker/issues/3330

This adds error handling on the agent's WaitStep function, on two sections where it could encounter a `panic: runtime error: invalid memory address or nil pointer dereference` in case it could no longer access complete information about a specific pod.

This error was found to happen if the node in which the pod was running was terminated during the step's execution.
spite active pipelines being executed on the node.

Now instead of a panic on the agent's logs and undefined behavior on the UI it will display a more helpful error message on the UI.

### Additional context

We observed the bug first on v2.1.1, but tested the fix internally on top of 2.3.0. 

image

### DIFF
--- a/pipeline/backend/kubernetes/kubernetes.go
+++ b/pipeline/backend/kubernetes/kubernetes.go
@@ -32,7 +32,6 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
-
 	// To authenticate to GCP K8s clusters
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"k8s.io/client-go/rest"
--- a/pipeline/backend/kubernetes/kubernetes.go
+++ b/pipeline/backend/kubernetes/kubernetes.go
@@ -264,17 +264,19 @@ func (e *kube) WaitStep(ctx context.Context, step *types.Step, taskUUID string) 
 	}
 
 	if isImagePullBackOffState(pod) {
-		return nil, fmt.Errorf("could not pull image for pod %s", pod.Name)
+		return nil, fmt.Errorf("could not pull image for pod %s", podName)
 	}
 
 	if len(pod.Status.ContainerStatuses) == 0 {
-		return nil, fmt.Errorf("no container statuses found for pod %s", pod.Name)
+		return nil, fmt.Errorf("no container statuses found for pod %s", podName)
 	}
 
 	cs := pod.Status.ContainerStatuses[0]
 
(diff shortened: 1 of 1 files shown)


jesielconceicao (approved):

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -266,8 +266,18 @@ func (e *kube) WaitStep(ctx context.Context, step *types.Step, taskUUID string)
...
 	}
 
+	if len(pod.Status.ContainerStatuses) == 0 {
+		return nil, fmt.Errorf("No container statuses found for pod %s", pod.Name)
+	}
+
+	cs := pod.Status.ContainerStatuses[0]
+
+	if cs.State.Terminated == nil {
+		return nil, fmt.Errorf("No terminated state found for container %s/%s", pod.Name, cs.Name)
```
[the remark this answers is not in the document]
fernandrone: A bit, yeah

I could add something like this

```
	if cs.State.Terminated == nil {
		err := fmt.Errorf("no terminated state found for container %s/%s", pod.Name, cs.Name)
		log.Error().Str("taskUUID", taskUUID).Interface("state", cs.State).Msg(err.Error())
		return nil, err
	}
```

Note that when I caught this issue the node had been killed mid-execution, and there were kubernetes events for terminating the pod correctly. So I guess it just never got "assigned" a termination status, but the pod was indeed terminated. In other words, whatever status it reported instead of terminated was certainly "bonkers".

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -263,11 +263,23 @@ func (e *kube) WaitStep(ctx context.Context, step *types.Step, taskUUID string)
...
+
+	if len(pod.Status.ContainerStatuses) == 0 {
+		return nil, fmt.Errorf("no container statuses found for pod %s", podName)
+	}
+
+	cs := pod.Status.ContainerStatuses[0]
+
+	if cs.State.Terminated == nil {
+		err := fmt.Errorf("no terminated state found for container %s/%s", podName, cs.Name)
+		log.Error().Str("taskUUID", taskUUID).Str("pod", podName).Str("container", cs.Name).Interface("state", cs.State).Msg(err.Error())
```
6543, line 277: Could you use `.Err(err)` and as msg that this is not expected or something more apropriated?

6543: Hmm never mind we generate the error two lines above

6543 (approved):

PR #3341, opened 2024-02-07
Add new experimental preprocessor

anbraten:
#2473

Can be enabled by using `WOODPECKER_EXPERIMENTAL_GO_PREPROCESSOR=true`
The legacy pre-processor will run as well. If it should be disabled this can be done by setting: `WOODPECKER_EXPERIMENTAL_DISABLE_LEAGAY_PREPROCESSOR=true`

Both settings are un-documented and this PR will be hidden from changelogs, so we can test them first with `ci.woodpecker-ci.org` and elaborate if we should continue this path.


6543:
This will break usage where we use e.g. string replace or shortening within yaml!

anbraten:
> This will break usage where we use e.g. string replace or shortening within yaml!

What do you mean. It will be opt-in!

anbraten:
> What do you think about this approach? Am I missing important downsides?

In general a interesting approach. Could be a bit tedious to implement for all properties / yaml value types.

Something I would really like to be able to do is a basic tenary operation to do sth like:

```yml
steps:
  - name: docker build
    image: buildx
    settings:
      dry-run: {{ if eq .env.CI_PIPELINE_EVENT "pull_request"  }}true{{ else }}false{{ end }}
```

But I can also image options where it would be helpful to write sth like:

```yml
steps:
  - name: docker build
    image: buildx
    {{ if eq .env.CI_PIPELINE_EVENT "pull_request"  }}
    depends_on:
      - test
    {{ end }}
    settings:
      ...
```

anbraten:
What do you think about adding this pre-processor behind an un-documented feature flag in addition to the `${{ ... }}` syntax for some time to test it, so we can simply remove it at any time again if we notice that it is too powerful / complicated, ...

lafriks:
Personally I'm very skeptical about this change, helm charts are using a similar approach and it has its problems, especially quirks with spaces etc, also it adds much complexity imho

Also it makes it invalid yaml file... For that should be used a different format, ex. starlak etc

anbraten:
The go templates can be provided using a config plugin. For the envsubst I am currently trying to create a version that would retain an env var if not set. Either based on https://github.com/a8m/envsubst or  https://github.com/drone/envsubst

Issue #3344, opened 2024-02-07
K8S backend new behaivor when storage rwx is `false`


eliasscosta:
### Clear and concise description of the problem

When we run a `service` step or use the `depends_on` (run parallel steps) and a storage RWX is set to `false`, sometimes pipelines get stuck when the steps (pods) run in another node. 

We get the error to mount the volumes.

```shell
  Warning  FailedMount     	19s	kubelet              	Unable to attach or mount volumes: unmounted volumes=[wp-01hmhhfewr9cwqsj2q4v440rwa-0-default], unattached volumes=[aws-iam-token wp-01hmhhfewr9cwqsj2q4v440rwa-0-default 
```

This is expected because of the way Kubernetes works.

### Suggested solution

My suggested solution is to add a `PodAffinity` in the podSpec configuration when the storage RWX is `false` it will force it to run the in the same node.

### Alternative

Another solution is to set the pods to use emptyDir volumes. 

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

pat-s:
@eliasscosta I recently configured an instance with `WOODPECKER_BACKEND_K8S_STORAGE_RWX: false` and couldn't replicate the issue so far. I see/acknowledge that it *might* potentially happen though.

Once a step is finished, the volume should be unmounted again, making it possible for the next step to mount it. No matter which node the step runs on.
You can run into issues if the volume is not in the same region as the node which should mount it, though this is actually a different issue. Or if the unmount process takes too long and then the volume is not available when the pod for the next step gets scheduled.

I understand that when steps run on different nodes with RWO, the time it takes to unmount and mount again might slow down the pipeline.

Regarding your suggested solution: I am not sure how this could be implemented. The affinity would need to be set dynamically after the first step in the pipeline got assigned. This is tricky as it is done by the k8s scheduler based on resources and placement restrictions.
To me, the only possible way to solve this more reliably would be to manually set selectors/affinity for each step, ensuring that all steps get scheduled to the same node.

@zc-devs @xoxys Might also be interesting to hear your thoughts on it!

zc-devs:
> RWX is set to false
[...]

PR #3345, opened 2024-02-07
chore: add podAffinity and labels to identify the pipeline

eliasscosta:
addressing https://github.com/woodpecker-ci/woodpecker/issues/3344

The idea is to run in the same node when we need to run parallel or service steps and have a storage RWX is false.

### DIFF
--- a/pipeline/backend/kubernetes/pod.go
+++ b/pipeline/backend/kubernetes/pod.go
@@ -31,8 +31,11 @@ import (
 )
 
 const (
-	StepLabel = "step"
-	podPrefix = "wp-"
+	StepLabel           = "step"
+	podPrefix           = "wp-"
+	PipelineNumberLabel = "pipeline_number"
+	RepositoryLabel     = "repository"
+	WorkflowNameLabel   = "workflow_name"
 )
 
 func mkPod(step *types.Step, config *config, podName, goos string, options BackendOptions) (*v1.Pod, error) {
@@ -43,7 +46,7 @@ func mkPod(step *types.Step, config *config, podName, goos string, options Backe
 		return nil, err
 	}
 
-	spec, err := podSpec(step, config, options)
+	spec, err := podSpec(step, config, affinityLabels(step.Environment), options)
 	if err != nil {
 		return nil, err
 	}
@@ -93,6 +96,14 @@ func podMeta(step *types.Step, config *config, options BackendOptions, podName s
 		meta.Labels[ServiceLabel] = step.Name
 	}
 
+	affinityLabels := affinityLabels(step.Environment)
+
+	if len(affinityLabels) > 0 && !config.StorageRwx {
+		meta.Labels[RepositoryLabel] = affinityLabels[RepositoryLabel]
+		meta.Labels[PipelineNumberLabel] = affinityLabels[PipelineNumberLabel]
+		meta.Labels[WorkflowNameLabel] = affinityLabels[WorkflowNameLabel]
+	}
+
 	meta.Annotations = config.PodAnnotations
 	if meta.Annotations == nil {
 		meta.Annotations = make(map[string]string)
@@ -113,7 +124,22 @@ func stepLabel(step *types.Step) (string, error) {
 	return toDNSName(step.Name)
 }
 
-func podSpec(step *types.Step, config *config, options BackendOptions) (v1.PodSpec, error) {
+func affinityLabels(env map[string]string) map[string]string {
+	affinityLabels := make(map[string]string)
+
+	if val, ok := env["CI_REPO_NAME"]; ok {
+		affinityLabels[RepositoryLabel] = val
+	}
+	if val, ok := env["CI_PIPELINE_NUMBER"]; ok {
+		affinityLabels[PipelineNumberLabel] = val
+	}
+	if val, ok := env["CI_WORKFLOW_NAME"]; ok {
+		affinityLabels[WorkflowNameLabel] = val
+	}
+	return affinityLabels
+}
+
+func podSpec(step *types.Step, config *config, affinityLabels map[string]string, options BackendOptions) (v1.PodSpec, error) {
 	var err error
 	spec := v1.PodSpec{
 		RestartPolicy:      v1.RestartPolicyNever,
@@ -124,6 +150,11 @@ func podSpec(step *types.Step, config *config, options BackendOptions) (v1.PodSp
 		Tolerations:        tolerations(options.Tolerations),
 		SecurityContext:    podSecurityContext(options.SecurityContext, config.SecurityContext),
 	}
+
+	if len(affinityLabels) > 0 && !config.StorageRwx {
+		spec.Affinity = podAffinity(affinityLabels)
+	}
+
 	spec.Volumes, err = volumes(step.Volumes)
 	if err != nil {
 		return spec, err
@@ -169,6 +200,37 @@ func podContainer(step *types.Step, podName, goos string, options BackendOptions
 	}
 
 	return container, nil
+}
+
+func podAffinity(affinityLabels map[string]string) *v1.Affinity {
+	return &v1.Affinity{
+		PodAffinity: &v1.PodAffinity{
+			RequiredDuringSchedulingIgnoredDuringExecution: []v1.PodAffinityTerm{
+				{
+					LabelSelector: &metav1.LabelSelector{
+						MatchExpressions: []metav1.LabelSelectorRequirement{
+							{
+								Key:      RepositoryLabel,
+								Operator: metav1.LabelSelectorOpIn,
+								Values:   []string{affinityLabels[RepositoryLabel]},
+							},
+							{
+								Key:      PipelineNumberLabel,
+								Operator: metav1.LabelSelectorOpIn,
+								Values:   []string{affinityLabels[PipelineNumberLabel]},
+							},
+							{
+								Key:      WorkflowNameLabel,
+								Operator: metav1.LabelSelectorOpIn,
+								Values:   []string{affinityLabels[WorkflowNameLabel]},
+							},
+						},
+					},
+					TopologyKey: v1.LabelHostname,
+				},
+			},
+		},
+	}
 }
 
 func volumes(volumes []string) ([]v1.Volume, error) {
--- a/pipeline/backend/kubernetes/pod_test.go
+++ b/pipeline/backend/kubernetes/pod_test.go
@@ -348,3 +348,357 @@ func TestFullPod(t *testing.T) {
 	ja := jsonassert.New(t)
 	ja.Assertf(string(podJSON), expected)
 }
+
+func TestTinyPodAffinity(t *testing.T) {
+	expected := `
+	{
+		"metadata": {
+			"name": "wp-01he8bebctabr3kgk0qj36d2me-0",
+			"namespace": "woodpecker",
+			"creationTimestamp": null,
+			"labels": {
+				"step": "build-via-gradle",
+				"pipeline_number": "2323",
+				"repository": "woodpecker",
+				"workflow_name": "02-tests"
+			}
+		},
+		"spec": {
+			"affinity": {
+				"podAffinity": {
+						"requiredDuringSchedulingIgnoredDuringExecution": [
+								{
+										"labelSelector": {
+												"matchExpressions": [
+														{
+																"key": "repository",
+																"operator": "In",
+																"values": [
+																		"woodpecker"
+																]
+														},
+														{
+																"key": "pipeline_number",
+																"operator": "In",
+																"values": [
+																		"2323"
+																]
+														},
+														{
+															"key": "workflow_name",
+															"operator": "In",
+															"values": [
+																	"02-tests"
+															]
+														}
+												]
+										},
+										"topologyKey": "kubernetes.io/hostname"
+								}
+						]
+				}
+			},
+			"volumes": [
+				{
+					"name": "workspace",
+					"persistentVolumeClaim": {
+						"claimName": "workspace"
+					}
+				}
+			],
+			"containers": [
+				{
+					"name": "wp-01he8bebctabr3kgk0qj36d2me-0",
+					"image": "gradle:8.4.0-jdk21",
+					"command": [
+						"/bin/sh",
+						"-c"
+					],
+					"args": [
+						"echo $CI_SCRIPT | base64 -d | /bin/sh -e"
+					],
+					"workingDir": "/woodpecker/src",
+					"env": [
+						"<<UNORDERED>>",
+						{
+							"name": "CI",
+							"value": "woodpecker"
+						},
+						{
+							"name": "HOME",
+							"value": "/root"
+						},
+						{
+							"name": "SHELL",
+							"value": "/bin/sh"
+						},
+						{
+							"name": "CI_SCRIPT",
+							"value": "CmlmIFsgLW4gIiRDSV9ORVRSQ19NQUNISU5FIiBdOyB0aGVuCmNhdCA8PEVPRiA+ICRIT01FLy5uZXRyYwptYWNoaW5lICRDSV9ORVRSQ19NQUNISU5FCmxvZ2luICRDSV9ORVRSQ19VU0VSTkFNRQpwYXNzd29yZCAkQ0lfTkVUUkNfUEFTU1dPUkQKRU9GCmNobW9kIDA2MDAgJEhPTUUvLm5ldHJjCmZpCnVuc2V0IENJX05FVFJDX1VTRVJOQU1FCnVuc2V0IENJX05FVFJDX1BBU1NXT1JECnVuc2V0IENJX1NDUklQVAoKZWNobyArICdncmFkbGUgYnVpbGQnCmdyYWRsZSBidWlsZAo="
+						},
+						{
+							"name": "CI_PIPELINE_NUMBER",
+							"value": "2323"
+						},
+						{
+							"name": "CI_REPO_NAME",
+							"value": "woodpecker"
+						},
+						{
+							"name": "CI_WORKFLOW_NAME",
+							"value": "02-tests"
+						}
+					],
+					"resources": {},
+					"volumeMounts": [
+						{
+							"name": "workspace",
+							"mountPath": "/woodpecker/src"
+						}
+					]
+				}
+			],
+			"restartPolicy": "Never"
+		},
+		"status": {}
+	}`
+
+	pod, err := mkPod(&types.Step{
+		Name:        "build-via-gradle",
+		Image:       "gradle:8.4.0-jdk21",
+		WorkingDir:  "/woodpecker/src",
+		Pull:        false,
+		Privileged:  false,
+		Commands:    []string{"gradle build"},
+		Volumes:     []string{"workspace:/woodpecker/src"},
+		Environment: map[string]string{"CI": "woodpecker", "CI_PIPELINE_NUMBER": "2323", "CI_REPO_NAME": "woodpecker", "CI_WORKFLOW_NAME": "02-tests"},
+	}, &config{
+		Namespace:  "woodpecker",
+		StorageRwx: false,
+	}, "wp-01he8bebctabr3kgk0qj36d2me-0", "linux/amd64", BackendOptions{})
+	assert.NoError(t, err)
+
+	podJSON, err := json.Marshal(pod)
+	assert.NoError(t, err)
+
+	ja := jsonassert.New(t)
+	ja.Assertf(string(podJSON), expected)
+}
+
+func TestFullPodAffinity(t *testing.T) {
+	expected := `
+	{
+		"metadata": {
+			"name": "wp-01he8bebctabr3kgk0qj36d2me-0",
+			"namespace": "woodpecker",
+			"creationTimestamp": null,
+			"labels": {
+				"app": "test",
+				"step": "go-test",
+				"pipeline_number": "2323",
+				"repository": "woodpecker",
+				"workflow_name": "02-tests"
+			},
+			"annotations": {
+				"apps.kubernetes.io/pod-index": "0"
+			}
+		},
+		"spec": {
+			"affinity": {
+				"podAffinity": {
+						"requiredDuringSchedulingIgnoredDuringExecution": [
+								{
+										"labelSelector": {
+												"matchExpressions": [
+														{
+																"key": "repository",
+																"operator": "In",
+																"values": [
+																		"woodpecker"
+																]
+														},
+														{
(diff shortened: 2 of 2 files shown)


jesielconceicao (approved):

pipeline/backend/kubernetes/pod.go
```diff
@@ -113,17 +120,49 @@ func stepLabel(step *types.Step) (string, error) {
...
-		Tolerations:        tolerations(step.BackendOptions.Kubernetes.Tolerations),
-		SecurityContext:    podSecurityContext(step.BackendOptions.Kubernetes.SecurityContext, config.SecurityContext),
+	spec := v1.PodSpec{}
+
+	repoLabel, hasLabelRepo := labels["repository"]
+	pipelineNumberLabel, hasLabelPipelineNumber := labels["pipeline_number"]
+
+	// force service and steps to run in the same node when rwx is false
+	// maybe in the future get theses values from backend configuration
+	if !config.StorageRwx && hasLabelRepo && hasLabelPipelineNumber {
```
6543, line 132: we should have tests for that :)

eliasscosta: done, do you see any other test we need to do?

pipeline/backend/kubernetes/pod.go
```diff
@@ -113,17 +120,49 @@ func stepLabel(step *types.Step) (string, error) {
...
+		spec.Affinity = &v1.Affinity{
```
zc-devs, line 133: Make separate method/function, please.

pipeline/backend/kubernetes/pod.go
```diff
@@ -113,17 +120,49 @@ func stepLabel(step *types.Step) (string, error) {
...
+			PodAffinity: &v1.PodAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: []v1.PodAffinityTerm{
+					{
+						LabelSelector: &metav1.LabelSelector{
+							MatchExpressions: []metav1.LabelSelectorRequirement{
+								{
+									Key:      "repository",
```
zc-devs, line 140: Make constant, please.

pipeline/backend/kubernetes/pod.go
```diff
@@ -113,17 +120,49 @@ func stepLabel(step *types.Step) (string, error) {
...
+									Operator: metav1.LabelSelectorOpIn,
+									Values:   []string{repoLabel},
+								},
+								{
+									Key:      "pipeline_number",
```
zc-devs, line 145: Constant

pipeline/backend/kubernetes/pod.go
```diff
@@ -113,17 +120,49 @@ func stepLabel(step *types.Step) (string, error) {
...
+									Operator: metav1.LabelSelectorOpIn,
+									Values:   []string{pipelineNumberLabel},
+								},
+							},
+						},
+						TopologyKey: "kubernetes.io/hostname",
```
zc-devs, line 151: https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetesiohostname

pipeline/backend/kubernetes/pod.go
```diff
@@ -93,6 +93,13 @@ func podMeta(step *types.Step, config *config, podName string) (metav1.ObjectMet
 		meta.Labels[ServiceLabel] = step.Name
 	}
 
+	if val, ok := step.Environment["CI_REPO_NAME"]; ok {
+		meta.Labels["repository"] = val
+	}
+	if val, ok := step.Environment["CI_PIPELINE_NUMBER"]; ok {
+		meta.Labels["pipeline_number"] = val
+	}
```
zc-devs, line 101: I propose to make separate map `affinityLabels` in `mlPod()` and supply it to `podMeta` and `podSpec` then.
Also, isn't there some pipeline ID besides repo-name/pipeline-number pair?

pipeline/backend/kubernetes/pod_test.go
```diff
@@ -351,3 +351,458 @@ func TestFullPod(t *testing.T) {
 	ja := jsonassert.New(t)
 	ja.Assertf(string(podJSON), expected)
 }
+
+func TestFullPodAffinity(t *testing.T) {
```
zc-devs, line 355: If you want separate test for affinity, then you don't need security context, ports, etc.

pipeline/backend/kubernetes/pod_test.go
```diff
@@ -351,3 +351,458 @@ func TestFullPod(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	podJSON, err := json.Marshal(pod)
+	assert.NoError(t, err)
+
+	ja := jsonassert.New(t)
+	ja.Assertf(string(podJSON), expected)
+}
+
+func TestFullPodWithoutAffinity(t *testing.T) {
```
zc-devs, line 596: Isn't it covered by `TinyPodTest`?

pipeline/backend/kubernetes/pod.go
eliasscosta: I don't get it, is better to do it in the podMeta? is related to metadata as well. Then podSpec receives the labels and mounts the affinity block.

zc-devs: Yeah, but podSpec receives all labels then... not just required affinity ones...
And also we can wrap labels creation in `if` and do not label Pod at all:
```
affinityLabels = createAffinityLabels(step)
podMeta(..., affinityLabels)
podSpec(..., affinityLabels)
```
```
func createAffinityLabels(step) map {
  affinityLabels : map
  if !config.StorageRwx { # or whatever condition, config maybe
    # fill affinity map from environment 
  }
 return affinityLabels
}
```
Now we always label Pod. Don't know, if it's bad or not, though.

pat-s (commented):
Affinity does not guarantee that a pod is placed on the same node, it is just a "recommendation", if possible. Why did you choose this one instead of `nodeSelector`?

What if the user adds manual affinity or nodeSelector settings into `backend_options`?

(We use spaces instead of tabs - this is also enforced by the linter in CI. Please see the editorconfig & friends.)

eliasscosta:
> Affinity does not guarantee that a pod is placed on the same node, it is just a "recommendation", if possible. Why did you choose this one instead of `nodeSelector`?
> 
Yes, we already use the `nodeSelector` and it selects a node from a pool with the same name/label, not the same node where the initial (service/step) starts to run.  Then when it happened we got the error to to mount the volume. The idea is to try to run a service or pod(steps) in parallel, to achieve it, we need to run the pods in the same node. 

> What if the user adds manual affinity or nodeSelector settings into `backend_options`?
> 

Today we don't have the configuration to change the affinity in the `backend_options`, if someone sets the nodeSelector it continues to work in the same way. Because this configuration I proposal is just when we have Storage RWX as `false`  

> (We use spaces instead of tabs - this is also enforced by the linter in CI. Please see the editorconfig & friends.)
>

I don't get it, do you have the link to it?

pipeline/backend/kubernetes/pod.go
```diff
@@ -31,8 +31,11 @@ import (
 )
 
 const (
-	StepLabel = "step"
-	podPrefix = "wp-"
+	StepLabel           = "step"
+	podPrefix           = "wp-"
+	PipelineNumberLabel = "pipeline_number"
```
zc-devs, line 36: As we use [dashes](https://github.com/woodpecker-ci/woodpecker/blob/134fb7900c514d4498ce02767715bb16d0d46f36/pipeline/backend/kubernetes/utils_test.go#L26), I suggest to change to `pipeline-number` and `workflow-name`.

pat-s:
> https://github.com/woodpecker-ci/woodpecker/pull/3252#discussion_r1462139468
As I understand. Suppose, we have selector

No, `nodeSelector` is a harder scheduling setting than `affinity`. 

- nodeSelector: pods are *only* required on nodes which match the key:value
- affinity: pods are *preferred* to run on nodes which match the key:value but won't be blocked to also run on other nodes

In the end, it all comes down to the key:value labels used and how "unique" they are to specific nodes.

> and have 50 nodes with AMD processors. Then volume and all steps may be scheduled on separate nodes anyway.

Yes, because the key:value applies to multiple nodes. Yet if any approach is desired to fully control pod scheduling then it should be `nodeSelector` and not `affinity` (regardless of the key:value pairs used).

pat-s (changes requested):
I've thought about this proposal again, and I am pretty sure this PR shouldn't be merged and labeled as 'wontfix'. Here's why:

- RWX is used primarily to solve exactly such cases, i.e. step pods are allowed to be scheduled on different nodes and RWX can be attached to different nodes, providing the required underlying sources to all nodes
- A RWO workaround is not really feasible since the only unique identifier for a specific single node is its IP. An alternative for this (beside `nodeSelector`) is to use [tains and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/), yet this is also just partly addressing the issue if there are multiple "CI nodes".

I don't see a dynamic workaround in which WP infers on which node it started and then uses the same one dynamically for future steps of the same pipeline (please correct me if I am overlooking something).

In the same way, I'd argue that having a RWX is a basic element of a cluster and that implementing complex RWO scheduling workarounds are not worth it in contrast to the potential gains.

zc-devs:
> to fully control pod scheduling then it should be nodeSelector and not affinity

Yes, but we (users) don't need to full control. We don't want to manually pick up the Node. We want `run this Pods on the same Node, where those Volume was created`.

This is how K3s Local Path Provisioner works:
1. Create PVC:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-test
  namespace: test-woodpecker-runtime
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1G
```
2. Kubernetes creates [PVC](https://github.com/woodpecker-ci/woodpecker/files/14222472/pvc-1.yaml.txt) in `pending` status.
3. Create Pod:
```
apiVersion: v1
kind: Pod
metadata:
  name: pod-test
  namespace: test-woodpecker-runtime
spec:
  containers:
    - name: test
      image: alpine
      command:
        - /bin/sh
        - "-c"
      args:
        - sleep 300
      volumeMounts:
        - name: vol
          mountPath: /woodpecker
  volumes:
    - name: vol
      persistentVolumeClaim:
        claimName: pvc-test
```
4. Kubernetes creates creates [PV](https://github.com/woodpecker-ci/woodpecker/files/14222498/pv-2.yaml.txt), [binds PVC](https://github.com/woodpecker-ci/woodpecker/files/14222499/pvc-2.yaml.txt)
5. and [creates Pod](https://github.com/woodpecker-ci/woodpecker/files/14222520/pod-2.yaml.txt)

All Pods [run on the same node](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) as Volume:
```
spec:
  nodeName: l7-1
```

I would like if it was implemented like above, except binding step:
1. Create Volume.
2. Get the Node.
3. Create steps/Pods on the Node from step 2.

zc-devs:
> I am pretty sure this PR shouldn't be merged and labeled as 'wontfix'

I disagree. I can translate current PR's implementation into `run whole pipeline on the same Node` requirements / feature request, which in itself deserves attention, although I prefer another solution.

pat-s:
> We want run this Pods on the same Node, where those Volume was created.

I understand what you want but still, `affinity` is the wrong approach for this as it is not at all guaranteeing this. When you're operating on more nodes, you have no guarantee for a pod to be scheduled on a specific node, yet at all to infer which one a specific PVC is attached to in a dynamic way. 

To achieve what you want, you likely want to consider running all steps in a single pod instead of creating a new pod for each step (this is not possible right now). This has been proposed somewhere else already (can't find it), and there was some discussion about it. I think this should be a way to go as it would also address some requests related to caching.

Going this way, the whole pipeline would run on the same node without a complex logic to dynamically infer where the next upcoming pod should go to due to the need of using a RWO.
This would also speed up k8s execution in general since the mount operation and pod creation would only happen once.

zc-devs:
> The only way I can think of would be

Seems, [you are right](https://stackoverflow.com/questions/62034907/how-to-find-my-persistent-volume-location) in case I proposed. If it looks too complicated, I'm fine with current PR's implementation in general.
However, we can get rid of rwx/rwo condition logic and just bring new kubernetes backend option like `try-to-run-on-same-node`. Does it suit you @pat-s, @eliasscosta?

> multi-container approach

This will implemented (if) a couple years from now. User has a problem now and even proposed a fix.

pat-s:
> This will implemented (if) in a couple years from now. User has a problem now and even proposed a fix.

Not if someone pushed into this direction. It is actually not too complicated to get this implemented. Definitely not more complicated than implementing the workaround of this PR. 

> Seems, [you are right](https://stackoverflow.com/questions/62034907/how-to-find-my-persistent-volume-location) in case I proposed. If it looks too complicated, I'm fine with current PR's implementation in general.
However, we can get rid of rwx/rwo condition logic and just bring new kubernetes backend option like try-to-run-on-same-node. Does it suit you @pat-s, @eliasscosta?

I am not fine with the current PR, though I am not in the position of solely blocking such PRs. But I see just this not being a solid solution for the issue at hand, it will cause additional issues in the future and makes use of `affinity` in a way it shouldn't be used. 
If this gets merged, it will likely not be removed anymore either and being workaround another time and so forth. If it works for OP and his setup, this might be a candidate for a fork which needs to be synced with upstream.

Also, there are additional solutions which are actually more "simple": use a RWX, which should be a basic component on every cluster. 

Overall, let me use a metaphor to describe the discussion here how it unfolds in my mind: If green and red cars exists, you also don't buy a red one and paint it green but rather sell the red one and buy a green one instead 🙃 (might not be a 100% fit but it should transport the message).

eliasscosta:
Thanks, everyone, it was a rich discussion and I decided to implement it outside to resolve our case. I already have an injector to add more labels for troubleshooting and then I will add affinity logic too. 

We use AWS to run the k8s and RWX isn't an option for us, our primary tests the steps run too slowly using (EFS) and if we use other solutions (like: nfs operator) inside of the cluster maybe turn too complex the cluster management

I'm not comfortable with the multi-container approach for my case, I guess is too complex for a simple pod creation. But is an option to think about. 

I try to follow the Kubernetes [doc](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#types-of-inter-pod-affinity-and-anti-affinity) and use the options we have when is not possible to use the RWX. 

Maybe If we have more labels on the pod and a way to set up the Affinity by an option in the agent would be great.

pat-s:
I wonder what you're doing when RWX is limiting WRT to performance. Judging from my experience within the last years, I usually get suspicious when RWX is being blamed to be the major issue unless it comes to webservers and other tool operating with many small files which require very small latency.

Note that you can also attach additional volumes to a pod which can be used for certain compute tasks. Or the node's tempdir. You can also copy the repo to this volume before performing any tasks and clean up afterward (or set a storageClass policy to delete the volume after use). Not everything must necessarily happen on the RWX holding the repo data.

eliasscosta:
> I wonder what you're doing when RWX is limiting WRT to performance. Judging from my experience within the last years, I usually get suspicious when RWX is being blamed to be the major issue unless it comes to webservers and other tool operating with many small files which require very small latency.
> 
> Note that you can also attach additional volumes to a pod which can be used for certain compute tasks. Or the node's tempdir. You can also copy the repo to this volume before performing any tasks and clean up afterward (or set a storageClass policy to delete the volume after use). Not everything must necessarily happen on the RWX holding the repo data.

Yes, the problem is the volume created by default for the pipeline, if we use a service the next pod try to get the same volume. If they don't run in the same node the pipeline get stuck running forever trying to mount.

fernandrone:
Thanks everyone for the comments. I'm a colleague to @eliasscosta so maybe I can bring some extra perspective.

> I wonder what you're doing when RWX is limiting WRT to performance. Judging from my experience within the last years, I usually get suspicious when RWX is being blamed to be the major issue unless it comes to webservers and other tool operating with many small files which require very small latency.

We don't have any benchmarks at hand but our issue with Amazon EFS is that it was painfully slow for many common i/o intensive tasks. That would be the likes of `npm install` (which I think that classifies as requiring "many small files"). We're talking about 3-4x slower when compared to our default 3000 iops gp3 disks, so even small projects would take 20 minutes to install their dependencies.

It's possible that there's more that we could've done on that front, and unfortunately that was some while ago so I don't think I can get exact numbers and configuration to help. That's what ultimately pushed us away from EFS, so I'm not sure that would ever be an alternative for many workloads.

> Note that you can also attach additional volumes to a pod which can be used for certain compute tasks. Or the node's tempdir. You can also copy the repo to this volume before performing any tasks and clean up afterward (or set a storageClass policy to delete the volume after use). Not everything must necessarily happen on the RWX holding the repo data.

Maybe I don't have the right picture of how this would work in practice, but if we assume that the NFS is unusably slow for many CI tasks, I can see this being a big burden on users:

1. project is cloned to NFS volume
2. copy project to separate ReadWriteOnce volume
3. run `npm install`
4. copy back to NFS so the result may be used by the next pod in the pipeline

And that assuming that the copy operation itself wouldn't be a bottleneck.

> Affinity does not guarantee that a pod is placed on the same node, it is just a "recommendation", if possible

@pat-s it seems part of the issue here (not everything, I understand that you also have valid criticisms against the design itself, not just the implementation) is this, that affinity supposedly does not _guarantee_ anything. But doesn't it?

I believe that `preferredDuringSchedulingIgnoredDuringExecution` indeed does not guarantee anything. However `requiredDuringSchedulingIgnoredDuringExecution`, as Elias implemented, should have guarantees, and will refuse to schedule if the rules are not met. It's explained here in the documentation:

image

The quote above is for the nodeAffinity rules, but still, the same logic applies for podAffinity. Am I missing something? I understand if you prefer another design but I believe this code guarantees the pods are scheduled on the same node 100% of the time (or else they're not scheduled, which is technically a possibility, i.e. if a node does not have enough resources...).

> What if the user adds manual affinity or nodeSelector settings into backend_options?

I agree this could conflict with this implementation. In fact, it seems this could cause a regression and break working pipelines. Thus if we were to move forward (and I have a vested interest in moving forward because I'd prefer to not have to run a fork) I do like the idea of an opt-in flag like `try-to-run-on-same-node`.

pat-s:
@fernandrone Thanks for the extensive reply, appreciated.

Yes, `requiredDuringSchedulingIgnoredDuringExecution` can used so that affinity behaves essentially like `nodeSelector`, yet this is usually not what it should be used for (even though, as noted, a more expressive syntax can be used).
Until now I was aware that this was the actual core idea of the implementation because when I read "affinity" I think of "soft affinity" and not the idea to invoke it as an implicit fixed `nodeSelector`.

### RWX

Yes, RWX can be slow when installing *many* packages like node or R packages. Yet if this happens in parallel, and we're not talking about 1k packages, it is usually not adding up to 20 mins. 
Of course the use case of copying to a RWO tempdir is not something that should become the default, yet in your case, I think it would help to workaround and provide a much easier solution than the proposed affinity solution.

### Using RWO in general

I see the advantages of having a RWO as the base PV. And we surely want to make everything better and not limit users or provide a bottlenecked system in any way. 
However, a workaround is often "quickly" implemented but not often the best overall solution.
I still think that using the "one-pod-many-containers" solution is by far the best solution for the underlying problems, and also comes with the benefit of less time spent between the pod creations. Which is why my proposal would be to invest time on this approach and not to implement the proposed solution of this PR as it would also interfere with user-provided `backend_options` overwrites.

Disclaimer: This is just my opinion, and I am only one maintainer. If others join the discussion and overwrite my vote, I am happy to pursue another way. Overall: Just here to help and make things better/less complicated in the long run, not to block anything from being improved. 🤞

zc-devs:
> The way I see it is, as of now, Woodpecker on a Kubernetes cluster doesn't work fully without a NFS volume: both services and depends_on (as well as the deprecated group) will not work properly and may altogether fail to start if the pods are scheduled on the wrong node. A NFS volume fixes this but from my limited experience introduces performance issues on common CI workloads that many teams will not find tolerable.

The simplest solution is using [Local Path Provisioner](https://github.com/rancher/local-path-provisioner). It provides local RWO Volumes and binds Pods to the same Node automagically. Works well in self-hosted environment. Don't know how it goes in a cloud, though. Seems, you have a reason not to use it...

pat-s:
@fernandrone My proposed workaround is of course nothing that scales and not a solution for many pipelines or general use. Yet, I didn't know your use cases and environment before.

> I really like the idea of running every containers on the same pod and FWIW that is how Tekton worked when I used it a couple of years ago. There seems to be a lot of advantages to that approach. However to me it also sounds like a much more complex change. I'm not sure if is is already something that is on the roadmap of the project and universally agreed upon.

So far it has only been a rough idea, also to tackle caching issues. The RWX approach in place right now has been there before I joined the project actively. I don't know why that decision was made, and hence can't comment on it. I am aware that Drone uses a single pod approach and that transitioning from Drone to WP would be more seamless if WP would do the same.

The only downside of that approach would be that resource-based scheduling would only be possible on the pipeline level and not step level, but I personally would see that rather as an improvement as it avoids having to define `backend_options` for every step.

> My suggestion would be that, unless the maintainers are committed to change the architecture from one-pod-per-step to one-pod-per-workflow (or are confident that this change can be made reasonably soon), it's reasonable to find a good enough solution within the current architecture 👍🏻

I see your need but I would still highly vote to push the "one-pod-many-containers" approach and stay a little patient until that is done (unless another maintainer is against it).
There could also be the option to keep the current model (for backward comp etc.) and add the "one-pod-many-containers" on the side with an option to switch to it. I would be happy to work on it, as it not only solves your problem, it simplifies caching and removes the need for a RWX/volume mount in the helm chart (as a pipeline could also just operate on `emptyDir` then within a pod).

Would you and your team be motivated to go this route? And to clarify again: I would not want to merge this one as it actually would imply updating the docs and the chart and then potentially revert everything if the "one-pod-many-containers" is put in place (as this one makes the other idea practically useless).

I am aware that you would like to see an instant merge to continue working - but until then, there's also your for which could be used in the meantime.

zc-devs:
[RFC: one Pod per Workflow (K8s)](https://github.com/woodpecker-ci/woodpecker/discussions/3366)

6543:
If there is a lot of arguing aganst & for it, i would say it should be a config option

pat-s:
Let's continue discussing in https://github.com/woodpecker-ci/woodpecker/discussions/3366 (thanks @zc-devs). 

@6543 With the little but important difference that it is not a yes/no discussion here but more about "which approach to use" and "not to hurry but avoiding adding a lot of code which then must be migrated/removed again for a (possibly) better solution".

Let's close here for now and find an agreement first which most/a majority is happy with.

PR #3348, opened 2024-02-07
Extend approval options

anbraten:
related to #336

To protect secrets and (agents) it should be possible to require approvals for potential harmful pipelines like PRs. To not require every pipeline to be approved the existing gated mode was replaced with 3 options addressing most common cases:

- allow every pipeline except PRs from forks
- require approval for all PRs (new default)
- require approval for every pipeline (current require approval / gated mode)

Those modes could later on be extended. Additionally a webhook extension could be added to get the approval state from an external service.

# Changes
- [x] replace `IsGated` with `RequireApproval`
- [x] migrate old settings to `RequireApproval`
- [x] adjust docs


xoxys:
> Can't you create an interface which just checks whether the pipeline needs approval or not so we can easily extend the feature by adding a new implementation? Maybe also with support for an external http extension?

If we need that level of flexibility, why not use something like [open policy agent](https://www.openpolicyagent.org/), as proposed in the issue? Repo rules can be written by users and added via UI by repo admins without writing external services and without blowing up the server code for every single use case that we have not yet thought about.

anbraten:
> In my opinion, we should support more here.

Sure, we can make it highly adjustable, but on the other hand, to keep things simple (which is often a quite import benefit / feature as well for the community) it's sometimes more helpful to support just the most common options IMO.

> If we need that level of flexibility, why not use something like [open policy agent](https://www.openpolicyagent.org/), as proposed in the issue? Repo rules can be written by users and added via UI by repo admins without writing external services and without blowing up the server code for every single use case that we have not yet thought about.

I don't think [open policy agent](https://www.openpolicyagent.org/) would be that helpful as we still need to provide context data (usernames, if user is admin / maintainer / ..., amount of pipeline executed for a context (like PR), ...) to be even able to write rules 
 as suggested here: https://github.com/woodpecker-ci/woodpecker/discussions/2293#discussioncomment-6774908

xoxys:
Sure, but implementing something like a `SecurityContext` would be required for external services as well.

> it's sometimes more helpful to support just the most common options IMO.

To be honest (please don't get me wrong) we also implement stuff like cli auto-updates cli auto-setup etc. just because it's convenient. This also adds code complexity, and we still do it even if it is easier to just let users download the binary from a release. Especially for security related features, flexibility is key IMO.

6543:
i still want to have the gated feature to be extendable

xoxys:
Yes, please... if we dont want to have more options in woodpecker core (I still dont understand why) please provide an http interface like drone ci does so users can implement validation.

lafriks:
There should be an option to not require approval from forks as this would be quite problematic for company internal instances imho

xoxys:
> There should be an option to not require approval from forks as this would be quite problematic for company internal instances imho

Fully agree. Otherwise, this PR is also breaking, as right now approvals can be disabled entirely.

cli/repo/repo_update.go
```diff
@@ -39,6 +39,10 @@ var repoUpdateCmd = &cli.Command{
 			Name:  "gated",
 			Usage: "repository is gated",
 		},
+		&cli.StringFlag{
+			Name:  "require-approval",
+			Usage: "repository require approval",
```
xoxys, line 44:
```suggestion
			Usage: "repository requires approval",
```

xoxys: Or

```suggestion
			Usage: "repositories require approval",
```

cli/repo/repo_update.go
```diff
@@ -87,8 +92,32 @@ func repoUpdate(ctx context.Context, c *cli.Command) error {
...
+			allEvents := "all_events"
+			patch.RequireApproval = &allEvents
+		} else {
+			forks := "forks"
+			patch.RequireApproval = &forks
+		}
+	}
+	if c.IsSet("require-approval") {
+		switch requireApproval {
+		case "forks", "pull_requests", "all_events":
```
xoxys, line 107: Can we use constants for these strings?

xoxys: Ive seen there is already `RequireApprovalAllEvents` etc. can we not use it here as well?

anbraten:
> There should be an option to not require approval from forks as this would be quite problematic for company internal instances imho

Was thinking about this. So you basically have one fork per employee?

cli/repo/repo_update.go
anbraten: I just copied the types from server/models to woodpecker-go now.

docs/docs/20-usage/project-settings.png
xoxys, line 1: On the screenshot is no radio button selected. Shouldn't "No approval required" selected by default? IMO, we should change the default in the upcoming v3 to "Pull request from forks" then.

anbraten: Updated it. Used `ci.woodpecker-ci.org` as api for the UI and that wasn't returning any value for the new setting 😅 

Currently it sets the level to `require approval for all pull-requests` for new repos and migrates existing ones like this:
- gated => still require approval for everything
- prs enabled flag checked => require approval for all prs
- else => require approval for forks (if someone as lafriks mentioned wants to completely disable approvals they would need to explicitly change this)

Should I change some part of this?

cli/repo/repo_update.go
```diff
@@ -87,8 +92,31 @@ func repoUpdate(ctx context.Context, c *cli.Command) error {
 	if c.IsSet("trusted") {
 		patch.IsTrusted = &trusted
 	}
+	// TODO: remove isGated in next major release
 	if c.IsSet("gated") {
-		patch.IsGated = &gated
+		if gated {
+			patch.RequireApproval = &woodpecker.RequireApprovalAllEvents
+		} else {
+			patch.RequireApproval = &woodpecker.RequireApprovalForks
```
xoxys, line 100: Changing the behavior for `gated == false` is also breaking and should be avoided if this PR should go into 2.7.x

xoxys: Lets discuss in the other thread.

docs/docs/20-usage/project-settings.png
xoxys: For existing repos we should not change the behavior in v2.7.x

> * gated => still require approval for everything

That's fine

> * prs enabled flag checked => require approval for all prs

That's breaking, we can and should do it that way but not in v2.7.x

> * else => require approval for forks (if someone as lafriks mentioned wants to completely disable approvals they would need to explicitly change this)

Same here.

pat-s:
The milestone was set to 2.7.2 - @anbraten I think this should be 3.0?

WRT to backporting to 2.x: is this really needed? This is quite a big feature and if at all, this should go into a point release but not a patch release. I see the tendency lately to start backporting more into 2.x and keeping that alive/around longer - is that the idea?

anbraten:
@pat-s I believe we should release this as soon as possible as a robust measure to address security risks. We’ve implemented multiple fixes recently - #3934, #4076, #3933, #3909, #4053, and #3918, to name a few - each balancing security, UX, and potential breaking changes. 

While researching how other CI systems manage cases like the potentially "risky" `PATH` environment variable 🙈, it’s clear they often avoid such restrictions entirely. The rationale? Strict limitations can hinder legitimate use cases, and restricting these features without impacting users is challenging. Consequently, they largely rely on not running risky pipelines in the first place - a pragmatic approach IMO.

The only reliable solution is to avoid executing potentially dangerous code from untrusted sources without prior review. To that end, I'd like to adjust our security approach to incorporate this chain of trust, addressing the risks outlined [here](https://github.com/woodpecker-ci/woodpecker/discussions/4234#discussioncomment-10990275):

- **Repo admins**: Ensure only "safe" code is executed. If trust is in question, conduct a review.
- **Agent admins**: Choose which repos/orgs to trust, and limit features to those within your control.

lafriks:
> > There should be an option to not require approval from forks as this would be quite problematic for company internal instances imho
> 
> Was thinking about this. So you basically have one fork per employee?

Mostly we have product teams but there could be coming fixes from other teams also and in such cases it would be fork

server/model/repo.go
```diff
@@ -20,6 +20,15 @@ import (
 	"strings"
 )
 
+type ApprovalMode string
+
+var (
+	RequireApprovalNone         ApprovalMode = "none"          // require approval for no events
+	RequireApprovalForks        ApprovalMode = "forks"         // require approval for PRs from forks (default)
+	RequireApprovalPullRequests ApprovalMode = "pull_requests" // require approval for all PRs
+	RequireApprovalAllEvents    ApprovalMode = "all_events"    // require approval for all events
```
6543, line 29:
```suggestion
	RequireApprovalAllEvents    ApprovalMode = "all_events"    // require approval for all external events
```

web/src/assets/locales/en.json
```diff
@@ -493,5 +489,14 @@
...
   "access_denied": "You are not allowed to access this instance",
-  "invalid_state": "The OAuth state is invalid"
+  "invalid_state": "The OAuth state is invalid",
+  "require_approval": {
+    "require_approval_for": "Require approval for",
+    "none": "No approval required",
+    "none_desc": "This setting can be dangerous and should only be used on private forges where all users are trusted.",
+    "forks": "Pull request from forked repositories",
+    "pull_requests": "All pull requests",
+    "all_events": "All events",
```
6543, line 499: all events description: all events that are external, so cron and manual events will run right away ...

6543:
> The milestone was set to 2.7.2 - @anbraten I think this should be 3.0?
> 
> WRT to backporting to 2.x: is this really needed? This is quite a big feature and if at all, this should go into a point release but not a patch release. I see the tendency lately to start backporting more into 2.x and keeping that alive/around longer - is that the idea?
> 
> Besides: great feature, thanks a lot for that! It will make administrating public instances a lot easier!

this can never be v2.7.x ... and we hold back v2.7.2 since for toooo long ...

If this should be released asap too ... just let us make a v2.8 release branch after v2.7.2 and continue on that ... also we could cut out v3 soonish - we wont get everything anyway .... even things I want to see, so just stick to do that into a new cycle ...

6543:
I still think we could add https://www.openpolicyagent.org/ or something alike .... but the current implementation is already better, and for better UX having simple pre-selected options is a good choice in any case

docs/docs/20-usage/project-settings.png
anbraten: I adjusted the logic when changing using the gated property via api to map `no gated` to `no approval required`. 

For the migration I changed it to:
- gated => approval for all events
- public repo => approval for forks (I would argue this is fine as the option was not available before, although everyone would have chosen it)
- private / internal repo & not gated => no approval required (this should allow teams as @lafriks mentioned to use internal forks without approval)

server/model/repo.go
```diff
@@ -41,8 +50,8 @@ type Repo struct {
 	Timeout                      int64                `json:"timeout,omitempty"               xorm:"timeout"`
 	Visibility                   RepoVisibility       `json:"visibility"                      xorm:"varchar(10) 'visibility'"`
 	IsSCMPrivate                 bool                 `json:"private"                         xorm:"private"`
-	Trusted                      TrustedConfiguration `json:"trusted"                         xorm:"json 'trusted'"`
-	IsGated                      bool                 `json:"gated"                           xorm:"gated"`
+	Trusted                      TrustedConfiguration `json:"trusted"                   xorm:"json 'trusted'"`
```
6543, line 53: fmt

server/model/repo.go
```diff
@@ -109,7 +118,9 @@ func (r *Repo) Update(from *Repo) {
 // RepoPatch represents a repository patch object.
 type RepoPatch struct {
 	Config                       *string                    `json:"config_file,omitempty"`
-	IsGated                      *bool                      `json:"gated,omitempty"`
+	IsTrusted                    *bool                      `json:"trusted,omitempty"`
```
6543, line 121:
```suggestion

```

6543:
It can be v2.8.0 ... once v2.7.2 got released ...

pat-s:
2.7.2 can just be a 2.8.0 then. Doesn't make much sense to chain a 2.8.0 shortly after 2.7.2?

server/model/repo.go
anbraten: adjusted

6543 (approved):

6543:
@anbraten I updated the migration, please have a look at it and :eyes:  :)

xoxys:
Looks like you broke the tests now

server/store/datastore/migration/017_gated_to_require_approval.go
```diff
@@ -0,0 +1,65 @@
...
+		if _, err := sess.Exec(
+			builder.Update(builder.Eq{"require_approval": "all_events"}).
+				From("repos").
+				Where(builder.Eq{"gated": true})); err != nil {
+			return err
+		}
+
+		// migrate public repos to new default require approval
+		if _, err := sess.Exec(
+			builder.Update(builder.Eq{"require_approval": "pull_requests"}).
```
6543, line 49: Me: from forks to pullrequest, as it is repo default on repo creation ... throug forks is also fine by me ...

@anbraten if you have a strong opinion just change it to forks

(It got updated by me while i rewrote the migration to not require pagination)

xoxys:
I'm confused about the milestone. Is this one targeted for v3 now or 2.8? And if the target for v3.0 is correct, why is there a `backport-required` label also present?

6543:
@anbraten I think you can merge it any time now :)

docs/docs/20-usage/75-project-settings.md
```diff
@@ -25,10 +25,9 @@ Only activate this option if you trust all users who have push access to your re
 Otherwise, these users will be able to steal secrets that are only available for `deploy` events.
 :::
 
-## Protected
+## Require approval for
 
-Every pipeline initiated by an webhook event needs to be approved by a project members with push permissions before being executed.
-The protected option can be used as an additional review process before running potentially harmful pipelines. Especially if pipelines can be executed by third-parties through pull-requests.
+To prevent malicious pipelines from extracting secrets or running harmful commands or to prevent accidental pipeline runs, you can require approval for an additional review process. If you enable this option, a pipeline will be put on hold after creation and will only continue after approval. The least restrictive setting is `Approvals for forked repositories`. Forked repositories will always require approval as they can't be trusted in general.
```
lafriks, line 30: This should be updated for no approval required mode, should probably update to say that default and recommended mode is to require approval from forks

6543:
conflicts :/

server/store/datastore/migration/017_gated_to_require_approval.go
6543: reverted the thing back to forks -> f54e288cc6fbdd0004c17a06f80eaa4463c387f1

docs/docs/20-usage/75-project-settings.md
6543: b17ce97da535553fa9fc4d8e0145eee5efd55817

6543 (approved):

6543:
please backport :)

xoxys:
> please backport :)

This is IMO pretty bad practice. If we merge the PR from someone else, we can't expect that the other one will be able to create the backport soon.

xoxys:
@6543 See discussion in https://github.com/woodpecker-ci/woodpecker/pull/3348#discussion_r1826043277

Migrating a public, not gated repo to RequiredForkApproval is by definition a breaking change as it changes the behavior for users. Yes, you can argue that for security reasons we accept this tradeoff, but then this should get a pretty clear note in the release notes and proper documentation.

Issue #906, opened 2022-05-10
Does not refresh Oauth token on webhook call


6543:
`cannot find config '' in 'refs/pull/1104/head' with user: '6543'`

is gone asap i log out and in again.
I would go the way (long term) to **not** store that token at all an create a woodpecker specific token, that will get recreated  if not exist ...

anbraten:
I am not sure if I understood you correctly, but IMO OAuth tokens are a good way here. Woodpecker is requesting resources on behalf of someone else which is exactly what OAuth was invented for.

Isn't the actual problem that Woodpecker is not detecting access-token expiry correctly and forgets to refresh that token by using the refresh-token?

PR #3368, opened 2024-02-11
Set correct link for commit

qwerty287:
Closes https://github.com/woodpecker-ci/woodpecker/issues/2657
Closes https://github.com/woodpecker-ci/woodpecker/issues/906

### DIFF
--- a/server/cron/cron_test.go
+++ b/server/cron/cron_test.go
@@ -48,8 +48,8 @@ func TestCreateBuild(t *testing.T) {
 	store.On("GetRepo", mock.Anything).Return(repo1, nil)
 	store.On("GetUser", mock.Anything).Return(creator, nil)
 	forge.On("BranchHead", mock.Anything, creator, repo1, "default").Return(&model.Commit{
-		SHA: "sha1",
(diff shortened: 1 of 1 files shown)


server/model/commit.go
```diff
@@ -0,0 +1,6 @@
+package model
+
+type Commit struct {
+	SHA      string
+	ForgeURL string
```
anbraten, line 5: So this is the url to the commit in the forge?

server/api/pipeline.go
```diff
@@ -87,8 +87,7 @@ func createTmpPipeline(event model.WebhookEvent, commitSHA string, repo *model.R
 		Author: user.Login,
 		Email:  user.Email,
 
-		// TODO: Generate proper URL to commit
-		ForgeURL: repo.ForgeURL,
+		ForgeURL: commit.ForgeURL,
```
anbraten, line 90: Should we maybe rename this to `ForgeContextUrl` as its not just the ForgeUrl, but the url to a PR / commit, tag, ...?

anbraten (approved):

PR #3374, opened 2024-02-12
Replace http types on forge interface

qwerty287:
It's hard to transfer types like `http.Request` or `http.ResponseWriter` via rpc, so I adjusted the forge interface to use regular structs.
I need this for #3268


server/forge/types/oauth.go
```diff
@@ -0,0 +1,8 @@
+package types
```
6543, line 1: copyright missing

6543 (approved):

Issue #3290, opened 2024-01-29
Case for secret names is altered


sschnabe:
### Component

server

### Describe the bug

After creating a repository secret with name `TF_VAR_github_username` the name is altered in secrets list to `tf_var_github_username`. In workflow the secret is accessible as `TF_VAR_GITHUB_USERNAME`.

IMHO the environment should stay `TF_VAR_github_username` ([terraform](https://github.com/hashicorp/terraform/issues/22671) is only an example here).

 

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.2.2"}


Database: [CloudSQL](https://cloud.google.com/sql/docs/postgres/editions-intro) Postgres 14
Woodpecker Server & Agent are running in k8s
```

### Additional context

Workaround:

```yaml
commands:
  - export TF_VAR_github_token=$TF_VAR_GITHUB_TOKEN
  - export TF_VAR_github_username=$TF_VAR_GITHUB_USERNAME
  - terraform plan -lock=false
```

instead of

```yaml
commands: [terraform plan -lock=false]
```

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

anbraten:
linking #3096

Issue #3096, opened 2023-12-31
Secrets are lowercased in list


CodeDoctorDE:
### Component

web-ui

### Describe the bug

After creating the secret in UPPERCASE, the list shows it in lowercase

### System Info

```shell
2.1.1
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

6543:
note: all secret names are handled upperscase (to match, as env var, ...)

anbraten:
We change the secrets to lower case on save and convert all secrets in the config to lowercase as well. So this is not a bug. Do you want to change it to be always uppercase?

CodeDoctorDE:
Yeah, this would be a good idea.
I spent plenty of hours figuring out why my bash step was failing

Issue #2368, opened 2023-09-04
Woodpecker UI isn't showing secrets's names with uppercase.


GamePlayer-8:
### Component

web-ui

### Describe the bug

Writing secrets with uppercase i.e. `OCI_TOKEN` is being interpreted as `oci_token` for the web UI although it stays as `OCI_TOKEN` in the workflow.

### System Info

```shell
`{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-5bd12aa646"}`
```

### Additional context

Might be related to https://github.com/woodpecker-ci/woodpecker/issues/2084

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

anbraten:
Woodpecker internally converts all secrets to lower case: https://github.com/search?q=repo%3Awoodpecker-ci%2Fwoodpecker%20.ToLower&type=code

I think the corresponding env var if `secrets:` is supplied is uppercase

GamePlayer-8:
the problem is that it's not converting uppercase env variable into lowercase for the workflow then.

GamePlayer-8:
Secret isn't displayed as uppercase in the UI of woodpecker after adding it, even if I type it all with uppercase.

GamePlayer-8:
I've tried to call the lowercase & I got no data while with uppercase I've got the secret. Shouldn't the environment variable inside the workflow be converted to lowercase to avoid confusion?

GamePlayer-8:
I'm using:

```
secrets:
 - xyz
 - XYZ
```

And only with the uppercase contains the actual secret.

anbraten:
secrets are always converted to lowercase. Env vars which the secret is exposed as if using `secrets: [secret_name]` are currently uppercase. I see the confusion. Actually full uppercase could be better.

PR #3375, opened 2024-02-12
Do not alter secret key upper-/lowercase

qwerty287:
Don't lowercase secret names and don't uppercase env vars.

Examples:

We have a secret `SECRET_UPPER` (it will show up exactly like this in UI).

- ```yaml
  from_secret: secret_upper
  ```
  
  Matching with `from_secret` is case-insensitive, this injects `SECRET_UPPER`.
- ```yaml
  secrets: [secret_upper]
  ```
  
  Matching these secrets is case-insensitive. The env var however looks exactly like the string here. In this example, `SECRET_UPPER` is injected as `secret_upper` env.
- ```yaml
  secrets:
    - source: secret_upper
      target: some_different_name
  ```
  
  Here `SECRET_UPPER` is injected as `some_different_name`, it is not uppercased.

Closes https://github.com/woodpecker-ci/woodpecker/issues/2368
Closes https://github.com/woodpecker-ci/woodpecker/issues/3290
Closes https://github.com/woodpecker-ci/woodpecker/issues/3096

### DIFF
--- a/cli/common/flags.go
+++ b/cli/common/flags.go
@@ -34,6 +34,11 @@ var GlobalFlags = append([]cli.Flag{
 		Usage:   "server address",
 	},
 	&cli.BoolFlag{
+		EnvVars: []string{"WOODPECKER_DISABLE_UPDATE_CHECK"},
+		Name:    "disable-update-check",
+		Usage:   "disable update check",
+	},
+	&cli.BoolFlag{
 		EnvVars: []string{"WOODPECKER_SKIP_VERIFY"},
 		Name:    "skip-verify",
 		Usage:   "skip ssl verification",
--- a/cli/common/hooks.go
+++ b/cli/common/hooks.go
@@ -1,0 +1,67 @@
+package common
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/rs/zerolog/log"
+	"github.com/urfave/cli/v2"
+
+	"go.woodpecker-ci.org/woodpecker/v2/cli/update"
+)
+
+var (
+	waitForUpdateCheck  context.Context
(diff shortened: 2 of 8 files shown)


6543:
If some consider it a gug or not ... it was intended as feature initialy ... to have keys upercase everywhere ...

6543:
To make it non breaking we just have to add a db migration final adjusting the values in DB and in upcomming versions including this patchset the user then can decide.

Please have a look at the docs again or are you sure you catched all places mentioning secrets?

6543:
no it just imitated what settings do for plugins. and try to guide the user so they do get it more often right. but if we completely get rid of the automatism, I'm fine with it too.  just mind the migration ;)

lafriks:
I'm not sure I like this change :/

6543:
If we encourage #3406  we can let it stay all uppercase or lowecase.

As you still kan have any key:

Environment:
  aNYkEy:
    from_secret: anykey

6543:
E.g. we should make it case insensitive to users :)

xoxys:
It's not only a debugging issue. There are cases where env vars cases matters. One example is Terraform/OpenTofu, see https://opentofu.org/docs/language/values/variables/#environment-variables

That means `TF_VAR_image_id=ami-abc123` is case-sensitive and there is currently no way to set such an env var from a secret.

xoxys (approved):

lafriks:
Why not set both uppercase and unchanged case env vars?

xoxys:
> Why not set both uppercase and unchanged case env vars?

https://github.com/woodpecker-ci/woodpecker/pull/3375#issuecomment-1949913789

lafriks:
just personal preference, not deal breaker 😉

PR #3378, opened 2024-02-12
Fix agent polling

### LINKED ISSUE #1990: agent option "stop agent from taking new tasks" not working
version: 8d625788461eacfbb3645a40c4ef099cb2737afb do not respect this option

anbraten:
Supersedes #3206
closes #1990 
closes https://github.com/woodpecker-ci/autoscaler/issues/50

This solves a bug that `NoSchedule` wasn't considered if the agent already had workers waiting for tasks in the queue by kicking those workers from the queue when updating the agents no-schedule flag via api. We now also kick the agents workers when deleting the agent.

In addition a timeout prevents agents with `NoSchedule` from spamming the server.

Removes errored tasks from the database queue.

I also removed the queue info completed property as it is never used.

### DIFF
--- a/server/grpc/rpc.go
+++ b/server/grpc/rpc.go
@@ -78,6 +78,28 @@ func (s *RPC) Next(c context.Context, agentFilter rpc.Filter) (*rpc.Workflow, er
 			return nil, nil
 		}
 
+		agent, err = s.getAgentFromContext(c)
+		if err != nil {
+			return nil, err
+		}
+
+		// the agent is meanwhile marked as no-schedule, push the task back to the queue
+		if agent.NoSchedule {
+			// we need to mark the task as done, otherwise it will be stuck in the queue
+			if err := s.queue.Error(c, task.ID, errors.New("agent is marked as no-schedule")); err != nil {
+				log.Error().Err(err).Msgf("mark task '%s' error failed", task.ID)
+			}
+
+			task.AgentID = -1
+			task.ID += "1"
+			if err := s.queue.Push(c, task); err != nil {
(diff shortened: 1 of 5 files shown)


anbraten:
Only issue left is as long as the agent is waiting for a task an update to `NoSchedule` will not be recognized.

server/grpc/rpc.go
```diff
@@ -56,28 +56,57 @@ func (s *RPC) Next(c context.Context, agentFilter rpc.Filter) (*rpc.Workflow, er
...
+
+			task.AgentID = -1
+			task.ID += "1"
+			if err := s.queue.Push(c, task); err != nil {
+				log.Error().Err(err).Msgf("push task '%s' back to queue failed", task.ID)
+			}
+
+			time.Sleep(1 * time.Second)
+			continue
+		}
```
anbraten, line 101: Not sure if this is the best approach, or if we should drop all workers with a specific agentID from the agent update api call.

anbraten: changed it to use a context which seems much cleaner

xoxys:
How does this PR supersede https://github.com/woodpecker-ci/woodpecker/pull/3206?

anbraten:
It fixes the long-polling loop

xoxys:
That was clear already, but wouldn't it still make sense to add a backoff to the agent polling?

anbraten:
> That was clear already, but wouldn't it still make sense to add a backoff to the agent polling?

I think we can consider that later on. Atm the `Poll` call is blocking and therefore an agent would request with x workers once and only does new requests as soon as one worker needs a new task. That shouldn't be a lot of requests.

xoxys (approved):

PR #3382, opened 2024-02-13
Cli updater

anbraten:
Shows info when cli update is available (update info is fetched in background with a maximal delay to the normal command of 250ms)

A command `update` with a flag `--force` was added to download and install the available update.


cli/update/command.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return err
+	}
+	dst := path.Join(pwd, path.Base(c.App.Name))
+
+	log.Debug().Msgf("Moving %s to %s", binFile, dst)
+
+	// if err := os.Rename(binFile, dst); err != nil {
+	// 	return err
+	// }
```
anbraten, line 65: TODO: get path of currently executed bin and replace it

cli/common/before.go
```diff
@@ -0,0 +1,63 @@
+package common
```
xoxys, line 1: Why is this file named `before` when it contains `Before` and `After`? Rename it to e.g. `common.go`?

cli/update/command.go
```diff
@@ -0,0 +1,66 @@
...
+	executablePath, err := os.Executable()
+	if err != nil {
+		return err
+	}
+
+	if err := os.Rename(binFile, executablePath); err != nil {
+		return err
+	}
+
+	log.Info().Msgf("woodpecker-cli %s has been updated successfully! Please restart the CLI.", newVersion.Version)
```
xoxys, line 63: What does `Please restart the CLI` mean? Isn't the CLI used for one-shot commands only?

cli/update/tar.go
```diff
@@ -0,0 +1,57 @@
...
+		case header == nil:
+			continue
+		}
+
+		target := filepath.Join(dst, header.Name)
+
+		switch header.Typeflag {
+		case tar.TypeDir:
+			if _, err := os.Stat(target); err != nil {
+				if err := os.MkdirAll(target, 0x755); err != nil {
```
xoxys, line 39: To avoid magic numbers can be use a var, e.g. `directoryPermission := 0x755`?

cli/update/command.go
```diff
@@ -0,0 +1,66 @@
...
+	log.Debug().Msgf("New version %s has been downloaded successfully! Installing ...", newVersion.Version)
+
+	binFile, err := extractNewVersion(tarFilePath)
+	if err != nil {
+		return err
+	}
+
+	log.Debug().Msgf("New version %s has been extracted to %s", newVersion.Version, binFile)
+
+	executablePath, err := os.Executable()
```
xoxys, line 54: As the current CLI binary could be a symlink, add validation step? From the [docs](https://pkg.go.dev/os#Executable):

> If a stable result is needed, path/filepath.EvalSymlinks might help.

cli/update/command.go
anbraten: Should I call it sth like `rerun` instead?

cmd/cli/app.go
```diff
@@ -37,11 +38,14 @@ import (
 func newApp() *cli.App {
 	app := cli.NewApp()
 	app.Name = "woodpecker-cli"
+	app.Description = "Woodpecker command line utility"
 	app.Version = version.String()
-	app.Usage = "command line utility"
+	app.Usage = "woodpecker [global options] command [command options] [arguments...]"
```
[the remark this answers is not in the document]
xoxys: The usage string should be autogenerated if not defined. If we use the default format `"woodpecker [global options] [command] [command options] [arguments...]"` anyway, why set it manually?

cli/update/command.go
xoxys: Personally, I would remove this part completely as obvious. If you want to keep it, yes, something like rerun makes more sense to me.

cmd/cli/app.go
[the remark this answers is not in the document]
anbraten: Perfect, its autogenerated. Just removed it.

cli/update/updater.go
```diff
@@ -0,0 +1,140 @@
...
+	log.Debug().Msgf("Latest version: %s", release.TagName)
+
+	fileRegex, err := regexp.Compile(fmt.Sprintf("^woodpecker\\-cli\\_%s_%s\\.tar\\.gz$", runtime.GOOS, runtime.GOARCH))
+	if err != nil {
+		return nil, err
+	}
+
+	assetURL := ""
+	for _, asset := range release.Assets {
+		if fileRegex.MatchString(asset.Name) {
```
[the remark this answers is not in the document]
anbraten: changed it

cli/common/before.go
```diff
@@ -0,0 +1,63 @@
...
+func Before(c *cli.Context) error {
+	if err := SetupGlobalLogger(c); err != nil {
+		return err
+	}
+
+	go func() {
+		// Don't check for updates when the update command is executed
+		if firstArg := c.Args().First(); firstArg == "update" {
+			return
+		}
```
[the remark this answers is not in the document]
anbraten: added

cli/update/command.go
anbraten: added symlink handling

cli/update/command.go
anbraten: removed it

xoxys (approved):

cli/common/flags.go
```diff
@@ -33,6 +33,11 @@ var GlobalFlags = append([]cli.Flag{
 		Aliases: []string{"s"},
 		Usage:   "server address",
 	},
+	&cli.BoolFlag{
+		EnvVars: []string{"DISABLE_UPDATE_CHECK"},
```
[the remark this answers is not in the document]
xoxys: Missed that too... @anbraten can you fix this?

PR #3384, opened 2024-02-13
Cli setup command

anbraten:
This PR adds a new cli command `setup`. It is also automatically invoked on the first start of the cli if no `server-url` is provided. Setup prompts the user to enter a server url and afterwards opens a browser to `<url-of-ui>/cli/auth` where the user can login or abort the login to the cli.

image

image


xoxys:
The server token is stored in clear text in a local file, right? Is this really a good idea at all and even make it simpler for users?

xoxys:
For a possible solution, you might want to take a look at https://docs.docker.com/engine/reference/commandline/login/#credential-stores and/or https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol

However, I'm still against implementing a solution that stores secrets unencrypted.

anbraten:
> The server token is stored in clear text in a local file, right? Is this really a good idea at all and even make it simpler for users?

Considering that most widely used tools like ssh keys, kubeconfig, docker-registry credentials normally follow a similar approach you can still use the cli without if you prefer.

xoxys:
SSH keys can use a password that is stored encrypred in the OS trust store. Docker registry logins can also use credential helpers to store in the the OS trust store as descibed in the linkes I have posted... Thats exactly my point, there **are** better solutions available while improving the ux.

anbraten:
> Is the screenshot still up-to-date? If yes, I think this page looks a bug ugly as it's that empty. Could we make it similar to the web ui login?

I just increased the text sizes in the meanwhile a bit. Wouldn't use the login page actually, as I think it is quite helpful to see the user currently being logged in. I tried a few things like adding the logo in the middle, the logo with a green checkmark / red cross corresponding if you allowed or denied, but that looked a bit weird as well.

anbraten:
This is how it currently looks like:

image

Some variants I've tried:

image

image

cli/internal/config/config.go
```diff
@@ -0,0 +1,131 @@
...
+
+	return nil
+}
+
+func getConfigPath(configPath string) (string, error) {
+	if configPath != "" {
+		return configPath, nil
+	}
+
+	configPath, err := xdg.ConfigFile("woodpecker/config.json")
```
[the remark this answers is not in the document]
anbraten: Not sure if I get you correctly, but `woodpecker/config.json` is used inside the xdg config folder, not sure how we could set that as default.

cli/setup/token_fetcher.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func setupRouter(tokenReceived chan string) *gin.Engine {
+	gin.SetMode(gin.ReleaseMode)
+	e := gin.New()
+	e.UseRawPath = true
+	e.Use(gin.Recovery())
+
+	e.Use(func(c *gin.Context) {
+		c.Writer.Header().Set("Access-Control-Allow-Origin", "*") // TODO: change to serverURL
```
xoxys, line 58: What's missing to add it before we merge it?

cli/setup/setup.go
```diff
@@ -0,0 +1,88 @@
...
+		if !setupAgain {
+			log.Info().Msg("Configuration skipped")
+			return nil
+		}
+	}
+
+	serverURL := c.String("server-url")
+
+	if serverURL == "" {
+		serverURL, err = ui.Ask("Enter the URL of the woodpecker server", "https://woodpecker-ci.org", true)
```
xoxys, line 50: Does it make sense to use a placeholder that is an existing URL but not a Woodpecker CI server? I would suggest using either `"https://ci.woodpecker-ci.org"` or `"https://ci.example.com"`.

anbraten:
```suggestion
		serverURL, err = ui.Ask("Enter the URL of the woodpecker server", "https://ci.woodpecker-ci.org", true)
```

cli/internal/config/config.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil
+	}
+
+	config, err := Get(c, c.String("config"))
+	if err != nil {
+		return err
+	}
+
+	if config == nil && !c.IsSet("server-url") && !c.IsSet("token") {
+		log.Info().Msg("The woodpecker-cli is not setup yet. Please run `woodpecker-cli setup`")
```
xoxys, line 32:
```suggestion
		log.Info().Msg("The woodpecker-cli is not yet set up. Please run `woodpecker-cli setup`")
```

cli/setup/token_fetcher.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/rs/zerolog/log"
+)
+
+func receiveTokenFromUI(c context.Context, serverURL string) (string, error) {
+	port := randomPort()
+
+	tokenReceived := make(chan string)
+
+	srv := &http.Server{Addr: fmt.Sprintf(":%d", port)}
```
anbraten, line 22:
```suggestion
	srv := &http.Server{Addr: fmt.Sprintf("127.0.0.1:%d", port)}
```

cli/setup/token_fetcher.go
anbraten: I think we can actually just ignore it when only listening on localhost.

```suggestion
		c.Writer.Header().Set("Access-Control-Allow-Origin", "*")
```

xoxys (approved):

6543:
->  #3499

Issue #3288, opened 2024-01-28
Multiple flaws when running pipelines with the same service (K8s)


zc-devs:
### Component

server, agent

### Describe the bug

User-1 have pipeline in repository `wp-test`, he [runs](https://woodpecker.test.smthd.com/repos/1/pipeline/139) `gitea-integration-test` branch:
```yaml
skip_clone: true
services:
  postgres:
    image: digitalocean/doks-debug
    commands:
      - echo 'This is Gitea Postgres test server' | nc -l -6 5432
    ports:
      - 5432
steps:
  gitea:
    image: digitalocean/doks-debug
    commands:
      - nc -v -6 -w 10 postgres 5432
```

User-2 have pipeline below in his `wp-test-2` repository and `woodpecker-integration-test` branch:
```yaml
skip_clone: true
services:
  postgres:
    image: digitalocean/doks-debug
    commands:
      - echo 'This is Woodpecker Postgres test server' | nc -l -6 5432
    ports:
      - 5432
steps:
  wp:
    image: digitalocean/doks-debug
    commands:
      - nc -v -6 -w 10 postgres 5432
```
When User-1's pipeline service and service pod was launched, User-2 [ran his own](https://woodpecker.test.smthd.com/repos/2/pipeline/13).

Bugs:
1. User-2's pipeline was cancelled  with `error":"services "postgres" already exists`.
2. User-1's Service and Service Pod were deleted because User-2's pipeline cleaned up resources and the names were the same - `postgres`.

### System Info

```shell
`next-0b5eef7d1e`, 1 Server, 1 agent, max workflows 2.
```

### Additional context

[woodpecker-agent.log](https://github.com/woodpecker-ci/woodpecker/files/14075623/woodpecker-agent.log)

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

Issue #3178, opened 2024-01-12
Pipeline with service doesn't work (K8s)


zc-devs:
### Component

server, agent

### Describe the bug

When running pipeline with service, it going to start, but there is no pods and even attempts to create them.
There are no errors and WP pods (Server, Agent) restarts.

Pipeline:
```yaml
skip_clone: true
services:
  server:
    image: digitalocean/doks-debug
    commands:
      - echo 'This is test server' | nc -l -6 1234
    ports:
      - 1234
steps:
  client:
    image: digitalocean/doks-debug
    commands:
      - nc -v -6 -w 10 server 1234
```

Result:
![Screenshot 2024-01-12](https://github.com/woodpecker-ci/woodpecker/assets/127358482/4c97de70-da33-4289-b720-808804fbbf16)

### System Info

```shell
`next-a5fa810bcc`
```

### Additional context

[woodpecker-server.log](https://github.com/woodpecker-ci/woodpecker/files/13920400/woodpecker-server.log)
[woodpecker-agent.log](https://github.com/woodpecker-ci/woodpecker/files/13920401/woodpecker-agent.log)

https://woodpecker.test.smthd.com/repos/1/pipeline/85

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

eliasscosta:
Hi, I am not able to use rwx volumes in my clusters, when I try with NFS services, the pipelines go to very slow to finish.

In my situation, the services start-up with a volume and lock it. It is blocking the other pod steps to continue. My suggestion is to create the pod service with an empty dir, without the volume used in the other pods steps. It's make the pods services isolated from the pv mount lock

zc-devs:
I don't think, that volume is the cause of this issue.

[...]

PR #3395, opened 2024-02-16
fix: can't run multiple services on k8s

eliasscosta:
Fix Issue: https://github.com/woodpecker-ci/woodpecker/issues/3288

The way the pod service starts up makes it impossible to run two or more pipelines at the same time when we have a service section. 

The idea is to set the name of the service in the same way we did for the pod name. 

Pipeline: 

```yaml

services:
  mydb:
    image: mysql
    environment:
      - MYSQL_DATABASE=test
      - MYSQL_ROOT_PASSWORD=example
    ports:
      - 3306/tcp
steps:
  get-version:
    image: ubuntu
    commands:
      - ( apt update && apt dist-upgrade -y && apt install -y mysql-client 2>&1 )> /dev/null
      - sleep 30s # need to wait for mysql-server init
      - echo 'SHOW VARIABLES LIKE "version"' | mysql -uroot -hmydb test -pexample
```

Running more than one pipeline result:

image


pipeline/backend/kubernetes/service_test.go
```diff
@@ -24,24 +24,24 @@ import (
...
-	assert.Equal(t, "awesome-service", name)
+	assert.Equal(t, "awesome-service-01he8bebctabr3kgk0qj36d2me-0", name)
 }
 
 func TestService(t *testing.T) {
 	expected := `
 	{
 	  "metadata": {
-	    "name": "bar",
+	    "name": "bar-01he8bebctabr3kgk0qj36d2me-0",
```
6543, line 44: the equestion is, how is the internal network "hostname" set ... as services should be reachable by name "bar" ...

pipeline/backend/kubernetes/service_test.go
```diff
@@ -24,24 +24,24 @@ import (
...
 	    "namespace": "foo",
 	    "creationTimestamp": null
 	  },
```
6543, line 47: i guess we should add "hostnames" key to spec ...!

6543: https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/#adding-additional-entries-with-hostaliases

eliasscosta: The hostname remains the same, just the `step.Name` to use in the pipeline. 

Name with UUID is just for not conflict with others k8s resources

pipeline/backend/kubernetes/service_test.go
eliasscosta: It's if the service step has the name bar, it's added in the hostAlias. https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/backend/kubernetes/kubernetes.go#L194

pipeline/backend/kubernetes/service_test.go
6543: yes if I'm correct the hostname was never set as it should be ...

.. I have to setup an k3s again to test services with cube ...

6543: the question is if they do not work, if you could also add the fix in here or create a new pull

if it works now just fine already - I'm happy :)

eliasscosta: In my tests, it already works as expected. 

Here I enter two different pod steps running they set `mydb` for each pipeline. 

image

zc-devs: > if I'm correct the hostname was never set as it should be

It worked before changes mentioned in #3178. Service name was unique, but hostalias (/etc/hosts) contained just name and IP, like `10.10.10.10 postgres`.

pipeline/backend/kubernetes/pod_test.go
```diff
@@ -51,7 +51,7 @@ func TestStepToPodName(t *testing.T) {
 	assert.EqualValues(t, "wp-01he8bebctabr3kg", name)
 	name, err = stepToPodName(&types.Step{UUID: "01he8bebctabr3kg", Name: "postgres", Type: types.StepTypeService})
 	assert.NoError(t, err)
-	assert.EqualValues(t, "postgres", name)
+	assert.EqualValues(t, "postgres-01he8bebctabr3kg", name)
```
zc-devs, line 54: I would make it suffix as it was before. Something like `wp-01he8bebctabr3kg-postgres` / `wp-01he8bebctabr3kg-svc-postgres` / `wp-01he8bebctabr3kg-service-postgres`.

pipeline/backend/kubernetes/service_test.go
```diff
@@ -24,24 +24,24 @@ import (
 )
 
 func TestServiceName(t *testing.T) {
-	name, err := serviceName(&types.Step{Name: "database"})
+	name, err := serviceName(&types.Step{Name: "database", UUID: "01he8bebctabr3kgk0qj36d2me-0"})
 	assert.NoError(t, err)
-	assert.Equal(t, "database", name)
+	assert.Equal(t, "database-01he8bebctabr3kgk0qj36d2me-0", name)
```
zc-devs, line 29: The same: I would make it suffix as it was before.

zc-devs (approved):

6543 (approved):

PR #3400, opened 2024-02-17
Apply `documentation`/`ui` label to corresponding renovate updates

qwerty287:
@woodpecker-ci/maintainers Does somebody know whether this works?
I.e.: Will
https://github.com/woodpecker-ci/woodpecker/blob/053197c9261a589901f24cf89fe6cbcbf6cbec8c/.github/renovate.json#L30-L35
overwrite
https://github.com/woodpecker-ci/woodpecker/blob/053197c9261a589901f24cf89fe6cbcbf6cbec8c/.github/renovate.json#L20-L24
? Or are they extended?

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -37,7 +37,7 @@
       "groupName": "docs npm deps non-major",
       "matchManagers": ["npm"],
       "matchUpdateTypes": ["minor", "patch"],
-      "matchFileNames": ["docs/**/package.json"],
+      "matchFileNames": ["docs/**/package.json"]
     },
     {
       "matchDatasources": ["docker"],


xoxys:
Works, all matching package rules are applied in order.

.github/renovate.json
```diff
@@ -18,18 +18,26 @@
...
       "matchFileNames": ["web/package.json"],
       "addLabels": ["ui"]
     },
+    {
+      "matchManagers": ["npm"],
+      "matchFileNames": ["docs/**/package.json"],
+      "addLabels": ["documentation"]
+    },
+    {
+      "groupName": "web npm deps non-major",
```
xoxys, line 31: This is the same rule as for docs non-major now. Use just one role with multiple matchFilenames?

xoxys: Sure but why would you keep a different group name? These group names are nor used elsewere so no grouping happens at all and we can just use a more generic group name as description.

xoxys: Im aware of it it just thought it doesnt matter if the npm non-major pr is for web or docs. If you want to keep it fine for me as well.

xoxys (approved):

PR #3404, opened 2024-02-17
Fix usage of WOODPECKER_DATABASE_DATASOURCE_FILE

### LINKED ISSUE #3389: Setting WOODPECKER_DATABASE_DATASOURCE_FILE does not work as expected when using docker image
### Component

server, other

### Describe the bug

Using this `docker-compose.yml` snippet

```yaml
  server:
    image: *woodpecker-server-image
    restart: unless-stopped
    expose:
      - 9000
    volumes:
      - ./woodpecker-server-data:/var/lib/woodpecker/
      - ./secrets/woodpecker/agent-secret.conf:/run/secrets/agent-secret.conf:ro
      - ./secrets/woodpecker/gitea-secret.conf:/run/secrets/gitea-secret.conf:ro
      - ./secrets/woodpecker/gitea-client.conf:/run/secrets/gitea-client.conf:ro
      - ./secrets/woodpecker/docker-config.json:/run/secrets/docker-config.json:ro
      - ./secrets/woodpecker/woodpecker-datasource.conf:/run/secrets/woodpecker-datasource.conf
    environment:
      - WOODPECKER_DATABASE_DRIVER=postgres
      - WOODPECKER_DATABASE_DATASOURCE_FILE=/run/secrets/woodpecker-datasource.conf
```

does not pick up the datasource secrets from the file, but will try to use `/var/lib/woodpecker/woodpecker.sqlite`

Log:

```txt
server-1  | 9:12AM TRC woodpecker/src/github.com/woodpecker-ci/woodpecker/cmd/server/setup.go:88 > setup datastore: store.Opts{Driver:"postgres", Config:"/var/lib/woodpecker/woodpecker.sqlite", XORM:store.XORM{Log:true, ShowSQL:true}}
```

The reason is, that the env var `WOODPECKER_DATABASE_DATASOURCE` is set in the Dockerfile and takes precedence.

To make it work one has to cancel this environment variable by using this:

```yaml
    environment:
      - WOODPECKER_DATABASE_DRIVER=postgres
      - WOODPECKER_DATABASE_DATASOURCE
      - WOODPECKER_DATABASE_DATASOURCE_FILE=/run/secrets/woodpecker-datasource.conf
```

**Note:** No equal sign behind `WOODPECKER_DATABASE_DATASOURCE` 

Had to find out the hard way. 

Should this be documented in https://github.com/woodpecker-ci/woodpecker/blob/5e0ec973baea3be1c4f6514b2ce778bfc0c961d6/docs/docs/30-administration/10-server-config.md?plain=1#L66-L102

If you agree, that is is only a documentation issue, I will provide a PR.

### System Info

```shell
Docker Image `woodpeckerci/woodpecker-server:v2.3.0`
```

### Additional context

_No response_

### Validations

smainz:
fixes #3389

Set variable to let server detect if it's deployed within a container image.
Set the default database connection based on this.


docker/Dockerfile.server.alpine.multiarch
```diff
@@ -6,7 +6,9 @@ ENV CA_CERTIFICATES_VERSION="20230506-r0"
 ARG TARGETOS TARGETARCH
 RUN apk add -U --no-cache ca-certificates=${CA_CERTIFICATES_VERSION}
 ENV GODEBUG=netdns=go
-ENV WOODPECKER_DATABASE_DATASOURCE=/var/lib/woodpecker/woodpecker.sqlite
+# create datasource config file
+RUN echo -n "/var/lib/woodpecker/woodpecker.sqlite" > /etc/woodpecker-datasource.conf
```
zc-devs, line 10: 1. Looks like a garbage file in case of using non-defaults (postgres, mysql)
2. There is nothing to hide in case of sqlite
3. A little inconsistency with [/etc/woodpecker/agent.conf](https://github.com/woodpecker-ci/woodpecker/blob/4e44dd0e76b516685bc79af4885bcfb739720986/cmd/agent/core/flags.go#L58)

zc-devs (commented):
What if get rid of `WOODPECKER_DATABASE_DRIVER` and `WOODPECKER_DATABASE_DATASOURCE` in Dockerfiles?
There are default values `sqlite3` and `woodpecker.sqlite` respectively. It covers default "test" deployment. If you want sqlite "prod" deployment you should mount volume anyway (set the path) in Docker and Kubernetes. So, there should not be a problem to set it twice. It increase transparency and you can write config in your preferred way (`*_FILE` or not).

xoxys:
> What if get rid of `WOODPECKER_DATABASE_DRIVER` and `WOODPECKER_DATABASE_DATASOURCE` in Dockerfiles?

I would prefer this way as well.

smainz:
Lets see, what the options are:

1. Do not change the code and just document, that you have to unset the env var `WOODPECKER_DATABASE_DATASOURCE` if you want to provide the datasource via _FILE:  **non breaking, but a bit ugly**
2. Change the way the default datasource config `woodpecker.sqlite` is overriden in the docker image as show in the first commit: **non breaking, not totaly obvious why the config is done in that way, leaves garbage file in the image, if not used**
3. Remove the env vars `WOODPECKER_DATABASE_DRIVER` and `WOODPECKER_DATABASE_DATASOURCE` completely in the Dockerfile: **probably breaks existing installs which did rely on this, Potential to loose data in this situation (new woodpecker.sqlite file wil be created in current directory), Possibly some frustration with new users which mount a volume to `/var/lib/woodpecker`, but do not define `WOODPECKER_DATABASE_DATASOURCE` or `WOODPECKER_DATABASE_DATASOURCE_FILE`**
4.  Remove the env vars `WOODPECKER_DATABASE_DRIVER` and `WOODPECKER_DATABASE_DATASOURCE` completely in the Dockerfile **and** change the hard coded default from `woodpecker.sqlite` to `/var/lib/woodpecker/woodpecker.sqlite`: **Will break installs not running in the container and relying on the default config (development and testing most probably)**

Number 3. would make the docker install compatible with the docs:
https://github.com/woodpecker-ci/woodpecker/blob/4e44dd0e76b516685bc79af4885bcfb739720986/docs/docs/30-administration/10-server-config.md?plain=1#L420-L422

As it looks, the intended place for a volume to be mounted is /var/lib/woodpecker as `XDG_CACHE_HOME` and 
`ENV XDG_DATA_HOME` also point to it.


Did I miss another option?

smainz:
Any ideas about the direction?

As it is planned for 2.4 it should not be breaking (for end-users) which makes it 2. or 4.

zc-devs:
> another option?

5. Make FR/PR/fork into/of `urfave/cli` with adjustable priorities and prefer `*_FILE` here :)

smainz:
Just collecting what needs to be done for 3.:

- [ ] Update documentation for docker install 
- [ ] Update dockercompose.example.yml 
- [ ] Create PR for [woodpecker/helm](https://github.com/woodpecker-ci/helm) to add ENV vars in values.yml
- [ ] Add info to migrations.md docs
- [ ] Change Dockerfile.server.xxx   (2 files)

All but the last are non breaking and could be done right now

smainz:
> this will end in each "important" setting being it's own file :/
> 
> we should rather switch to a lib that supports config files written in yaml, toml or json...

It is not about having a config file for settings, the existing means are well suited. The only thing I want to achieve is to be able to keep secrets out of the docker-compose or .env file  (kept in a git repo) but to mount them into the container.

Could be compared to 

```yml
  environment:
    DOCKER_PASSWORD:
       from_secret docker_password
```
in a pipeline.

6543:
well that is all fine ... but I would then implement the `WOODPECKER_DATABASE_DATASOURCE_FILE` as how all other`_FILE` things are implemented and dont change the dockerfile ...

6543:
#3419

6543:
PS: still thanks for contributing you valuable time and try to fix it! :)

smainz:
> PS: still thanks for contributing you valuable time and try to fix it! :)

I should say thank you for woodpecker-ci.

smainz:
PR updated to use the same config file location as the agent does.

I still prefere this PR over #3419, as it keeps the same semantic as all other "_FILE" env vars which is:
>  Keep in mind that when the original environment variable gets specified at the same time it will override the value read from the file.

anbraten:
I would also stick to not adding the `_FILE` cli-argument again. Creating the folder is a bit "ugly" on the other hand. What do you think about setting sth like `ENV IN_CONTAINER=true` and in case of that use a different default in the go?

smainz:
> I would also stick to not adding the `_FILE` cli-argument again. Creating the folder is a bit "ugly" on the other hand. What do you think about setting sth like `ENV IN_CONTAINER=true` and in case of that use a different default in go?

If this is the option everyone agrees to, i can implement it that way:
If IN_CONTAINER is set, the default will be `/var/lib/woodpecker/woodpecker.sqlite`, otherwise it will be `woodpecker.sqlite`.

anbraten:
Yes that's what I had in mind. @6543 What do you think?

6543:
Ack

smainz:
`WOODPECKER_IN_CONTAINER` is not documented in server settings as I consider this an implementation detail.

cmd/server/flags.go
```diff
@@ -473,3 +473,14 @@ var flags = append([]cli.Flag{
 		Usage:   "Flag to decrypt all encrypted data and disable encryption on server",
 	},
 }, logger.GlobalLoggerFlags...)
+
+// If woodpecker is running inside a container the default value for
+// the datasource is different from running outside a container.
+func datasourceDefaultValue() string {
+	_, found := os.LookupEnv("WOODPECKER_IN_CONTAINER")
+	if found {
+		return "/var/lib/woodpecker/woodpecker.sqlite"
```
zc-devs, line 482: This should be by default. Especially not in container.
https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

cmd/server/flags.go
```diff
@@ -473,3 +473,14 @@ var flags = append([]cli.Flag{
...
+	} else {
+		return "woodpecker.sqlite"
```
zc-devs, line 484: On the other hand, containers doesn't follow standard sometimes and place data in `/data`, `/bitnami/mongodb`, etc. directories.

cmd/server/flags.go
```diff
@@ -473,3 +473,14 @@ var flags = append([]cli.Flag{
 		Usage:   "Flag to decrypt all encrypted data and disable encryption on server",
 	},
 }, logger.GlobalLoggerFlags...)
+
+// If woodpecker is running inside a container the default value for
+// the datasource is different from running outside a container.
+func datasourceDefaultValue() string {
+	_, found := os.LookupEnv("WOODPECKER_IN_CONTAINER")
```
zc-devs, line 480: What if mainframe or Kubernetes? Two additional flags?

cmd/server/flags.go
```diff
@@ -473,3 +473,14 @@ var flags = append([]cli.Flag{
 		Usage:   "Flag to decrypt all encrypted data and disable encryption on server",
 	},
 }, logger.GlobalLoggerFlags...)
+
+// If woodpecker is running inside a container the default value for
+// the datasource is different from running outside a container.
```
zc-devs, line 478: Why they should differ?

6543:
> `WOODPECKER_IN_CONTAINER` is not documented in server settings as I consider this an implementation detail.

you can even "hide" it in the cli help menue too ... (flag option) ... and document it in code why it's hidden

cmd/server/flags.go
smainz: Yes, but the current default is current directory/woodpecker.sqlite and changing that would break e.g. development.
This option has been neglected.

cmd/server/flags.go
smainz: The current "default" when running inside the container is like it is and I do not want to change that.

cmd/server/flags.go
smainz: The flag is only set in the Dockerfile when woodpecker-server is packaged into an image. So the same setting applies for kubernetes.
Backward compatibility.

cmd/server/flags.go
smainz: Only for backward compatibility.
Current working directory is not a sensible option for containers.

docker/Dockerfile.server.alpine.multiarch
```diff
@@ -6,8 +6,8 @@ ENV CA_CERTIFICATES_VERSION="20230506-r0"
 ARG TARGETOS TARGETARCH
 RUN apk add -U --no-cache ca-certificates=${CA_CERTIFICATES_VERSION}
 ENV GODEBUG=netdns=go
-ENV WOODPECKER_DATABASE_DATASOURCE=/var/lib/woodpecker/woodpecker.sqlite
-ENV WOODPECKER_DATABASE_DRIVER=sqlite3
+# Signal that woodpecker is running inside a container
+ENV WOODPECKER_IN_CONTAINER=true
```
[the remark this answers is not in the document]
smainz: In the docs it is stated, that the default is different when running inside a container from running ad an plain executable. This is only used to"signal, that it is running inside container"

6543 (approved):

docker/Dockerfile.server.alpine.multiarch
anbraten, line 10:
```suggestion
# Internal setting do NOT change! Signals that woodpecker is running inside a container
ENV WOODPECKER_IN_CONTAINER=true
```

docker/Dockerfile.server.multiarch
```diff
@@ -3,8 +3,8 @@ FROM --platform=$BUILDPLATFORM docker.io/golang:1.22 AS certs
 FROM scratch
 ARG TARGETOS TARGETARCH
 ENV GODEBUG=netdns=go
-ENV WOODPECKER_DATABASE_DATASOURCE=/var/lib/woodpecker/woodpecker.sqlite
-ENV WOODPECKER_DATABASE_DRIVER=sqlite3
+# Signal that woodpecker is running inside a container
+ENV WOODPECKER_IN_CONTAINER=true
```
anbraten, line 7:
```suggestion
# Internal setting do NOT change! Signals that woodpecker is running inside a container
ENV WOODPECKER_IN_CONTAINER=true
```

docker/Dockerfile.server.alpine.multiarch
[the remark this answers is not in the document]
anbraten: it is an internal setting to detect if woodpecker is executed as part of our container setup.

cmd/server/flags.go
anbraten: I think this is totally fine as `./woodpecker.sqlite` is mainly used for development and similar. Most users are using the container images or some dpkg, nixpack, ... defining other default paths anyways.

cmd/server/flags.go
anbraten: works for all container setups.

anbraten (approved):

6543 (approved):

PR #3406, opened 2024-02-18
Deprecate alternative names on secrets

qwerty287:
Closes https://github.com/woodpecker-ci/woodpecker/discussions/2274

# deprecation of alternative names

Instead of
```yaml
secrets:
  - source: some_secret
    target: some_env
```
you now write:
```yaml
environment:
  some_env:
    from_secret: some_secret
```

Also, it's possible to use complex yaml objects in `environment`, they're turned into json (just like `settings`).


docs/docs/20-usage/60-services.md
```diff
@@ -53,30 +53,6 @@ Service containers generally expose environment variables to customize service s
      image: redis
 ```
```
6543, line 55: Could you move that into it's own pull?

docs/docs/91-migrations.md
```diff
@@ -8,6 +8,9 @@ Some versions need some changes to the server configuration or the pipeline conf
 - Removed `WOODPECKER_ROOT_PATH` and `WOODPECKER_ROOT_URL` config variables. Use `WOODPECKER_HOST` with a path instead
 - Pipelines without a config file will now be skipped instead of failing
 - Deprecated `includes` and `excludes` support from **event** filter
+- Deprecated alternative names for secrets, use `environment` with `from_secret`
+- Deprecated slice definition for labels and env vars
+- Deprecated `detached` in favor of services
```
6543, line 13: I dont think we should depricate detatched. There is no equivalent ... services wont do as they start before while detatched can start and use e.g. artifacts build by a step before

6543 (changes requested):
Detatched need discussion

docs/docs/20-usage/40-secrets.md
```diff
@@ -26,20 +26,20 @@ once their usage is declared in the `secrets` section:
 +    secrets: [ docker_username, docker_password ]
 ```
 
-### Use secrets in settings
+### Use secrets in settings and environment
 
-Alternatively, you can get a `setting` from secrets using the `from_secret` syntax.
-In this example, the secret named `secret_token` would be passed to the setting named `token`, which will be available in the plugin as environment variable named `PLUGIN_TOKEN`. See [Plugins](./51-plugins/20-creating-plugins.md#settings) for details.
+Alternatively, you can get a setting or environment from secrets using the `from_secret` syntax. This also allows using a different name for the environment variable.
```
6543, line 31:
```suggestion
You can set an setting or environment value from secrets using the `from_secret` syntax.
```

pipeline/frontend/yaml/compiler/compiler.go
```diff
@@ -178,11 +178,15 @@ func (c *Compiler) Compile(conf *yaml_types.Workflow) (*backend_types.Config, er
 		if c.metadata.Curr.Event == metadata.EventTag {
 			cloneSettings["tags"] = "true"
 		}
+		env := map[string]any{}
```
6543, line 181: isnt there a maps.Copy func?

pipeline/frontend/yaml/types/base/map.go
```diff
@@ -21,13 +21,13 @@ import (
 )
 
 // SliceOrMap represents a map of strings, string slice are converted into a map
-type SliceOrMap map[string]string
+type SliceOrMap map[string]any
```
6543, line 24: is slice or map only used at environment? else I dont want to loosen other datatypes unrelated to this change

6543: If we dont throw an error out already ... we should not silently drop it within one version

First depricate 
next error
then remove

6543: yes but Labels are affected ... anyway I guess it's fine as we did deprecated them a long time ago

pipeline/frontend/yaml/types/workflow.go
```diff
@@ -22,15 +22,15 @@ import (
...
-		Labels    base.SliceOrMap `yaml:"labels,omitempty"`
-		DependsOn []string        `yaml:"depends_on,omitempty"`
-		RunsOn    []string        `yaml:"runs_on,omitempty"`
-		SkipClone bool            `yaml:"skip_clone"`
+		When      constraint.When   `yaml:"when,omitempty"`
+		Workspace Workspace         `yaml:"workspace,omitempty"`
+		Clone     ContainerList     `yaml:"clone,omitempty"`
+		Steps     ContainerList     `yaml:"steps,omitempty"`
+		Services  ContainerList     `yaml:"services,omitempty"`
+		Labels    map[string]string `yaml:"labels,omitempty"`
```
6543, line 30: this is a breaking unrelated change ...

pipeline/frontend/yaml/types/base/map.go
6543: e.g. we have to copy the `SliceOrMap` type to ~ `SliceOrMapAny` and still use the old SliceOrMap for Labels

pipeline/frontend/yaml/types/workflow.go
6543: ah !!! nice

6543 (approved):

6543:
https://github.com/woodpecker-ci/woodpecker/pull/3406#discussion_r1499569399

yes - if it's not in the docs it's about to be subject to change at any time - :dagger:

PR #3407, opened 2024-02-18
More `when` filters

qwerty287:
to not run empty pipelines


6543:
Did you checked zje workflows do not depend on each other ... tjat was a regression in the last releace cycles ...

6543 (approved):

Issue #3342, opened 2024-02-07
Org name doesn't changes after User is renamed


zc-devs:
### Component

server

### Describe the bug

Local Gitea `usr2` exists
```
woodpecker_test=# select user_id, user_login, user_org_id from users order by user_id;
 user_id | user_login | user_org_id
---------+------------+-------------
       3 | usr2       |           3
(5 rows)

woodpecker_test=# select * from orgs;
 id | name | is_user | private
----+------+---------+---------
  3 | usr2 | t       | f
(5 rows)
```

Go to Gitea Site Administration -> Identity and Access -> User Accounts. Edit `usr2`. Change Username to `usr2-gitea-local`.
Relogin in Gitea and Woodpecker by `usr2-gitea-local`.
```
woodpecker_test=# select user_id, user_login, user_org_id from users order by user_id;
 user_id |    user_login    | user_org_id
---------+------------------+-------------
       3 | usr2-gitea-local |           3
(5 rows)

woodpecker_test=# select * from orgs;
 id | name | is_user | private
----+------+---------+---------
  3 | usr2 | t       | f
```
User login was changed, but not organization name.
Now, if we go to User's profile page in Woodpecker we get
```
An unknown error occurred
GET
https://woodpecker.test.smthd.com/api/orgs/3/secrets?page=1
[HTTP/2 403  61ms]
```

### System Info

```shell
Woodpecker 2.3.0, Gitea 1.21.5, Postgres 16
```

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

zc-devs:
https://github.com/woodpecker-ci/woodpecker/issues/3328#issuecomment-1951384828

```
users.user_login=MrMeeb
orgs.name=mrmeeb
```

@MrMeeb, try to rename user to `mrmeeb` or organization to `MrMeeb` and test again.

MrMeeb:
Updated `orgs.name` to `MrMeeb` and now it seems to work. Want to be clear that I did not rename my user to cause this though, this was the out of the box behaviour. All I did was sign in to Woodpecker and try to create secrets.

pat-s:
[...]

PR #3409, opened 2024-02-18
Update org name on login

qwerty287:
and create a new org once a user is created (logged in for the first time)

closes https://github.com/woodpecker-ci/woodpecker/issues/3342

### DIFF
--- a/cmd/server/flags.go
+++ b/cmd/server/flags.go
@@ -438,6 +438,43 @@ var flags = append([]cli.Flag{
 		Usage:   "gitlab skip ssl verification",
 	},
 	//
+	// Bitbucket DataCenter/Server (previously Stash)
+	//
+	&cli.BoolFlag{
+		EnvVars: []string{"WOODPECKER_BITBUCKET_DC"},
+		Name:    "bitbucket-dc",
+		Usage:   "Bitbucket DataCenter/Server driver is enabled",
+	},
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_BITBUCKET_DC_URL"},
+		Name:    "bitbucket-dc-server",
+		Usage:   "Bitbucket DataCenter/Server server address",
+	},
+	&cli.StringFlag{
+		EnvVars:  []string{"WOODPECKER_BITBUCKET_DC_CLIENT_ID"},
+		Name:     "bitbucket-dc-client-id",
+		Usage:    "Bitbucket DataCenter/Server OAuth 2.0 client id",
+		FilePath: os.Getenv("WOODPECKER_BITBUCKET_DC_CLIENT_ID_FILE"),
+	},
+	&cli.StringFlag{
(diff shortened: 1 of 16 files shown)


6543:
I think we should start writing tests for this function :sweat_smile:

server/api/login.go
```diff
@@ -119,6 +119,34 @@ func HandleAuth(c *gin.Context) {
...
+				_ = c.AbortWithError(http.StatusInternalServerError, err)
+				return
+			}
+			org = &model.Org{
+				Name:    u.Login,
+				IsUser:  true,
+				Private: false,
+			}
+			if err := _store.OrgCreate(org); err != nil {
+				log.Error().Err(err).Msgf("on user creation, could not mark org as user")
```
anbraten, line 133:
```suggestion
				log.Error().Err(err).Msgf("on user creation, could create org for user")
```

server/api/login.go
```diff
@@ -119,6 +119,34 @@ func HandleAuth(c *gin.Context) {
...
 			}
+		} else {
+			if err != nil && !errors.Is(err, types.RecordNotExist) {
+				_ = c.AbortWithError(http.StatusInternalServerError, err)
+				return
+			}
+			org = &model.Org{
+				Name:    u.Login,
+				IsUser:  true,
+				Private: false,
```
anbraten, line 130: What does a public org allow? List all public repos?

server/api/login.go
```diff
@@ -119,6 +119,34 @@ func HandleAuth(c *gin.Context) {
 			if err := _store.OrgUpdate(org); err != nil {
```
anbraten, line 119:
```suggestion
      u.OrgID = org.ID
			if err := _store.OrgUpdate(org); err != nil {
```

server/api/login.go
```diff
@@ -119,6 +119,34 @@ func HandleAuth(c *gin.Context) {
...
+			if err := _store.OrgCreate(org); err != nil {
+				log.Error().Err(err).Msgf("on user creation, could not mark org as user")
+			}
+			u.OrgID = org.ID
+		}
+	}
+
+	// update org name
+	if u.Login != tmpuser.Login {
+		org, err := _store.OrgGet(u.OrgID)
```
anbraten, line 141: Can org-id be null / `0` for users created before we added the user-orgs?

6543:
> That's really hard to do as it requires a lot of mocking. It needs the gin context with store and a forge.

we have a mock for both already ... anyway this is not blocking, just a todo to be done

server/api/login.go
```diff
@@ -116,9 +116,38 @@ func HandleAuth(c *gin.Context) {
...
+		if err != nil {
+			log.Error().Err(err).Msgf("cannot get org %s", u.Login)
+			c.Redirect(http.StatusSeeOther, server.Config.Server.RootPath+"/login?error=internal_error")
+			return
+		}
+		org.Name = u.Login
```
6543, line 148: if org.Name != u.Login ... onyl then update

6543 (approved):

Issue #2495, opened 2023-09-23
Allow to assign steps to services


qwerty287:
### Clear and concise description of the problem

Services always run, even if the pipeline fails before a step runs that needs the service

### Suggested solution

Add a parameter like `steps: [ s1, s2 ]` to services that allows to control for which steps services must start.
Don't start the step before it is clear that the assigned steps will run, and shut it down after the steps has finished.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

anbraten:
You could currently use a step with the detached keyword.

6543:
well if we implement it the right way ... you can create a e.g. build-service step

then a test-with-service step that depends on the build-service step and make the service.steps=test-with-service

two notes: serivce.steps should be a list and as soon as one of the mentioned steps is going to be executed, the service is started first (note we als should wayt until the service reports to be healty if an api was set in the OCI) and then start the step.

also we should not tear it down until all the listed steps are finished exec, if the condition is not met its saver to run till workflow teardown

PR #3411, opened 2024-02-19
Add `needs` for services and deprecate `detached`

qwerty287:
related to #2495

Treat services just like steps to support `depends_on` (including all directions: step -> service, step -> step, service -> service, service -> step)

### DIFF
--- a/cli/repo/repo_update.go
+++ b/cli/repo/repo_update.go
@@ -36,8 +36,8 @@ var repoUpdateCmd = &cli.Command{
 			Usage: "repository is trusted",
 		},
 		&cli.BoolFlag{
-			Name:  "gated",
-			Usage: "repository is gated",
+			Name:   "gated", // TODO: remove in next release
+			Hidden: true,
 		},
 		&cli.StringFlag{
 			Name:  "require-approval",
@@ -82,7 +82,6 @@ func repoUpdate(ctx context.Context, c *cli.Command) error {
 		config          = c.String("config")
 		timeout         = c.Duration("timeout")
 		trusted         = c.Bool("trusted")
-		gated           = c.Bool("gated")
 		requireApproval = c.String("require-approval")
 		pipelineCounter = int(c.Int("pipeline-counter"))
 		unsafe          = c.Bool("unsafe")
@@ -92,28 +91,17 @@ func repoUpdate(ctx context.Context, c *cli.Command) error {
 	if c.IsSet("trusted") {
 		patch.IsTrusted = &trusted
 	}
-	// TODO: remove isGated in next major release
+
+	// TODO: remove in next release
 	if c.IsSet("gated") {
-		if gated {
-			patch.RequireApproval = &woodpecker.RequireApprovalAllEvents
-		} else {
-			patch.RequireApproval = &woodpecker.RequireApprovalNone
-		}
+		return fmt.Errorf("'gated' option has been set in version 2.8, use 'require-approval' in >= 3.0")
 	}
+
 	if c.IsSet("require-approval") {
 		if mode := woodpecker.ApprovalMode(requireApproval); mode.Valid() {
 			patch.RequireApproval = &mode
 		} else {
 			return fmt.Errorf("update approval mode failed: '%s' is no valid mode", mode)
-		}
-
-		// TODO: remove isGated in next major release
-		if requireApproval == string(woodpecker.RequireApprovalAllEvents) {
-			trueBool := true
-			patch.IsGated = &trueBool
-		} else if requireApproval == string(woodpecker.RequireApprovalNone) {
-			falseBool := false
-			patch.IsGated = &falseBool
 		}
 	}
 	if c.IsSet("timeout") {
--- a/docs/docs/30-administration/11-forges/100-addon.md
+++ b/docs/docs/30-administration/11-forges/100-addon.md
@@ -38,7 +38,7 @@ Addons use RPC to communicate to the server and are implemented using the [`go-p
 
 This example will use the Go language.
 
-Directly import Woodpecker's Go packages (`go.woodpecker-ci.org/woodpecker/woodpecker/v2`) and use the interfaces and types defined there.
+Directly import Woodpecker's Go packages (`go.woodpecker-ci.org/woodpecker/v2`) and use the interfaces and types defined there.
 
 In the `main` function, just call `"go.woodpecker-ci.org/woodpecker/v2/server/forge/addon".Serve` with a `"go.woodpecker-ci.org/woodpecker/v2/server/forge".Forge` as argument.
 This will take care of connecting the addon forge to the server.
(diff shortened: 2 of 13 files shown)


6543:
it does not solve the thing detatched was able to and is now supposed to remove :/

I still if following the KISS principe detatched is the best way to do so.

yes the proposed feature here would have other benefits through ;)

PS: if service is not registered at a single step -> always run; else only if step is also exec ...

6543 (changes requested):
I'm still against it!

as long as we have no way to run something that first has to be build as service.

we can do so if services can access the workspace too and also can depend on steps ...

... but we have to implement it first

6543:
in exactly this way: https://github.com/woodpecker-ci/woodpecker/issues/2495#issuecomment-2194557622 ?

as far as i remember the backend engine do not do that exactly this way ...

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -80,11 +79,11 @@ func (c *Compiler) createProcess(container *yaml_types.Container, stepType backe
 
 	environment["CI_WORKSPACE"] = path.Join(c.base, c.path)
 
-	if stepType == backend_types.StepTypeService || container.Detached {
-		detached = true
+	if container.Detached {
```
6543, line 82: until we dont have detatched removed from the frontend we should not remove that!!

else a user can create two services with the same hostname

pipeline/frontend/yaml/types/container.go
```diff
@@ -36,18 +36,19 @@ type (
...
-		Image          string             `yaml:"image,omitempty"`
-		Name           string             `yaml:"name,omitempty"`
-		Pull           bool               `yaml:"pull,omitempty"`
-		Settings       map[string]any     `yaml:"settings"`
-		Volumes        Volumes            `yaml:"volumes,omitempty"`
-		When           constraint.When    `yaml:"when,omitempty"`
-		Ports          []string           `yaml:"ports,omitempty"`
-		DependsOn      base.StringOrSlice `yaml:"depends_on,omitempty"`
+		// TODO deprecated remove in 3.0
+		Detached  bool               `yaml:"detach,omitempty"`
```
6543, line 40: you want to depricated directory and name too ?!?

6543:
be aware that steps now can have the same name and its not that uncommon to have a service and a step called the same

pipeline/frontend/yaml/types/container.go
6543: yes so you should move it form the main list below

6543:
> Yes having both step/service with the same name could be possible but this has caused so many problem already...

well no we fixed all of them and now the code works better!!!

the (pipeline-)backend should not care about the name at all as long as it is no service!

6543:
why not add a new keyword specific to services: `needs`  so we don't have any collision

pipeline/frontend/yaml/compiler/convert.go
6543: well i have a solution: prefix the step name with the uuid ... this way you can not guess the hostname (as it was with non service-steps that are detached) and they cant clash

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -114,8 +113,8 @@ func (c *Compiler) createProcess(container *yaml_types.Container, stepType backe
 		return secret.Value, nil
 	}
 
-	// TODO: why don't we pass secrets to detached steps?
-	if !detached {
+	// TODO: why don't we pass settings to services?
+	if stepType != backend_types.StepTypeService {
```
6543, line 117: good question ... I would pass them down if they exist ... but we can address that in a new pull that also alters the json schema and docs

6543:
asap https://github.com/woodpecker-ci/woodpecker/pull/3411#discussion_r1745488047 this is addressed It  looks realy good.

I just want to make use one usecase works:

if a service is defined but non of the steps need it (ignoring all when filters) it runs as the current modell all the time and starts before all steps ...

6543:
Yep that should fix it :)

docs/docs/20-usage/60-services.md
```diff
@@ -37,6 +37,8 @@ services:
       - 51820/udp
 ```
 
+You can use any step property, including `depends_on`. Services can depend on other services and steps.
```
6543, line 40: Still up to date?

6543:
Thanks for resolving ... i realy do have to make miselfe time to properly review it 😓

6543:
added two TODOs ...

I'll have a look in addressing them and streamline the code to better to maintain.
Also i will create more test coverage and try to cover the edgecases

## TODO:
- [ ] make generic service (currently how services work) work
- [ ] explizite document how services work
- [ ] document what can depend upon what
- [ ] document what can need upon what
- [ ] more tests
- [ ] better maintainable code (with look into using an dag in the backend) (might be done before merging this or at least prepare here)
- [ ] create generic DAG type we can use
- [ ] allow need on sequential steps

6543:
I'll created an todo list, feel free to start addressing them ... else I'll catch them on one by one ...

If it's easyer for you i can create pull requests onto your feature branch, so just tell me, else I'll push short commits

docs/docs/20-usage/20-workflow-syntax.md
```diff
@@ -502,17 +502,27 @@ steps:
 
 :::
 
-### `volumes`
+### `needs`
 
-Woodpecker gives the ability to define Docker volumes in the YAML. You can use this parameter to mount files or folders on the host machine into your containers.
+Just like [`depends_on`](#depends_on), but for services.
```
zc-devs, line 507: Why new keyword? Why not `depends_on`?

zc-devs: What is the probability, that a service (noun like postgres, mongo) would have name like a step (verb like prepare, build, test, build-image, push-image)? And in those rare cases, when they do, we can throw error like `ambigous names in depends_on, rename step or service`.
If this is the only reason, I propose to use only `depends_on`.

zc-devs: Clean your pipeline: rename steps to `test-mysql`, `test-postgres` or (`unit-test`, `build`) `integration-test-sqlite`, `integration-test-mysql`, `integration-test-postgres`. No need in different keyword.
Let me ask some questions: what is the behavior, if I define multiple steps with the same name? What if I define multiple services with the same name? Apply this behavior to inter-group naming.
If I can make both of them with the same names _and_ you do not want throwing error on the same names when `depends_on` is used, alright - do not fail, but depend on all of them. Except on step itself, which defines `depends_on`, obviously. Again, no need in separate keyword, IMO.

pat-s:
Given the open discussions and todos, I'll remove this from the 3.0 milestone.

lafriks:
Damn, I was waiting for this 🙃

PR #3419, opened 2024-02-20
Correctly prefer datasource env var over file content

6543:
close  #3404

fixes https://github.com/woodpecker-ci/woodpecker/issues/3389

### DIFF
--- a/cmd/server/flags.go
+++ b/cmd/server/flags.go
@@ -224,7 +224,7 @@ var flags = append([]cli.Flag{
 	&cli.StringFlag{
 		EnvVars: []string{"WOODPECKER_DATABASE_DATASOURCE_FILE"},
 		Name:    "datasource-file",
-		Usage:   "database driver configuration stored in a file, if set will always be prefered",
+		Usage:   "database driver configuration stored in a file, if set will always be preferred over simple configuration string",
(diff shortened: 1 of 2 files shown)


smainz:
That was option 5 :-), which I did not want to propose in #3404, due to consistency reasons. Now there are two different precendence rules for environment vars in place.

You still need to fix the documentation here:
https://github.com/woodpecker-ci/woodpecker/blob/364d708923fafcb93f0b656324a7a7e10618863c/docs/docs/30-administration/10-server-config.md?plain=1#L72

This is true for all but `DATABASE_DATASOURCE_FILE`

My referred solution was to stay with the preference rules stated above but use the **least precedence way** of setting the datasource in the docker image.

Anyway, thanks for taking the issue seriously.

6543:
The cli lib we use is working on a v3 ... i have a look if there it is posible to specify the prio. If not atm you can suggest that :)

smainz:
If you use a file for overriding the default value instead of setting the env var directly, it is the least precedence way.

So the only problem is how we set up the Dockerfile. This makes a setting, which is meant to be overridden a de facto hard coded value thus eliminating the flexibility urfave/cli provides.

PR #3437, opened 2024-02-25
Simplify store interfaces

qwerty287:
Use `store.Store` interface if possible.


server/store/store.go
```diff
@@ -92,6 +86,7 @@ type Store interface {
 	// CreatePipeline creates a new pipeline and steps.
 	CreatePipeline(*model.Pipeline, ...*model.Step) error
 	// UpdatePipeline updates a pipeline.
+	// TODO
```
anbraten, line 89: What do you want to change here? Could you add a note to the `TODO`,

anbraten (approved):

PR #3440, opened 2024-02-25
Migrate to bun

anbraten:
closes #2769


zc-devs:
Does it work [on Windows](https://bun.sh/blog/bun-v1.0#bun-more-thing)?

anbraten:
> Does it work [on Windows](https://bun.sh/blog/bun-v1.0#bun-more-thing)?

Sounds like it's still experimental: https://bun.sh/docs/installation#windows

zc-devs:
> still experimental

Perhaps, docs should be adjusted in case of WSL.

In case of native support, I'm concerned more about 

> only the JavaScript runtime is supported; the package manager, test runner, and bundler have been disabled

than performance and excessive debug logs.

P.S.: I'm not a professional in JS stuff :)

anbraten:
replaced by #3676

PR #3451, opened 2024-02-27
Step logs removing API and Button

zc-devs:
Closes #3444
closes #1272

---

Admin:
Screenshot 2024-02-27 1
Screenshot 2024-02-27 2
Screenshot 2024-02-27 3

User:
Screenshot 2024-02-27 4

PR #3458, opened 2024-03-01
CLI: remove step logs

zc-devs:
On top of #3451, addresses [PR note](https://github.com/woodpecker-ci/woodpecker/pull/3451#discussion_r1505438144)

Not tested.

### DIFF
--- a/woodpecker-go/woodpecker/interface.go
+++ b/woodpecker-go/woodpecker/interface.go
@@ -116,7 +116,7 @@ type Client interface {
 	Deploy(repoID, pipeline int64, env string, params map[string]string) (*Pipeline, error)
 
 	// LogsPurge purges the pipeline logs for the specified pipeline.
-	LogsPurge(repoID, pipeline int64) error
+	LogsPurge(repoID, pipeline int64, step int64) error
 
 	// Registry returns a registry by hostname.
 	Registry(repoID int64, hostname string) (*Registry, error)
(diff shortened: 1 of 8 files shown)


woodpecker-go/woodpecker/interface.go
```diff
@@ -116,7 +116,7 @@ type Client interface {
 	Deploy(repoID, pipeline int64, env string, params map[string]string) (*Pipeline, error)
 
 	// LogsPurge purges the pipeline logs for the specified pipeline.
-	LogsPurge(repoID, pipeline int64) error
+	LogsPurge(repoID, pipeline int64, step int64) error
```
6543, line 119: is breaking ...

you could make step optional via:
```suggestion
	LogsPurge(repoID, pipeline int64, step ...int64) error
```

zc-devs: Thanks, that is what I meant.

6543:
Well you still have to adjust the code as `step` is now not int64 but of type []int64 ... 

So instead of:
```
if step != 0 {
...
````

use
```
if len(step) != 0 {
  stepID := step[0]
  ...
````

**Comment by zc-devs** [comment]
TLDR: Made separate library call. Should not be a breaking too.

---

```
LogsPurge(repoID, pipeline int64, step ...int64) error
```
Looks like it is not many step IDs, but any optional argument. So, I could make it
```
LogsPurge(repoID, pipeline int64, optArg ...int64) error
```
Next question is why are optional args integers only? Then it should be
```
LogsPurge(repoID, pipeline int64, optArg ...Interface{}) error
```
Then we should distinct that arguments somehow. So, it probably would be a map (argName:argValue).
Looks over complicated to me.

zc-devs:
Let's try. Add `build` label, please.

PR #3462, opened 2024-03-05
feat: Remove user part of http clone url

### LINKED ISSUE #3443: Bitbucket Datacenter forge does not use provided git credentials
### Component

server, agent

### Describe the bug

When using bitbucket-dc forge, git pull is performed with the format https://user@repourl instead of using the service account credentials specified in the forge documentation.

+ git init -b develop
Initialized empty Git repository in /woodpecker/src/xxxx.xxxx.it/projects/OP/repos/expor-argocd-app/browse/.git/
+ git config --global --replace-all safe.directory /woodpecker/src/xxx.xxx.it/projects/OP/repos/expor-argocd-app/browse
+ git remote add origin https://xxxxx@bitbucket.xxxx.xxxx.it/scm/op/expor-argocd-app.git
+ git fetch --no-tags --depth=1 --filter=tree:0 origin +da3bb[2](https://peck.xxx.it/repos/1/pipeline/10#L2)[3](https://peck.municipia.eng.it/repos/1/pipeline/10#L3)ebbf90[4](https://peck.municipia.eng.it/repos/1/pipeline/10#L4)26fcdf840472[5](https://peck.municipia.eng.it/repos/1/pipeline/10#L5)18094d7ba2e2f:
fatal: could not read Password for 'https://xxxx@xxx.xxxx.it': No such device or address
exit status 128

### System Info

```shell
helm chart 1.1.1 image next-c3e4c14c23
```

### Additional context

_No response_

### Validations

langecode:
Some configuration of Bitbucket Datacenter seems to return username as part of the clone URL. This needs to be removed for `.netrc` to work properly.

Closes #3443

### DIFF
--- a/server/forge/bitbucketdatacenter/convert.go
+++ b/server/forge/bitbucketdatacenter/convert.go
@@ -16,6 +16,7 @@ package bitbucketdatacenter
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


langecode:
> I assume you tested?

Yes, it is tested :-)

PR #3469, opened 2024-03-06
WIP: Add command grouping and render closable log output per each command

lafriks:
- [x] Refactor command output and change to more unique command prefix
- [ ] Render executed command as summary for each command


anbraten:
Was thinking about doing this mainly in the UI by detecting a string like [githubs `::group::{title}` ](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines) and replacing it with html like `<details><summary>{{summary}}</summary>` and `::endgroup::` with `</details>`. Just not sure how we could tell vue to add a html element without having the closing part immediately.

lafriks:
Yes, I do

PR #3376, opened 2024-02-12
Kubernetes documentation improvements

zc-devs:

PR #3471, opened 2024-03-08
Update prettier version

qwerty287:
Use latest to fix issues in #3376


anbraten (approved):

pat-s:
We could maintain the version, and it's updates via renovate - should we?

Issue #3473, opened 2024-03-08
Support for setting runtime class on Kubernetes backend


stevapple:
### Clear and concise description of the problem

Woodpecker can support setting the `runtimeClassName` configuration for a pipeline step Pod. `runtimeClassName` specifies the alternative container runtime to use, which is useful when we need different runtime features, most notably CUDA support from NVIDIA’s runtime distributions.

### Suggested solution

Add a configurable `runtimeClass` property in `steps[].backend_options.kubernetes`. When this property is specified, use it as the `spec.runtimeClassName` of the created Pod.

### Additional context

Runtime Class | Kubernetes: https://kubernetes.io/docs/concepts/containers/runtime-class/

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

PR #3474, opened 2024-03-08
Add `runtimeClassName` in Kubernetes backend options

stevapple:
Reaolves #3473


zc-devs:
Schema suggestion: [fixed_schema.patch](https://github.com/woodpecker-ci/woodpecker/files/14547275/fixed_schema.patch)

stevapple:
> Schema suggestion: [fixed_schema.patch](https://github.com/woodpecker-ci/woodpecker/files/14547275/fixed_schema.patch)

Looks quite reasonable, will take it. Thanks!

Should I do the same gardening for `serviceAccountName`, or leave it to another PR?

zc-devs:
> Should I do the same gardening for serviceAccountName, or leave it to another PR?

I like the idea. I would do it in separate PR.

stevapple:
Ping for review/merge

stevapple:
> Should this be available to all repos/pipelines?

I think it should.

> Or can it be used to configure stuff that could be dangerous and therefore is only allowed in trusted repos?

At least this was clarified in the KEP proposal, that security is a non-goal for this feature, and `RuntimeClass` is not restricted by the Kubernetes RBAC system.

> RuntimeClass should not be used a replacement for PodSecurityPolicy. The use cases for defining multiple RuntimeClasses for the same underlying runtime implementation should be extremely limited (generally only around updates & rollouts). To enforce this, no authorization or restrictions are placed directly on RuntimeClass use; in order to restrict a user to a specific RuntimeClass, you must use another policy mechanism such as PodSecurityPolicy.

PR #3482, opened 2024-03-13
Improve security context handling

anbraten:


lafriks (approved):

xoxys:
I've added another comment to the private discussion.

everflux:
I wonder if this change could have let to no able to use the buildx plugin like before, as I am quite sure that this worked with WP 1.3 and Kubernetes backend.

```
  publish:
    settings:
      repo: *repo
      tags: 8h
```

I am still investigating before I am confident to open an issue.

Issue #2634, opened 2023-10-23
Feature to let agents require labels for pipelines


simcop2387:
### Clear and concise description of the problem

Right now if you set `WOODPECKER_FILTER_LABELS=...` on an agent, if a pipeline doesn't set any labels (or only sets labels that matches that agent) then it will accept all pipelines, even if they don't have a label that you defined for them

### Suggested solution

I'm proposing a way to set on the agent side something like this:

```bash
# define the labels and their values for the agent
WOODPECKER_FILTER_LABELS=customlabel1=foobar,...
# define which labels *MUST* be set on a pipeline to be picked up and must match the value the agent has for them
WOODPECKER_REQUIRE_LABELS=customlabel1,...
```

The idea being here then that you could say require that the hostname label is set for a pipeline to run on a specific agent,
and that jobs that don't have a hostname set, are ineligible to run on that host.

### Additional context

The motivation here being that I've got multiple systems I want to run agents on with different resources (disk space, cpu, etc.)
which makes them unsuitable to use in the general case but there are some jobs I want to run there.

Some more specific examples below

    build_server - Major build/job server, has lots of ram and disk space available and isn't used interactively so it can just chug along
    ai_server - Another server that runs some AI jobs.  It has GPUs, CPUs, etc. but not much disk space since it accesses things over the network.  It's also not as power efficient so it's more expensive to run jobs here that aren't AI related.
    riscv_board - Small SBC that can run stuff, has 8gb of ram but only 8gb of local disk also.  Shouldn't run anything except stuff explicitly asking for RISC-V platform
    arm_board - Similar to the riscv board above

So for those systems I'd set build_server to be unrestricted, since it's purpose is to be the general job runner.
The AI server I'd allow jobs that need the gpus or other specific things there, and only those, so set `WOODPECKER_REQUIRE_LABELS=hostname`
And then the arm and riscv SBCs would get `WOODPECKER_REQUIRE_LABELS=platform` since a specific board isn't needed (I have several of each) but
because they're much smaller they shouldn't run anything that could be bigger or that doesn't specifically need them, my plan there is
to have them around for running integration tests on projects on other architectures.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
[...]

PR #3483, opened 2024-03-14
Force agent labels to be set on pipelines as well

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/2634


xoxys:
Yes it would. We can still decide to introduce another config option but is said in the issue the current behavior doesnt make sense to me.

xoxys:
What's the best way to make it none-breaking?

What I tried is to add a flag `WOODPECKER_FILTER_LABELS_REQUIRED=true/false` but how to pass it to the `grpc.Next` method? I think the parameter can be added to the `rpc.Filter` struct, but either way this would require changes to protobuf or am I wrong?

xoxys:
Sure we could, I still think this should be better configurable but open for other decisions.

6543:
this is not easy as the labels are currently used to let agents "advertice capabilitys/propertys" the pipelines can target.

so make them mandatory by default will break all existing pipelines and add unnecessary complexytiy to the config.

to make an label required should be optional!

I think we could either add some prefix like `!`  to indicate such .... etc

6543:
image

ontop of https://github.com/woodpecker-ci/woodpecker/pull/4141:
```diff
 ### `WOODPECKER_AGENT_LABELS`
 
 > Default: empty
 
 Configures custom labels for the agent, to let workflows filter by it.
 Use a list of key-value pairs like `key=value,second-key=*`. `*` can be used as a wildcard.
+If you use `!` as key prefix it is mandatory for the workflow to have that label set (without !) set and matched.
 By default, agents provide three additional labels `platform=os/arch`, `hostname=my-agent` and `repo=*` which can be overwritten if needed.
 To learn how labels work, check out the [pipeline syntax page](../20-usage/20-workflow-syntax.md#labels).
```

Issue #3494, opened 2024-03-16
Duplicate step/service names results in conflicting container names


xoxys:
### Component

server

### Describe the bug

If a workflow contains multiple steps or services with the same name (one service and one step with the same name is affected as well) it will result in a race condition during the container creation.

![image](https://github.com/woodpecker-ci/woodpecker/assets/3391958/82cb05e8-9993-4d3d-ac95-ef6da3acd228)

![image](https://github.com/woodpecker-ci/woodpecker/assets/3391958/09586881-66b5-4b9c-a578-3a1e7cdacd9f)
![image](https://github.com/woodpecker-ci/woodpecker/assets/3391958/f1d2edc2-20c1-4df2-a136-452ee3b772f7)
 

### System Info

```shell
next-5d98a717ac
```

### Additional context

caused by https://github.com/woodpecker-ci/woodpecker/pull/2757/files#diff-6f809c0a6ea3b1924d8664a9f0dba9e308c15a5915d6444a18f0cb913809e26aR90

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

xoxys:
Wondering a bit how this can happen with ULID + StepID in https://github.com/woodpecker-ci/woodpecker/blob/main/server/pipeline/stepbuilder/stepBuilder.go#L288

xoxys:
What's also strange is that it's called `StepID` https://github.com/woodpecker-ci/woodpecker/blob/main/server/pipeline/stepbuilder/stepBuilder.go#L289 while the `WorkflowID` is passed to the method https://github.com/woodpecker-ci/woodpecker/blob/main/server/pipeline/stepbuilder/stepBuilder.go#L166

xoxys:
I'm a bit confused. While the description of `compiler.WithPrefix` is `WithPrefix configures the compiler with the prefix. The prefix is used to prefix container, volume and network names to avoid collision at runtime.` it looks like the container name is __not__ using this prefix. Instead, `"wp_" + step.ID` is used :thinking:

xoxys:
[...]

PR #3497, opened 2024-03-18
Temp rename CI services to avoid duplicate name issues

xoxys:
Workaround for: https://github.com/woodpecker-ci/woodpecker/issues/3494


anbraten (approved):

6543:
I dont like it but ... as workaround we can merge it

PS: the ci fail we have is fixed by  #3499

PR #3498, opened 2024-03-18
Fix skip setup for some general cli commands

anbraten:

PR #3499, opened 2024-03-18
Require login at cli only for subcommands who depend on it

6543:
Fix regression of  #3384 and also fix https://ci.woodpecker-ci.org/repos/3780/pipeline/13941/24


xoxys:
I dont get that change was already fixed in https://github.com/woodpecker-ci/woodpecker/pull/3498

xoxys (changes requested):
See comment

6543:
well ... yes looks like we worked on it at the same time

PR #3500, opened 2024-03-18
Use map on all environment keys in our config

6543:
and also fix https://ci.woodpecker-ci.org/repos/3780/pipeline/13941/24

### DIFF
--- a/.woodpecker/test.yaml
+++ b/.woodpecker/test.yaml
@@ -38,6 +38,8 @@ steps:
     image: *golang_image
     commands:
       - go run go.woodpecker-ci.org/woodpecker/v2/cmd/cli lint
+    environment:
+      WOODPECKER_DISABLE_UPDATE_CHECK: true
     when:
       - event: pull_request
         path:
@@ -93,7 +95,7 @@ steps:
       - vendor
     image: *golang_image
     environment:
-      - WOODPECKER_DATABASE_DRIVER=sqlite3
+      WOODPECKER_DATABASE_DRIVER: sqlite3
     commands:
       - make test-server-datastore-coverage
     when:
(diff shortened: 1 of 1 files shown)


.woodpecker/test.yaml
```diff
@@ -104,8 +106,8 @@ steps:
       - vendor
     image: *golang_image
     environment:
-      - WOODPECKER_DATABASE_DRIVER=postgres
-      - WOODPECKER_DATABASE_DATASOURCE=host=sv-postgres user=postgres dbname=postgres sslmode=disable
+      WOODPECKER_DATABASE_DRIVER: postgres
+      WOODPECKER_DATABASE_DATASOURCE: 'host=postgres user=postgres dbname=postgres sslmode=disable'
```
xoxys, line 110: This reverts the required hostname change related to the service names. Same for mysql.

.woodpecker/test.yaml
```diff
@@ -141,18 +143,18 @@ steps:
...
     ports: ['5432']
     environment:
-      - POSTGRES_USER=postgres
-      - POSTGRES_HOST_AUTH_METHOD=trust
+      POSTGRES_USER: postgres
+      POSTGRES_HOST_AUTH_METHOD: trust
     when: *when
 
-  sv-mysql:
+  mysql:
```
xoxys, line 154: This reverts the step name to temp fix the duplicate names.

6543: ah yes ... that was picked form my old pull and did not adjust to the current conf ...

xoxys (approved):

wxiaoguang:
It is a breaking change .....

Issue #1068, opened 2022-08-03
Delete old pipeline logs after X days or Y new runs


qwerty287:
### Clear and concise description of the problem

From https://codeberg.org/Codeberg-CI/feedback/issues/63:

It would be nice to have an option to delete old logs of pipeline entries. Either automatic (delete logs of succesful runs after X days or Y new runs?) or with some sort of checkbox selection and action?

### Suggested solution

An option that automatically deletes old pipeline logs

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

maxkratz:
Indeed, an option to manually delete pipeline runs (for example, after debugging the CI config on a testing branch) would be nice to have.

anbraten:
I would like to add support for automatic deletion of old pipelines.
Therefore I would suggest the "keep max last x pipelines" approach as it seems to be covering the issue (I want to keep the database size to a minimum by deleting unused content) well and seems to be the easiest solution to implement keeping things simple. This would make special api endpoints as requested in #3506 obsolete for now.

My plan would be the following:

- add UI option to set max amount of pipelines to keep per repo (we could later on add a server max amount as well)
- everytime a new pipeline is created we simply check the amount and delete the oldest ones to get back to amount x specified

xoxys:
I don't get it. Why not keep the API first approach? Why would you hardcode this feature instead of a reusable API?

anbraten:
> I don't get it. Why not keep the API first approach? Why would you hardcode this feature instead of a reusable API?

How should any other users benefit from this feature, will you open-source it as a service? If the plan is to introduce it as an internal feature later on wouldn't those api endpoints be unused by the internal service anyways? I guess we don't plan to call the api endpoints from the code?

xoxys:
[...]

PR #3506, opened 2024-03-18
Add DeletePipeline API

xoxys:
This is just a first step, the final goal is to have an API endpoint to prune Repo Pipelines older than the given date. 

@woodpecker-ci/maintainers Can I get some feedback if this is the right direction?


server/store/datastore/pipeline.go
```diff
@@ -52,9 +52,18 @@ func (s storage) GetPipelineLastBefore(repo *model.Repo, branch string, num int6
 		Get(pipeline))
 }
 
-func (s storage) GetPipelineList(repo *model.Repo, p *model.ListOptions) ([]*model.Pipeline, error) {
+func (s storage) GetPipelineList(repo *model.Repo, p *model.ListOptions, f *model.FilterOptions) ([]*model.Pipeline, error) {
 	pipelines := make([]*model.Pipeline, 0, 16)
-	return pipelines, s.paginate(p).Where("pipeline_repo_id = ?", repo.ID).
+
+	if f != nil {
+		return pipelines, s.paginate(p).Where(builder.Eq{"pipeline_repo_id": repo.ID}).
```
6543, line 59: i would create a new condition with builder and set `builder.Eq{"pipeline_repo_id": repo.ID}` on the beginning

then have an if that will enable filters based on options:
```go
f.After != 0 
  cond = cond.And(builder.Gt{"pipeline_started": f.After}
}
```
...

6543: PS: defining an filter type to pass into the store func is the right way yes :)

xoxys: Thanks!

server/model/pipeline.go
```diff
@@ -53,6 +53,11 @@ type Pipeline struct {
 	IsPrerelease        bool                    `json:"is_prerelease,omitempty"     xorm:"is_prerelease"`
 } //	@name Pipeline
 
+type FilterOptions struct {
```
6543, line 56: PipelineFilter... ?

server/store/datastore/pipeline.go
```diff
@@ -52,9 +52,21 @@ func (s storage) GetPipelineLastBefore(repo *model.Repo, branch string, num int6
 		Get(pipeline))
 }
 
-func (s storage) GetPipelineList(repo *model.Repo, p *model.ListOptions) ([]*model.Pipeline, error) {
+func (s storage) GetPipelineList(repo *model.Repo, p *model.ListOptions, f *model.FilterOptions) ([]*model.Pipeline, error) {
 	pipelines := make([]*model.Pipeline, 0, 16)
-	return pipelines, s.paginate(p).Where("pipeline_repo_id = ?", repo.ID).
+
+	cond := builder.NewCond()
+	cond = cond.And(builder.Eq{"pipeline_repo_id": repo.ID})
```
6543, line 59:
```suggestion
	cond := builder.NewCond().And(builder.Eq{"pipeline_repo_id": repo.ID})
```

server/store/datastore/pipeline.go
```diff
@@ -52,9 +52,21 @@ func (s storage) GetPipelineLastBefore(repo *model.Repo, branch string, num int6
...
+
+	if f != nil {
+		cond = cond.And(builder.Gt{"pipeline_started": f.After})
```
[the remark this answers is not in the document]
xoxys: This will still result in a segfault as `f` can be nil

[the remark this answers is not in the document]
xoxys: Adjusted

xoxys:
Will add some more tests for the delete API later today.

xoxys:
Should we prevent deleting active pipelines for `api.DeletePipelines`?

6543:
:point_up: yes & pending ... as they have to be removed from task queue

server/router/api.go
```diff
@@ -94,6 +94,7 @@ func apiRoutes(e *gin.RouterGroup) {
 
 					repo.GET("/pipelines", api.GetPipelines)
 					repo.POST("/pipelines", session.MustPush, api.CreatePipeline)
+					repo.DELETE("/pipelines", api.DeletePipelines)
```
6543, line 97: must session.MustRepoAdmin()

xoxys:
> ☝️ yes & pending ... as they have to be removed from task queue

I have added a helper function to check the status and applied it to DeletePipelineLogs as well. `model.StatusRunning, model.StatusPending, model.StatusBlocked` are blocked now, for both delete actions.

anbraten:
@xoxys Is your plan to have something like requested in #1068? (seems to have quite a few 👍🏾)  So currently this would be done by calling the api, should we add some cron / ui button as well?

xoxys:
> Is your plan to have something like requested in https://github.com/woodpecker-ci/woodpecker/issues/1068
In general, yes. My DB is growing constantly already.

>  should we add some cron / ui button as well?
I would like to separate this. First, let's get the API changes merged. After that, we can add UI components, CLI, etc.
Personally, I don't need a UI, I will just use the CLI to run a scheduled job to clean up old pipelines.

Note: These APIs are about deleting the entire pipeline, not only it logs.

xoxys:
In addition to this, an API `DeletePipelineLogs` exists already and there is a PR https://github.com/woodpecker-ci/woodpecker/pull/3451 to add the ability to delete logs of particular steps.

anbraten:
One think I am wondering is, would it be cleaner to have a delete-pipeline endpoint instead of delete-pipelines (as that could be used to delete a single pipeline) similar to delete pipeline-logs of a single pipeline etc. And then the get endpoint with time filters could be used to get all pipelines before and after x and delete them with a script etc by using their id?

xoxys:
I would like to avoid extra scripting around it. However, I can still add a DeletePipeline API that just calls `store.DeletePipeline` to delete a single pipeline by ID. Or is there any downside to have both API endpoints?

xoxys:
@woodpecker-ci/maintainers Anything missing for this?

anbraten:
> I would like to avoid extra scripting around it. 

I see, but shouldn't we then integrate a cron job doing that based on a setting into wp directly then?

> However, I can also add DeletePipeline API that just calls store.DeletePipeline to delete a single pipeline by ID. Or is there any downside to have both API endpoints?

It feels quite powerful to me having an endpoint that could delete all pipelines of a repo.

xoxys:
> It feels quite powerful to me having an endpoint that could delete all pipelines of a repo.

Please make a decision and I'll implement it that way then.

> I see, but shouldn't we then integrate a cron job doing that based on a setting into wp directly then?

Again, yes we can add client integrations (UI, CLI) later please let's get the API changes merged first.

xoxys:
There is a `DeleteRepo` API exposed already, this will delete everything from the DB as well. To be honest, I don't see why we should not allow repo admins (they should know what they are doing) to delete all pipelines from a repo.

xoxys:
@woodpecker-bot Can we make a decision please? I would like to move on and get it merged. A rapidly growing DB is a real issue for frequently used instances...

xoxys:
As discussed in this thread and in https://github.com/woodpecker-ci/woodpecker/issues/1068, I would still like to have an API endpoint for deleting pipelines. Since @anbraten prefers to have an endpoint to delete a single pipeline instead of multiple, I have refactored the PR.

PR #3509, opened 2024-03-19
Post on mastodon when releasing a new version

anbraten:
Needs: https://codeberg.org/woodpecker-plugins/mastodon-post/pulls/6


6543 (approved):

dunn:
not sure about having a tool that only produces accurate text by accident in charge of public announcements of changes to the software ...

6543:
well if you revere to using llms to summarize the changelog ... I'm personally consider myselve not the best in writing, so this tradeoff is not bad for me personally. in the worst case we have to delete and toot a new post. PS: i hope it would just haras people etc ... but that should be ensured by the ai providers etc ...

6543:
PS: PS: if openai would be deterministic, we could have a similar proces like the changelog, where a programm generates it but a person do have to validate things before actions are done ...

Issue #3445, opened 2024-02-26
Support GitHub Deployments 'task' parameter


fernandrone:
### Clear and concise description of the problem

As a user of Woodpecker I want the `deployment` event to support the `task` parameter. I'm open to submitting a pull request if this is a desired feature and we can agree on a design.

The `deployment` event is [a Github-only event](https://woodpecker-ci.org/docs/administration/forges/overview#supported-features) that can currently be filtered by an [environment](https://woodpecker-ci.org/docs/usage/workflow-syntax#environment-1). However the [GitHub API](https://docs.github.com/en/rest/deployments/deployments?apiVersion=2022-11-28#create-a-deployment) also accepts a `task` parameter which has the intention of further "filtering" the execution path of a pipeline:

> The `task` parameter is used by the deployment system to allow different execution paths. In the web world this might be `deploy:migrations` to run schema changes on the system. In the compiled world this could be a flag to compile an application with debugging enabled.

At the moment, Woodpecker does not parse this parameter, so it is discarded. However, if it were to be parsed, it would allow greater flexibility when using the deployment event.

For a more specific use case, we've adopted Woodpecker and we're using GitHub Deployments to trigger pipelines on demand. Although the `manual` event can be used we like many of the features that come out of the box with GitHub deployments (such as deployment queues as well as a page on your repository with all the past deployments, sorted by environment). The extra variable would allow more fine-grained control, that today can only be achieved with workarounds such as custom branch prefixes or tag names; for example, to build only one specific module, or only run the "migration" step.

### Suggested solution

For example, the proposed syntax below could be used to run only database migrations in the production environment.

```
when:
  - event: deployment
     environment: production
     task: migration
```

Or, in a monorepo with multiple applications, deploy a single application.

```yaml
when:
  - event: deployment
     environment: staging
     task: deploy:app1
```

A new environment variable would likely be required, seeing that `CI_PIPELINE_DEPLOY_TARGET` refers to the value of environment I think a `CI_PIPELINE_DEPLOY_TASK` would likely be a good candidate (and a `CI_PREV_PIPELINE_DEPLOY_TASK` as well).

From what I can see the implementation should be straightforward, as it's an existing field on a GitHub event, you could add a `hook.GetDeployment().GetTask()` on parseDeployHook: https://github.com/woodpecker-ci/woodpecker/blob/main/server/forge/github/parse.go#L122-L133

### Validations

[...]

PR #3512, opened 2024-03-19
feat: support github deploy task

fernandrone:
Fixes https://github.com/woodpecker-ci/woodpecker/issues/3445

Support GitHub Deployments 'task' parameter (ref: https://docs.github.com/en/rest/deployments/deployments?apiVersion=2022-11-28#create-a-deployment).


docs/versioned_docs/version-2.1/20-usage/20-workflow-syntax.md
```diff
@@ -803,6 +803,17 @@ when:
...
 
+### `task`
+
+Execute a step for deployment events matching the target deployment task:
+
+```diff
+when:
+  task: migration
+  event: deployment
+```
```
anbraten, line 814: Not sure if we really need a separate when element here. To not bloat the list I think it would be better to just use `evaluate: 'CI_PREV_PIPELINE_DEPLOY_TASK == "my-task"'` instead.

fernandrone:
I've manually this on top of Woodpecker 2.3 on our internal infrastructure and it's working as I'd expect.

server/api/pipeline.go
```diff
@@ -409,6 +409,9 @@ func PostPipeline(c *gin.Context) {
 	// make Deploy overridable
 	pl.Deploy = c.DefaultQuery("deploy_to", pl.Deploy)
 
+	// make Deploy task overridable
+	pl.DeployTask = c.DefaultQuery("deploy_task", pl.DeployTask)
```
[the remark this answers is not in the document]
fernandrone: A few considerations:

* Not sure if the "task" input required or not. The "environment" field is required (which is a good idea; the GitHub API will default it to "production" if its unset so it's good to prompt users to be explicit). The "task" field will default to "deploy" if unset so I see it as less critical... As I'm unsure I've copied the environment field and made it required but could be seem as a breaking change (we're asking users to input something they didn't have to before).

* I've added translations for a few languages I have some familiarity, but I'm not sure what's the recommended approach here.

fernandrone:
Updated 👍🏻

docs/versioned_docs/version-2.1/20-usage/20-workflow-syntax.md
anbraten: Would still prefer the environment variable version as this is a quite specific feature for github only and that way it wont be to present in the docs / yaml autocompletion etc. But I wont block it and leave that up to you.

anbraten: > I'm totally fine with not adding a specific yaml key, but then I'd say we should also remove environment for consistency reasons there.

True. That would be breaking, so we would need to deprecate it first?

fernandrone:
Sorry, I had seen the discussion but I thought you were talking about hypotheticals 😅 yeah, sure, I should be able to remote the task key during the weekend.

pat-s:
@fernandrone To be able to make it into 2.5.0 (which is already overdue) the open points should be resolved within the next days.

fernandrone:
Here it is -- sorry for the delay. I'm not 100% sure I removed the right ammount of code. I may be able to ship and validate this change in a test environment by Friday but if you also have the means to validate this quickly on your side that'd be appreciated.

Dan6erbond:
I landed here from the [Here is Woodpecker 2.5.0](https://woodpecker-ci.org/blog/release-v250#support-for-github-deploy-tasks) blog post and while I really like this feature, it's exactly what I needed to allow us to switch from Drone to Woodpecker, it would really be great if the docs also explain a bit clearer how deployments work! The [Forges overview](https://woodpecker-ci.org/docs/next/administration/forges/overview) also specifically seems to imply that deployments aren't available with Gitea, and I couldn't find docs describing the manual deployment strategy via the UI so I had to run an instance myself just to figure this out. 

To be clear, the [Workflow syntax](https://woodpecker-ci.org/docs/next/usage/workflow-syntax) page mentions `deployment` as an event, but even there it claims "[only available for GitHub](https://woodpecker-ci.org/docs/next/usage/workflow-syntax#event)" which doesn't actually seem to be true given the deploy UI:

image

pat-s:
Gitea doesn't support the "deployment" even but this is different to "deployments" for pipelines. I also fell over this a few months ago and was confused.

So essentially "forge events" != "workflow triggers".

Issue #3470, opened 2024-03-07
Provide a separate WOODPECKER_GITEA_API_URL for easier local setup


everflux:
### Clear and concise description of the problem

When running inside kubernetes/docker without externally reachable domains, the .local and .localhost domains can be used for browser based access.
These point to 127.0.0.1 which does not work for inter-container connectivity.

### Suggested solution

An optional, separate env for the API of gitea server for OAuth and API requests that is separate from the URL that is accessed by the browser of the user would fix that.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

everflux:
I looks promising, I did not have time for testing, though. If it indeed solves the issue it might be very helpful to add it to the documentation.

everflux:
It does not seem to work for my environment. (Context: I want to give woodpecker/gitea based trainings using a minikube cluster with no external domains)

This is my setup:
WOODPECKER_HOST http://woodpecker.localhost
WOODPECKER_GITEA_URL http://gitea-http.gitea:3000
WOODPECKER_DEV_OAUTH_HOST http://gitea.localhost

The oauth redirect is performed to http://gitea-http.gitea:3000 it seems the dev setting is overridden by the gitea url.

anbraten:
As a side note, we have a gitpod setup (hopefully working atm) which we used quite successful for workshop purposes some time ago.

Issue #3510, opened 2024-03-19
Parsing error for alternative secret names


pat-s:
### Component

server

### Describe the bug

Setting something like

```
    environment:
      ENV:
        from_secret: secret
```

yields

`Incorrect type. Expected one of boolean, string, number.` as an inline lint schema error.

Additionally in the server

```
[linter]
release: steps
Must validate one and only one schema (oneOf)
[linter]
release: steps.release-chart.environment
Must validate one and only one schema (oneOf)
[linter]
release: steps.release-chart.environment.ENV
Invalid type. Expected: [boolean,string,number], given: object
```

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-e57a09a404"}

2024-03-18
```

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

PR #3513, opened 2024-03-20
Allow separate gitea oauth URL 

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/3470

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,7 +53,7 @@ ci:
   autofix_prs: true
   autoupdate_branch: ''
   autoupdate_commit_msg: '[pre-commit.ci] pre-commit autoupdate'
-  autoupdate_schedule: monthly
+  autoupdate_schedule: quarterly
   # NB: hadolint not included in pre-commit.ci
   skip: [check-hooks-apply, check-useless-excludes, hadolint, prettier]
   submodules: false
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ check-xgo: ## Check if xgo is installed
 
 install-tools: ## Install development tools
 	@hash golangci-lint > /dev/null 2>&1; if [ $$? -ne 0 ]; then \
-		go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.56.2 ; \
+		go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest ; \
 	fi ; \
 	hash gofumpt > /dev/null 2>&1; if [ $$? -ne 0 ]; then \
 		go install mvdan.cc/gofumpt@latest; \


xoxys:
IMO, this should also be named `_DEV_` and outlined that this is not required in most production setup. I would also prefer `OAuthURL` over `OAuth2URL` but no strong feelings about this.

closes https://github.com/woodpecker-ci/woodpecker/issues/3510 seems to be a drive-by fix that might better go in its own PR.

xoxys:
> I don't think so as it's not a dev flag. Maybe we can add some note in docs that most won't need it?

But for what real world scenario this is required? It just applies for dev/test setups, otherwise you have an accessible external domain anyway, or am I wrong?

anbraten:
If I didn't missed anything `WOODPECKER_DEV_OAUTH_HOST` is exactly doing the same already. At least I am using it successfully locally with gitea during development.

xoxys:
Yes me too, and I was wondering as well but had no time to check the details yet.

xoxys:
Yes, the WOODPECKER_GITEA_URL needs to be accessible from the container and from the browser. If you can't find a value that fits both this new option is required. Makes sense to me.

server/forge/gitea/gitea.go
```diff
@@ -63,6 +62,7 @@ type Gitea struct {
 // Opts defines configuration options.
 type Opts struct {
 	URL        string // Gitea server url.
+	OAuth2URL  string // Gitea server url.
```
xoxys, line 65: Better description?

server/forge/gitea/gitea_test.go
```diff
@@ -62,10 +62,6 @@ func Test_gitea(t *testing.T) {
 				g.Assert(f.url).Equal("http://localhost:8080")
 				g.Assert(f.SkipVerify).Equal(true)
 			})
-			g.It("Should handle malformed url", func() {
```
xoxys, line 65: Why was this removed?

cmd/server/flags.go
```diff
@@ -386,6 +386,11 @@ var flags = append([]cli.Flag{
 		Name:    "gitea-skip-verify",
 		Usage:   "gitea skip ssl verification",
 	},
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_DEV_GITEA_OAUTH_URL"},
+		Name:    "gitea-oauth-server",
+		Usage:   "user-facing gitea server address",
```
xoxys, line 392:
```suggestion
		Usage:   "user-facing gitea server url for oauth",
```

xoxys (approved):

cmd/server/flags.go
xoxys: Just aligned it to the new description. Feel free to skip this change.

PR #3514, opened 2024-03-20
Fix env schema

### LINKED ISSUE #3510: Parsing error for alternative secret names

qwerty287:
closes #3510

### DIFF
--- a/pipeline/frontend/yaml/linter/schema/schema.json
+++ b/pipeline/frontend/yaml/linter/schema/schema.json
@@ -630,7 +630,7 @@
         {
           "type": "object",
           "additionalProperties": {
-            "type": ["boolean", "string", "number"]
+            "type": ["boolean", "string", "number", "array", "object"]
           }
         }
       ]


anbraten (approved):

6543:
We might specify object bit stricter to only allow secrets ...

PR #3516, opened 2024-03-20
Fix uppercased env

### LINKED ISSUE #3515: Environment variable names are silently changed to SCREAMING_SNAKE_CASE
### Component

agent

### Describe the bug

Our app uses the environment variable naming schema `application_name.property_name`, e.g. `matrix.environment='test'`. In 2.3.0 this worked fine, but as of 2.4.0 this is being silently changed to `MATRIX_ENVIRONMENT='test'`, which our app isn't recognizing. I can't find any option to disable this behavior in the documentation.

While I do understand that the screaming snake case is much more common than what we're using, ours is still a valid environment variable name and silently correcting it like this is causing us a ton of headache.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.4.0"}
```

### Additional context

_No response_

### Validations

qwerty287:
closes #3515 

I think after this is fixed, we should publish a new release as this can be quite important.

### DIFF
--- a/pipeline/frontend/yaml/compiler/settings/params.go
+++ b/pipeline/frontend/yaml/compiler/settings/params.go
@@ -26,7 +26,7 @@ import (
 
 // ParamsToEnv uses reflection to convert a map[string]interface to a list
 // of environment variables.
-func ParamsToEnv(from map[string]any, to map[string]string, prefix string, getSecretValue func(name string) (string, error)) (err error) {
+func ParamsToEnv(from map[string]any, to map[string]string, prefix string, upper bool, getSecretValue func(name string) (string, error)) (err error) {
 	if to == nil {
 		return fmt.Errorf("no map to write to")
 	}
@@ -34,7 +34,7 @@ func ParamsToEnv(from map[string]any, to map[string]string, prefix string, getSe
 		if v == nil || len(k) == 0 {
 			continue
 		}
-		to[sanitizeParamKey(prefix, k)], err = sanitizeParamValue(v, getSecretValue)
+		to[sanitizeParamKey(prefix, upper, k)], err = sanitizeParamValue(v, getSecretValue)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


xoxys:
Uh... Do you know which PR has introduced it?

pipeline/frontend/yaml/compiler/settings/params.go
```diff
@@ -26,15 +26,15 @@ import (
 
 // ParamsToEnv uses reflection to convert a map[string]interface to a list
 // of environment variables.
-func ParamsToEnv(from map[string]any, to map[string]string, prefix string, getSecretValue func(name string) (string, error)) (err error) {
+func ParamsToEnv(from map[string]any, to map[string]string, prefix string, upper bool, getSecretValue func(name string) (string, error)) (err error) {
```
xoxys, line 29: Can we call it `toUpper`? Sorry, I know a bit nitpicky.

lafriks (approved):

pipeline/frontend/yaml/compiler/settings/params.go
xoxys: While I fully disagree about this way of communication, I don't care much enough to further discuss it...

xoxys:
> Lint fail is unrelated

I can take care in a dedicated PR if you like.

pipeline/frontend/yaml/compiler/settings/params.go
lafriks: I thumb'ed it down because in Go function param naming should be short (you will rarely find anything more then few chars let a lone multiple words with uppercase characters)

lafriks: sorry if I offended it with it, was not my intention

xoxys: No worries. I would just like to learn/hear opinions, so a plain thumbs-down says nothing to me. Thanks for adding context.

PR #3518, opened 2024-03-20
Fix cli version comparison and improve setup

anbraten:


xoxys:
This version check implementation has the same issue as the one in the UI. If a "dev" version from a commit is used, it complains. Is this also fixed? Or expected?

anbraten:
For development versions it will skip the check. For pre-releases next versions it will show you the hint to update to the latest stable, that should be okay however, I guess.

xoxys:
Thanks :+1:

xoxys:
Should answer "What happens with next- version?"

> For pre-releases next versions it will show you the hint to update to the latest stable

anbraten:
Adjusted the code to use our `version.json` file. We could potentially run into a 404 error as we don't know if the binary is actually uploaded (yet) if we see a new version, but I think that shouldn't be an issue and the user will simply get an error in that case.

anbraten:
Fixed the wrong url string in the code and added a test. Also did a test again. Seems to work now.

cli/common/hooks.go
```diff
@@ -58,7 +58,7 @@ func After(_ *cli.Context) error {
 		select {
 		case <-waitForUpdateCheck.Done():
 		// When the actual command already finished, we still wait 250ms for the update check to finish
```
xoxys, line 60: Desc is wrong now.

PR #3521, opened 2024-03-20
Fix golangci-lint version pinning

xoxys:


lafriks:
Can be reverted as issue was fixed and new version released: https://github.com/golangci/golangci-lint/pull/4552

xoxys:
Will you take care or should I?

PR #3522, opened 2024-03-20
Only allow to deploy from push, tag and release

qwerty287:


lafriks:
is this already limited in UI this way?

6543 (approved):

PR #3524, opened 2024-03-20
Add option to set the local repository path to the cli command exec.

manuelluis:
The cli command exec not handle in a coherent way the repository path when a directory or filename is given for the pipeline.

` woodpecker-cli exec [command options] [path/to/.woodpecker.yaml]`

If the path to the pipeline is a file in the `.woodpecker` directory, for example: `.woodpecker/pipeline.yaml`, 
the repository path will be: `.woodpecker`

If the path to the pipeline yaml is a directory with more than one level, for example `ci/woodpecker/`,
the repository path will be:  `ci`

In order not to break the old behavior we added a new option to put the root directory of the repository:
~~~
woodpecker-cli exec --local --repo-path . --pipeline-event manual .woodpecker/build.yml 
~~~

### DIFF
--- a/docs/versioned_docs/version-2.4/40-cli.md
+++ b/docs/versioned_docs/version-2.4/40-cli.md
@@ -280,6 +280,8 @@ execute a local pipeline
 
 **--repo-clone-url**="":
 
+**--repo-path**="": path to local repository
+
 **--repo-private**="":
 
 **--repo-remote-id**="":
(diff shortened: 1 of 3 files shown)


lafriks (approved):

docs/versioned_docs/version-2.4/40-cli.md
```diff
@@ -280,6 +280,8 @@ execute a local pipeline
 
 **--repo-clone-url**="":
 
+**--repo-path**="": path to local repository
```
6543, line 283: are we going to backport it ?!?

lafriks: why should we? 🤔

lafriks: ohhhh...  versioned docs....

Issue #3537, opened 2024-03-22
Woodpecker 2.4 breaks privileged steps/plugins with Kubernetes backend


everflux:
This worked with WP 2.3 and Kubernetes backend:

```
  publish:
    image: woodpeckerci/plugin-docker-buildx
    settings:
      repo: *repo
      tags: 8h
```

Using WP 2.4 the docker daemon does not start.

Debugging the Pod manifest I see an empty security context and the docker daemon does not start.

This works with WP 2.4.1 but is not so user friendly:

```
  publish:
    image: woodpeckerci/plugin-docker-buildx
    privileged: true
    backend_options:
      kubernetes:
        securityContext:
          privileged: true
    settings:
      repo: *repo
      tags: 8h
      daemon.debug: "true"
```

See also: https://github.com/woodpecker-ci/woodpecker/issues/3482#issuecomment-2015672185_

zc-devs:
Comments almost useless, but...

1. I think you are talking about `2.3.x`, `2.4.x` versions.
2. One of the options should be enough. Pipeline's `privileged` takes precedence over backend's option, as I remember.
```yaml
    privileged: true
    backend_options:
      kubernetes:
        securityContext:
          privileged: true
```

everflux:
You are correct. I actually meant version 2.4, I linked the comment as I suspect this commit might be the cause.

For me it did not work to set privileged only on pipeline level.
With the old woodpecker version 2.3, it was not necessary to set the option at all.
I am quite sure this is because this image is recognized against list of hardcoded images to be run in privileged mode when used as plugin.

zc-devs:
> I am quite sure this is because this image is recognized against list of hardcoded images to be run in privileged mode when used as plugin.

Seems, it sets up `privileged` on step level.

> With the old woodpecker version 2.3, it was not necessary to set the option at all.

Backend option was checked firstly. If was not set, step-level option was checked (was set by trusted plugin logic) => security context was set.

> For me it did not work to set privileged only on pipeline level.

Now, it is `and` logic: [security context sets from backend option if step option is set too](https://github.com/woodpecker-ci/woodpecker/blob/9db9c7116fc7f85042c103a29f1883a34cce3c69/pipeline/backend/kubernetes/pod.go#L410).
So, as pipeline-level option is set by plugin logic, you have to set backend one as well.

cc @anbraten

anbraten:
[...]

Issue #3164, opened 2024-01-10
Add the ability to override default env variables for Kubernetes pipeline steps


dominic-p:
### Clear and concise description of the problem

I'm trying to run a buildah container in my pipeline which requires the `HOME` variable to be set to `/home/build`. Currently, the `HOME` variable seems to be hardcoded to `/root` [here](https://github.com/woodpecker-ci/woodpecker/blob/768fd71841ff348518c2988bb2d0278f59dc690e/pipeline/backend/common/script.go#L31).

### Suggested solution

It would be nice if we could use the standard pipeline environment variable config to override the `HOME` (and other) variables like so:

```yaml
steps:
  build:
    image: quay.io/buildah/stable:v1.33
    environment:
      - HOME=/home/build
```

### Alternative

I'm not sure if running a command before my main command like `export HOME=/home/build` will work. I haven't tested it yet. If that works, it would be a nice workaround. But, it would still be good to be able to set the variables directly, in my opinion.

__Update:__ I have confirmed that I can run this at the beginning of my build scripts, and it works. So, we have a good workaround. I still do think it would be nice to be able to specify it in the pipeline config with other env variables.

### Additional context

This is part of my ongoing efforts to get buildah running on Woodpecker (#2545). This should be the last piece of the puzzle once the AppArmor PR lands!

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

maltegrosse:
@dominic-p I got buildah up and running (somehow) on k8s.... just have a look here: https://github.com/maltegrosse/woodpecker-buildah (please consider the performance vs. security aspect carefully)

appreciate feedback / improvements  :)

dominic-p:
Thanks for sharing this, @maltegrosse; that looks really cool!

I'm curious how you are passing in registry credentials (see #3582). That's pretty much the last issue I'm having with my cobbled together setup.

[...]

PR #3538, opened 2024-03-23
Using repository trust in security contexts

zc-devs:
Addresses https://github.com/woodpecker-ci/woodpecker/issues/3537#issuecomment-2016480296

Require a trusted repo to be able to configure the security context similar to how the docker backend requires a trusted repo to set options like volumes, privileged, ...


zc-devs:
As it uses step's `privileged`, should close #3537.

6543:
@zc-devs bugfix or enhancement :)

zc-devs:
> bugfix or enhancement

As a user I would call #3537 a bug => bugfix.

---

BTW, I noticed the same [error](https://ci.woodpecker-ci.org/repos/3780/pipeline/14323/9) as I almost always face. This occurs not only with `fr.js`. Does someone know what is it and how to get rid of? Now, I'm restarting a pipeline until it works.

everflux:
Happy to test it out, do you have a pullable image?

pipeline/backend/kubernetes/pod.go
```diff
@@ -464,6 +480,15 @@ func mapToEnvVars(m map[string]string) []v1.EnvVar {
 	return ev
 }
 
+func isRepoTrusted(step *types.Step) bool {
+	// TODO: probably, should be stored in the model (pipeline/backend/types/config.go)
+	trustedRepo, err := strconv.ParseBool(step.Environment["CI_REPO_TRUSTED"])
```
xoxys, line 485: Is it safe to read it from the env? It can't be overwritten by users?

zc-devs: > Is it safe to read it from the env?

¯\_(ツ)_/¯
Elaborate, please.
Do you suggest to use dedicated variable in `pipeline/backend/types/config.go`?
What is the difference between env var (one map's entry) vs dedicated variable?

> It can't be overwritten by users?

Seems, not for now, at least: #3164.

xoxys: >  Elaborate, please.

Right now I have not suggested anything :) I just asked if it is safe to read a security-related setting from an env var. If only the server can set it, and it can't be overwritten by the pipeline config, I'm fine with it, but it might increase the possibilities to become an issue in the future.

zc-devs: Yes, might be an issue. Depends on #3164 implementation.
So, what are we gonna do?

zc-devs:
> do you have a pullable image?

[There](https://hub.docker.com/r/woodpeckerci/woodpecker-server/tags?page=1&name=3538)

zc-devs:
Pipeline:
```yaml
skip_clone: true
steps:
  server:
    image: alpine
    commands:
      - echo Hello
      - sleep 60
    backend_options:
      kubernetes:
        securityContext:
          privileged: true
          runAsUser: 0
          runAsGroup: 0
```

[woodpecker-agent.log](https://github.com/woodpecker-ci/woodpecker/files/14735890/woodpecker-agent.log)
[pod-untrusted.yaml](https://github.com/woodpecker-ci/woodpecker/files/14735887/pod-untrusted.yaml.txt)
[pod-trusted.yaml.txt](https://github.com/woodpecker-ci/woodpecker/files/14735889/pod-trusted.yaml.txt)

everflux:
Thanks for looking into it this quickly.
I have currently only arm machine at my disposal I plan to verify tomorrow evening.

everflux:
Successfully tested the prebuilt image with buildx plugin. Only setting "trusted repository" and no additional "privileged" setting on the workflow or workflow.backend level. Nice!

stevapple:
> Only setting "trusted repository" and no additional "privileged" setting on the workflow or workflow.backend level.

IIRC running a trusted plugin (with a privileged container) shouldn't require the repository to be trusted?

everflux:
I would be really happy if this could be included in 2.5.0 and any further changes could be discussed separately. (For example should escalated plugins be treated as trusted implicitly, would handling of volumes differently from privileged make sense etc)

zc-devs:
> Only setting "trusted repository"

Now, Buildx plugin should work  without "trusted repository" as it was before, on `2.3.x`.

everflux:
Any chance this could be merged into a 2.4.x release?

pipeline/backend/kubernetes/pod.go
xoxys: IMO, the clean solution would be to pass down the repo settings to the backend similar to `Environment` e.g. 
`Step.RepoSettings`.

everflux:
Too bad, I have a training planned on April 9th...

xoxys:
We can add a label to build an image for this PR and you can use that one for your training.

everflux:
Thanks for the proposal, but that won't suffice (mixed user base with apple arm64 and amd64 architectures). No worries, though, I can use drone instead, the concepts are similar enough. And I don't want to cause additional effort in any way, I was just hoping for a point release before 2.5.

pat-s:
I would be ok merging this to allow @everflux to possible use WP on April 9th (if not too late). Besides, I am seeing the same issues on our PROD instance with more than 50+ files I would need to adapt on a mixed cluster, hence I am also having a strong need for a fix that restores 2.3.x functionality.

If not everything is working percectly yet we can still iterate until 2.5.x but in the meantime allow for a workaround via `next` for mixed cluster users.

@zc-devs Merging main caused some build failures now. In addition, `cspell.json` needs a fix. I just a few minutes right now but can look into it later this evening.

xoxys:
Not perfectly working and introducing another security issue again is a difference 🙂 if we are sure reading the repo trust setting from the env var and it cant be manipulated by end users feel free to go on.

zc-devs:
> Merging main caused some build failures now

Fixed.

> reading the repo trust setting from the env var...

> the clean solution would be to pass down the repo settings to the backend similar to Environment e.g.
Step.RepoSettings

TBH, I don't want to put my dirty hands in the "core". And definitely won't before majority agreed on implementation or at least said something on topic.

xoxys:
Fully agree on that and the PR should not be merged untill the majority agreed as well.

pat-s:
@zc-devs @xoxys It seems one conversation from you is still open (see above). How can we make progress there?

If this doesn't make it into 2.5.0 (release date tomorrow), 2.5.0 will also be unusable as 2.4.x for folks coming from 2.3.x without touching many workflow files.

The current situation is therefore a bit suboptimal. If we can't find an agreement, I'd also consider reverting the backend change in 2.4 and restore the 2.3 behavior and include the proposed changes of this PR in a new/updated approach?

zc-devs:
I think, Xoxys and I have met agreement. Moreover, I'm on his side in general (otherwise didn't leave the TODO).

> How can we make progress there?
> the majority agreed

> release date tomorrow

It can be merged as is by deadline. At least I cannot develop in next few days.

> reverting the backend change in 2.4

You [cannot](https://github.com/woodpecker-ci/woodpecker/issues/3537#issuecomment-2016464281).

pat-s:
> You https://github.com/woodpecker-ci/woodpecker/issues/3537#issuecomment-2016464281.

I know about the issue but that doesn't mean it is not revertable at all. Anyone can still run and install any WP version (even the ones including the mentioned flaws). But anyway, as this is not the main goal here, let's focus on getting this one here merged and move onward.

everflux:
Perhaps the release could wait for resolving this issue?

zc-devs:
As bug is fixed in #3711 and there is no solution for review's concern, I'm closing the PR for now.
The code can be obtained as described in [Modifying an inactive pull request locally](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally#modifying-an-inactive-pull-request-locally).

Issue #1816, opened 2023-06-03
add scopes for agents (-> user agents)


6543:
and write checks for all grpc actions that agents can only change what is in there scope

anbraten:
link #267

PR #3539, opened 2024-03-23
Implement org/user agents

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/267
close https://github.com/woodpecker-ci/woodpecker/issues/1816

TODOs:
- [x] api to register agent for orgs/users
- [x] secure all rpc endpoints
- [x] secure set of labels for non-admins
- [x] WebUI


anbraten:
Nice. 

One step would be to add `orgId: <orgId>`, `repoId: <repoId>` labels, so we can later on filter by that context, so an agent could be registered for a repo, org or the whole instance.

6543:
:point_up: that's the idea ... I just wrote down my starting point while chating @ c-base in berling ... and my ideas after brainstorming :D

this will take shape next week(_s^tm_)

server/model/agent.go
```diff
@@ -37,3 +39,20 @@ func (Agent) TableName() string {
...
+	if filters == nil {
+		filters = make(map[string]string)
+	}
+
+	// enforce filters for user and organization agents
+	if a.IsSystemAgent() {
+		filters["repo"] = "*"  // allow all repos by default
+		filters["owner"] = "*" // allow all owners by default
+	} else {
+		filters["owner"] = "*" // we dont have org agents implemented jet
```
anbraten, line 54:
```suggestion
		filters["owner"] = "*" // we don't have org agents implemented yet
```

cmd/agent/core/agent.go
```diff
@@ -187,7 +187,6 @@ func run(c *cli.Context, backends []types.Backend) error {
 		"hostname": hostname,
 		"platform": engInfo.Platform,
 		"backend":  backendEngine.Name(),
-		"repo":     "*", // allow all repos by default
```
anbraten, line 190: Might be a breaking change as agent owners were allowed to filter for repos as well. We could solve this by adding the new enforced filters with sth like `internal_repo_id` and `internal_org_id` and leave `repo` untouched

anbraten (approved):

anbraten:
@6543 I've added a commit with the enforced labels and checks I had in mind. What do you think about that approach? 

(For the permission checks it currently does a few db selects currently like select workflow, select pipeline, select repo which could be replaced by a single select with joins for performance improvements.)

server/grpc/rpc.go
```diff
@@ -95,11 +105,47 @@ func (s *RPC) Wait(c context.Context, id string) error {
...
+	if err != nil {
+		log.Error().Err(err).Msgf("cannot find pipeline with id %d", workflow.PipelineID)
+		return err
+	}
+
+	repo, err := s.store.GetRepo(currentPipeline.RepoID)
+	if err != nil {
+		log.Error().Err(err).Msgf("cannot find repo with id %d", currentPipeline.RepoID)
+		return err
+	}
```
anbraten, line 134: I've added quite a few selects to simply get the repo atm.

server/model/agent.go
```diff
@@ -14,19 +14,26 @@
...
+	OrgID       int64  `json:"owner_id"      xorm:"'owner_id'"` // TODO: rename to org_id
 	Token       string `json:"token"         xorm:"token"`
 	LastContact int64  `json:"last_contact"  xorm:"last_contact"`
 	Platform    string `json:"platform"      xorm:"VARCHAR(100) 'platform'"`
 	Backend     string `json:"backend"       xorm:"VARCHAR(100) 'backend'"`
 	Capacity    int32  `json:"capacity"      xorm:"capacity"`
 	Version     string `json:"version"       xorm:"'version'"`
 	NoSchedule  bool   `json:"no_schedule"   xorm:"no_schedule"`
+	// Server side enforced agent filters
+	Filters map[string]string `json:"filters" xorm:"'filters' json"`
```
anbraten, line 36: What kind of filters do you have in mind that should be set through the UI instead of the labels env variable config of the agent itself?

server/model/agent.go
```diff
@@ -35,5 +42,37 @@ func (Agent) TableName() string {
...
+	}
+
+	// enforce filters for user and organization agents
+	if a.IsSystemAgent() {
+		filters["org-id"] = "*"  // allow all orgs
+		filters["repo-id"] = "*" // allow all repos
+	} else if a.OrgID > 0 {
+		filters["org-id"] = fmt.Sprintf("%d", a.OrgID)
+		filters["repo-id"] = "*" // allow all repos of the org
+	}
```
anbraten, line 63: This is what I had in mind to set agent access labels by filters ☝🏾

server/pipeline/queue.go
```diff
@@ -57,6 +57,13 @@ func queuePipeline(ctx context.Context, repo *model.Repo, pipelineItems []*stepb
 	return server.Config.Services.Queue.PushAtOnce(ctx, tasks)
 }
 
+func enforcedLabels(task *model.Task, repo *model.Repo) {
```
anbraten, line 60:
```suggestion
func enforceLabels(task *model.Task, repo *model.Repo) {
```

server/grpc/auth_server.go
```diff
@@ -89,5 +89,8 @@ func (s *WoodpeckerAuthServer) getAgent(agentID int64, agentToken string) (*mode
 	if err != nil && errors.Is(err, types.RecordNotExist) {
 		return nil, fmt.Errorf("individual agent not found by token: %w", err)
 	}
+
+	// TODO: check if an agent can still pretend to be an other one
+
```
6543, line 94:
```suggestion
```

server/model/agent.go
6543: anything u can think of and users find probably more use-cases ...

6543:
I can confirm the backend already works as intended :)

Only the UI part is missing ...

6543:
technically user agents are just org agents of orgs that are of type user

yes the frontend is dublicated code but I somehow got it to work at all ... if you want to have it sooner, feel free to push changes to the WebUI, ... else it has to wait till i have time to look into it again

server/grpc/filter.go
```diff
@@ -39,9 +42,11 @@ func createFilterFunc(agentFilter rpc.Filter) queue.FilterFn {
 				continue
 			}
 
-			if taskLabelValue != agentLabelValue {
-				return false
+			match, err := doublestar.Match(taskLabelValue, agentLabelValue)
```
anbraten, line 45: Could you give an example how / for what you would use a glob pattern here. Should also be added to the docs.

server/grpc/rpc.go
```diff
@@ -69,6 +67,20 @@ func (s *RPC) Next(c context.Context, agentFilter rpc.Filter) (*rpc.Workflow, er
 		return nil, nil
 	}
 
+	// enforce server set agent filters ...
+	agentServerFilters, err := agent.GetServerFilters()
```
anbraten, line 71: We now have two ways for setting filters. IMO we should decided for one: UI or agent flags.
I would currently favor to keep agent flags as those can be easily set by deployment tools like helm, ansible. The ability to switch labels on the fly (without restarting the agent) is interesting, but honestly not really necessary as I can't image someone updating those regularly.

server/model/agent.go
```diff
@@ -28,13 +35,59 @@ type Agent struct {
...
+	// OrgID is counted as unset if set to -1, this is done to ensure a new(Agent) still enforce the OrgID check by default
+	OrgID int64 `json:"org_id"        xorm:"INDEX 'org_id'"`
+	// RepoID is counted as unset if set to -1, this is done to ensure a new(Agent) still enforce the OrgID check by default
+	RepoID int64 `json:"repo_id"       xorm:"INDEX 'repo_id'"`
+	// Server side enforced agent filters
+	Filters map[string]string `json:"filters" xorm:"'filters' json"`
 } //	@name Agent
 
+const (
+	SystemAgentOwnerID = -1
```
anbraten, line 47: As it is used to signal repo / org is not set as well sth like `IDNotSet` would be better:

```suggestion
	IDNotSet = -1
```

server/model/task.go
```diff
@@ -41,6 +41,19 @@ func (t *Task) String() string {
 	return sb.String()
 }
 
+func (t *Task) CalcLabels(r *Repo) (map[string]string, error) {
```
anbraten, line 44:
```suggestion
func (t *Task) GetLabels(r *Repo) (map[string]string, error) {
```

server/router/api.go
```diff
@@ -137,6 +142,15 @@ func apiRoutes(e *gin.RouterGroup) {
 					repo.PATCH("/cron/:cron", session.MustPush, api.PatchCron)
 					repo.DELETE("/cron/:cron", session.MustPush, api.DeleteCron)
 
+					agentBase := repo.Group("/agents")
```
anbraten, line 145: Do we really need repo agents? As you can apply filters to org / (user org) agents for specific repos I don't think we need them TBH.

server/grpc/filter.go
6543: I think this pull is already much I'll move that onto the next one

6543: -> #4122

server/pipeline/queue.go
```diff
@@ -31,13 +32,14 @@ func queuePipeline(ctx context.Context, repo *model.Repo, pipelineItems []*stepb
...
-		task.ID = fmt.Sprint(item.Workflow.ID)
-		task.Labels = map[string]string{}
-		for k, v := range item.Labels {
-			task.Labels[k] = v
+		task := &model.Task{
+			ID:     fmt.Sprint(item.Workflow.ID),
+			Labels: make(map[string]string),
+		}
+		maps.Copy(task.Labels, item.Labels)
+		if _, err := task.CalcLabels(repo); err != nil {
```
6543, line 40:
```suggestion
		if _, err := task.GetLabels(repo); err != nil {
```

server/store/datastore/migration/015_add_org_and_repo_agents.go
```diff
@@ -0,0 +1,52 @@
...
+	ID: "add-org-and-repo-agents",
+	MigrateSession: func(sess *xorm.Session) (err error) {
+		if err := sess.Sync(new(agentV015)); err != nil {
+			return fmt.Errorf("sync models failed: %w", err)
+		}
+
+		// Explizite make current existing agents to global agents
+		_, err = sess.Cols("org_id", "repo_id").Update(&model.Agent{
+			OrgID:  model.SystemAgentOwnerID,
+			RepoID: model.SystemAgentOwnerID,
```
6543, line 48:
```suggestion
			OrgID:  model.IDNotSet,
			RepoID: model.IDNotSet,
```

server/api/agent.go
```diff
@@ -172,11 +180,12 @@ func PostAgent(c *gin.Context) {
 
 	agent := &model.Agent{
 		Name:       in.Name,
-		NoSchedule: in.NoSchedule,
 		OwnerID:    user.ID,
-		Token: base32.StdEncoding.EncodeToString(
-			securecookie.GenerateRandomKey(32),
-		),
+		OrgID:      model.SystemAgentOwnerID,
+		RepoID:     model.SystemAgentOwnerID,
```
6543, line 185:
```suggestion
		OrgID:      model.IDNotSet,
		RepoID:     model.IDNotSet,
```

server/api/agent.go
```diff
@@ -227,3 +236,397 @@ func DeleteAgent(c *gin.Context) {
...
+	if err != nil {
+		c.String(http.StatusBadRequest, err.Error())
+		return
+	}
+
+	agent := &model.Agent{
+		Name:       in.Name,
+		OwnerID:    user.ID,
+		OrgID:      orgID,
+		RepoID:     model.SystemAgentOwnerID,
```
6543, line 276:
```suggestion
		RepoID:     model.IDNotSet,
```

server/model/agent.go
```diff
@@ -28,13 +35,59 @@ type Agent struct {
...
+		filters = make(map[string]string)
+	}
+
+	// enforce filters for user and organization agents
+	if a.OrgID != SystemAgentOwnerID {
+		filters[agentFilterOrgID] = fmt.Sprintf("%d", a.OrgID)
+	} else {
+		filters[agentFilterOrgID] = "*"
+	}
+	if a.RepoID != SystemAgentOwnerID {
```
6543, line 77:
```suggestion
	if a.OrgID != IDNotSet {
		filters[agentFilterOrgID] = fmt.Sprintf("%d", a.OrgID)
	} else {
		filters[agentFilterOrgID] = "*"
	}
	if a.RepoID != IDNotSet {
```

server/model/agent.go
```diff
@@ -28,13 +35,59 @@ type Agent struct {
...
+	return filters, nil
+}
+
+func (a *Agent) CanAccessRepo(repo *Repo) bool {
+	if a.IsSystemAgent() {
+		return true
+	}
+
+	return a.RepoID != SystemAgentOwnerID && a.RepoID == repo.ID && (a.OrgID == SystemAgentOwnerID || a.OrgID == repo.OrgID) ||
+		a.OrgID != SystemAgentOwnerID && a.OrgID == repo.OrgID && (a.RepoID == SystemAgentOwnerID || a.RepoID == repo.ID)
```
6543, line 92:
```suggestion
	return a.RepoID != IDNotSet && a.RepoID == repo.ID && (a.OrgID == IDNotSet || a.OrgID == repo.OrgID) ||
		a.OrgID != IDNotSet && a.OrgID == repo.OrgID && (a.RepoID == IDNotSet || a.RepoID == repo.ID)
```

server/model/agent_test.go
```diff
@@ -0,0 +1,149 @@
...
+		assert.Equal(t, map[string]string{
+			agentFilterRepoID: "0",
+			agentFilterOrgID:  "0",
+		}, filters)
+	})
+
+	t.Run("SystemAgent", func(t *testing.T) {
+		agent := &Agent{
+			OrgID:  SystemAgentOwnerID,
+			RepoID: SystemAgentOwnerID,
```
6543, line 47:
```suggestion
			OrgID:  IDNotSet,
			RepoID: IDNotSet,
```

server/model/agent_test.go
```diff
@@ -0,0 +1,149 @@
...
+		assert.Equal(t, map[string]string{
+			agentFilterOrgID:  "*",
+			agentFilterRepoID: "*",
+		}, filters)
+	})
+
+	t.Run("OrgAgent", func(t *testing.T) {
+		agent := &Agent{
+			OrgID:  123,
+			RepoID: SystemAgentOwnerID,
```
6543, line 60:
```suggestion
			RepoID: IDNotSet,
```

server/model/agent_test.go
```diff
@@ -0,0 +1,149 @@
...
+		assert.NoError(t, err)
+		assert.Equal(t, map[string]string{
+			agentFilterOrgID:  "123",
+			agentFilterRepoID: "*",
+		}, filters)
+	})
+
+	t.Run("RepoAgent", func(t *testing.T) {
+		agent := &Agent{
+			OrgID:  SystemAgentOwnerID,
```
6543, line 72:
```suggestion
			OrgID:  IDNotSet,
```

server/model/agent_test.go
```diff
@@ -0,0 +1,149 @@
...
+	t.Run("EmptyAgent", func(t *testing.T) {
+		agent := &Agent{}
+		assert.False(t, agent.CanAccessRepo(repo))
+	})
+
+	t.Run("SystemAgent", func(t *testing.T) {
+		agent := &Agent{
+			OwnerID: SystemAgentOwnerID,
+			OrgID:   SystemAgentOwnerID,
+			RepoID:  SystemAgentOwnerID,
```
6543, line 110:
```suggestion
			OwnerID: IDNotSet,
			OrgID:   IDNotSet,
			RepoID:  IDNotSet,
```

server/model/agent_test.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+
+		assert.True(t, agent.CanAccessRepo(repo))
+	})
+
+	t.Run("OrgAgent", func(t *testing.T) {
+		agent := &Agent{
+			OrgID:  12,
+			RepoID: SystemAgentOwnerID,
```
6543, line 119:
```suggestion
			RepoID: IDNotSet,
```

server/model/agent_test.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+		assert.True(t, agent.CanAccessRepo(repo))
+		assert.False(t, agent.CanAccessRepo(otherRepo))
+	})
+
+	t.Run("RepoAgent", func(t *testing.T) {
+		agent := &Agent{
+			OrgID:  SystemAgentOwnerID,
```
6543, line 127:
```suggestion
			OrgID:  IDNotSet,
```

server/router/api.go
6543: it does not add more complexity and allow a even more granular management ...
... one of the things i want to do next is to have a prioritized queue and with such it should priorize the lowest hirarchy always if it can ...

beside that, you could be in an org and be an repo admin but not an org admin ... this would cover the case

anbraten: > it does not add more complexity and allow a even more granular management ...

It does add more code and thereby complexity ... 😉 As the benefit isn't granted / rather unclear atm we should not add them for now IMO

server/api/agent.go
```diff
@@ -97,7 +99,10 @@ func GetAgentTasks(c *gin.Context) {
 	info := server.Config.Services.Queue.Info(c)
 	for _, task := range info.Running {
 		if task.AgentID == agent.ID {
-			tasks = append(tasks, task)
+			t := *task
+			// we copy the task and remove sensitive data
+			t.Data = nil
+			tasks = append(tasks, &t)
```
anbraten, line 105: already done by json flag `-`

server/model/agent.go
```diff
@@ -14,12 +14,19 @@
...
+	"github.com/gorilla/securecookie"
+)
+
 type Agent struct {
 	ID          int64  `json:"id"            xorm:"pk autoincr 'id'"`
 	Created     int64  `json:"created"       xorm:"created"`
 	Updated     int64  `json:"updated"       xorm:"updated"`
 	Name        string `json:"name"          xorm:"name"`
-	OwnerID     int64  `json:"owner_id"      xorm:"'owner_id'"`
+	OwnerID     int64  `json:"owner_id"      xorm:"INDEX 'owner_id'"`
```
anbraten, line 29: Doesn't really seem that we query by the `owner_id` so the index should not be necessary. In general the `owner_id` might be redundant to `org_id` & `repo_id` now.

web/src/components/admin/settings/AdminQueueTab.vue
```diff
@@ -110,7 +110,12 @@ const tasks = computed(() => {
     _tasks.push(...queueInfo.value.waiting_on_deps.map((task) => ({ ...task, status: 'waiting_on_deps' })));
   }
 
-  return _tasks.sort((a, b) => a.id - b.id);
+  return _tasks
+    .map((task) => ({
+      ...task,
+      labels: Object.fromEntries(Object.entries(task.labels).filter(([key]) => key !== 'org-id' && key !== 'repo-id')),
+    }))
+    .sort((a, b) => a.id - b.id);
```
anbraten, line 118: `sort` modifies the var.

```suggestion
    .toSorted((a, b) => a.id - b.id);
```

server/store/datastore/migration/015_add_org_and_repo_agents.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+var addOrgAndRepoAgents = xormigrate.Migration{
+	ID: "add-org-and-repo-agents",
+	MigrateSession: func(sess *xorm.Session) (err error) {
+		if err := sess.Sync(new(agentV015)); err != nil {
+			return fmt.Errorf("sync models failed: %w", err)
+		}
+
+		// Explizite make current existing agents to global agents
```
anbraten, line 45:
```suggestion
		// Update all existing agents to be global agents
```

server/queue/queue.go
```diff
@@ -28,6 +28,9 @@ var (
 
 	// ErrNotFound indicates the task was not found in the queue.
 	ErrNotFound = errors.New("queue: task not found")
+
+	// ErrAgentMissMatch indicates an agent does not mat to the task.
+	ErrAgentMissMatch = errors.New("task has not expected agent id")
```
anbraten, line 33:
```suggestion
	// ErrAgentMissMatch indicates a task is assigned to a different agent
	ErrAgentMissMatch = errors.New("task assigned to different agent")
```

server/model/task_test.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+func TestTask_CalcLabels(t *testing.T) {
+	t.Run("Nil Repo", func(t *testing.T) {
+		task := &Task{}
+		labels, err := task.CalcLabels(nil)
+
+		assert.Error(t, err)
+		assert.Nil(t, labels)
+		assert.EqualError(t, err, "repo is nil but needed for calc task labels")
```
anbraten, line 30:
```suggestion
		assert.EqualError(t, err, "repo is nil but needed to get task labels")
```

server/model/task.go
```diff
@@ -41,6 +41,19 @@ func (t *Task) String() string {
 	return sb.String()
 }
 
+func (t *Task) GetLabels(r *Repo) (map[string]string, error) {
+	if r == nil {
+		return nil, fmt.Errorf("repo is nil but needed for calc task labels")
```
anbraten, line 46:
```suggestion
		return nil, fmt.Errorf("repo is nil but needed to get task labels")
```

server/grpc/rpc.go
```diff
@@ -429,6 +488,43 @@ func (s *RPC) ReportHealth(ctx context.Context, status string) error {
...
+			return err
+		}
+	}
+
+	if !agent.CanAccessRepo(repo) {
+		msg := fmt.Sprintf("agent '%d' is not allowed to interact with repo[%d] '%s'", agent.ID, repo.ID, repo.FullName)
+		log.Error().Int64("repoId", repo.ID).Msg(msg)
+		return errors.New(msg)
+	}
+	return nil
```
anbraten, line 525: For a security check it would be better to fail by default and `return nil` if everything is fine.

server/api/agent.go
6543: yes did not merged latest into this ... jet

server/grpc/rpc.go
6543: fd6a92e07c577ee084d2fb4d8391e426b2444bab

server/router/api.go
6543: still even in terms of UX it's better to have the option to let users just attach agents to repos.

And one of the next things I'll imediatly work on in terms of features (before the the compile step) will be to have a priorized queue, so that if an repo agent exist, it will always be prefered over org or instance wide agents. does this do it?

server/grpc/rpc.go
6543: moved to https://github.com/woodpecker-ci/woodpecker/pull/4131 to not hold this pull back

6543:
> Do we really need repo agents? As you can apply filters to org / (user org) agents for specific repos I don't think we need them TBH.

>> it does not add more complexity and allow a even more granular management ...
... one of the things i want to do next is to have a prioritized queue and with such it should priorize the lowest hirarchy always if it can ...
>>beside that, you could be in an org and be an repo admin but not an org admin ... this would cover the case

>>> it does not add more complexity and allow a even more granular management ...

>>>> It does add more code and thereby complexity ... 😉 As the benefit isn't granted / rather unclear atm we should not add them for now IMO

_Originally posted by @anbraten in https://github.com/woodpecker-ci/woodpecker/pull/3539#discussion_r1765367322_
            

well lets say you have an organization that contains ~100 repos with pipelines.
and you want one repo to have a dedicated agent:
## with repo agents:
you simply add one agent to that repo (and if it should be exclusive add an unique filter to both agent and pipeline lables)
## without:
you have to add lable filters to all other 99 pipelines(multiplied by the workflow number...) and to the agents you wana still use there. so that you could exclude execution.  (And then still, if it should be exclusive add an unique filter to both agent and pipeline lables)

server/router/api.go
6543: move conversation to main chat: https://github.com/woodpecker-ci/woodpecker/pull/3539#issuecomment-2366869300

6543:
@anbraten added user agents tab and deduped the code

anbraten:
> without:

Shouldn't make a difference if server enforces a specific `repo-id: 123` label or you set it manually. In general how often are those specific agents needed. As said before we can always add such things later on. Removing such features is always a pain therefore I still prefer to start with org agents for.

anbraten:
@6543 Could you extract repo agents for now, so we can test org agents for the beginning.

6543:
I'm realy against it but i want to get this feature done :(

6543:
@anbraten now the repo stuff is ripped out

anbraten:
@6543 One last question. I had a quick look in the code, couldn't we use `OrgID = -1` (or `OrgID = -1 && RepoID = -1`) as check for system agents and drop `OwnerID`.

We only check for ownerid / system agents here (which could be done by using `OrgID = -1` as well):
https://github.com/6543-forks/woodpecker/blob/dc67296cdee9b63f69442775e2cbe03c1badacf3/server/grpc/rpc.go#L472

6543:
no owner represent the user who created the agent ... you want to know that!

6543:
@anbraten my first patch did used this but I removed it because it was just a cause of lot of bugs poping up

anbraten (approved):

6543:
as mentioned the priority agent selection by the queue:  #4201

6543:
if admins want to disable it for hardening porpuses:  #4206

PR #3542, opened 2024-03-24
Remove `esbuild-loader`

qwerty287:
It seems that `esbuild-loader` is not necessary - everything's working fine. Or is there something I missed there?
However, it breaks the updates in #3541 and #3485.


6543 (approved):

anbraten:
I've added it as it drastically improve the build time.

PR #3546, opened 2024-03-24
Reenable `esbuild-loader`

qwerty287:
This reverts commit dfa7c36a64d7dd70bf2d39e36493f85a05d2afe2.

And update to latest version, I found the reason why it fails.


xoxys:
Can you explain why this change is required to get it working, or what was causing the issue with the latest esbuild-loader version? Or is there an upstream issue with context?

xoxys (approved):

PR #3568, opened 2024-03-29
Allow to store logs in files

qwerty287:
https://github.com/woodpecker-ci/woodpecker/discussions/2278

### DIFF
--- a/server/services/log/file/file.go
+++ b/server/services/log/file/file.go
@@ -21,7 +21,7 @@ func NewLogStore(base string) (log.Service, error) {
 		return nil, fmt.Errorf("file storage base path is required")
 	}
 	if _, err := os.Stat(base); err != nil && os.IsNotExist(err) {
-		err = os.MkdirAll(base, 0600)
+		err = os.MkdirAll(base, 0o600)
 		if err != nil {
 			return nil, err
 		}
@@ -62,7 +62,7 @@ func (l logStore) LogFind(step *model.Step) ([]*model.LogEntry, error) {
 }
 
 func (l logStore) LogAppend(logEntry *model.LogEntry) error {
-	file, err := os.OpenFile(l.filePath(logEntry.StepID), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
+	file, err := os.OpenFile(l.filePath(logEntry.StepID), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o600)
 	if err != nil {
 		return err
 	}


zc-devs:
> Do we need native support for something like S3 or can we just ask users to mount it directly into the FS?

> in object storage

Native S3.

1. I believe, environments exist, where you cannot mount arbitrary FUSE (AWS EKS, Windows, I guess).
2. I'm not sure if FUSE S3 can patch a file. If can't, then this is not well from performance and maybe costs perspective.

---

Sadly, there is no description in the issue, but I'm guessing it is related to [growing Database](https://github.com/woodpecker-ci/woodpecker/pull/3506#issuecomment-2022159070). If so, [tiering across storage mediums](https://min.io/product/automated-data-tiering-lifecycle-management) could be implemented.
1. Some bunch of data stores in DB.
2. After some retention condition old data archives to S3 (#1068).

anbraten:
Could we try to merge #3722 first or do you think they wouldn't conflict that much?

server/services/log/file/file.go
```diff
@@ -0,0 +1,79 @@
...
+		err = os.MkdirAll(base, 0600)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return logStore{base: base}, nil
+}
+
+func (l logStore) filePath(id int64) string {
+	return filepath.Join(l.base, strconv.Itoa(int(id))+".json")
```
anbraten, line 33:
```suggestion
	return filepath.Join(l.base, fmt.Sprintf("%d.json", id))
```

anbraten, line 33: Would it make sense to include the repo id in the path to improve grouping in the folder?

server/services/log/file/file.go
```diff
@@ -0,0 +1,79 @@
...
+	file, err := os.ReadFile(l.filePath(step.ID))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	var entries []*model.LogEntry
+	for _, j := range bytes.Split(file, []byte("\n")) {
```
anbraten, line 46: Would be better to use a scanner instead I think.

cmd/server/server.go
```diff
@@ -276,6 +276,11 @@ func setupEvilGlobals(c *cli.Context, s store.Store) error {
 	}
 	server.Config.Services.Manager = serviceManager
 
+	server.Config.Services.LogStore, err = setupLogStore(c, s)
+	if err != nil {
+		return fmt.Errorf("could not setup log store manager: %w", err)
```
anbraten, line 281:
```suggestion
		return fmt.Errorf("could not setup log store: %w", err)
```

cmd/server/setup.go
```diff
@@ -154,3 +156,12 @@ func setupMetrics(g *errgroup.Group, _store store.Store) {
 		}
 	})
 }
+
+func setupLogStore(c *cli.Context, s store.Store) (logService.Service, error) {
+	switch c.String("log-store") {
+	case "file":
+		return file.NewLogStore(c.String("log-store-file-base"))
```
anbraten, line 163:
```suggestion
		return file.NewLogStore(c.String("log-store-file-path"))
```

docs/docs/30-administration/10-server-config.md
```diff
@@ -543,6 +543,18 @@ Enable the Swagger UI for API documentation.
...
+
+> Default: `database`
+
+Where to store logs. Possible values: `database` or `file`.
+
+### `WOODPECKER_LOG_STORE_FILE_PATH`
+
+> Default empty
+
+Directory to store logs if [`WOODPECKER_LOG_STORE`](#woodpecker_log_store) is `file`.
```
anbraten, line 556:
```suggestion
Directory to store logs in if [`WOODPECKER_LOG_STORE`](#woodpecker_log_store) is `file`.
```

anbraten (approved):

PR #3584, opened 2024-04-02
Remove datastore testfiles

xoxys:
Ref: https://github.com/woodpecker-ci/woodpecker/pull/3227/files


xoxys:
@qwerty287 If this file is required for testing and was added intentionally, we can close this PR. To me, that looks like an oversight.

PR #3588, opened 2024-04-03
Temp exclude huh/spinner from renovate

xoxys:
Ref: https://github.com/woodpecker-ci/woodpecker/pull/3585

There seems to be an issue in renovate when updating modules from a subdirectory of another module. I'll create an upstream issue/discussion later.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,6 +3,10 @@
   "extends": ["github>woodpecker-ci/renovate-config"],
   "packageRules": [
     {
+      "matchPackageNames": ["github.com/charmbracelet/huh/spinner"],
+      "enabled": false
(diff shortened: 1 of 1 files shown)


xoxys:
> Actually that's kind of a renovate bug

There is a long discussion about it at https://github.com/renovatebot/renovate/discussions/21656

> Can you add github.com/distribution/reference too?

If valid updates are excluded from the renovation, there is a possibility that they will be completely forgotten. What about adding [a config](https://docs.renovatebot.com/configuration-options/#dependencydashboardapproval) for this package to enforce an approval? This way the package is still tracked by renovate, it's listed in the dashboard but will not be included in automatic deps updates.

PR #3609, opened 2024-04-12
Ability to set pod annotations and labels from step

### LINKED ISSUE #3188: Add the ability to set arbitrary annotations on Kubernetes pods

Right now, I don't see a way to set Pod level annotations for pipeline build steps in Kubernetes. I need this to allow Pods in my pipeline to gain access to `/dev/fuse`. CRI-O includes a handy feature to enable this via annotations described in [detail here](https://www.opensourcerers.org/2023/10/16/eclipse-che-openshift-dev-spaces-podman-with-fuse-overlay/).

The upshot is that after you have configured CRI-O you just need to add annotations like the following to your Pod and the /dev/fuse device from the host will be passed through:

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    my.workload.activation: ""
    io.kubernetes.cri-o.Devices: /dev/fuse
```

### Suggested solution

The ideal solution for me would be to allow arbitrary Pod annotations to be configured in the pipeline like so:

```yaml
steps:
  test:
    image: alpine
    commands:
      - echo Hello world
    backend_options:
      kubernetes:
        annotations:
          io.kubernetes.cri-o.Devices: /dev/fuse
          another.annotation: ""
```

### Alternative

For this particular issue, I can use another means of passing `/dev/fuse` into the container. For instance, there is a generic method that uses `resources.limits` which are [already supported](https://woodpecker-ci.org/docs/administration/backends/kubernetes#resources). That said, that options requires running a DaemonSet on the cluster referencing a 3rd party image which is not ideal.

It would be a lot cleaner/safer to use the runtime's built in features to do this. Plus, allowing generic annotations will likely enable other features/integrations down the line.

### Additional context

This is part of my attempt to get buildah to run on Woodpecker. This _should_ be the last piece of the puzzle now that `securityContext` and AppArmor and sorted. But, I've said that before.

### Validations

zc-devs:
Closes #3188

### DIFF
--- a/pipeline/backend/kubernetes/flags.go
+++ b/pipeline/backend/kubernetes/flags.go
@@ -46,14 +46,26 @@ var Flags = []cli.Flag{
 	&cli.StringFlag{
 		EnvVars: []string{"WOODPECKER_BACKEND_K8S_POD_LABELS"},
 		Name:    "backend-k8s-pod-labels",
-		Usage:   "backend k8s additional worker pod labels",
+		Usage:   "backend k8s additional Agent-wide worker pod labels",
 		Value:   "",
+	},
+	&cli.BoolFlag{
+		EnvVars: []string{"WOODPECKER_BACKEND_K8S_POD_LABELS_ALLOW_FROM_STEP"},
+		Name:    "backend-k8s-pod-labels-allow-from-step",
+		Usage:   "whether to allow using labels from step's backend options",
+		Value:   false,
 	},
 	&cli.StringFlag{
 		EnvVars: []string{"WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS"},
 		Name:    "backend-k8s-pod-annotations",
-		Usage:   "backend k8s additional worker pod annotations",
+		Usage:   "backend k8s additional Agent-wide worker pod annotations",
 		Value:   "",
+	},
+	&cli.BoolFlag{
+		EnvVars: []string{"WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS_ALLOW_FROM_STEP"},
+		Name:    "backend-k8s-pod-annotations-allow-from-step",
+		Usage:   "whether to allow using annotations from step's backend options",
+		Value:   false,
 	},
 	&cli.BoolFlag{
 		EnvVars: []string{"WOODPECKER_BACKEND_K8S_SECCTX_NONROOT"},
--- a/pipeline/backend/kubernetes/backend_options.go
+++ b/pipeline/backend/kubernetes/backend_options.go
@@ -11,6 +11,8 @@ type BackendOptions struct {
 	Resources          Resources         `mapstructure:"resources"`
 	RuntimeClassName   *string           `mapstructure:"runtimeClassName"`
 	ServiceAccountName string            `mapstructure:"serviceAccountName"`
+	Labels             map[string]string `mapstructure:"labels"`
+	Annotations        map[string]string `mapstructure:"annotations"`
 	NodeSelector       map[string]string `mapstructure:"nodeSelector"`
 	Tolerations        []Toleration      `mapstructure:"tolerations"`
 	SecurityContext    *SecurityContext  `mapstructure:"securityContext"`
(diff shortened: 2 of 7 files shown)


pipeline/backend/kubernetes/flags.go
```diff
@@ -46,15 +46,27 @@ var Flags = []cli.Flag{
...
 		Name:    "backend-k8s-pod-annotations",
-		Usage:   "backend k8s additional worker pod annotations",
+		Usage:   "backend k8s additional Agent-wide worker pod annotations",
 		Value:   "",
 	},
+	&cli.BoolFlag{
+		EnvVars: []string{"WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS_ALLOW_FROM_STEP"},
+		Name:    "backend-k8s-pod-annotations-allow-from-step",
+		Usage:   "whether to allow using annotations from step's backend options",
+		Value:   false,
```
pat-s, line 68: What are the possible security risks when allowing this by default?

zc-devs: > gain access to /dev/fuse

Don't know, what is this device about, but also don't want to check it out [again](https://github.com/woodpecker-ci/woodpecker/issues/3537#issuecomment-2016464281).

zc-devs:
@qwerty287, maybe build images here too and @dominic-p could test?

dominic-p:
Thanks for the work on this @zc-devs! I'm ready to test as soon as the pipelines run. It looks like they are pending approval at the moment.

zc-devs:
This one is [ready](https://hub.docker.com/r/woodpeckerci/woodpecker-server/tags?page=&page_size=&ordering=&name=pull_3609)

dominic-p:
Ok, I was able to test this tonight, and it's working as expected for me! If you would like, I can take a stab at a PR for the docs. Just let me know.

zc-devs:
> I can take a stab at a PR for the docs

It would be great! Thanks in advance.

pipeline/backend/kubernetes/flags.go
```diff
@@ -46,15 +46,27 @@ var Flags = []cli.Flag{
...
+	},
 	&cli.StringFlag{
 		EnvVars: []string{"WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS"},
 		Name:    "backend-k8s-pod-annotations",
-		Usage:   "backend k8s additional worker pod annotations",
+		Usage:   "backend k8s additional Agent-wide worker pod annotations",
 		Value:   "",
 	},
+	&cli.BoolFlag{
+		EnvVars: []string{"WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS_ALLOW_FROM_STEP"},
```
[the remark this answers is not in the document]
zc-devs: First off, I don't care too much, can name it whatever you (someone) want. I'm just curious, why?

1. In regards #3655
The idea was the same: allowing to use secrets from steps. I left the room for someone would want to set Agent-wide secrets.
There wasn't Agent-wide functionality and was renamed to `WOODPECKER_BACKEND_K8S_ALLOW_NATIVE_SECRETS`. I can read it as "woodpecker k8s backend feature gate to allow native secrets".
There is a word "allow" and some meaning: Do we allow native secrets? Yes/No.
2. `WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS_STEP`
There is no "allow". So, what is the _step of pod annotations_ we are talking about?
`WOODPECKER_BACKEND` is 41% of length. Maybe we should work on _it_? Keycloak project [uses short prefix](https://www.keycloak.org/server/configuration), for example. Seems, they don't care about length also ([KC_CACHE_EMBEDDED_MTLS_KEY_STORE_PASSWORD](https://www.keycloak.org/server/all-config)).
3. Is there some restrictions to length of env var on some platform/whatever? Maybe some guidelines? How much is "too long"?
4. Configuration process
```yaml
kind: ConfigMap
metadata:
  name: agent
data:
  WOODPECKER_SERVER: woodpecker-server:9000
  WOODPECKER_BACKEND_K8S_NAMESPACE: woodpecker
  WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS_ALLOW_FROM_STEP: true
```
How did I craft this? Opened documentation, for each var did: {double-click -> Ctrl+C -> Ctrl+V -> set the value} -> done. There is no difference in var length in that process.

---

I would not rename it, but as I've said, I don't care too much. Let me know, if I should, I'll do it.

PR #3624, opened 2024-04-18
Use our github-release plugin

qwerty287:
instead of the drone one


.woodpecker/binaries.yaml
```diff
@@ -90,11 +90,10 @@ steps:
   release:
     depends_on:
       - checksums
-    image: docker.io/plugins/github-release
-    secrets:
-      - source: github_token
-        target: github_release_api_key
+    image: woodpeckerci/plugin-github-release:1.1.2
```
xoxys, line 93: Maybe adding the image to `variables:` as all other images might make sense.

xoxys (approved):

Issue #3627, opened 2024-04-18
Dashboard links to wrong section for deprecated secret alternative names


Andre601:
### Component

web-ui

### Describe the bug

When using the deprecated secret alternative names will the Dashboard mention the deprecation of this option, but link to a completely unrelated section.

It currently links to the [`event` section](https://woodpecker-ci.org/docs/usage/workflow-syntax#event) within the Workflow Syntax page, when it shoudl actually link to the [`Use secrets in settings and environemt` section](https://woodpecker-ci.org/docs/usage/secrets#use-secrets-in-settings-and-environment) of the Secrets page.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.4.1"}
```

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

Issue #3607, opened 2024-04-11
Confusing [bad_habit] Please set an event filter on all when branches message


oliverpool:
### Component

web-ui

### Describe the bug

Initially reported on https://codeberg.org/Codeberg-CI/feedback/issues/170

When a `when` block has no `event` field, a `[bad_habit] Please set an event filter on all when branches message` warning is shown, introduced by #3222.

![image](https://github.com/woodpecker-ci/woodpecker/assets/3864879/a4d94a63-7caf-4c51-b8cb-ad8fbbabc5ab)

This is not really actionable for the user:
- no idea which part of the pipeline is faulty
- no much idea how to solve it.

### System Info

```shell
source	"https://github.com/woodpecker-ci/woodpecker"
version	"2.4.1"
```

### Additional context

Having the step involved would be nice (it is recorded, but not displayed).

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

zc-devs:
> warning is shown

even though [there is no "when" directive at all](https://github.com/woodpecker-ci/woodpecker/pull/3269#issuecomment-2016852418).

anbraten:
If I am correct the idea was basically to encourage the user to add at least some kind of when filters as otherwise in the case we add more functionality like additional events the existing pipelines would also be triggered by those new events when updating.

I guess we can simply improve the message a bit and additionally show the workflow file it was detected in.

Issue #3630, opened 2024-04-19
unclear workflow lint


wez:
### Component

web-ui

### Describe the bug

This set of lints provides no actionable context to understand the issue or how to resolve it; it just says that there is a bad habit.
In addition, because the English language is what it is, the phrasing is extra difficult to parse and understand:

`Please set an event filter on all when branches`

I'd recommend quoting the `"when"` so that the grammar is easy to understand.  Should `"event"` also be quoted? I'm not sure what the linter is looking for.

![image](https://github.com/woodpecker-ci/woodpecker/assets/117777/5d765293-d170-4c2b-8b0d-eb1c82b0ec02)

It's also unclear to me why 6 warnings are generated from the input (see below).

What I think would make things easier for users is:

* It should at least reference the file name and line number like some of the other lints
* Have the lint message suggest what should be added to resolve it. Should there be an `event:` field added? If so, can we make the lint message say that plainly?
* The documentation should be clearer about the recommended syntax and schema for workflow level conditions

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.4.1"}
```

### Additional context

Here's the `when` portion of my workflow:

```yaml
when:
  - event: [push, pull_request, manual]
  - path:
      include:
        - .woodpecker/docs.yml
        - docs/**
        - mkdocs-base.yml
        - stylua.toml
```

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

anbraten:
duplicate of #3630

PR #3634, opened 2024-04-21
Deprecate environment filter and improve errors

qwerty287:
As of https://github.com/woodpecker-ci/woodpecker/pull/3512/files#r1566753245

Closes https://github.com/woodpecker-ci/woodpecker/issues/3607
Closes https://github.com/woodpecker-ci/woodpecker/issues/3627

### DIFF
--- a/docs/docs/20-usage/20-workflow-syntax.md
+++ b/docs/docs/20-usage/20-workflow-syntax.md
@@ -744,7 +744,7 @@ Workflows that should run even on failure should set the `runs_on` tag. See [her
 Woodpecker gives the ability to configure privileged mode in the YAML. You can use this parameter to launch containers with escalated capabilities.
 
 :::info
-Privileged mode is only available to trusted repositories and for security reasons should only be used in private environments. See [project settings](./71-project-settings.md#trusted) to enable trusted mode.
+Privileged mode is only available to trusted repositories and for security reasons should only be used in private environments. See [project settings](./75-project-settings.md#trusted) to enable trusted mode.
 :::
 
 ```diff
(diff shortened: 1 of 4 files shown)


pipeline/frontend/yaml/linter/linter.go
```diff
@@ -351,7 +383,7 @@ func (l *Linter) lintBadHabits(config *WorkflowConfig) (err error) {
 			if field != "" {
 				err = multierr.Append(err, &errorTypes.PipelineError{
 					Type:    errorTypes.PipelineErrorTypeBadHabit,
-					Message: "Please set an event filter on all when branches",
+					Message: "Please set an event filter for all steps/whole workflow on all when branches",
```
zc-devs, line 386: #3630

1. What is the `when branches`? Is it ['when' block with 'branch' filter](https://woodpecker-ci.org/docs/2.3/usage/workflow-syntax#branch)?
2. If I don't have 'event' filter, but 50 steps, will I have 50 `bad_habit` warnings?
3. Should warning contain a link to the docs?
```
Please set an event filter for all [steps](https://woodpecker-ci.org/docs/2.3/usage/workflow-syntax#event) / whole [workflow](https://woodpecker-ci.org/docs/2.3/usage/workflow-syntax#event-1)
```

anbraten: > has two "branches": one for event = push and branch = main and one for event = tag.

I think the word `branches` for this is confusing as its mixed up with a git branch. 

So something as suggested by @zc-devs like `Please set a some filter for the [workflow](https://woodpecker-ci.org/docs/2.3/usage/workflow-syntax#event-1) or all [steps](https://woodpecker-ci.org/docs/2.3/usage/workflow-syntax#event).` might be better.

zc-devs: > No - it's every "subfilter"

then maybe

> ... on all items in 'when' block.

if you want to have this end of sentence.

But short version with links is enough, I think.

anbraten: We could also start a new page in the docs similar to how eslint is explaining rules that describes the warning in more detail and gives positive and negative examples.

PR #3648, opened 2024-04-25
Add make target for spellcheck

xoxys:


xoxys:
I tried it, but no make.

PR #3654, opened 2024-04-27
Compare to pipeline created timestamp while using before/after filter

xoxys:
`pipeline_started` and `pipeline_finished` can be `0` for various reasons and therefor is unreliable for timestamp filters. Using `pipeline_created` might be the better option anyway.


xoxys:
Thanks. As I try improve my own label handling, when do we choose bug vs regression?

Issue #3582, opened 2024-04-01
Reference secret values from Kubernetes secrets


dominic-p:
### Clear and concise description of the problem

It would be nice to be able to leverage [Kubernetes native Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) inside of Woodpecker pipelines.

### Suggested solution

I'm imagining a workflow definition that looks like this:

```yaml
steps:
  - name: docker
    image: docker
    commands:
      - echo $docker_username
      - echo $DOCKER_PASSWORD
    secrets:
    - docker_username
    - name: DOCKER_PASSWORD
      kubernetesRef:
        name: my-docker-secret
        key: password
```

So, the `docker_username` secret would be referenced as it normally is today. The value for the `DOCKER_PASSWORD` secret would come from a Kubernetes Secret called `my-docker-secret` that is defined in the same namespace that pipeline runs. That secret is expected to have a key named `password` that contains the value that we want to use for the secret.

This is loosely based on [the way Drone does it](https://blog.drone.io/drone-kubernetes-secrets/) and the way that Kubernetes allows you to reference secrets for env variables.

The exact syntax is, of course, flexible, but the idea is to allow for other external secret providers as well (see #929).

### Alternative

Right now, it seems like the only way to [define secrets](https://woodpecker-ci.org/docs/usage/secrets#adding-secrets) is via the UI or the CLI. This works, but this isn't ideal if you already have the secretes defined in your cluster.

Another alternative would be to store the information in [global env variables](https://woodpecker-ci.org/docs/usage/environment#global-environment-variables), but that doesn't seem ideal from a security standpoint.

Another option would be to expand the Kubernetes backend [volume mounting logic](https://woodpecker-ci.org/docs/administration/backends/kubernetes#volumes) to allow mounting Secrets in addition to PVs.

### Additional context

This is part of my ongoing effort to get a container build system (specifically buildah) to run on Woodpecker on a CRI-O based Kubernetes cluster. I currently have it up and working, but I need a way to pass registry credentials to it. Otherwise, I can build images, but I can't push them to my repo when I'm done.

There may be a simpler way to do this that I'm just missing at the moment.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
[...]

PR #3349, opened 2024-02-07
Allow to configure a config extension per repo

anbraten:
closes #783

Allow users to set / use service extensions (for this PR just config, secrets / registries could be added later on) on a per repo basis. Those extensions are little external webservice which can implement a set of endpoints to replace woodpeckers internal functionality like:
- pipeline config processing

# TODO
- [x] add config & secrets extension service urls to repo settings  
- [x] allow to change extensions urls for repo
- [x] add docs
- [x] restrict http call locations (prevent local ips)

image

PR #3655, opened 2024-04-27
K8s secrets reference from step

zc-devs:
Part of #3582

Pipeline:
```yaml
skip_clone: true
steps:
  secret-test:
    image: alpine
    commands:
      - echo $$AWS_ACCESS_KEY_ID
      - echo $$AWS_SECRET_ACCESS_KEY
    backend_options:
      kubernetes:
        secretNames:
          - test-secret
```

Test secret:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-secret
  namespace: test-woodpecker-runtime
data:
  AWS_ACCESS_KEY_ID: N0lrZHNzb0xleGFtcGxlNkpJVnZDRXE=
  AWS_SECRET_ACCESS_KEY: TXk0WE5BYXNleGFtcGxldXRKRHNUWHc=
type: Opaque
```

1. Default Agent config.

In log I see warning:
```
{"level":"debug","time":"2024-04-27T18:17:49Z","caller":"/src/pipeline/backend/kubernetes/pod.go:166","message":"Secret names were defined in backend options, but its using disallowed by instance configuration "}
```

WP output:
```
+ echo $AWS_ACCESS_KEY_ID

+ echo $AWS_SECRET_ACCESS_KEY

```

2. Allowing native secrets.

Agent config:
```
data:
  WOODPECKER_BACKEND_K8S_NATIVE_SECRETS_ALLOW_FROM_STEP: 'true'
  ...
```

WP output:
```
+ echo $AWS_ACCESS_KEY_ID
7IkdssoLexample6JIVvCEq
+ echo $AWS_SECRET_ACCESS_KEY
My4XNAasexampleutJDsTXw
```


zc-devs:
> if you use internal secrets and print them out to the logs, they're replaced with ********

I know, i know... Also thought about it and decided to warn in the docs too.
Now we don't have secrets to find them in the logs. However, when [second part](https://github.com/woodpecker-ci/woodpecker/issues/3582#issuecomment-2033256484) is implemented, this is achievable.

> WOODPECKER_BACKEND_K8S_ALLOW_NATIVE_SECRETS

Will take it.

> http service extension doing this?

Extension or external service?
I don't like extensions :) but like idea of external (micro)service. It might be useful in terms of duplication of masking secrets logic, at least.
Is there some example / skeleton of service? Seems, I cannot manage it alone.

zc-devs:
> But probably this feature will take some more time to complete

Let's give this PR a try then. Could you add `build_image` label, please?

@dominic-p, could you test then?

dominic-p:
Thanks for working on this! I'm ready to test this PR as soon as we get an image published.

zc-devs:
And this [too](https://hub.docker.com/r/woodpeckerci/woodpecker-server/tags?page=&page_size=&ordering=&name=pull_3655)

pipeline/backend/kubernetes/pod.go
```diff
@@ -236,6 +244,25 @@ func containerPort(port types.Port) v1.ContainerPort {
 	}
 }
 
+func containerSecrets(secretNames []string) []v1.EnvFromSource {
+	if secretNames == nil || len(secretNames) == 0 {
```
zc-devs, line 248:
```suggestion
	if len(secretNames) == 0 {
```

pipeline/backend/kubernetes/pod.go
```diff
@@ -159,6 +159,14 @@ func podContainer(step *types.Step, podName, goos string, options BackendOptions
 
 	container.Env = mapToEnvVars(step.Environment)
 
+	if len(options.SecretNames) > 0 {
+		if config.NativeSecretsAllowFromStep {
+			container.EnvFrom = containerSecrets(options.SecretNames)
+		} else {
+			log.Debug().Msg("Secret names were defined in backend options, but its using disallowed by instance configuration ")
```
dominic-p, line 166: Minor nit, this doesn't read quite right to me. I would go with:

> Secret names were defined in backend options, but secret access is disallowed by instance configuration.

dominic-p:
Ok, I was able to test this tonight, and it seems to be working as expected! I was able to successfully build and push my first image to my local repo! 🎉🎉🎉

I did run into one snag. As I mentioned, my main use case for this feature is to provide container image registry credentials to my build script. These credentials are currently stored in a Kubernetes [registry cred secret](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials). This kind of secret uses the hardcoded key `.dockerconfigjson` to reference its data. The issue is that with the current implementation we wind up with an environment variable with a dot in its name. I had a hard time using it since env variable names really aren't supposed to have dots in them.

I was able to work around the issue with `printenv`, but it's kind of awkward. I'm not sure if this problem is worth fixing in the current iteration of this feature or not. Maybe we can just add a short blurb to the documentation about it? Or, a simple solution might be to replace any special characters in the secret key names with underscores and then document that convention?

zc-devs:
Could you provide an example?
1. Secret.
2. Build tool (kaniko, buildx, etc.) and its configuration.

I use kaniko debug image and plain secret: 
```
    secrets:
      - docker_usr
      - docker_pwd
```

kaniko [requires Docker auth file](https://github.com/GoogleContainerTools/kaniko?tab=readme-ov-file#pushing-to-docker-hub)  in `/kaniko/.docker/config.json`, so I [create this file](https://gitea.smthd.com/mirko/woodpecker-agents-sweeper/src/branch/main/.cicd/set-docker-auth.sh) manually:
```
DOCKER_AUTHS=$(printf '{"auths":{"%s":{"auth":"%s"}}}' "$REGISTRY" "$AUTH")
printf '%s' "$DOCKER_AUTHS" > "/kaniko/.docker/config.json"
```

In my use-case it could be requirements:
1. Existing `dockerconfigjson` secret
2. Mount as file to `/kaniko/.docker/config.json`

Current implementation treat `.dockerconfigjson` as env var. Suppose it is `DOCKER_AUTHS` instead. What do you do next? Do you write it in a file and point your build tool to it?

dominic-p:
Thanks for the detailed feedback.

Here's an example secret from my cluster:

```yaml
apiVersion: v1
type: kubernetes.io/dockerconfigjson
kind: Secret
metadata:
  name: reg-cred
  namespace: woodpcker
data:
  .dockerconfigjson: <base64 encoded JSON auth file here>
```

And, here's a portion of the build script that would use it. I'm using [buildah](https://buildah.io/) to build my container images via a standard shell script.

```sh
#!/bin/sh

# Create a working container
dev=$(buildah from docker.io/alpine)

# ... build the image

# Save credentials in temporary file

# You would like to do something like this, but, of course, it won't work
# echo "$.dockerconfigjson" > /tmp/.dockerconfigjson

printenv ".dockerconfigjson" > /tmp/.dockerconfigjson

# Save and push the image to our local registry
buildah commit $dev "$name"

buildah push --authfile /tmp/.dockerconfigjson "$name" "docker://$registry_url/$name"
```

So, the simplest workflow for me is to simply dump the reg-cred secret data into a temporary file and then reference it as shown above.

Of course, I could use a a regular Opaque secret and build the docker config JSON as you showed in your example (or reference the username and password with `buildah push --creds`), but since I already have the JSON in an existing secret it would be nice to leverage it. One benefit of doing it this way is that Kubernetes requires you to use this kind of secret in [`spec.imagePullSecrets`](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/). So, I can use the same secret to push to the local registry and pull from it.

At the end of the day, this is not a big deal. There are many usable workarounds. I just wanted to point out the awkwardness that you get when you wind up with an env variable that has a dot in its name.

zc-devs:
Thanks for explanation. This PR have to be reworked then. Meantime I move it to a draft.

dominic-p:
Sure thing. My personal preference would be to release this as-is and then iterate from there. Even in its current form it's enough for my use case. But, I can understand if you want to wait until it's a bit more complete.

zc-devs:
> iterate from there

Sure. But we have to decide on [syntax](https://github.com/woodpecker-ci/woodpecker/issues/3582#issuecomment-2096782023).

dominic-p:
Just to share my perspective here, my use case is a very private CI environment. Adding manual config in the UI to enable these features would add a bit of friction for me.

Also, as a counterpoint, it feels like the env variables can allow for a more fine grained approach than a blanket trusted switch. For example, a repo may need to set security context but may not need secret access. In that case, we can have tighter controls using env vars.

All of that said, I totally understand that the security implications here need to be taken seriously, and making all the backends work in a similar fashion definitely seems like a good idea.

dominic-p:
Thanks @qwerty287 I'll chime in there. :)

zc-devs:
BTW, [new image is ready](https://hub.docker.com/layers/woodpeckerci/woodpecker-agent/pull_3655/images/sha256-a7b47e8b023738c8c2308b0cc9218f9cc5f44b3682ea4acabb76df67469a2827?context=explore). @dominic-p, could you test it?

dominic-p:
Thanks for the heads up @zc-devs. Can you give me an idea of what I'm testing in this new image?

Looking at the most recent comment vs the original, it looks like the syntax changed from: `kubernetes.secretNames[]` to `kubernetes.secrets[]`. Is that the only change? I'm guessing you haven't implemented all of the [variations here](https://github.com/woodpecker-ci/woodpecker/issues/3582#issuecomment-2096782023) yet.

zc-devs:
> syntax changed

Yes, that is all. I think it would be great, if someone independent test it before merge.

However, I doubt a little now... Is there more expected from me in this PR or we can merge it and do other tasks in separate PRs?

dominic-p:
Ok, I was able to test the new version tonight, and it still seems to be working as expected. Thanks!

zc-devs:
If there are no comments/notes, we are ready to merge.

dominic-p:
I can take a stab at some documentation, but I thought the plan was to wait on the docs until some of the follow up features landed.

zc-devs:
Yes, I would postpone writing docs.
I've updated [TODO list](https://github.com/woodpecker-ci/woodpecker/issues/3582#issuecomment-2096736572).

zc-devs:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-secret
data:
  AWS_ACCESS_KEY_ID: N0lrZHNzb0xleGFtcGxlNkpJVnZDRXE=
  AWS_SECRET_ACCESS_KEY: TXk0WE5BYXNleGFtcGxldXRKRHNUWHc=
type: Opaque
```
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-file-secret
data:
  tls.crt: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K
  tls.key: >-
    LS0tLS1CRUdJTiBSU0EgUFJJVkFURSBLRVktLS0tLQotLS0tLUVORCBSU0EgUFJJVkFURSBLRVktLS0tLQo=
type: Opaque
```

1.
```yaml
  simple-secret:
    image: alpine
    commands:
      - echo $$AWS_ACCESS_KEY_ID
      - echo $$AWS_SECRET_ACCESS_KEY
    backend_options:
      kubernetes:
        secrets:
          - name: test-secret
```
```
+ echo $AWS_ACCESS_KEY_ID
7IkdssoLexample6JIVvCEq
+ echo $AWS_SECRET_ACCESS_KEY
My4XNAasexampleutJDsTXw
```
1.
```yaml
  advanced-secret-key:
    image: alpine
    commands:
      - echo $$AWS_ACCESS_KEY_ID
      - echo $$AWS_SECRET_ACCESS_KEY
    backend_options:
      kubernetes:
        secrets:
          - name: test-secret
            key: AWS_ACCESS_KEY_ID
```
```
+ echo $AWS_ACCESS_KEY_ID
7IkdssoLexample6JIVvCEq
+ echo $AWS_SECRET_ACCESS_KEY

```
1.
```yaml
  advanced-secret-key-mapping:
    image: alpine
    commands:
      - echo $$AWS_ACCESS_KEY_ID
      - echo $$AWS_SECRET_ACCESS_KEY
    backend_options:
      kubernetes:
        secrets:
          - name: test-secret
            key: AWS_ACCESS_KEY_ID
            target:
              env: AWS_SECRET_ACCESS_KEY
```
```
+ echo $AWS_ACCESS_KEY_ID

+ echo $AWS_SECRET_ACCESS_KEY
7IkdssoLexample6JIVvCEq
```
1.
```yaml
  file-secret:
    image: alpine
    commands:
      - cat /data/ca.crt
    backend_options:
      kubernetes:
        secrets:
          - name: test-file-secret
            key: tls.crt
            target:
              file: /data/ca.crt
```
```
+ cat /data/ca.crt
-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
```

pipeline/backend/kubernetes/secrets.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	for _, secret := range nsp.secrets {
+		if secret.isSimple() {
+			simpleSecret, err := secret.toEnvFromSource()
+			if err != nil {
+				return err
+			}
+			nsp.envFromSources = append(nsp.envFromSources, simpleSecret)
+
```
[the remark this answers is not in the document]
zc-devs: I would not remove this as it separates different branches.

dominic-p:
Sure thing. It looks like the pipelines are currently waiting for approval. Once we have a new image built with the latest version of the PR I can test it.

dominic-p:
I'm not sure why the docker publish failed.

> Canceled: grpc: the client connection is closing

Maybe a network error of some kind?

zc-devs:
> File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(go.woodpecker-ci.org/woodpecker) --custom-order (gci)

Which command should I run?

6543 (approved):

lafriks:
Might be a bit late to the party but just a question why we do not implement defining such secrets in woodpecker secret section where instead of value there would be option to provide where to get secret value from (be it k8s secret, hashicorp vault etc) and then in pipeline just define secret as we already do.
With adding more and more runner specific options to pipeline syntax it will become less portable and harder to use

6543:
:point_up: if there is an api / interface for it - that would be best

I remember to have created an issue: "add more secret provider" or so ...

6543:
anyway that's still a good enouth solution for now, and later we can move this into a more generic solution if it exists

anbraten:
>☝️ if there is an api / interface for it - that would be best

#3349

lafriks:
In this case this probably can't be received at server side, it would still be for agent to bind when starting step so process would still be the same, my main point is that adding such definitions to pipeline makes them hard to maintain if things changes in future, so my suggestion would be just moving definition to server side but resolving value in this case would be still be on specific agent runner (like docker would not still be able to resolve k8s secrets)

PR #3670, opened 2024-05-01
Cleanup server env settings

anbraten:
ref #3669 

Following the sequence from the image (a rather advanced setup, for simple setups most options could be avoided) all possibly needed config values are:

- red: `WOODPECKER_HOST` address of woodpecker UI / api (needs to be accessible by browser)
- blue: `WOODPECKER_OAUTH_HOST` address of the forge (needs to be accessible by browser)
- green: `WOODPECKER_FORGE_URL` address of the forge (needs to be accessible by woodpecker-server)
- yellow: `WOODPECKER_WEBHOOK_HOST` address of woodpecker UI / api (needs to be accessible by forge) 

Namings of the env variables are not fixed yet and suggestions if they should have a prefix like `DEV_` / `INTERNAL_` or not are very welcome (we just need to add old namings to not add breaking changes).

Untitled-2024-04-30-0827


server/forge/gitea/gitea.go
```diff
@@ -95,12 +88,16 @@ func (c *Gitea) URL() string {
 }
 
 func (c *Gitea) oauth2Config(ctx context.Context) (*oauth2.Config, context.Context) {
+	publicFacingOAuthURL := server.Config.Server.ForgeOAuthHost
+	if publicFacingOAuthURL == "" {
+		publicFacingOAuthURL = c.url
+	}
```
anbraten, line 94: Should we add this for all selfhostable forges like gitlab?

docs/docs/91-migrations.md
```diff
@@ -12,6 +12,15 @@ Some versions need some changes to the server configuration or the pipeline conf
...
 - Deprecated `environment` filter, use `when.evaluate`
+- Deprecated some forge specific configs in favor of `WOODPECKER_FORGE_...`:
+  - Use `WOODPECKER_FORGE_URL` instead of	`WOODPECKER_GITHUB_URL`, `WOODPECKER_GITLAB_URL`, `WOODPECKER_GITEA_URL`, `WOODPECKER_BITBUCKET_URL`
+  - Use `WOODPECKER_FORGE_CLIENT` instead of `WOODPECKER_GITHUB_CLIENT`, `WOODPECKER_GITLAB_CLIENT`, `WOODPECKER_GITEA_CLIENT`, `WOODPECKER_BITBUCKET_CLIENT`, `WOODPECKER_BITBUCKET_DC_CLIENT_ID`
+  - Use `WOODPECKER_FORGE_SECRET` instead of `WOODPECKER_GITHUB_SECRET`, `WOODPECKER_GITLAB_SECRET`, `WOODPECKER_GITEA_SECRET`, `WOODPECKER_BITBUCKET_SECRET`, `WOODPECKER_BITBUCKET_DC_CLIENT_SECRET`
+  - Use `WOODPECKER_FORGE_SKIP_VERIFY` instead of `WOODPECKER_GITHUB_SKIP_VERIFY`, `WOODPECKER_GITLAB_SKIP_VERIFY`, `WOODPECKER_GITEA_SKIP_VERIFY`, `WOODPECKER_BITBUCKET_SKIP_VERIFY`
+- Use `WOODPECKER_DEV_OAUTH_HOST` instead of `WOODPECKER_DEV_GITEA_OAUTH_URL`
+- Replace `WOODPECKER_DEV_OAUTH_HOST` with `WOODPECKER_HOST` (public facing address) and `WOODPECKER_WEBHOOK_HOST` (internal webhook address)
+- Deprecated `WOODPECKER_WEBHOOK_HOST` in favor of `WOODPECKER_DEV_WEBHOOK_HOST`
```
zc-devs, line 22: Why? I use it not in dev/test env.

docs/docs/91-migrations.md
```diff
@@ -12,6 +12,15 @@ Some versions need some changes to the server configuration or the pipeline conf
...
 - Deprecated slice definition for env vars
 - Deprecated `environment` filter, use `when.evaluate`
+- Deprecated some forge specific configs in favor of `WOODPECKER_FORGE_...`:
+  - Use `WOODPECKER_FORGE_URL` instead of	`WOODPECKER_GITHUB_URL`, `WOODPECKER_GITLAB_URL`, `WOODPECKER_GITEA_URL`, `WOODPECKER_BITBUCKET_URL`
+  - Use `WOODPECKER_FORGE_CLIENT` instead of `WOODPECKER_GITHUB_CLIENT`, `WOODPECKER_GITLAB_CLIENT`, `WOODPECKER_GITEA_CLIENT`, `WOODPECKER_BITBUCKET_CLIENT`, `WOODPECKER_BITBUCKET_DC_CLIENT_ID`
+  - Use `WOODPECKER_FORGE_SECRET` instead of `WOODPECKER_GITHUB_SECRET`, `WOODPECKER_GITLAB_SECRET`, `WOODPECKER_GITEA_SECRET`, `WOODPECKER_BITBUCKET_SECRET`, `WOODPECKER_BITBUCKET_DC_CLIENT_SECRET`
+  - Use `WOODPECKER_FORGE_SKIP_VERIFY` instead of `WOODPECKER_GITHUB_SKIP_VERIFY`, `WOODPECKER_GITLAB_SKIP_VERIFY`, `WOODPECKER_GITEA_SKIP_VERIFY`, `WOODPECKER_BITBUCKET_SKIP_VERIFY`
+- Use `WOODPECKER_DEV_OAUTH_HOST` instead of `WOODPECKER_DEV_GITEA_OAUTH_URL`
+- Replace `WOODPECKER_DEV_OAUTH_HOST` with `WOODPECKER_HOST` (public facing address) and `WOODPECKER_WEBHOOK_HOST` (internal webhook address)
```
zc-devs, line 21: I don't understand, how webhook host is related to oAuth and public hosts?

docs/docs/91-migrations.md
anbraten: It basically means to newbies of woodpecker don't touch when doing your first setup. It often seams that new users quickly get lost in a jungle of endless guessing when touching more advanced settings.

docs/docs/91-migrations.md
anbraten: Before you've added #1869 we always told people to change `WOODPECKER_HOST` to the internal url and `WOODPECKER_DEV_OAUTH_HOST` to the public address. This was for example also done in the gitpod setup. Now after #1869 there is no reason anymore to have `WOODPECKER_HOST` different from `WOODPECKER_DEV_OAUTH_HOST` if you can also change the webhooks with `WOODPECKER_WEBHOOK_HOST`

docs/docs/91-migrations.md
xoxys: Name them `WOODPECKER_INTERNAL_`? I agree in general to prefix them somehow.

xoxys:
I added a comment about the naming. Did you decided against it or missed it?

docs/docs/91-migrations.md
anbraten: Sth like `WOODPECKER_INTERNAL_` or `WOODPECKER_ADVANCED_` would also be fine for me as well. We should just decide on sth unique we can then use for all of those kind of configs.

6543:
we should update/ehnhance https://woodpecker-ci.org/docs/development/ui ?

server/forge/gitea/gitea.go
xoxys: Yes, I think that would make sense.

server/forge/gitea/gitea.go
```diff
@@ -95,12 +88,16 @@ func (c *Gitea) URL() string {
 }
 
 func (c *Gitea) oauth2Config(ctx context.Context) (*oauth2.Config, context.Context) {
+	publicFacingOAuthURL := server.Config.Server.ForgeOAuthHost
```
xoxys, line 91: I would shorten the var name a bit and use e.g. `publicOAuthURL` but just a personal preference.

xoxys (approved):
LGTM in general, personally I would prefer an `INTERNAL_` prefix instead of `DEV_`.

docs/docs/91-migrations.md
zc-devs: I'm for `internal` then. It is more aligned with deployment architecture, I think about it like "internal network".

zc-devs:
I've drawn a diagram. Feel free to edit.
[wp-schema.drawio.txt](https://github.com/woodpecker-ci/woodpecker/files/15209483/wp-schema.drawio.txt)

docs/docs/91-migrations.md
anbraten: renamed to internal

server/forge/gitea/gitea.go
anbraten: changed it

server/forge/gitea/gitea.go
anbraten: added it

cmd/server/server.go
```diff
@@ -317,11 +317,14 @@ func setupEvilGlobals(c *cli.Context, s store.Store) error {
...
-	if c.IsSet("server-dev-oauth-host") {
-		server.Config.Server.OAuthHost = c.String("server-dev-oauth-host")
+	if c.IsSet("server-dev-oauth-host-deprecated") {
+		server.Config.Server.OAuthHost = c.String("server-dev-oauth-host-deprecated")
 	} else {
 		server.Config.Server.OAuthHost = serverHost
 	}
+	if c.IsSet("server-forge-oauth-host") {
+		server.Config.Server.ForgeOAuthHost = c.String("server-forge-oauth-host")
+	}
```
anbraten, line 327:
```suggestion
  server.Config.Server.ForgeOAuthHost = c.String("server-forge-oauth-host")
```

anbraten, line 327:
```suggestion
  server.Config.Server.ForgeOAuthHost = c.String("server-forge-oauth-host")
```

docs/docs/91-migrations.md
```diff
@@ -12,6 +12,8 @@ Some versions need some changes to the server configuration or the pipeline conf
 - Deprecated alternative names for secrets, use `environment` with `from_secret`
 - Deprecated slice definition for env vars
 - Deprecated `environment` filter, use `when.evaluate`
+- Use `WOODPECKER_INTERNAL` instead of `WOODPECKER_DEV_GITEA_OAUTH_URL` or `WOODPECKER_DEV_OAUTH_HOST`
```
anbraten, line 15:
```suggestion
- Use `WOODPECKER_INTERNAL_OAUTH_HOST` instead of `WOODPECKER_DEV_GITEA_OAUTH_URL` or `WOODPECKER_DEV_OAUTH_HOST`
```

xoxys:
I can give it a quick test with my local dev setup.

xoxys:
I'm a bit confused: `WOODPECKER_DEV_GITEA_OAUTH_URL` was "user-facing **gitea server url** for oauth" and  `WOODPECKER_DEV_OAUTH_HOST` was "**server fully qualified url** used for oauth redirect"

Now we have `WOODPECKER_INTERNAL_FORGE_OAUTH_HOST` that still holds the **"server fully qualified url**. Why is it named _FORGE_OAUTH_HOST then? IMO, that's confusing and should be `WOODPECKER_INTERNAL_OAUTH_HOST`.

cmd/server/flags.go
```diff
@@ -324,6 +319,11 @@ var flags = append([]cli.Flag{
 		Usage:   "skip ssl verification",
 		EnvVars: []string{"WOODPECKER_FORGE_SKIP_VERIFY", "WOODPECKER_GITHUB_SKIP_VERIFY", "WOODPECKER_GITLAB_SKIP_VERIFY", "WOODPECKER_GITEA_SKIP_VERIFY", "WOODPECKER_BITBUCKET_SKIP_VERIFY"},
 	},
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_INTERNAL_FORGE_OAUTH_HOST", "WOODPECKER_DEV_GITEA_OAUTH_URL"}, // TODO: remove WOODPECKER_DEV_GITEA_OAUTH_URL in next major release
+		Name:    "forge-oauth-host",
+		Usage:   "fully qualified url (<scheme>://<host>[/<prefixpath>]). use it if your forge url WOODPECKER_FORGE_URL or WOODPECKER_GITEA_URL, ... isn't a public url",
```
xoxys, line 325:
```suggestion
		Usage:   "fully qualified forge url (<scheme>://<host>[/<prefixpath>]). use it if your forge url WOODPECKER_FORGE_URL or WOODPECKER_GITEA_URL, ... isn't a public url",
```

cmd/server/flags.go
```diff
@@ -410,10 +405,19 @@ var flags = append([]cli.Flag{
 		Usage:   "serve the website by using a proxy (used for development)",
 		Hidden:  true,
 	},
+	//
+	// internal flags
+	//
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_INTERNAL_WEBHOOK_HOST", "WOODPECKER_WEBHOOK_HOST"}, // TODO: remove WOODPECKER_WEBHOOK_HOST in next major release
+		Name:    "server-webhook-host",
+		Usage:   "server fully qualified url called by forge's webhooks (<scheme>://<host>[/<prefixpath>])",
```
xoxys, line 414:
```suggestion
		Usage:   "fully qualified woodpecker server url called by forge's webhooks (<scheme>://<host>[/<prefixpath>])",
```

xoxys:
Ok, nevermind. Looks like `WOODPECKER_DEV_OAUTH_HOST` becomes `WOODPECKER_INTERNAL_WEBHOOK_HOST`.

xoxys:
Should we deprecate `WOODPECKER_DEV_WWW_PROXY` as well and rename it to `WOODPECKER_INTERNAL_WWW_PROXY`?

xoxys (changes requested):
`WOODPECKER_INTERNAL_FORGE_OAUTH_HOST` doesn't seem to work. I have set:

```
WOODPECKER_FORGE_URL: http://localhost:3000
WOODPECKER_INTERNAL_FORGE_OAUTH_HOST: http://woodpecker-gitea:3000
```

but woodpecker still tries to use `WOODPECKER_FORGE_URL`:

```
woodpecker-server   | {"level":"error","error":"Post \"http://localhost:3000/login/oauth/access_token\": dial tcp [::1]:3000: connect: connection refused","time":"2024-05-13T08:32:36Z","caller":"/build/server/api/login.go:64","message":"cannot authenticate user"}
```

anbraten:
> Should we deprecate WOODPECKER_DEV_WWW_PROXY as well and rename it to WOODPECKER_INTERNAL_WWW_PROXY?

The `WOODPECKER_DEV_WWW_PROXY` setting only makes sense for developers and therefore seems to be sth different to me compared to the `INTERNAL_` settings which can make sense for users, but should normally not be touched.

anbraten:
> Now we have WOODPECKER_INTERNAL_FORGE_OAUTH_HOST that still holds the "server fully qualified url. Why is it named _FORGE_OAUTH_HOST then? IMO, that's confusing and should be WOODPECKER_INTERNAL_OAUTH_HOST.

I named it `WOODPECKER_INTERNAL_OAUTH_HOST` at the beginning as well. However now as querty requested it to be an actual forge option I've added the `FORGE_` prefix to be consistent will the other forge options.

xoxys:
With the latest changes, it's completely broken for me:

```
      WOODPECKER_HOST: http://localhost:8000
      WOODPECKER_GITEA: true
      WOODPECKER_FORGE_URL: http://localhost:3000
      WOODPECKER_INTERNAL_FORGE_OAUTH_HOST: http://woodpecker-gitea:3000
      WOODPECKER_INTERNAL_WEBHOOK_HOST: http://woodpecker-server:8000
```

After clicking on the login button in Woodpecker, it tries to redirect to `http://woodpecker-gitea:3000/login/oauth/authorize` now.

anbraten:
> After clicking on the login button in Woodpecker, it tries to redirect to http://woodpecker-gitea:3000/login/oauth/authorize now.

That seems to be correct: `fully qualified forge url (<scheme>://<host>[/<prefixpath>]). use it if your forge url WOODPECKER_FORGE_URL or WOODPECKER_GITEA_URL, ... isn't a public url`.
You probably want it the other way around:

```
      WOODPECKER_HOST: http://localhost:8000
      WOODPECKER_GITEA: true
      WOODPECKER_FORGE_URL: http://woodpecker-gitea:3000 # forge url used by woodpecker to call forge
      WOODPECKER_INTERNAL_FORGE_OAUTH_HOST: http://localhost:3000 # public forge url
      WOODPECKER_INTERNAL_WEBHOOK_HOST: http://woodpecker-server:8000
```

anbraten:
`INTERNAL` was introduced instead of `DEV_`  to let the user now it is a setting that should normally not be touched. It's not about the forge-oauth-host being internal or something like that.

xoxys:
OK, let me rephrase it. `WOODPECKER_FORGE_URL` should be the URL that is used by regular users. In most cases, e.g. gitea.example.com. This url can be resolved by the container __and__ browser, everything works as expected.

 If we now have an environment where the url needed for browser and container access is not the same, I need to flip it? That's IMO a bad UX.

xoxys:
Some more tests:

```
      WOODPECKER_HOST: http://localhost:8000
      WOODPECKER_INTERNAL_WEBHOOK_HOST: http://woodpecker-server:8000
      WOODPECKER_FORGE_URL: http://woodpecker-gitea:3000
      WOODPECKER_INTERNAL_FORGE_OAUTH_HOST: http://localhost:3000
```
- Login flow works
- Trigger CI run by webhook works
- Clone failed as it tries to clone from http://localhost:3000

```
fatal: unable to access 'http://localhost:3000/testorg/test.git/': Failed to connect to localhost port 3000 after 0 ms: Couldn't connect to server
```

---

```
      WOODPECKER_HOST: http://localhost:8000
      WOODPECKER_INTERNAL_WEBHOOK_HOST: http://woodpecker-server:8000
      WOODPECKER_FORGE_URL: http://localhost:3000
      WOODPECKER_INTERNAL_FORGE_OAUTH_HOST: http://woodpecker-gitea:3000
```
- Login flow broken as oauth redirect URL is now http://woodpecker-gitea:3000 which doesn't work in the browser

---

Requirements:

- a way to access the woodpecker-server from browser
- a way to access the woodpecker-server from forge container
- a way to access forge-server from browser
- a way to access forge-server from woodpecker container

Question:

What's the correct configuration?

anbraten:
> Clone failed as it tries to clone from http://localhost:3000/

That url is coming from gitea itself. You have to set the correct public url in gitea.

zc-devs:
>INTERNAL was introduced instead of DEV_ https://github.com/woodpecker-ci/woodpecker/pull/3670#discussion_r1587553355 to let the user now it is a setting that should normally not be touched.

If it is in the docs, then doesn't look "internal" to me anymore (https://github.com/woodpecker-ci/woodpecker/pull/3670#discussion_r1586593999).

> It's not about the forge-oauth-host being internal or something like that.

Well, [I treat it like that](https://github.com/woodpecker-ci/woodpecker/pull/3670#discussion_r1589965162). That was the meaning of `WOODPECKER_WEBHOOK_HOST`.

Seems, it is still confusing...

> //localhost

Doesn't look like public faced URL... https://github.com/woodpecker-ci/woodpecker/discussions/3669#discussioncomment-9284761

---

Draw what you want to achieve, then it can be discussed. It difficult to understand stuff like `localhost`, `woodpecker-server`, `woodpecker-gitea`. My understanding is in https://github.com/woodpecker-ci/woodpecker/pull/3670#issuecomment-2094123336.

Though, requirements section in https://github.com/woodpecker-ci/woodpecker/pull/3670#issuecomment-2111844743 seems right to me.

xoxys:
Looks like we have (again) a communication issue. It's probably up to me then, however, no idea how to improve it. From my perspective, I have clearly communicated what my problem is and what I would expect from these settings. 

> That url is coming from gitea itself. You have to set the correct public url in gitea as well.

OK, in that case the env vars seem to work as expected then. The rest of the issue seems to be related to my localhost dev setup, can solve that somehow...

anbraten:
As other users might be confused by the `INTERNAL_` prefix, I've renamed all of them to `EXPERT_`, so users will hopefully know that they are only required in complex setups and will normally be avoided.

xoxys:
IMO, `EXPERT_` sound a bit weird. I would keep `INTERNAL_` and just extend the documentation if required. However, no strong feelings.

xoxys (approved):

Issue #1689, opened 2023-04-02
build: let the tag pipeline announce new release


6543:
we missed the release announcement on discord for  ~3weeks

so next time the tagging pipeline should create a message into our chat announce channels if all succeeded

https://discordapp.com/channels/838698813463724034/885877564449624095/1092157720683090042

qwerty287:
Once the bsky plugin is working we can close this.

PR #3678, opened 2024-05-03
Add discord announcements

qwerty287:
see https://github.com/woodpecker-ci/woodpecker/issues/1689

@woodpecker-ci/owner We need a webhook and the ID/token for the `announcements` channel.

### DIFF
--- a/.woodpecker/social.yaml
+++ b/.woodpecker/social.yaml
@@ -26,7 +26,7 @@ steps:
   - name: discord
     image: docker.io/appleboy/drone-discord:1.3.1
     settings:
-      webhook_id: TODO
+      webhook_id: 1236558396820295711
       webhook_token:
         from_secret: discord_token
       message: '**{{ build.tag }}** was released: <https://github.com/woodpecker-ci/woodpecker/releases/tag/{{ build.tag }}>'
--- a/.woodpecker/social.yaml
+++ b/.woodpecker/social.yaml
@@ -6,7 +6,7 @@ when:
   - event: tag
 
 steps:
-  mastodon-toot:
+  - name: mastodon-toot
     image: docker.io/woodpeckerci/plugin-mastodon-post
     settings:
       server: https://floss.social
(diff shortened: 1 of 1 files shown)


.woodpecker/social.yaml
```diff
@@ -16,10 +16,17 @@ steps:
       ai_token:
         from_secret: openai_token
       ai_prompt: |
-        We want to present the next version of our app on Twitter.
+        We want to present the next version of our app on Mastodon.
```
anbraten, line 19: Used Twitter for the prompt to get a shorter and more precise result.

.woodpecker/social.yaml
```diff
@@ -16,10 +16,17 @@ steps:
...
         Therefore we want to post a catching text, so users will know why they should
         update to the newest version. If there is no special feature included
         just summarize the changes in a few sentences. Use #WoodpeckerCI, #release and
         additional fitting hashtags and emojis to make the post more appealing.
 
         The changelog entry: {{ changelog }}
+  - name: discord
+    image: docker.io/appleboy/drone-discord:1.3.1
+    settings:
+      webhook_id: TODO
```
anbraten, line 29:
```suggestion
      webhook_id: 1236558396820295711
```

anbraten (approved):

.woodpecker/social.yaml
```diff
@@ -16,10 +16,17 @@ steps:
       ai_token:
         from_secret: openai_token
       ai_prompt: |
-        We want to present the next version of our app on Twitter.
+        We want to present the next version of our app on Mastodon.
         Therefore we want to post a catching text, so users will know why they should
         update to the newest version. If there is no special feature included
         just summarize the changes in a few sentences. Use #WoodpeckerCI, #release and
         additional fitting hashtags and emojis to make the post more appealing.
```
anbraten, line 23:
```suggestion
        We want to present the next version of our app on Mastodon.
        Therefore we want to post a catching text, so users will know why they should
        update to the newest version. Highlight the most special features. If there is no special feature 
        included just summarize the changes in a few sentences. The whole text should not be longer than 240
        characters. Avoid naming contributors from. Use #WoodpeckerCI, #release and 
        additional fitting hashtags and emojis to make the post more appealing.
```

.woodpecker/social.yaml
anbraten: I still think it is quite nice to use an llm to summarize as if done correctly it can phrase a quite nice text (most of the time at a native speaker level and without spelling mistakes). Having the option to highlight specific changes will probably enhance the generation and til then I would just use what we currently have.

PR #3679, opened 2024-05-03
Clarify info on failing workflows/Steps

Andre601:
I felt like the info at the start of the workflow syntax page was a bit misleading.
It mentioned that a failed step cancels the full workflow, not mentioning that steps with a `status: [failed]` will actually execute under such circumstances.

I've replaced that part with a `:::note` block, informing about this special case, while also linking to the option itself.
Hope this is a good change.

### DIFF
--- a/docs/docs/20-usage/20-workflow-syntax.md
+++ b/docs/docs/20-usage/20-workflow-syntax.md
@@ -1,6 +1,10 @@
 # Workflow syntax
 
-The workflow section defines a list of steps to build, test and deploy your code. Steps are executed serially, in the order in which they are defined. If a step returns a non-zero exit code, the workflow and therefore all other workflows and the pipeline immediately aborts and returns a failure status.
+The workflow section defines a list of steps to build, test and deploy your code. Steps are executed serially, in the order in which they are defined.
+
+:::note
+Should a step exit with a non-zero code, the workflow and all remaining steps will abort and return a failure status. An exception to this rule are steps with a [`status: [failure]`](#status) condition, which will execute them on a failed run.
+:::
 
 Example steps:
 
--- a/docs/docs/20-usage/20-workflow-syntax.md
+++ b/docs/docs/20-usage/20-workflow-syntax.md
@@ -1,11 +1,9 @@
 # Workflow syntax
 
-The workflow section defines a list of steps to build, test and deploy your code. Steps are executed serially, in the order in which they are defined.
-
-If a step returns a non-zero exit code, the workflow and therefore the whole pipeline immediately aborts and returns a failure status.
+The Workflow section defines a list of steps to build, test and deploy your code. The steps are executed serially in the order in which they are defined. If a step returns a non-zero exit code, the workflow and therefore the entire pipeline terminates immediately and returns an error status.
 
 :::note
-An exception to this rule are steps with a [`status: [failure]`](#status) condition, which will execute them on a failed run.
+An exception to this rule are steps with a [`status: [failure]`](#status) condition, which ensures that they are executed in the case of a failed run.
 :::
 
 Example steps:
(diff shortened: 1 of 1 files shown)


docs/docs/20-usage/20-workflow-syntax.md
```diff
@@ -1,6 +1,10 @@
 # Workflow syntax
 
-The workflow section defines a list of steps to build, test and deploy your code. Steps are executed serially, in the order in which they are defined. If a step returns a non-zero exit code, the workflow and therefore all other workflows and the pipeline immediately aborts and returns a failure status.
+The workflow section defines a list of steps to build, test and deploy your code. Steps are executed serially, in the order in which they are defined.
+
+:::note
+Should a step exit with a non-zero code, the workflow and all remaining steps will abort and return a failure status. An exception to this rule are steps with a [`status: [failure]`](#status) condition, which will execute them on a failed run.
+:::
```
zc-devs, line 7:
```suggestion
If a step returns a non-zero exit code, the workflow and therefore the whole pipeline immediately aborts and returns a failure status.

:::note
An exception to this rule are steps with a [`status: [failure]`](#status) condition, which will execute them on a failed run.
:::
```

docs/docs/20-usage/20-workflow-syntax.md
```diff
@@ -1,6 +1,12 @@
 # Workflow syntax
 
-The workflow section defines a list of steps to build, test and deploy your code. Steps are executed serially, in the order in which they are defined. If a step returns a non-zero exit code, the workflow and therefore all other workflows and the pipeline immediately aborts and returns a failure status.
+The workflow section defines a list of steps to build, test and deploy your code. Steps are executed serially, in the order in which they are defined.
+
+If a step returns a non-zero exit code, the workflow and therefore the whole pipeline immediately aborts and returns a failure status.
```
xoxys, line 5:
```suggestion
The Workflow section defines a list of stepsto build, test and deploy your code. The steps are executed serially in the order in which they are defined. If a step returns a non-zero exit code, the workflow and therefore the entire pipeline terminates immediately and returns an error status.
```

docs/docs/20-usage/20-workflow-syntax.md
```diff
@@ -1,6 +1,12 @@
...
+
+:::note
+An exception to this rule are steps with a [`status: [failure]`](#status) condition, which will execute them on a failed run.
```
xoxys, line 8:
```suggestion
An exception to this rule are steps with a [`status: [failure]`](#status) condition, which ensures that they are also executed in the case of a failed run.
```

docs/docs/20-usage/20-workflow-syntax.md
Andre601: That has some typos in it, like `stepsto`

docs/docs/20-usage/20-workflow-syntax.md
Andre601: I dislike this change.
Mainly because the `also` implies that `status: [failure]` would make the pipeline execute on failure AND success, when it actually only executes on failure.

xoxys: Changed

docs/docs/20-usage/20-workflow-syntax.md
xoxys: Fixed

xoxys:
@Andre601 can you check it again?

Andre601:
> @Andre601 can you check it again?

Done. Checked on mobile and there you can't commit suggestions :/

xoxys (approved):

Andre601:
Looks like some CI runs encounter propblems (pr/docs seems to have a 403 error).

Issue #3680, opened 2024-05-04
The value of CI_PIPELINE_PARENT is the ID of the parent pipeline


yitsushi:
### Component

server

### Describe the bug

Based on the documentation:
- CI_PIPELINE_NUMBER: pipeline number
- CI_PIPELINE_PARENT: number of parent pipeline

My understanding is that the `CI_PIPELINE_PARENT` should be the `CI_PIPELINE_NUMBER` of the parent pipeline.

In reality it references to the `id` of the pipeline and not the actual number.

**Expected Behaviour**

The `CI_PIPELINE_PARENT` value is `2` in the `deploy` pipeline triggered from the pipeline with `CI_PIPELINE_NUMBER=2`.

**Why it's an issue?**

In my notifications (Matrix channel for example), I want to see clearly what was the build that was deployed. Now it's a bit messy as I can see:
```
✅ Build #2 of org/repo succeeded.
```
and then
```
✅ Promotion of build #135 to production on org/repo succeeded.
```

**Example**

```yaml
# build.yaml
when:
  event:
    - push

steps:
  - image: alpine
    commands:
      - "echo 'Current: ${CI_PIPELINE_NUMBER}'"
      - "echo 'Parent: ${CI_PIPELINE_PARENT}'"
```

```yaml
# deploy.yaml
when:
  environment: production
  event: deployment

steps:
  - image: alpine
    commands:
      - "echo 'Current: ${CI_PIPELINE_NUMBER}'"
      - "echo 'Parent: ${CI_PIPELINE_PARENT}'"
      - "echo 'Target: ${CI_PIPELINE_DEPLOY_TARGET}'"
```

**Output**

```
# build
+ echo 'Current: 2'
Current: 2
+ echo 'Parent: 0'
Parent: 0
```

```
# deploy (from the previous pipeline)
+ echo 'Current: 3'
Current: 3
+ echo 'Parent: 139'
Parent: 139
+ echo 'Target: production'
Target: production
```

I checked in the database and indeed, the `build` pipeline's ID is `139`.

### System Info

```shell
❯ curl https://cicd.efertone.me/version
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.4.1"}

❯ docker stack ps woodpecker
ID             NAME                      IMAGE                                   NODE                         DESIRED STATE   CURRENT STATE           ERROR     PORTS
[...]

PR #3683, opened 2024-05-06
Fix parent pipeline env

qwerty287:
Closes https://github.com/woodpecker-ci/woodpecker/issues/3680

### DIFF
--- a/server/pipeline/restart.go
+++ b/server/pipeline/restart.go
@@ -64,7 +64,7 @@ func Restart(ctx context.Context, store store.Store, lastPipeline *model.Pipelin
 	}
 
 	newPipeline := createNewOutOfOld(lastPipeline)
-	newPipeline.Parent = lastPipeline.ID
+	newPipeline.Parent = lastPipeline.Number
 
 	err = store.CreatePipeline(newPipeline)
 	if err != nil {


anbraten (approved):

xoxys:
Not that simple. I had this issue quite often with Drone and Woodpecker while using the autoscaler. Summary from my observation:

- Some US cloud platforms to heavy IP based geo/- spam blocking
- With an autoscaler setup you get a random IP from the Hetzner pool 
- If a former owner of the IP has used it for bad habit, they get tainted and blocked by some providers

Solutions:

- Quick and dirty: Change the Hetzer DC. IP ranges seems to be different for DCs so temp switching to e.g. Nue, Fsn or Hel was working for me in most cases.
- Use an HTTP proxy server with a static IP and set HTTP_PROXY/HTTPS_PROXY env on the affected steps. However, most HTTP_SERVERS can't be used with authentication, binding them to a public IP will allow everyone to use your proxy.
- Create a wiregurad network and connect all autoscalers to this network. Use a persistent host with a static IP as wiregurad gateway and route the entire internet traffic over this gateway.

xoxys:
> Would caching using sth like the s3-cache help with this at least for the majority of executions

In a lot of cases, no.

> Otherwise as this problem seems to just occur rarely we are probably best to just throw away the agent and try again for now.

Keep in mind that you will get the same IP address assigned again if they were not assigned to someone else already.

PR #3692, opened 2024-05-11
Skip golangci in pre-commit.ci

qwerty287:
It looks like something has changed for golangci-lint and it breaks our current workflows. See for example https://results.pre-commit.ci/run/github/410589193/1714881301.sLyiCpR2QZODkjw5ZZUs7w. Probably, you have to download go modules but I don't really know whether that's possible.


xoxys:
This must be some pre-commit cloud foo... Executing pre-commit locally with this version works as expected. I still think we should just use Woodpecker instead of pre-commit cloud.

xoxys (approved):

PR #3695, opened 2024-05-12
Use IDs for tokens

anbraten:
We currently used user and repo names for tokens. Users will need to login again. Repo hooks don't need to be updated as we never actually check the tokens content, but just use the token as hash.


xoxys:
Why is it better to use IDs? What's the motivation for this change?

anbraten:
Repos can be renamed and multiple users with the same username can exist when using multiple forges.

shared/token/token.go
```diff
@@ -145,13 +159,16 @@ func keyFunc(token *Token, fn SecretFunc) jwt.Keyfunc {
...
 		token.Kind, _ = kind.(string)
 
-		// extract the token value and cast to
-		// expected type.
-		text, ok := claims["text"]
-		if !ok {
-			return nil, jwt.ErrInvalidType
+		// extract the data claims
+		for k, v := range claims {
+			if k == "type" || k == "exp" || k == "nbf" || k == "iat" || k == "aud" || k == "iss" || k == "sub" || k == "jti" {
```
lafriks, line 164: imho we should do opposite to check only supported ones:

```suggestion
			if k != "text" && k != "user-id" && k != "repo-id" {
```

anbraten: I would like to keep it this way as we otherwise have to add our custom claims to the list before we could use `token.Get` or `token.Set` correctly. The list is from the official spec: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1 and should not change. I've added the source as a comment as well.

lafriks: I was looking this more from security/performance perspective (not that there are any risks currently I can think of) as allow list is usually best way to go than deny list but anyway I'm not blocking

PR #3698, opened 2024-05-12
Create devcontainer.json

StemmlerSisters:


6543:
links to some documentation would also be useful

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,5 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
```
lafriks, line 2: I would prefer we would use `mcr.microsoft.com/devcontainers/go:1.22-bullseye` with features required for woodpecker development like `ghcr.io/devcontainers/features/node:1` with same node version we use in pipelines etc

lafriks (changes requested):

lafriks:
> Hello! Could you explain why this change is necessary and what it's doing?

This would allow developing woodpecker without needing to install development tools (go, node etc) on your pc, VSCode would detect and ask to switch to devcontainer mode where it starts vscode instance in container that is build based on this file.

More info in <https://containers.dev/>

Would be nice so that would also allow setting up test gitea instance etc to work with but this is not prerequesite for this PR

anbraten:
> Would be nice so that would also allow setting up test gitea instance etc to work with but this is not prerequesite for this PR

Would be great to have the same setup as we use for gitpod. Shouldn't be actually to tough to achieve using devcontainers with a docker-compose file.

6543:
can containers use nix flakes?

if so, we can use the flake to maintain the dependencys

lafriks:
> Do we really need yet another dev environment? We just had that discussion in #3702.

devcontainers are supported out of the box in vscode (github codespaces and other services also use it), it has more compatibility between different tools not like gitpod or nix that works only with specific environment

devcontainers can be used in win/linux/mac and does not depend on any external service like gitpod

lafriks:
Personally I don't use gitpod as I don't see any reason for me to use it (it's external commercial cloud, requires special plugin and registration etc).

Devcontainers can be used in GitHub to do quick development and contributions thus more easy to use:
image

VSCode does not require any external services at all (just VSCode and Docker), have not used but I think IntelliJ also supports devcontainers out of the box.

There are also multiple open source tools that allow to use devcontainers having just docker and this tool installed (ex [daytona.io](https://www.daytona.io/) or [devpod.sh](https://devpod.sh/) )

Nix is nice but also requires to have nix/nixos and works only for linux but it's small config so not that it's hard to maintain.

So I would vote for droping gitpod and replacing it with devcontainers

6543:
Well we can move forwarx with this and remove you custom made 'in_docker' makefile target https://github.com/woodpecker-ci/woodpecker/blob/402e357fe8754067a91c958e1b819d13b9faf6a1/Makefile#L44

Also the related dockerfile to build the dev env can be used here for this ...

6543:
and we could reuse nix: https://github.com/xtruder/nix-devcontainer

lafriks:
> and we could reuse nix: https://github.com/xtruder/nix-devcontainer

does not look very standard way to use devcontainers and not sure it works in github codespaces

6543:
Well it either works or the devcontaiberprommise is broken

lafriks:
in cloud based environments there could be some problems with specific used mounts/volumes and security options used there

6543:
I'll make a test :)

anbraten:
Started some testing here: https://github.com/woodpecker-ci/woodpecker/compare/main...anbraten:woodpecker:dev-container Mainly missing the env var setup

PR #3700, opened 2024-05-13
use wayback machine to fix offline link

6543:


6543:
that's a codeberg-pages problem and i dont whant to let our ci fail because of that

so until that's fixed I stick to the pined wayback machine ... no harm in having the wayback link

PR #3701, opened 2024-05-13
docs: bump deps

6543:


6543:
I can not get that sec issue fixed on our side :/

it's docusaurus who need to do that

Issue #3705, opened 2024-05-14
Linter: cSpell add golang dic


6543:
we should use `go` and `node` dics

https://cspell.org/docs/dictionaries

to be able to ignore less paths

for that we have to find a way to easy install mentioned dictionaries for ci as via make

6543:
https://github.com/woodpecker-ci/woodpecker/blob/42f2734308f2cb3810edfd4d3048ce0beb50eed1/.cspell.json#L178

Now is only missing

anbraten:
and `docs/` and `_test.go`

6543:
Do test files need to be checked for misspells?!?

anbraten:
I would say so. They are not generated and include text that could have mistakes.

PR #3706, opened 2024-05-14
cspell lint go code

6543:
address  #3705


xoxys:
In general, I like that we add spellchecking to more areas. But is there no way to handle excludes in a central place? E.g. most imports could be excluded by a `ignoreRegExpList` setting?

6543:
Well it is handled in one place at .cspell.json , or what do you intend?

xoxys:
I'm talking about the `// cspell:` annotation behind a lot of import URLs.

6543:
well yes because the usernames are mostly not in a dictionary and I don't want to add them to ours as we don't need them globaly

xoxys:
That's why I asked for a central `ignoreRegExpList` config and not dict entry. Basically something like:

```
"ignoreRegExpList": [
    ".*github.com/.*"
]
```

URLs should not be linted at all.

xoxys:
Here is an example config https://gist.github.com/r3code/21d1e9a3f862ad865808f07225b59068

.woodpecker/social.yaml
```diff
@@ -25,7 +25,7 @@ steps:
 
         The changelog entry: {{ changelog }}
   - name: discord
-    image: docker.io/appleboy/drone-discord:1.3.2
+    image: docker.io/appleboy/drone-discord:1.3.2 # cspell:disable-line
```
xoxys, line 28: Can we add a regex for container image urls aswell?

.woodpecker/test.yaml
```diff
@@ -74,7 +74,7 @@ steps:
   lint-license-header:
     image: *golang_image
     commands:
-      - go install github.com/google/addlicense@latest
+      - go install github.com/google/addlicense@latest # cspell:words addlicense
```
xoxys, line 77: Same here, exlude all go install commands by a regex

agent/runner.go
```diff
@@ -50,11 +50,11 @@ func NewRunner(workEngine rpc.Peer, f rpc.Filter, h string, state *State, backen
 	}
 }
 
-func (r *Runner) Run(runnerCtx context.Context) error { //nolint:contextcheck
+func (r *Runner) Run(runnerCtx context.Context) error { //nolint:contextcheck // cspell:words contextcheck
```
xoxys, line 53: All nolint annotations should be exckuded by a regex

pipeline/backend/kubernetes/utils.go
```diff
@@ -65,23 +65,23 @@ func isImagePullBackOffState(pod *v1.Pod) bool {
 	return false
 }
 
-// getClientOutOfCluster returns a k8s clientset to the request from outside of cluster.
+// getClientOutOfCluster returns a k8s client set to the request from outside of cluster.
 func getClientOutOfCluster() (kubernetes.Interface, error) {
-	kubeConfigPath := os.Getenv("KUBECONFIG")
+	kubeConfigPath := os.Getenv("KUBECONFIG") // cspell:words KUBECONFIG
```
xoxys, line 70: Common, could be added to dictionary.

pipeline/backend/kubernetes/utils.go
```diff
@@ -23,7 +23,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
+	client_cmd "k8s.io/client-go/tools/clientcmd" // cspell:disable-line
```
xoxys, line 26: Why does this line need to be excluded? Shouldnt it be covered by the regex exclude for import statements?

pipeline/backend/local/command.go
```diff
@@ -50,11 +50,12 @@ func (e *local) genCmdByShell(shell string, cmds []string) (args []string, err e
 		return []string{"/c", cmd.Name()}, nil
 	case "fish":
 		script := ""
-		for _, cmd := range cmds {
+		for _, cmd := range cmdList {
 			script += fmt.Sprintf("echo %s\n%s || exit $status\n", strings.TrimSpace(shellescape.Quote("+ "+cmd)), cmd)
 		}
 		return []string{"-c", script}, nil
-	case "powershell", "pwsh":
+	case "powershell", "pwsh": // cspell:words pwsh
```
xoxys, line 57: pwsh is afaik also common and could be added to the dict.

.woodpecker/test.yaml
6543: yes but the next lines not

pipeline/backend/kubernetes/utils.go
6543: image

not realy common i would argue

xoxys (approved):

PR #3708, opened 2024-05-14
Setting for empty commits on path condition

da-Kai:
Allow to set the behavior on empty commits evaluating `path` conditions.

default value is true, to preserve current behavior.

### DIFF
--- a/docs/docs/20-usage/20-workflow-syntax.md
+++ b/docs/docs/20-usage/20-workflow-syntax.md
@@ -402,7 +402,7 @@ when:
 
 You can use [glob patterns](https://github.com/bmatcuk/doublestar#patterns) to match the changed files and specify if the step should run if a file matching that pattern has been changed `include` or if some files have **not** been changed `exclude`.
 
-The behavior on an empty commit can be set using `fail_on_empty`. A value of `false`(default) will evaluate this condition to true.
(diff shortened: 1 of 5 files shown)


da-Kai:
Hello @qwerty287,

I expanded the documentation a bit. I hope it is now a little clearer. 😊

I added `on_empty` to `when-path-include-exclude` in
[pipeline/frontend/yaml/linter/schema/.woodpecker/test-when.yaml](https://github.com/da-Kai/woodpecker/blob/4a6f5fbd4256d8530bf675d6c9e3c726808c5149/pipeline/frontend/yaml/linter/schema/.woodpecker/test-when.yaml#L113-L122)

anbraten:
> If I get the code correctly, settings on_empty: true will run the step (except other conditions are false) and on_empty: false will not run it?
> Can you somehow make this more clear in the docs? 

@qwerty287 As both of use were not sure about this option when looking at it for the first time, might it be a good idea to call it sth like `ignore_when_no_file_changed: true` instead?

@da-Kai What do you think?

da-Kai:
@anbraten @qwerty287 
I agree that `on_empty` is not very clear. But I also didn't want to use a name that was too long.

Maybe one of these 2?
- `pass_on_empty_commit`
- `ignore_empty_commit`

PR #3710, opened 2024-05-15
Replace duplicated strings

### LINKED ISSUE #1347: Remove redundancy in translation entrys
especially the settings object is 3x in the locales witch most stuff is just identical!

the redundant translation entry's is unrelated to this but need to be solved ... in an other pull

_Originally posted by @6543 in https://github.com/woodpecker-ci/woodpecker/issues/1344#issuecomment-1292860196_

qwerty287:
Mainly the secrets strings.

Is there a way how we can not lose the translations?

Closes #1347

### DIFF
--- a/web/src/components/admin/settings/AdminSecretsTab.vue
+++ b/web/src/components/admin/settings/AdminSecretsTab.vue
@@ -6,12 +6,7 @@
     :warning="$t('admin.settings.secrets.warning')"
   >
     <template #titleActions>
-      <Button
-        v-if="selectedSecret"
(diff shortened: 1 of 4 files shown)


anbraten:
> Is there a way how we can not lose the translations?

I think the [translation memory](https://docs.weblate.org/en/weblate-4.18.2/admin/memory.html) should do that.

anbraten (approved):
code lgtm. not 100% sure if all translations exists (might be a nice idea to add https://eslint-plugin-vue-i18n.intlify.dev/rules/no-missing-keys.html)

PR #3711, opened 2024-05-15
Fix privileged steps in kubernetes

### LINKED ISSUE #3537: Woodpecker 2.4 breaks privileged steps/plugins with Kubernetes backend

anbraten:
closes #3537


xoxys:
Yep, it looks like it ignores the entire discussion from the linked PR.

anbraten:
Step privileged is only set in case its a plugin and in the list of privilege plugins:

https://github.com/woodpecker-ci/woodpecker/blob/75803dba41a0a276af34606f6330fedf944873b9/pipeline/frontend/yaml/compiler/convert.go#L130

Or if the `privileged: true` option is set. 

In case a user tries to set that option without having a trusted repo he will get a linter error from:

https://github.com/woodpecker-ci/woodpecker/blob/b2cfa37682b002360a63b65213e473d290a32e7d/pipeline/frontend/yaml/linter/linter.go#L115

https://github.com/woodpecker-ci/woodpecker/blob/b2cfa37682b002360a63b65213e473d290a32e7d/pipeline/frontend/yaml/linter/linter.go#L153

anbraten:
Setting the security-context / user / group to `0` (root) or fs-group etc is checked here and will only work for privileged steps which can only be allowed by trusted plugins or by setting `privileged: true` which can only be done if the repo is trusted: 

https://github.com/woodpecker-ci/woodpecker/blob/9cc5d0875c9172c0c7ffd09e150d42bed94c9e3d/pipeline/backend/kubernetes/pod.go#L391

https://github.com/woodpecker-ci/woodpecker/blob/9cc5d0875c9172c0c7ffd09e150d42bed94c9e3d/pipeline/backend/kubernetes/pod.go#L445

anbraten:
> Yep, it looks like it ignores the entire discussion from the linked PR.

It solves the issue that the `woodpeckerci/plugin-docker-buildx` plugin can only be used if a `securityContext` is explicitly configured. For already privileged steps (trusted plugins) this wont be necessary anymore now.

anbraten:
My goal was to simply fix the actual issue #3537 first. Changing the logic to require a trusted repo for the `buildx` plugin is somehow a different change IMO (at least that's how the docker backend works and the docs explain it to the users) . 

Requiring a trusted repo to manually set the `securityContext` could be done separately.

xoxys:
I would suggest to involve the people that have created, contributed and tested the original PR. This entire discussion should have been done in the existing PR instead of just starting the a new one...

anbraten:
Thanks for approving.

> Please use a new discussion to revert these changes and make it how it should work ideally, but the bug should be fixed asap.

We can just move on with #3538 as this is part of it.

PR #3717, opened 2024-05-22
Remove discord as official community channel

6543:
as we mostly already stop-ed using discord, we should not encourage to new users


6543:
well i would first start with just make it unofficial ...

anbraten:
I would just leave the general channel. Seems like some users still use it and help each other.

lafriks (approved):

pat-s (approved):

6543:
> I would just leave the general channel. Seems like some users still use it and help each other, so why should we force them to use matrix.

I think we should not force - the discord server is still able to be discovered via discord and I wont delete the server - e.g. it's now just unofficial but we can & still should have a look at it from time to time

6543:
I dont see harm in having release notes there ... untill it breaks at that point we can remove it

6543:
ok could you submit a pull to rm it then?

Issue #3311, opened 2024-02-01
Agents environment variables


biolds:
### Clear and concise description of the problem

Hello,

I'm trying set per agent environment variable but could not find a way to do it. I want to use them to set ip addresses to access external resources that varies by agent.
So far, I have tried without success:
- passing `--env=WOODPECKER_ENV_FILE=/etc/drone.conf` to docker, like it was done with Drone
- passing `--env=WOODPECKER_ENVIRONMENT=extern_ip:1.2.3.4`, but this one is server only it seems
- passing `--env=WOODPECKER_HOSTNAME=my_agent`, this variable is supported by Woodpecker, but it seems like it's not available from the pipelines

Did I miss a way to pass agent specific data to the steps, or a way to identify which worker is running the test ?

Thanks for your effort on this great software!

### Suggested solution

Support one of the three way I tried.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

silverwind:
It's not implemented yet it seems. My propsal from https://github.com/woodpecker-ci/woodpecker/issues/3762 is:

> Support a option `WOODPECKER_AGENT_ENV_FILE` that works like [DRONE_RUNNER_ENV_FILE](https://docs.drone.io/runner/docker/configuration/reference/drone-runner-env-file/) which can be mounted into the agent which would make the variables available to all pipelines.

zc-devs:
> it could possibly done in the agent config ui too

Sounds similar to https://github.com/woodpecker-ci/woodpecker/pull/3723, but for/per Agent.

xoxys:
Setting `WOODPECKER_ENVIRONMENT` on the server works for me. Env variable are passed to every pipeline automatically. Only tested with`next`.

PR #3723, opened 2024-05-26
Implement repo level, user level and global environment variables support

### LINKED ISSUE #3661: Global `environment` option to set env variables across all steps

Right now, there only seems to be a way to define global env variables for the entire CI, rather than a single pipeline itself.
In addition is this from what I understand only something an admin can do, meaning users on a shared instance may not have a way to easily define an environment that is accessible across all steps (See "Alternatives" for the ones I know of and their (mostly) cons).

### Suggested solution

A global `environment` option should be implemented, which allows you to define the environment variables you want to use across all steps.

I'm unsure if this global one should be overridable by a step-level environment setting with the same name.

Example:
```yaml
when:
  - event: push

environment:
  FOO: BAR # Set a global env here.

steps:
  - name: 'echo'
    image: bash
    commands:
      - echo ${FOO}
  - name: 'echo again'
    image: bash
    environment:
      FOO: NOT_BAR # Should this be possible?
    commands:
      - echo ${FOO}
```

### Alternative

From what I understand and gathered are there these possible alternatives, which all have some downsides to them:

- Set a secret.
  - `-` Using a secret for sharing a non-secret value globally seems wrong.
  - `-` doesn't allow using other env variables to be processed(? Unsure about this one).
  - `+` It's available across all steps and can also be used in plugin settings.
- Create a step that saves a value to an envvars file and then later get it.
  - `-` Depending on the image set may require different commands to set and get.
  - `-` You can't pull the env in a plugin (unless there is an aproach to run commands before the image execution somehow).
  - `-` Requires to pull the envvars file on every step to make it available.
  - `+` Above negative allows only exposing the env to specific steps.

### Additional context

_No response_

### Validations

dvjn:
Fixes #3661


dvjn:
Hi @6543 @anbraten @qwerty287 I have done initial UI changes for this, please review the proposal and screenshots and let me know if I should continue working on this and implement the rest.


**Screenshots:**

Repo Settings:
repo settings

User Settings:
user settings

Admin Settings:
admin settings

Add Variable Screen:
add variable

anbraten:
As variables are not secret anyways there was a suggestion somewhere to have a special config file like `.woodpecker/config.yaml` in the repo which can be used for such settings. This would be quite handy as it would also allow to use those variables in the `cli exec` command as well and allows easy copies of that config to other woodpecker instances etc. Should those variables then be shown in the UI as well as read-only or should it only be the config file? What do you think?

zc-devs:
JFYI, how it is implemented in [Gitea](https://blog.gitea.com/release-of-1.22.0/#rocket-support-instance-level-actions-variables-28115).

dvjn:
> As variables are not secret anyways there was a suggestion somewhere to have a special config file like `.woodpecker/config.yaml` in the repo which can be used for such settings.

I think this is not as useful as it seems, since you can already use variables in a workflow file. This `config.yaml` approach is mainly beneficial when you need these variables in multiple workflows within the same repository.
However, the most important use-cases of this PR, in my opinion, are: first, the ability to manage global variables using the admin UI, and second, the capability to share variables across repositories under an organization or user.

FYI, the solution I'm proposing is heavily inspired by Gitea.

dvjn:
@anbraten @qwerty287 @zc-devs
Do let me know your final thoughts on this and let me know if I should move forward with implementing it.

dvjn:
I can also add agent variables to address #3311

zc-devs:
> your final thoughts

I like the idea.

> if I should move forward with implementing it

I am not a right guy :) Let's hear from maintainers.

> I can also add agent variables

That would be great!

dvjn:
> However, you write "Fixes #3661" in the desc which is not correct (I believe there's an issue about your thing too). #3661 is about workflow-level vars and this should definitely be defined in the yaml.

My bad, will remove that issue from the description.

itsTurnip:
I'm really interested in this PR. Keep moving! Thanks for your work.

dvjn:
I have completed the code changes in the server and web frontend. The docs and cli changes are still pending. Planning to do those after an initial round of code review.

It would be great if someone could review the code.

Summary of Changes:
- Duplicated code for secrets (removing events and images).
- Added support for `from_variable` and `variables` (similar to `from_secret` and `secrets`) in pipeline config.
- Added new table in database `variables`, with migrations.
- Copied some secrets test cases and used for variables.

lafriks:
Wasn't there an idea to drop `secrets` section from step? Maybe let's do not add variables section and keep only one way to use them with `from_variable`?

dvjn:
> Wasn't there an idea to drop `secrets` section from step?

I am not aware of it. If it is just an idea without a proper discussion, maybe let's not act on that?

> Maybe let's do not add variables section and keep only one way to use them with `from_variable`?

I have kept feature parity with secrets, to not cause confusion. If we want to remove it, wouldn't it be better to remove both `secrets` and `variables` together in a separate PR.

Just my thoughts, let me know if we still want to remove it.

zc-devs:
If I set some variable
333869343-cacbe4cd-f92a-4fcb-8287-8972ea97df60
and run pipeline
```yaml
skip_clone: true
steps:
  glob-vars:
    image: alpine
    commands:
      - echo $$COMPANY
```
what would I see in steps (UI) logs?

--- 

Maybe add `build_pr_images` label?

dvjn:
> what would I see in steps (UI) logs?

It would be empty as you haven't explicitly added those variables in your step either by using `variables` or `from_variable`.

zc-devs:
As this is not a secret value, it doesn't make sense to restrict this values by a step (to me).
So, I would expect it to be printed without `explicitly added those variables in your step`.
In case if I want to map "global" var to different name in a step, I can use `from_variable`.

dvjn:
@zc-devs

That is a good point. That would make this feature much easier to use.

Although, I feel that this kind of implicit behaviour might cause some confusion. It won't always be obvious where an environment variable is coming from.

Also, it could be confusing for a user who has push permissions on a repo, but cannot change the repo settings. This kind of user won't be able to view where the variables are automatically getting set from.

dvjn:
I understand your point, but I'm still afraid of adding this new implicit behavior.

Can we get another opinion if that's okay?

dvjn:
Hi maintainers, it would be great to get a review from y'all.

lafriks:
I don't think setting them explicitly is a good idea as that could have unforeseen consequences that would be not so easy to track down both from usability and security perspective

lafriks:
> > Wasn't there an idea to drop `secrets` section from step?
> 
> I am not aware of it. If it is just an idea without a proper discussion, maybe let's not act on that?
> 
> > Maybe let's do not add variables section and keep only one way to use them with `from_variable`?
> 
> I have kept feature parity with secrets, to not cause confusion. If we want to remove it, wouldn't it be better to remove both `secrets` and `variables` together in a separate PR.
> 
> Just my thoughts, let me know if we still want to remove it.

Personally I don't see benefit of having multiple ways of using same feature, just adds confusion.

Also if we add it and later on remove we will have to again create more breaking changes that can be prevented without adding that in the first place

lafriks:
I don't think setting them explicitly is a good idea as that could have unforeseen consequences that would be not so easy to track down both from usability and security perspective

lafriks:
Also would be confusing if both variables and secrets would be used by the same name on what takes preference etc.

I only briefly skimmed through file changes but imho this introduces back security vulnerability with using `variables` section with plugins

dvjn:
Thank you for the comments.

So from what I understood, I'll just keep one way to use variables, i.e., using `from_variable`, and remove support for `variables` in pipeline. Right?

Then I'll complete the Cli and Docs changes for the final review.

dvjn:
@lafriks I have removed the support for using `variables`, and completed the go client, cli and doc changes.

This PR is ready to be reviewed and merged.

dvjn:
Hi maintainers, I have finished working on this PR. This is ready for review.

6543:
before reading more into this pull code and issue, tow important questions:
- why did you touch the `pipeline/*` code at all? I thought we already allow to have environment viariables
- if variables dont have a prefix, all currently steps counted as plugins wo have variables set should not be contued as plugins!

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -98,6 +98,16 @@ func (c *Compiler) createProcess(container *yaml_types.Container, stepType backe
 
 	workingDir = c.stepWorkingDir(container)
 
+	getVariableValue := func(name string) (string, error) {
+		name = strings.ToLower(name)
```
6543, line 102: we switched back to case sensitive for secrets ... so we should stay consistent

web/src/assets/locales/en.json
```diff
@@ -255,6 +255,9 @@
       "secrets": {
         "desc": "Organization secrets can be passed to all organization's repository's pipeline steps as environmental variables."
       },
+      "variables": {
+        "desc": "Organization variables can be passed to all organization's repository's pipeline steps as environmental variables."
```
6543, line 259:
```suggestion
        "desc": "Organization variables can be passed to all organization's repository's pipeline steps (e.g. as environmental variables)."
```

web/src/assets/locales/en.json
```diff
@@ -409,6 +419,22 @@
       }
     }
   },
+  "variables": {
+    "variables": "Variables",
+    "desc": "Variables can be passed to individual pipeline steps at runtime as environmental variables.",
```
6543, line 424:
```suggestion
    "desc": "Variables can be passed to individual pipeline steps at execution start.",
```

6543: clear differentiate to variables as e.g. github-actions uses them

pipeline/frontend/yaml/compiler/convert.go
dvjn: I can still see `name = strings.ToLower(name)` in the `getVariableValue` even after merging main, should I still proceed to remove them from variables?

6543: let me have a look at the last pulls in this topic ...

dvjn: Any updates on this?

6543:
uh damed i got requested but it fall under my todo list :?

pipeline/frontend/yaml/compiler/compiler.go
```diff
@@ -107,6 +113,7 @@ func New(opts ...Option) *Compiler {
 		env:                 map[string]string{},
 		cloneEnv:            map[string]string{},
 		secrets:             map[string]Secret{},
+		variables:           map[string]Variable{},
```
6543, line 116: I think a simple `map[string]string` should be sufficient enouth ...

pipeline/frontend/yaml/compiler/convert.go
6543: yes case sensitive it is

pipeline/frontend/yaml/compiler/compiler.go
6543: bf7b2996c4c530a4e2d8cc95d12f37f4eec56caf

pipeline/frontend/yaml/compiler/settings/params.go
```diff
@@ -160,10 +160,10 @@ func handleComplex(v any, getSecretValue func(name string) (string, error)) (str
 	return string(out), nil
 }
 
-// injectSecret probes if a map is a from_secret request.
+// injectVariableOrSecret probes if a map is a from_secret request.
 // If it's a from_secret request it either  returns the secret value or an error if the secret was not found
 // else it just indicates to progress normally using the provided map as is.
-func injectSecret(v map[string]any, getSecretValue func(name string) (string, error)) (string, bool, error) {
+func injectVariableOrSecret(v map[string]any, getVariableValue, getSecretValue func(name string) (string, error)) (string, bool, error) {
 	if secretNameI, ok := v["from_secret"]; ok {
```
6543, line 167: we might move those two keys into const

6543: 4c4221e80f0f4651bb00f69b0a4f8fa9932c3701

6543:
Review process:
 - [x] docs
 - [x] pipeline engine
 - [ ] api
 - [ ] server datastore
 - [ ] cli
 - [ ] webui

6543:
NOTE: write into docs that it is case-sensitive

6543:
# TODO:
 - [ ] The Variable views should addopt KeyValueEditor (https://github.com/woodpecker-ci/woodpecker/pull/4412)
 - [ ] variable creation via WebUI should work again

6543:
@dvjn using secrets as template is fine ... but there are a lot of copy-paste errors :/

pat-s:
> It would be empty as you haven't explicitly added those variables in your step either by using variables or from_variable.

If that would be the case, most of the implementation would be meaningless as then one could simply use `secrets`, as already mentioned.

The real value of this PR IMO lies in the possibility to (easily) set global/org-wide variables which are not sensitive. Without having to repeat them within the YAML again.
For every variable there could/should be an informative log line telling that this variable has been set by a global/org-wide one.

Also, I am not sure if "variables" is the best wording if the values are mapped as env vars in the end. Then, it could directly be named "Environment". This might avoid confusion if there is a difference between a "variable" and an "env var".

PR #3726, opened 2024-05-26
Fix oauth file envvars

### LINKED ISSUE #1479: the `*_FILE` values are not sanitized/cleaned sending control characters to upstream oAuth servers
### Component

server, web-ui

### Describe the bug

When providing the forge client and secret via file (likely also other values, but I've stumbled across these) control characters are not stripped, this causes oAuth to fail as there is as %0A in the id / secret passed to the upstream oAuth server.

I've hit this problem while using `WOODPECKER_GITEA_CLIENT_FILE` and `WOODPECKER_GITEA_SECRET_FILE`, while one can work around this by doing an `echo -n "value" > file`, when adding manually, many secrets providers and tools always have at least the %0A (\0) at the end of string (even when it's just a single line/string), also good luck when preparing these files from a non *nix os.

### System Info

```shell
next-a3f4f305 on Linux
```

### Additional context

When using editors / tools providing the file:

File:
```text
<client-id stripped>

```

Results in
`https://codeberg.org/login/oauth/authorize?client_id=<client-id stripped>%0A%0A&redirect_uri=...`

File:
```text
<client-id stripped>
```
Results in
`https://codeberg.org/login/oauth/authorize?client_id=<client-id stripped>%0A&redirect_uri=...`

---
When doing the `echo -n "" >file` or `cat createdWithEditorOrTool | awk '{printf $1}' > file` trick, there is no control character at the end, and it works as intended.

### Validations

danial23:
Fixes a bug (introduced with #1417) where setting environment variables of the form WOODPECKER_\<forge\>\_CLIENT_FILE and WOODPECKER_\<forge\>\_SECRET_FILE had no effect.

Also fixes #1479.

### DIFF
--- a/server/services/setup.go
+++ b/server/services/setup.go
@@ -21,6 +21,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/rs/zerolog/log"
 	"github.com/urfave/cli/v2"
@@ -114,8 +115,8 @@ func setupForgeService(c *cli.Context, _store store.Store) error {
 		_forge.AdditionalOptions = make(map[string]any)
 	}
 
-	_forge.Client = c.String("forge-oauth-client")
-	_forge.ClientSecret = c.String("forge-oauth-secret")
+	_forge.Client = strings.TrimRight(c.String("forge-oauth-client"), "\r\n")
+	_forge.ClientSecret = strings.TrimRight(c.String("forge-oauth-secret"), "\r\n")
 	_forge.URL = c.String("forge-url")
 	_forge.SkipVerify = c.Bool("forge-skip-verify")
 	_forge.OAuthHost = c.String("forge-oauth-host")
--- a/server/services/setup.go
+++ b/server/services/setup.go
@@ -115,8 +115,8 @@ func setupForgeService(c *cli.Context, _store store.Store) error {
 		_forge.AdditionalOptions = make(map[string]any)
 	}
 
-	_forge.Client = strings.TrimSpace(c.String("forge-oauth-client"), "\r\n")
-	_forge.ClientSecret = strings.TrimSpace(c.String("forge-oauth-secret"), "\r\n")
+	_forge.Client = strings.TrimSpace(c.String("forge-oauth-client"))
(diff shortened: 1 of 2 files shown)


server/services/setup.go
```diff
@@ -114,8 +115,8 @@ func setupForgeService(c *cli.Context, _store store.Store) error {
 		_forge.AdditionalOptions = make(map[string]any)
 	}
 
-	_forge.Client = c.String("forge-oauth-client")
-	_forge.ClientSecret = c.String("forge-oauth-secret")
+	_forge.Client = strings.TrimRight(c.String("forge-oauth-client"), "\r\n")
```
anbraten, line 118: Mabye even `TrimSpace` to remove unexpected `\t` and ` ` as well.

server/services/setup.go
```diff
@@ -114,8 +115,8 @@ func setupForgeService(c *cli.Context, _store store.Store) error {
...
+	_forge.ClientSecret = strings.TrimRight(c.String("forge-oauth-secret"), "\r\n")
```
anbraten, line 119:
```suggestion
	_forge.Client = strings.TrimSpace(c.String("forge-oauth-client"), "\r\n")
	_forge.ClientSecret = strings.TrimSpace(c.String("forge-oauth-secret"), "\r\n")
```

anbraten (approved):

server/services/setup.go
```diff
@@ -114,8 +115,8 @@ func setupForgeService(c *cli.Context, _store store.Store) error {
 		_forge.AdditionalOptions = make(map[string]any)
 	}
 
-	_forge.Client = c.String("forge-oauth-client")
-	_forge.ClientSecret = c.String("forge-oauth-secret")
+	_forge.Client = strings.TrimSpace(c.String("forge-oauth-client"), "\r\n")
+	_forge.ClientSecret = strings.TrimSpace(c.String("forge-oauth-secret"), "\r\n")
```
anbraten, line 119:
```suggestion
	_forge.Client = strings.TrimSpace(c.String("forge-oauth-client"))
	_forge.ClientSecret = strings.TrimSpace(c.String("forge-oauth-secret"))
```

PR #3741, opened 2024-05-31
Fix spelling in README

qwerty287:


anbraten (approved):
isn't cspell checking it?

PR #3745, opened 2024-06-01
Add vue module declaration

qwerty287:
see https://github.com/vuejs/vetur/blob/71aa326b93b152c02e886750d75eefdcb1db83bc/docs/guide/setup.md#vue3

### DIFF
--- a/web/pnpm-lock.yaml
+++ b/web/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '9.0'
+lockfileVersion: '6.0'
 
 settings:
   autoInstallPeers: true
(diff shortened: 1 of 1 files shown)


anbraten (changes requested):
Adding a shim file like this was done in the past, but is not needed anymore and has the disadvantage that all components (things like props) aren't typed anymore. Instead of Vetur, volar is used in here

anbraten:
> What would be affected by this? The vue-tsc or other tools?

All typescript related setups would not use `.vue` files for the types, but instead use the shim file and therefore would have sth like `props: any, emits: any`.

This might be interesting for you:
https://github.com/vuejs/language-tools?tab=readme-ov-file#community-integration
https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/volar.lua

PR #3747, opened 2024-06-01
Add documentation for KUBERNETES_SERVICE_HOST in Agent

jcgl17:
This PR adds some documentation that was missing (from my perspective as a new user). I am running the Agent outside of Kubernetes, and therefore didn't have the benefit of KUBERNETES_SERVICE_HOST already being set. Hopefully this documentation addition can help other newcomers get up and running without also needing to drill into the actual code :)

P.S. I know that this env var isn't a `WOODPECKER_BACKEND_K8S_`-prefixed one, but I figured that this was still the appropriate place to document it. Please let me know if there's a better place to put this information!

### DIFF
--- a/docs/docs/30-administration/22-backends/40-kubernetes.md
+++ b/docs/docs/30-administration/22-backends/40-kubernetes.md
@@ -231,6 +231,12 @@ See [this issue](https://github.com/woodpecker-ci/woodpecker/issues/2510) for mo
 
 These env vars can be set in the `env:` sections of the agent.
 
+### `KUBERNETES_SERVICE_HOST`
+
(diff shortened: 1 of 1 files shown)


jcgl17:
It's not Woodpecker-specific in the sense that Woodpecker is its only user or original user. However, [Woodpecker itself does explicitly use it in its own way](https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/backend/kubernetes/kubernetes.go#L133) rather than the variable being used only implicitly by one of Woodpecker's dependencies.

That being said, I've now moved it into the Tips and Tricks section.

PR #3756, opened 2024-06-03
Fix repo creation

anbraten:


anbraten:
@qwerty287 This should fix the issue you've mentioned. I haven't tested it yet.

anbraten:
tested, works fine

PR #3764, opened 2024-06-06
Fix config loading of cli

### LINKED ISSUE #3739: woodpecker-cli requires dbus-launch?
### Component

cli

### Describe the bug

```console
$ ./woodpecker-cli --config woodpecker-config.json user ls
6:19PM FTL error running cli error="exec: \"dbus-launch\": executable file not found in $PATH"
```
this is from running on a alpine 3.19.1 machine.

### Steps to reproduce

1. on a Alpine linux machine, download woodpecker-cli (I tried both the executable from the releases tag on github, as well as extracting the binary from a woodpecker-cli:v.2.4.1-alpine container).
2. create woodpecker-config.json with "server-url" and "token" keys.

### Expected behavior

Not relying on dbus-launch? It feels like a heavy dependency for a cli utility, no?

### System Info

```shell
Alpine 3.19.1
```

### Additional context

I did search org:woodpecker-ci on github for any dbus-launch reference, and found none, so I guess this is pulled in by some dependency?

### Validations

### LINKED ISSUE #3763: `docker run woodpeckerci/woodpecker-cli setup` tries to run xdg-open to spawn a web browser
### Component

cli

### Describe the bug

The docker version of the CLI wants to spawn a web browser via a launcher of some kind, which will always fail in the context of docker, because the image cannot include a browser.

### Steps to reproduce

```console
$ docker run --rm -it woodpeckerci/woodpecker-cli setup --server-url https://some.host
1:01AM FTL error running cli error="exec: \"xdg-open\": executable file not found in $PATH"
```

### Expected behavior

Rather than raising an error for the not-found case, this logic:

https://github.com/woodpecker-ci/woodpecker/blob/964860434c7548b73778476bf64125e6713d6b0a/cli/setup/token_fetcher.go#L115-L124

could print the URL to stdout so that the user can click on it, or copy and paste it to open it that way.
That would also help in the "unsupported platform" case.

### System Info

```shell
$ docker run --rm woodpeckerci/woodpecker-cli --version
woodpecker-cli version 2.5.0
```

### Additional context

_No response_

### Validations

anbraten:
closes #3739 

closes #3763

- fix config merging
- disable update check in docker image

### DIFF
--- a/cli/common/flags.go
+++ b/cli/common/flags.go
@@ -28,16 +28,16 @@ var GlobalFlags = append([]cli.Flag{
 		Usage:   "path to config file",
 	},
 	&cli.StringFlag{
-		EnvVars: []string{"WOODPECKER_TOKEN"},
-		Name:    "token",
-		Aliases: []string{"t"},
-		Usage:   "server auth token",
-	},
-	&cli.StringFlag{
 		EnvVars: []string{"WOODPECKER_SERVER"},
 		Name:    "server",
 		Aliases: []string{"s"},
 		Usage:   "server address",
+	},
+	&cli.StringFlag{
+		EnvVars: []string{"WOODPECKER_TOKEN"},
+		Name:    "token",
+		Aliases: []string{"t"},
+		Usage:   "server auth token",
 	},
 	&cli.BoolFlag{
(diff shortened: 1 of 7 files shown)


anbraten:
Tested again locally and with docker and now works for me correctly.

PR #3768, opened 2024-06-06
Serve plugin icons locally

### LINKED ISSUE #1771: Docs: Serve plugin icons locally
To prevent connections to external servers and to improve usability if a server is unreachable, download the plugin icons when docs are built and store and serve them from the docs server.

qwerty287:
As docusaurus does not allow plugins to store static files, the icons are stored as base64 in the HTML.

Closes #1771


docs/plugins/woodpecker-plugins/src/index.ts
```diff
@@ -29,6 +29,23 @@ async function loadContent(): Promise<Content> {
           return undefined;
         }
 
+        let pluginIcon: string | undefined;
+        let pluginIconType: string | undefined;
+        if (docsHeader.icon) {
+          try {
+            const response = await axios(docsHeader.icon, {
+              responseType: 'arraybuffer',
+            });
```
anbraten, line 38: What do you think about storing the image as data-url directly?

```suggestion
            const response = await axios(docsHeader.icon, {
              responseType: 'blob',
            });
            
            pluginIconDataUrl = URL.createObjectURL(response.data);
```

anbraten (approved):

PR #3771, opened 2024-06-06
Remove web.archive.org workaround in docs

pat-s:
reverts #3700 as it caused a CI failure in https://ci.woodpecker-ci.org/repos/3780/pipeline/17019/27 (#3722)

The original site at https://jan.wildeboer.net/2022/07/Woodpecker-CI-Jekyll/ is up.


pat-s:
Yes. 

CB pages got a little better but it is still not yet that great. I wouldn't host a private website on right now but we are not here to judge this part.
What we can do is either

- accept the current state
- exclude the URL entirely from being checked
 
The workaround caused issues now, so I doesn't really look like a solid alternative either.

PR #3781, opened 2024-06-11
Remove misleading crontab guru suggestion from docs

wez:
Since the underlying syntax is non-standard, it is very confusing and misleading to suggest using a tool with different syntax!


zc-devs:
https://sysctl.io/crontab-generator
https://www.freeformatter.com/cron-expression-generator-quartz.html
https://www.programmertools.online/generator/cron_expression.html

zc-devs:
> Your first link gives me a 404

Yeah, it is working in my general browser (maybe cached), but not working in other or in incognito. I hope it is temporary issue.
Anyway, it is self-hosted [it-tools](https://github.com/CorentinTh/it-tools) project. The same is deployed at https://it-tools.tech/.

PR #3785, opened 2024-06-13
Migrate to maintained cron lib and remove seconds

qwerty287:
Use https://github.com/gdgvda/cron and remove seconds from cron syntax to use standard cron.


zc-devs:
If you build an image, I can test.

server/store/datastore/migration/031_cron_without_sec.go
```diff
@@ -0,0 +1,48 @@
...
+			return fmt.Errorf("sync new models failed: %w", err)
+		}
+
+		var crons []*model.Cron
+		if err := sess.Find(&crons); err != nil {
+			return err
+		}
+
+		for _, c := range crons {
+			c.Schedule = strings.SplitN(c.Schedule, " ", 2)[1]
```
anbraten, line 40: Needs to check for amount of spaces to support `@daily` etc.

anbraten:
https://github.com/go-co-op/gocron is using github.com/robfig/cron internally

xoxys:
> What do you mean by "reliable"?

My main concern was the reputation of the owner/org. The repo has not many stars, the maintainer has nearly no activity on GitHub, same applies to the org. 

If we just use the Parser, why not embed this part in the WP code directly? The parser code wasn't touched since 4 years, even in the new fork. However, I don't have a strong opinion on this and won't block you, I'm just a bit more careful about unpopular repos/authors.

anbraten:
> If we just use the Parser, why not embed this part in the WP code directly? The parser code wasn't touched since 4 years, even in the new fork. However, I don't have a strong opinion on this and won't block you, I'm just a bit more careful about unpopular repos/authors.

We could also just stick to the lib as it works AFAIK and just change the format to the one without seconds.

pat-s:
If we would remove seconds, this would be a breaking change and should go into 3.0.

lafriks (approved):

pat-s (approved):
Migrating to a maintained lib is certainly a good idea - in addition to removing seconds (which are of no practical use anyway).

PR #3809, opened 2024-06-21
Remove `unplugin-icons`

qwerty287:
Instead, directly import icons.


web/src/components/user/UserGeneralTab.vue
```diff
@@ -28,6 +28,8 @@ import Settings from '~/components/layout/Settings.vue';
 import { setI18nLanguage } from '~/compositions/useI18n';
 import { useTheme } from '~/compositions/useTheme';
 
+import InputField from '../form/InputField.vue';
```
anbraten, line 31:
```suggestion
import InputField from '~/components/form/InputField.vue';
```

anbraten (approved):

anbraten:
From the docs its says you have to add this to the tsconfig:

```
// tsconfig.json
{
  "compilerOptions": {
    "types": [
      "unplugin-icons/types/vue",
    ]
  }
}
```

anbraten:
Would actually try to stick to `iconify` as its pretty handy and flexibel. You can just change the import from `@iconify/json` to `@iconify-json/mdi`

anbraten:
> I can revert it - but I don't really understand what you mean be "flexible". It's only flexible as long as you install all sets and that's explicitly what I not want to do

`unplugin-icons` has a quite unified way for importing different iconsets and seems to be well maintained in comparison to `@jamescoyle/vue-icon` (last update 3 years ago). I don't see a reason why we should replace it. 🤔 (Also really like the [index](https://icones.js.org/))

web/src/components/atomic/Button.vue
```diff
@@ -28,7 +28,7 @@
           'bg-wp-control-error-200': color === 'red',
         }"
       >
-        <Icon name="loading" class="animate-spin" />
+        <Icon name="spinner" fill="#FFFFFF" />
```
6543, line 31: hardcoded collor?

lafriks (approved):

PR #3814, opened 2024-06-22
Allow to set dist dir using env var

6543:
just move some hard-coded string behind an env var as per https://www.bestpractices.dev/en/projects/5309 suggested


Makefile
```diff
@@ -8,6 +8,8 @@ ifeq ($(TARGETOS),windows)
 	BIN_SUFFIX := .exe
 endif
 
+DESTDIR ?= dist
```
anbraten, line 11:
```suggestion
DIST_DIR ?= dist
```

We should remove the file from `cspell`s ignore list

xoxys (approved):

PR #3816, opened 2024-06-22
Document how to pitch ideas & proposals

6543:


pat-s (commented):
Isn't this already covered by the existing issue templates and the "discussions" option?

I also wouldn't mention the "closed" matrix group here, it does not add any value. 

Overall I am not sure this section is really needed, "everyone" knows nowadays where/how to contribute?

6543:
well I'm currently going though https://www.bestpractices.dev/en/projects/5309 ... and thought it was missing

but some could argue the issue templates are reference enouth :D

PR #3829, opened 2024-06-24
Let webhook pass on pipeline parsion error

6543:
Instead of show in e.g. gitea a 500, we get an 200 response and the actual pipeline info

this helps to debug and easyer to integrate ...


6543:
juts test it ... update a pipeline definition to an invalid format ... e.g. non valid json

Before you got an 500 ... now you get 200

An other status is fine by me too but not 500

PR #3831, opened 2024-06-24
Exclude mocks from release build

6543:
- Makefile pass TAGS env var into go tests and builds
- put mockery generated code behind "test" tag
- adjust linting and testing to use the "test" tag


6543:
moved unrelated changes of  #3820 into this pull

.golangci.yaml
```diff
@@ -184,3 +184,6 @@ issues:
 
 run:
   timeout: 15m
+  build-tags:
+    - test
+    - integration
```
6543, line 189:
```suggestion
```

PR #3840, opened 2024-06-26
Update forgejo sdk

qwerty287:
It looks like renovate is unable to do this automatically?


6543 (approved):

pat-s:
Worth inspecting the renovate logs: https://ci.codeberg.org/repos/12974

6543:
either rerun when less resources are used or we have to deploy agents on VMs that have more ram ...

pat-s:
Digests are always a bit problematic. Maybe it works better now with a semver version!

PR #3849, opened 2024-06-27
Fix cancel states

anbraten:
closes #2875
closes #833
closes #2062
closes #2911
closes #2197

PR #3850, opened 2024-06-28
Cleanup state reporting

anbraten:
Extracted from #3849 to just have all the renamings and logging cleanups for the beginning

# Changes
- [x] use separate WorkflowState and StepState types (quite some attributes were assigned but never used)
- [x] rename vars to improve readability (exp id => workflowID)
- [x] as WorkflowStates never handled exit-codes apart from logging them, the error handling in the runnger.go has slightly changed


agent/runner.go
```diff
@@ -123,76 +122,66 @@ func (r *Runner) Run(runnerCtx context.Context) error { //nolint:contextcheck
...
-		}
+
+	if errors.Is(err, pipeline.ErrCancel) {
+		canceled = true
+	} else if canceled {
+		err = errors.Join(err, pipeline.ErrCancel)
+	}
+
+	if err != nil {
+		state.Error = err.Error()
```
anbraten, line 165: Note for reviewers: this is the only bigger change in this PR. As workflow state updates never considered exit-codes in the actual logic and just used them for logging, I've removed them. Instead `state.Error` is now set respectively.

server/grpc/rpc.go
```diff
@@ -220,24 +220,25 @@ func (s *RPC) Init(c context.Context, id string, state rpc.State) error {
...
 	s.updateForgeStatus(c, repo, currentPipeline, workflow)
+
 	return nil
 }
 
-// Done implements the rpc.Done function.
-func (s *RPC) Done(c context.Context, id string, state rpc.State) error {
-	workflowID, err := strconv.ParseInt(id, 10, 64)
+// Done marks the workflow with the given ID as done.
+func (s *RPC) Done(c context.Context, _workflowID string, state rpc.WorkflowState) error {
```
lafriks, line 233: please do not use param/variable names with underscore, would be better to use `workflowID` here and `id, err := strconv...` line below

lafriks (commented):
looks good from code side beside underscored param naming

server/grpc/rpc.go
anbraten: Wanted to not use `id` to avoid confusion with step.ID or pipeline.ID. Maybe `strWorkflowID`? Whats the issue with leading `_`s?

agent/rpc/client_grpc.go
```diff
@@ -277,17 +271,16 @@ func (c *client) Extend(ctx context.Context, id string) (err error) {
...
 	req.Id = id
-	req.State = new(proto.State)
-	req.State.Error = state.Error
-	req.State.ExitCode = int32(state.ExitCode)
-	req.State.Exited = state.Exited
-	req.State.Finished = state.Finished
-	req.State.Started = state.Started
+	req.State = new(proto.StepState)
 	req.State.StepUuid = state.StepUUID
+	req.State.Started = state.Started
```
6543, line 280: Missing exited

anbraten: Do you have an idea what exited is actually used for? Could we just use `Finished != 0`?

6543: well yes but I would rather have it removed in an dedicated pull to be able to git-bisect if things go wrong ...

anbraten: Readded it for now

server/grpc/rpc.go
lafriks: it's not really how naming is usually done in Go, there shouldn't really be confusion if function param name would be named workflowID and in function body just use id as that variable is used only once and are 2 lines apart from converting from string to number so there shouldn't really be any confusion what it represents

xoxys (approved):

server/grpc/rpc.go
anbraten: changed the vars to not start with `_`

PR #3877, opened 2024-07-07
Exclude dummy backend in production

qwerty287:
folow-up https://github.com/woodpecker-ci/woodpecker/pull/3820

Really exclude the backend instead of replacing with a noop one.


docs/pnpm-lock.yaml
```diff
@@ -14,22 +14,22 @@ importers:
     dependencies:
       '@docusaurus/core':
```
6543, line 15: Unrelated

pipeline/backend/dummy/dummy.go
```diff
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build test
```
6543, line 15: I would still let the tag stay

pipeline/backend/dummy/dummy_test.go
```diff
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build test
```
6543, line 15: could you also re-add the tags here please?

anbraten: Same happens as mocks are excluded since #3831

6543 (approved):

PR #3880, opened 2024-07-08
Fix vscode settings

anbraten:


6543:
@anbraten what error do you get?

anbraten:
Sth about build tags are configured twice.

6543:
could you provide logs or an screenshot ...

anbraten:
image

image

As the docs state this option is basically not needed as "go.buildTags" is already set.

6543 (approved):

6543:
must be my IDE who is in the broken state then ...

lafriks (approved):

PR #3887, opened 2024-07-12
docs: Fix spelling

finga:
Use "an" instead of "a" for "`.env` file" and use the correct tense.


zc-devs:
An environment file or a dot-env file? 😆

finga:
True xD

6543 (approved):

PR #3895, opened 2024-07-13
Gracefully shutdown agent

6543:
# agent in idle
image

# agent while exec an workflow
image
_(we now get the logs and cancel status reported back to the server if an agent gets stopped gracefully)_


6543:
we now get the logs and cancel status reported back to the server if an agent gets stopped gracefully

lafriks (approved):

PR #3897, opened 2024-07-13
Convert to openapi 3.0

anbraten:
"Migrate" is currently done by converting the swagger 2.0 to an openapi 3.0 spec using `github.com/getkin/kin-openapi/openapi2conv` as https://github.com/swaggo/swag/issues/386 itself is not having 3.0 support and alternatives for a code first approach using 3.0 don't seem to exist.


lafriks:
There are alternatives (I have been using this one in my projects for quite some time) - https://github.com/lafriks-fork/goas but it does have quite different syntax for comments

anbraten:
> There are alternatives (I have been using this one in my projects for quite some time) - https://github.com/lafriks-fork/goas but it does have quite different syntax for comments

Sure long term goal would be to have some first class support. Just wanted to migrate for now to be able to continue with generated types for typescript.

6543 (approved):

silverwind:
This would be great to have in Gitea as well in case you are up for a PR there 😉.

PR #3918, opened 2024-07-16
Remove all default 3rd party privileged plugins

6543:
As woodpecker do assume a lot of trust to those plugins, if we have no influence we should not **by default** have them as trusted.

#4056 will future enhance the security model of it,

close  #4053


6543:
well now we can argue about tradeoffs: harden but make it inconfinient and harder for beginners, or more confinient and beginner freandly but not hardened.

about the security implications, if it whould be that bad to have privileged plugin we should go and delete our buildx anyway ...
e.g. what I try to say instead of fearing that buildx might do things wrong, improve and harden the buildx plugin itselve

docs/docs/91-migrations.md
```diff
@@ -11,6 +11,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 
 ## `next`
 
+- Remove `plugins/docker`, `plugins/gcr` and `plugins/ecr` from the default list of privileged plugins ([change via config](./30-administration/10-server-config.md#woodpecker_escalate)).
```
6543, line 14:
```suggestion
- Remove `plugins/docker`, `plugins/gcr` and `plugins/ecr` from the default list of privileged plugins ([modify the list via config if needed](./30-administration/10-server-config.md#woodpecker_escalate)).
```

xoxys:
There are a LOT of other ways to build container without privileged containers, so not everyone needs this plugin out of the box. However, there are still valid reasons to use buildkit, but I would not escalate any plugin by default. This should be a decision made by the server admin, and the admin should know the impact before enabling it globally. Setting one option more with a simple list is not inconvenient IMO.

lafriks:
This PR deals with risk that 3rd-party plugins are trusted by default, I think we should keep our plugin buildx trusted for now by default until there is better plugin for building docker images that does not require special privileges

lafriks (approved):

xoxys (changes requested):
Ok then I would like to block this PR until we discussed it further. For me, this doesn't make sense, the buildx plugin is a high risk plugin and as the next release will be a breaking release anyway we should drop it now all together.

zc-devs:
> better 

Too abstract.

> plugin for building docker images that does not require special privileges

They all require _special_ [privileges](https://github.com/containers/buildah/blob/main/docs/tutorials/05-openshift-rootless-build.md).
But if we are talking about [privileged](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#securitycontext-v1-core) alone, then there is a choice:
https://github.com/woodpecker-ci/plugin-kaniko
https://github.com/maltegrosse/woodpecker-buildah
https://codeberg.org/Taywee/woodpecker-buildah/

xoxys:
We have 3:2 votes for removing all default trusted plugins. I don't see why we should keep it? It's pretty simple to enable it server-wide again if (and only if a server admin is aware of the impact). There are way more high impact changes in the upcoming v3 that this one.

lafriks:
But I think most instances will still need to add it anyway as nowadays containers are pretty much a way to go for deployment, so I don't really think this will improve anything apart to make 90% (wild guess and might be exaggeration 😁) this change that they will find out mostly by builds not working without really improving security as with our current limitations set by server there is not really a known way of abusing buildx plugin

pat-s (dismissed):
I am fine with this change going into 3.0 **IF** we better communicate what needs to be done to get the functionality back for internal instances. 
We already communicated the removal of the `environment:` section in plugins badly lately and caused a lot of trouble.
Such PRs should IMO also include changes to the docs and refs to other PRs which update the docs there (e.g. here to an update in the buildx plugin which explains this change vice-versa).
-> I am blocking this PR as well until we get better at cross-updating all places which are affected by such changes instead of only changing the core codebase.

I can see why it should be not enabled/trusted by default and having a focus on security is surely not a bad thing. At the same time, having a focus often comes with inconvenience as, this is not just an issue of WP or CI in general.

zc-devs:
https://github.com/woodpecker-ci/woodpecker/issues/3997#issuecomment-2263646174

Seems, this was not deprecated in [2.7.0](https://github.com/woodpecker-ci/woodpecker/releases/tag/v2.7.0) and 

> can't be removed in 3.0 directly

Or why can it? Because of `security` label?

lafriks:
But what will we really achieve by that apart by forcing pretty much everyone to add that setting manually?

As I mentioned previously there are no alternatives to build docker images without privileged plugins and there are no known security risks for our controlled plugin but forcing users to choose we might be putting them in more risks than having this default.

So what exact security risk are we addressing by not having our trusted code plugin as default?

xoxys:
> forcing pretty much everyone
Still just a personal assumption.

> As I mentioned previously there are no alternatives to build docker images without privileged

I don't get this point. The dangerous part is the Docker engine `[privileged](https://learn.snyk.io/lesson/container-runs-in-privileged-mode/)` mode. Almost every other agentless container build tool than docker doesn't require the privileged flag (Kaniko, Ko, Buildah, ...), so there _are_ a lot of other solutions that don't required the container build step to be running in privileged mode.

lafriks:
How exactly is that dangerous if you can't really abuse it?

So then why are we not using them if there are better alternatives?

zc-devs:
> They all require special privileges

Correction: they all require a _part_ of root privileges. In Buildah rootless example it is `cap_setgid` and `cap_setuid` only.

> if we are talking about privileged alone

Correction: privileged  = god mod = full access = disabled isolation = disabled SELinux (at least partly)
^ that is what Docker requires, but not Kaniko (for example).
Not only all build images via DinD (I'm not).

xoxys:
> How exactly is that dangerous if you can't really abuse it?

You are keen enough to assume we covered everything this time? Container running in privileged mode that runs with root inside the container are effectively root on the entire host. What could go wrong :shrug: 

You can use privileged container safely even in CIs if e.g. the pipeline config is protected against malicious changes. For the Woodpecker server, this is done by the gated feature. Every CI run needs to be approved by a trusted maintainer.

> So then why are we not using them if there are better alternatives?

Because nobody was motivated enough to work on it?

lafriks:
> > How exactly is that dangerous if you can't really abuse it?
> 
> You are keen enough to assume we covered everything this time? Container running in privileged mode that runs with root inside the container are effectively root on the entire host. What could go wrong :shrug: 
> 
> You can use privileged container safely even in CIs if e.g. the pipeline config is protected against malicious changes. For the Woodpecker server, this is done by the gated feature. Every CI run needs to be approved by a trusted maintainer.

Currently we have fixed all known issues and I'm not saying that we should not remove it but not know.

> > So then why are we not using them if there are better alternatives?
> 
> Because nobody was motivated enough to work on it?

And that's the reason to first build one and when there is decent alternative that does not require privileged mode we can remove that default, otherwise as I already mentioned we will just force to add that and even when alternative is made it will be already added to most instances manually and we can't really remove it anymore and it will just stay there for most users, so it will be even worse situation from security perspective

lafriks:
I just want to have more gradual plan to have more control and not break everything at once, we can add recommendations to disable that if you don't need it to improve security and minimize risks for instances that don't use privileged mode plugins

june128:
Just chiming to mention that kaniko doesn't require *any* privileges and we use it to successfully build images.
However it can't do multi-arch by itself, so that's a trade-off, but ye.

zc-devs:
> doesn't require any privileges

Then run it in restricted PSS.

june128:
> Then run it in restricted PSS.

restricted PSS?

zc-devs:
https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities

june128:
@zc-devs Ahh okay, yeah, fair enough. Let me rephrase then: kaniko doesn't require to run in a privileged container then, so it doesn't need any additional privileges.

6543:
-> https://github.com/woodpecker-ci/woodpecker/pull/4053 ?

6543:
I have created https://github.com/woodpecker-ci/woodpecker/pull/3918 witch should eliminate most concerns as we now will allow-list only by tag ...

If we want we could even go so far as to pin the hash but I'm not sure if it will work  with  multi arch images ....

xoxys:
> I have created https://github.com/woodpecker-ci/woodpecker/pull/4056 witch should eliminate most concerns as we now will allow-list only by tag ...

I don't know what concern you are referring to. My concern is: Woodecker developers should not make the decision to enable privileged mode for containers by default. It doesn't matter if the default containers on the allowlist are now pinned by the version and/or digest. The argumentation that this setting makes it more complicated for new users is still a strange one. This setting is as complicated as setting `WOODPECKER_ADMIN`. It's a very simple comma-separated list of strings... Document it and add a hint to the `buildx` plugin docs, and you are done. Creating an OAuth client and connecting it to Woodpecker is way more complicated than that...

xoxys:
With this approach, I can't even opt-out as a server admin.

xoxys:
Ok, if that works 👍 but I thought `urvafe/cli` flags will fall back to the default even if the flag is set to an empty string.

6543:
> Ok, if that works 👍 but I thought `urvafe/cli` flags will fall back to the default even if the flag is set to an empty string.

let me lookup this ...

I think that we can agree on remove at least this much for now ... so if it's possible to set it to `blank` I'll merge this pull and we can discusst the rest on https://github.com/woodpecker-ci/woodpecker/pull/4053 ...

6543:
`WOODPECKER_ESCALATE=''` did not work in the past ... now it does  
`WOODPECKER_ESCALATE=' '` would have done it too...

also we clearly indicate via an lint error when the pipelines are still in usage, so users migrating from drone or with old config don't debug hours and hours

cmd/server/flags.go
```diff
@@ -161,7 +161,7 @@ var flags = append([]cli.Flag{
 	&cli.StringSliceFlag{
 		Sources: cli.EnvVars("WOODPECKER_ESCALATE"),
 		Name:    "escalate",
-		Usage:   "images to run in privileged mode",
+		Usage:   "Allow plugins to get access to docker socket (privileged mode), if defined but empty  there will be none",
```
6543, line 164:
```suggestion
		Usage:   "Allow plugins to get access to docker socket (privileged mode), if environment variable is defined but empty there will be none",
```

6543:
@xoxys is your concern addressed?

6543:
@pat-s added the linter error so users get notified first hand

xoxys:
> I'll merge this pull and we can discusst the rest on

Why do you make a poll if you then decide on personal preferences? However, I dont have anything to add to the discussion.

cmd/server/flags.go
```diff
@@ -161,7 +161,7 @@ var flags = append([]cli.Flag{
 	&cli.StringSliceFlag{
 		Sources: cli.EnvVars("WOODPECKER_ESCALATE"),
 		Name:    "escalate",
-		Usage:   "images to run in privileged mode",
+		Usage:   "Allow plugins to get access to docker socket (privileged mode), if environment variable is defined but empty there will be none",
```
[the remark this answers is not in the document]
6543: well could you suggest a better wording please

6543:
> Why do you make a poll ...

that is currently `+3 & -2` vs `+3 & -3` ... so that's no clear answer at all ...

cmd/server/flags.go
6543, line 164:
```suggestion
		Usage:   "Allow plugins to run in privileged mode, if environment variable is defined but empty there will be none",
```

Daniel451:
@6543 I am late to the party and since I also just very recently started using Woodpecker, I would opt to not vote anyways and leave this to maintainers / long-term users.

However, I wanted to share something important about that: the one reason I migrated to Woodpecker (except being open-source) is: simplicity. It looked nice, setup was easy, and migration was easy.

I cannot speak for other / new users and increasing security is always a good idea but I would really like to see Woodpecker keeping setup, configuration, ... as easy as possible. In my use-case, for example, I am running the agent on a separated server that literally runs the Woodpecker agents _only_, thus security there is not a big concern. **Easy setup & configuration**, on the other hand, is something that will _always_ impact _all_users.

Just my 2 cents :)

xoxys:
As long as the dedicated node used as agent runner are not isolated from the rest of you network and disconnected from the internet, security should be a concern.

We were talking about full node takeover incl. roor access here.

6543:
Well normal abuse is _only_ about resource abuse ... but asap you can extract secrets or take over a whole host we should have good defaults in place.

And nothing beside building docker is affected ... I get that it's inconvinient :/

And I dont have a better option atm.

As soon as we can do the same without having to forward an docker socket into an step, the problem will be solved for all users, so we should focus on that

PR #3921, opened 2024-07-17
Add docs for v2.7.0

6543:


6543:
we should write a script for this to not forget things!

docs/docusaurus.config.ts
```diff
@@ -236,25 +236,27 @@ const config: Config = {
...
           versions: {
             current: {
               label: 'Next 🚧',
               banner: 'unreleased',
             },
+            '2.7': {
+              label: '2.7.x',
+            },
             '2.6': {
               label: '2.6.x',
```
lafriks, line 253: 💀

6543, line 253:
```suggestion
              label: '2.6.x 💀',
```

PR #3925, opened 2024-07-18
fix: internal server error when listing PRs for empty repositories

hhamalai:
Empty Gitea repositories return error 404 when there are no commits in the repository, which is translated to 500 Internal Server Error on woodpecker.

While personally I would like to see API improvements on Gitea side, Woodpecker should not report 500s for pretty common use cases like this.


6543 (approved):

6543:
@hhamalai please format the code so our linter is pleased :)

PR #3909, opened 2024-07-14
Disalow to set arbitrary environments for plugins

6543:
also split the different step types in our json schema to lint accordingly

PR #3939, opened 2024-07-19
Add quirk of PULLREQUEST_DRONE_PULL_REQUEST not to be ""

6543:
https://github.com/woodpecker-ci/woodpecker/pull/3909#issuecomment-2239163099

cc @wez are you sure its `PULLREQUEST_DRONE_PULL_REQUEST` and not `DRONE_PULL_REQUEST`

### DIFF
--- a/pipeline/frontend/metadata/drone_compatibility.go
+++ b/pipeline/frontend/metadata/drone_compatibility.go
@@ -58,6 +58,16 @@ func SetDroneEnviron(env map[string]string) {
 	// misc
 	copyEnv("CI_SYSTEM_HOST", "DRONE_SYSTEM_HOST", env)
 	copyEnv("CI_STEP_NUMBER", "DRONE_STEP_NUMBER", env)
+
+	// some quirks
+
+	// Legacy env var to prevent the plugin from throwing an error
+	// when converting an empty string to a number
+	//
+	// plugins affected: "plugins/manifest"
+	if env["CI_COMMIT_PULL_REQUEST"] == "" {
+		env["PULLREQUEST_DRONE_PULL_REQUEST"] = "0"
+	}
 }
 
 func copyEnv(woodpecker, drone string, env map[string]string) {


wez:
@6543 yeah, that env var is definitely the setting I use to get this particular thing to work

6543:
@wez does the server run on x86 ?

wez:
I run this particular configuration on amd64 and aarch64; so intel and arm, but 64-bit only

wez:
For clarity, I have an amd64 agent and an aarch64 agent. They both build their respective platform packages and docker image, and they both have a step that tries to merge the other architectures together into a multiarch image with `plugins/manifest`, so I have `merge-docker-image-arm` and `merge-docker-image` step as platform-dependent steps in that particular matrix workflow, and they "race" to merge the manifest together. They are set to `failure: ignore` because one of them will always lose the race.

6543:
Well in this case the server image `woodpeckerci/woodpecker-server:pull_3939` (x84 only) in combination with the newest 2.7.0 agent images should do it

6543:
or do you men to do:

```diff
 if env["CI_COMMIT_PULL_REQUEST"] == "" { 
 	env["PULLREQUEST_DRONE_PULL_REQUEST"] = "0"
+	env["DRONE_PULL_REQUEST"] = "0"
 }
```
?

6543:
ah @wez you struggle with the fact that if you set environment, the plugin will not be considered as plugin anymore ... (security stuff)

wez:
I'm not sure that I understand what you mean there.

I have this in my config, on 2.7.1:

```yaml
  merge-docker-image-arm:
    depends_on: [build-docker-image-arm]
    when:
      - evaluate: 'platform == "linux/aarch64" && UBUNTU_IMAGE == "ubuntu:22.04" && CI_PIPELINE_EVENT == "tag"'
    image: plugins/manifest
    failure: ignore
    environment: &docker_manifest_env
      # Legacy env var to prevent the plugin from throwing an error
      # when converting an empty string to a number
      PULLREQUEST_DRONE_PULL_REQUEST: 0
    settings: &docker_manifest_settings_tag
      <<: *docker_credentials
      target: "ghcr.io/kumocorp/kumomta-dev"
      template: "ghcr.io/kumocorp/kumomta:${CI_COMMIT_TAG}-ARCH"
      tags:
        - latest
        - ${CI_COMMIT_TAG}
      platforms:
        - linux/amd64
        - linux/arm64
```

the presence of the `environment` block yields a lot of warnings in the UI, but the step does run successfully.

I understand that the removal of the `environment` block is to address a potential security issue, but it doesn't appear to block me from running this config.

PR #3953, opened 2024-07-21
Make cli plugin log purge recognize steps by name

smainz:
Give a little love to `woodpecker-cli pipeline log purge`:

- Make it recognize steps by name
- Improve error messages in case of invalid/missing args
- Add step name to message if only logs of a single stepare purged


smainz:
yes, forgot to add this.

PR #3976, opened 2024-07-25
Two row layout for title and context of pipeline list

anbraten:
This PRs adds the two row layout for the pipeline list again to allow to show long messages and pr titles.

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -242,9 +242,7 @@
       "errors": "Errors ({count})",
       "warnings": "Warnings ({count})",
       "show_errors": "Show errors",
-      "we_got_some_errors": "Oh no, we got some errors!",
-      "duration": "Pipeline duration",
-      "created": "Created: {created}"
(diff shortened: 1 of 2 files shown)


zc-devs:
It turns out you're [ahead of me](https://github.com/woodpecker-ci/woodpecker/discussions/2162#discussioncomment-10040399), that is cool 👍 
Could you add some [branch-mark](https://github.com/user-attachments/assets/df7d3ac8-c038-40c7-b86f-a53fab7e954f)?
Add `build-image` label, please.

xoxys:
Does it work on the mobile view? Can you add an example screenshot for that as well?

lafriks:
Maybe we can we just show PR number before title in first row and in second only commit?

lafriks:
I would probably do the other way around as the PR title is more likely to be more descriptive and is the primary info

xoxys:
As you requested a review, there is still feedback that was not addressed yet.

anbraten:
closing for now as stale

PR #3994, opened 2024-08-01
Mark "deployment" event as supported on Gitea and Forgejo

pat-s:
At least on Forgejo they are working, just executed one recently on the CB instance. I assume that it is also working on Gitea?


pat-s:
Ah hm - but isn't the "deploy" event aimed to be executed manually? How would this be triggered by the forge?

Anyhow, we should add more information to the table then, as it is not clear what the features are referring to then (regardless of the outcome of this one).

lafriks:
Gitea does not support deployments

lafriks:
Closing this PR as Gitea and Forgejo does not support triggering deployment events

PR #4006, opened 2024-08-05
Improved Local backend detection

### LINKED ISSUE #3996: Disable auto-fallback to Local backend (v3)

For example, automatic detection doesn't work for Podman. This causes strange errors like `git executable doesn't exist` in the `clone` step.
This issue was posted in the Matrix, but there was something like that before, as I remember.

### Suggested solution

Disable automatic fallback to the `local` backend.

### Alternative

_No response_

### Additional context

_No response_

### Validations

zc-devs:
Automatically detect Local backend only if Agent is NOT running in container.
Closes #3996.


6543 (approved):

6543:
> This actually makes it impossible to use the local backend in containers. Maybe the better solution is to just drop the `auto-detect` option and set `docker` as default value?

oh right ... let me fix it

6543 (approved):

zc-devs:
```go
backend == "local" || !inContainer
```

That is exactly what I've just wanted to write! :)

PR #4009, opened 2024-08-06
Add examples of CI environment variable values

hg:
Not sure if you're interested in this; if not, I'll just keep it for my own reference.

I find it useful when writing build scripts because sometimes the exact way a value is formatted is not obvious, and you have to run a test job to find out.

Drone has something similar, but in a much less convenient form, spread over dozens of pages:

- https://docs.drone.io/pipeline/environment/reference/drone-commit/
- https://docs.drone.io/pipeline/environment/reference/

etc.


pat-s (commented):
I like the idea in general. 
I don't think we should format all values as code, I'd prefer to have them as plain text instead.

pat-s:
Then let's hear more opinions - @woodpecker-ci/maintainers

PR #4010, opened 2024-08-06
[Bitbucket Datacenter] Return empty list instead of null

jenrik:
This change make the Bitbucket Datacenter forge correctly return empty list instead of nil which indirectly fixes the "Run pipeline" button in the repository view.

Previously you where unable to manually run a pipeline because the frontend code expects the API to return a empty JSON array for the last page instead of the null value that where actually returned.


lafriks (approved):

jenrik:
@qwerty287 rebase completed

PR #4016, opened 2024-08-08
Remove old pipeline options

qwerty287:
- secret `target` (alternative names)
- uppercasing secrets
- slice env definitions
- environment `when` filter
- `exclude` in `when`
- `group`


zc-devs:
Even though I do not use plugins, I wonder if there is an alternative now to the removed `secret target (alternative names)`.

lafriks (approved):

PR #4017, opened 2024-08-08
Remove various deprecations

qwerty287:


lafriks (approved):

sloonz:
Why is Configuration in config service removed here ?

https://woodpecker-ci.org/docs/administration/advanced/external-configuration-api does not mention it is deprecated. It explicitly says "Before the run or restart of any pipeline Woodpecker will make a POST request to an external HTTP API sending the current repository, build information **and all current config files retrieved from the repository**". I don’t see much of a point of a configuration service which cannot see the original files ?

fernandrone:
> Why is Configuration in config service removed here ?
> 
> https://woodpecker-ci.org/docs/administration/advanced/external-configuration-api does not mention it is deprecated. It explicitly says "Before the run or restart of any pipeline Woodpecker will make a POST request to an external HTTP API sending the current repository, build information **and all current config files retrieved from the repository**". I don’t see much of a point of a configuration service which cannot see the original files ?

Same. This sees to have broken an internal configuration plugin we had after the 3.0 upgrade.

qwerty287:
The netrc data that is available allows you to query the forge directly.

And I think in https://github.com/woodpecker-ci/woodpecker/pull/3349 this is added back to the per-repo option.

fernandrone:
Thanks for the quick answer!

> The netrc data that is available allows you to query the forge directly.

I'd like to avoid using netrc to pull github data because of https://github.com/woodpecker-ci/woodpecker/issues/4052. It's been hard to manage rate limit with woodpecker already in some repos that have multiple workflows. If we have to pull the source data with external config files then we are just duplicating the issue (and duplicating the code, since woodpecker has already pulled the configdata).

> And I think in #3349 this is added back to the per-repo option.

If it does it's better and we can move to that when/if it's merged (although we'd still need to configure it in a per-repo basis which is less than ideal for our use cases).

Is it possible to revert this? The way I see it this broke a feature (which is [still implied to be working](https://woodpecker-ci.org/docs/administration/advanced/external-configuration-api)) and I did not see this documented in the release.

At the moment we're trying to work around this by building an internal patch for our custom plugins so they use Woodpecker API to pull the config data but seems less than ideal.

qwerty287:
> If it does it's better and we can move to that when/if it's merged (although we'd still need to configure it in a per-repo basis which is less than ideal for our use cases).

It also supports a global option, not only per repo. Not sure if it will be merged in the next release, but I hope it's ready soon. I'll see if I can review it again in the next days.

> Is it possible to revert this? The way I see it this broke a feature (which is still implied to be working) and I did not see this documented in the release.

I wouldn't do a revert.

This was done 100% on purpose. Anything in the docs (your link is broken) that refers to something different is just wrong and old.

fernandrone:
Sorry, here's the correct link https://woodpecker-ci.org/docs/3.8/administration/configuration/server#external-configuration-api. The examples have been updated but the text still implies Woodpecker makes a POST with all the config:

> Before the run or restart of any pipeline Woodpecker will make a POST request to an external HTTP API sending the current repository, build information and all current config files retrieved from the repository. 

To be fair, over a year later it's unlikely a revert would be realistic. Still, for all intents and purposes this is a breaking change and should be listed on https://woodpecker-ci.org/migrations#300 right?

qwerty287:
> Still, for all intents and purposes this is a breaking change and should be listed on https://woodpecker-ci.org/migrations#300 right?

Yes, that's where it should be listed.

PR #4027, opened 2024-08-11
Warn if using secrets/env with plugin

qwerty287:
address https://github.com/woodpecker-ci/woodpecker/discussions/3959

If `secrets` is set, it's not a plugin anymore.

However, only warn about such cases, but do not fail.

This PR should be backported.

### DIFF
--- a/docs/docs/20-usage/51-plugins/51-overview.md
+++ b/docs/docs/20-usage/51-plugins/51-overview.md
@@ -50,9 +50,10 @@ steps:
 Plugins are just pipeline steps. They share the build workspace, mounted as a volume, and therefore have access to your source tree.
 While normal steps are all about arbitrary code execution, plugins should only allow the functions intended by the plugin author.
 
-So there are a few limitations, like the workspace base is always mounted at `/woodpecker`, but the working directory is dynamically adjusted accordingly. So as user of a plugin you should not have to care about this.
-
-Also instead of using environment variables the plugin should only care about one prefixed with `PLUGIN_` witch are the internal representation of the **settings** ([read more](./20-creating-plugins.md)).
+That's why there are a few limitations. The workspace base is always mounted at `/woodpecker`, but the working directory is dynamically
+adjusted accordingly, as user of a plugin you should not have to care about this. Also, you cannot use the plugin together with `commands`
+or `entrypoint` which will fail. Using `secrets` or `environment` is possible, but in this case, the plugin is internally not treated as plugin
+anymore. The container then cannot access secrets with plugin filter anymore and the containers won't be privileged without explicit definition.
 
 ## Finding Plugins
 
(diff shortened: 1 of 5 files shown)


lafriks:
Can we use settings with non-plugins?

zc-devs:
> Can we use settings with non-plugins?

I vote for `No`.
Settings are for plugins; envs and secrets are prohibited.
Envs (and secrets*) are for general steps; (secrets*) settings are prohibited. * depends on #4015 resolution.

zc-devs:
Title
> Disallow secrets for plugins

> for plugins envs and secrets are prohibited

> This pr changes it, they're no longer prohibited

I'm lost.

zc-devs:
So, now using `settings` with `environment`, `secrets`, `commands`, `entrypoint` in plugins would rise warnings instead of errors and won't fail pipeline anymore, right? It is like deprecation? The reason behind this is we don't want to introduce breaking changes in pipeline syntax in upcoming `3.0`, right?

If `yes`, I would rename to `Deprecate general step syntax in plugins` with according record in the docs.

zc-devs:
I would switch this PR to `release/v2.7` branch.

And for `3.0` I would keep errors and disallow secrets (hard-way with pipeline fail). Although it breaks pipeline syntax, security should get priority (definitely for major releases, for minor and patch versions it is an open question). Otherwise we would live with `not fixed security issue yet` badge for a year.

lafriks:
> If you set `commands`, `environment` etc. on a plugin, the step is not treated as plugin internally anymore and will therefore not be given additional privileges without explicit definition (which requires the repo to be trusted). It will just be like any other step, so there isn't any security problem here.

Also secrets that are allowed only for specified images won't be allowed?

zc-devs:
I understand that it works, and why. Thank you for explanation. 

> The reason behind this (changing to warnings) is we don't want to introduce breaking changes in pipeline syntax in upcoming 3.0, right?

Is this temporary solution? Do you try to trick the system?

zc-devs:
I like fail-fast, strict approach, which was before.

zc-devs:
I feel I make too much efforts to defend something, that I do not really use. It makes me tired. Probably, you too.

> Failing at commands and entrypoint is probably better 

I think so.

In the end, I just wanna highlight another thing. It is just to think (or not :) about, no need to reply.
https://github.com/woodpecker-ci/woodpecker/discussions/3959#discussion-6964412
Did the example work as plugin? I assume, yes.
Now, does it execut as plugin? I assume, no. Can it lead to some plugin unexpected behavior from a user point of view. How `janydoe` should have known about it? Can it lead to endless questions / bug reports?
That is how magic works.

6543 (approved):

lafriks (approved):

PR #4028, opened 2024-08-11
Updated Secrets image filter docs

zc-devs:
Renamed secrets image->plugins filter, improved description


zc-devs:
> Does webkate import these changes too?

I know nothing about weblate.

> And can you make clear that a secret with plugin filter will not be available to regular steps? (Just generalize the commands part)

Where, in docs or UI?
Docs mentions plugins. After that general steps. 
UI contains `only` and also mentions general step in description.
Could you suggest changes?

zc-devs:
@qwerty287

> just revert the changes to the locales

Done.

PR #4044, opened 2024-08-16
move arbitrary number into const

6543:


server/grpc/rpc.go
```diff
@@ -511,7 +513,7 @@ func (s *RPC) getHostnameFromContext(ctx context.Context) (string, error) {
 
 func (s *RPC) updateAgentLastWork(agent *model.Agent) error {
 	// only update agent.LastWork if not done recently
-	if time.Unix(agent.LastWork, 0).Add(1 * time.Minute).Before(time.Now()) {
+	if time.Unix(agent.LastWork, 0).Add(updateAgentLastWorkDelay).Before(time.Now()) {
```
anbraten, line 516: Why though?

6543: better to maintain

PR #4045, opened 2024-08-16
Process workflow logs in batches

### LINKED ISSUE #3999: Inefficient handling of verbose log output
### Component

server, agent

### Describe the bug

Hello. I recently migrated some CI jobs from Drone and noticed their running time increasing from 4 seconds to about two minutes.

The reason seems to be rather inefficient handling of long job output. Moving it from PostgreSQL into JSON files (`WOODPECKER_LOG_STORE`) improved running time by ~10%, but it's still much slower than it should be.

The agent is running on an adjacent VM, there's pretty much no network latency between the server and the agent.

Disabling detailed logging reduced their running time to the same few seconds. The actual jobs run `rsync --itemize-changes` on large PHP projects (~15-20k files/ea, one line of output for each file) or running tests with very detailed logging, but it's easier to reproduce with a synthetic benchmark:

```yaml
# woodpecker
when:
  event: push

steps:
  log:
    image: alpine:3.20
    commands:
      - seq 0 50000
```

```yaml
# drone
kind: pipeline
type: docker
name: test

trigger:
  event: push

steps:
  - name: log
    image: alpine:3.20
    commands:
      - seq 0 50000
```

On my CI server, Drone runs this in less than two seconds, Woodpecker finishes in just under *five minutes*. Both are using the same VM and the same instance of Docker.

As Woodpecker is much nicer overall and is actually FOSS, I have no desire to go back. Any way we can fix this?

I don't write Go daily, so instead of profiling properly I decided to simply [check] if the bottleneck is in the storage procedure. While it helped with a synthetic benchmark:

```
BenchmarkLogAppend-16  269067   4411 ns/op   // before
BenchmarkLogAppend-16  2715338  442.3 ns/op  // after
```

the more realistic test described previously was sped up by only a few seconds.

Any thoughts? Maybe I'm doing something very wrong and missed some knob in the docs?

There was a similar problem in #2064 and it's marked as "fixed".

[check]: https://github.com/hg/woodpecker/commit/d256f2ab752978340dd1145081c4ee13c23ab9f1

---

server:

```yaml
  woodpecker:
    image: woodpeckerci/woodpecker-server
    restart: always
    environment:
      - WOODPECKER_OPEN=true
      - WOODPECKER_HOST=https://woodpecker.example
      - WOODPECKER_GITEA=true
      - WOODPECKER_GITEA_URL=https://gitea.example
      - WOODPECKER_GITEA_CLIENT=xxx
      - WOODPECKER_GITEA_SECRET=xxx
      - WOODPECKER_AGENT_SECRET=xxx
      - WOODPECKER_DATABASE_DRIVER=postgres
      - WOODPECKER_DATABASE_DATASOURCE=postgres://woodpecker:xxx@db:5432/woodpecker?sslmode=disable
      - WOODPECKER_LOG_STORE=file
      - WOODPECKER_LOG_STORE_FILE_PATH=/var/log/ci-logs
      - WOODPECKER_ADMIN=xxx
      - WOODPECKER_GRPC_SECRET=xxx
    networks:
      - woodpecker
      - db
    volumes:
      - ./data/woodpecker/log:/var/log/ci-logs
      - ./data/woodpecker/data:/var/lib/woodpecker
    ports:
      - 10.24.0.1:9000:9000
```

agent:

```bash
docker run --detach \
  --restart=always \
  --volume=/var/run/docker.sock:/var/run/docker.sock \
  --env=WOODPECKER_SERVER=10.24.0.1:9000 \
  --env=WOODPECKER_AGENT_SECRET=xxx \
  --name=woodpecker \
  --volume=woodpecker-agent-config:/etc/woodpecker \
  woodpeckerci/woodpecker-agent
```

### Steps to reproduce

See repro above.

### Expected behavior

Simple jobs with tens of thousands of lines of output should finish in 1-2 seconds.

### System Info

```shell
- 2.7.0, the latest `next`, and built from trunk
- both agent and server are running under Docker 20.10.24
- Debian 12 amd64
```

### Additional context

_No response_

### Validations

hg:
Hopefully this removes the ~~first~~ bottleneck.

~~The benchmark mentioned in the issue now finishes in 2 seconds because its output fits into one batch. Anything that spills over into more batches hangs waiting for the server to write out the previous ones.~~

edit: original description, not relevant anymore.

Closes #3999
Ref: #2072
( Close https://github.com/woodpecker-ci/woodpecker/issues/2064 )


6543:
in itselfe looks god, we just a:
- make sure on graceful stop the last logs are commited
- line numbers are still working correctly

6543:
yes drone let the agent stream the log and use that stream just for live logs and then at the end send the whole log again as blob to be stored. We want to have a single stream to both store and view the logs at the same time ... as it ensures that what you see is what is actually happening and later on you still can ensure that all seen logs are stored also the same.

In the end your pull finish what i tested and then had in mind at #2072 but never go to implement it ...

6543:
well it does not enhance the storing logs part... but that's as you wrote up for the next pull ...

server/api/stream.go
```diff
@@ -213,17 +214,23 @@ func LogStreamSSE(c *gin.Context) {
...
 	go func() {
-		err := server.Config.Services.Logs.Tail(ctx, step.ID, func(entries ...*model.LogEntry) {
-			for _, entry := range entries {
-				select {
-				case <-ctx.Done():
-					return
-				default:
-					ee, _ := json.Marshal(entry)
-					logChan <- ee
+		batches := make(logging.LogChan, 100)
```
zc-devs, line 217: Is Server limited by 100 workflow execution simultaneously? Why 100? Should it be a constant?

server/logging/log.go
```diff
@@ -92,15 +93,20 @@ func (l *log) Write(ctx context.Context, stepID int64, logEntry *model.LogEntry)
...
 
 	s.Lock()
-	s.list = append(s.list, logEntry)
+	s.list = append(s.list, entries...)
 	for sub := range s.subs {
-		go sub.handler(logEntry)
+		select {
+		case sub.receiver <- entries:
+		default:
+			logger.Info().Msgf("subscriber channel is full -- dropping logs for step %d", stepID)
```
zc-devs, line 101: Shouldn't we wait for channel be empty?

server/services/log/file/file.go
```diff
@@ -61,19 +62,30 @@ func (l logStore) LogFind(step *model.Step) ([]*model.LogEntry, error) {
...
 	}
-	_, err = file.Write(append(jsonData, byte('\n')))
-	if err != nil {
-		return err
+
+	if _, err = file.Write(bytes); err != nil {
+		logger.Error().Err(err).Msg("could not write out log entries")
 	}
+
 	return file.Close()
```
zc-devs, line 89: Defer?

6543:
if you need help to resolve the conflict just ask :)

server/grpc/rpc.go
```diff
@@ -368,7 +353,34 @@ func (s *RPC) Log(c context.Context, rpcLogEntry *rpc.LogEntry) error {
 		return err
 	}
 
-	return server.Config.Services.LogStore.LogAppend(logEntry)
+	var logEntries []*model.LogEntry
+
+	for _, rpcLogEntry := range rpcLogEntries {
+		if rpcLogEntry.StepUUID != stepUUID {
+			panic(fmt.Sprintf("expected step UUID %s, got %s", stepUUID, rpcLogEntry.StepUUID))
```
6543, line 360: please not panic

server/logging/log_test.go
```diff
@@ -39,28 +39,37 @@ func TestLogging(t *testing.T) {
 		context.Background(),
 	)
 
+	receiver := make(LogChan, 10)
+
+	go func() {
+		for {
+			<-receiver
```
6543, line 46: that looks like there is no proper cleanup ...

6543:
beside this two nits it looks good to go :tada:

server/logging/log.go
6543: well should be fine as the data is still in DB and can be retrieved again if needed

6543 (approved):

6543:
thanks for the awesome work!!!

6543:
In this case you might DM me via matrix https://matrix.to/#/@marddl:obermui.de if you find a timeslot, so i can onbord u.

If you dont have an account, just use the matrix.org to create one ...

6543:
-> https://github.com/woodpecker-ci/woodpecker/pull/4356

PR #4048, opened 2024-08-18
Read long log lines from file storage correctly

hg:
Log lines may contain up to 1 MB of text, but with default settings `bufio.Scanner` only reads lines up to 64k in length and fails on anything larger, rendering the `Scanner` unusable.

As a result, the user won't see anything after the first very long line, including the line itself.

```yaml
when:
  event: push

steps:
  logs:
    image: ubuntu:24.04
    commands:
      - seq 1 3
      - tr -dc '[:alnum:]' </dev/urandom | head -c70k
      - echo
      - seq 4 6
```

The output should be:

```
1
2
3
zkwmLrpYmk6XcgJZiYqzNO3hsIV26HJfVQfyvSo1ccrMqP ... very long line of random text
4
5
6
```

Instead, Woodpecker cuts everything after `3`:

```
1
2
3
(end)
```

I tested this on jobs with up to 20 MBs of output (simply update `70k` with something like `20M`).

### DIFF
--- a/cli/exec/exec.go
+++ b/cli/exec/exec.go
@@ -281,8 +281,7 @@ func convertPathForWindows(path string) string {
 	return filepath.ToSlash(path)
 }
 
-const maxLogLineLength = 1024 * 1024 // 1mb
 var defaultLogger = pipeline.Logger(func(step *backend_types.Step, rc io.ReadCloser) error {
 	logWriter := NewLineWriter(step.Name, step.UUID)
-	return pipelineLog.CopyLineByLine(logWriter, rc, maxLogLineLength)
+	return pipelineLog.CopyLineByLine(logWriter, rc, pipeline.MaxLogLineLength)
 })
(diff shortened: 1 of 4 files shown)


cli/exec/exec.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/urfave/cli/v3"
 
+	"go.woodpecker-ci.org/woodpecker/v2/agent"
```
6543, line 31: if you need to import the agent then the const is in the wrong place and should be moved elsewhere ...!

I suggest [go.woodpecker-ci.org/woodpecker/v2/pipeline](https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/const.go)

6543:
please dont forcepush if you can! it makes reviews harder and we squashmerge anyway ...

lafriks (approved):

6543:
https://ci.woodpecker-ci.org/repos/3780/pipeline/19786/24 ... well

6543:
thanks for not running at the dedicated pull :/

6543:
If you d'like to create a pull?

PR #4053, opened 2024-08-25
Remove all default privileged plugins

6543:
re-add plugins to the list via config if needed

close #3918

### DIFF
--- a/docs/docs/91-migrations.md
+++ b/docs/docs/91-migrations.md
@@ -4,6 +4,7 @@ Some versions need some changes to the server configuration or the pipeline conf
 
 ## `next`
 
+- Remove all default privileged plugins ([re-add plugins to the list via config if needed](./30-administration/10-server-config.md#woodpecker_escalate)).
 - Removed `WOODPECKER_DEV_OAUTH_HOST` and `WOODPECKER_DEV_GITEA_OAUTH_URL` use `WOODPECKER_EXPERT_FORGE_OAUTH_HOST`
 - Compatibility mode of deprecated `pipeline:`, `platform:` and `branches:` pipeline config options are now removed and pipeline will now fail if still in use.
 - Removed `steps.[name].group` in favor of `steps.[name].depends_on` (see [workflow syntax](./20-usage/20-workflow-syntax.md#depends_on) to learn how to set dependencies)
--- a/pipeline/frontend/yaml/utils/image_test.go
+++ b/pipeline/frontend/yaml/utils/image_test.go
@@ -199,6 +199,11 @@ func Test_matchImage(t *testing.T) {
 			to:   "gcr.io/golang",
 			want: false,
 		},
+		{
+			from: "woodpeckerci/plugin-docker-buildx",
+			to:   "docker.io/woodpeckerci/plugin-docker-buildx",
+			want: true,
+		},
 	}
 	for _, test := range testdata {
 		assert.Equal(t, test.want, MatchImage(test.from, test.to))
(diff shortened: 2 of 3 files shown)


6543:
I'm personally in fafour of https://github.com/woodpecker-ci/woodpecker/pull/3918

6543:
#4056 in combination with https://github.com/woodpecker-ci/woodpecker/pull/3918 should be good enouth in my poinion ...

pat-s (commented):
For the overall issue: I don't see why we should keep only our plugins and I am in favor of removing all from the list.

Besides, similar to my comment in #3918, I think that we are lacking documentation about this change and it should be included in the PR that will be chosen in the end.
As this happened a few times in the past for user-visible changes, I would be really happy to see this being addressed this time.

Last, I think that `WOODPECKER_ESCALATED` is a strange var name as it doesn't really indicate what "exactly" the setting is doing. Yes, it "escalates" something but that word per se has multiple meanings and is often rather used in a different context.
Why not something more specific like `WOODPECKER_ALLOW_PRIVILEGED` or `WOODPECKER_PLUGINS_PRIVILEGED`?

6543:
created https://github.com/woodpecker-ci/infrastructure/pull/91

cmd/server/flags.go
```diff
@@ -159,10 +159,10 @@ var flags = append([]cli.Flag{
 		Value:   time.Hour * 72,
 	},
 	&cli.StringSliceFlag{
-		Sources: cli.EnvVars("WOODPECKER_ESCALATE"),
-		Name:    "escalate",
+		Sources: cli.EnvVars("WOODPECKER_PLUGINS_PRIVILEGED", "WOODPECKER_ESCALATE"),
```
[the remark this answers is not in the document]
6543: https://github.com/woodpecker-ci/woodpecker/pull/4053/commits/dc78c9161666bbe70e3e17dd7dd353d7d580835a

cli/exec/flags.go
```diff
@@ -58,9 +56,10 @@ var flags = []cli.Flag{
 		Hidden:  true,
 	},
 	&cli.StringSliceFlag{
-		Name:  "privileged",
-		Usage: "privileged plugins",
-		Value: constant.PrivilegedPlugins,
+		Sources: cli.EnvVars("WOODPECKER_PLUGINS_PRIVILEGED"),
+		Name:    "plugins-privileged",
+		Aliases: []string{"privileged"}, // TODO: remove in next version
```
6543, line 61:
```suggestion
```

docs/docs/91-migrations.md
```diff
@@ -4,9 +4,10 @@ Some versions need some changes to the server configuration or the pipeline conf
 
 ## `next`
 
+- Rename server environment variable `WOODPECKER_ESCALATE` to `WOODPECKER_PLUGINS_PRIVILEGED`
+- Remove all default privileged plugins ([re-add plugins to the list via config if needed](./30-administration/10-server-config.md#woodpecker_plugins_privileged)).
```
anbraten, line 8: This is confusing as they should not be removed by the admin. Please adjust accordingly

6543: ? what exactly do you mean?

anbraten: The migrations list is containing steps the admin is supposed to do.  ` Remove all plugins` doesn't make sense as we already removed all plugins. What you probably want to say is something like `All privileged plugins were removed. Add those like buildx which are required by your setup`. In general provide some more details for such a breaking change.

6543: Uh sure >D

6543: -> https://github.com/woodpecker-ci/woodpecker/pull/4088

anbraten: thanks

PR #4056, opened 2024-08-26
By default enforce check of clone image via allow-list by image name and tag

6543:
we currently allow all tags for our clone image. this change it to a an default allow list  we maintain.


shared/constant/constant.go
```diff
@@ -38,6 +43,41 @@ const (
 )
 
 var TrustedCloneImages = []string{
```
6543, line 45: we should also lint and create a warning so users will be notifyed if the clone step image is not trusted ...

6543: 5dfed307f6a2708552ae11018029c013d7dbd7a6

pipeline/frontend/yaml/linter/linter.go
```diff
@@ -94,6 +100,23 @@ func (l *Linter) lintFile(config *WorkflowConfig) error {
...
 }
 
+func (l *Linter) lintCloneSteps(config *WorkflowConfig) error {
+	if len(config.Workflow.Clone.ContainerList) == 0 {
+		return nil
+	}
+	var linterErr error
+	for _, container := range config.Workflow.Clone.ContainerList {
+		if !utils.MatchImageExact(container.Image, constant.TrustedCloneImages...) {
+			linterErr = multierr.Append(linterErr,
```
zc-devs, line 110: Suppose, `woodpeckerci/plugin-git:2.5.3` has been released after WP `3.0` release. Will I be able to use it in my pipeline `clone` step?

6543: that's the whole point ... it will just print out an warning if you dont use the allow-listed images, that you wont get the credentials injected for cloning the repo

6543: if you dont specify the clone step, this message should never apear or we have a serious bug in the default clone step generation. it just informs users who define an own clone step

zc-devs: > you wont get the credentials injected for cloning the repo

So, won't work, right? But it's just new version of official plugin.

> it will just print out an warning

Not related to this PR, but it would be great, if there were two constructors for lint issues: `newLinterWarning` and `newLinterError`.

shared/constant/constant.go
```diff
@@ -19,8 +19,13 @@ var PrivilegedPlugins = []string{
 	"plugins/docker",
 	"plugins/gcr",
 	"plugins/ecr",
-	"woodpeckerci/plugin-docker-buildx",
-	"codeberg.org/woodpecker-plugins/docker-buildx",
+	// latest buildx
+	"docker.io/woodpeckerci/plugin-docker-buildx:latest",
+	"codeberg.org/woodpecker-plugins/docker-buildx:latest",
+	// renovate: datasource=docker depName=docker.io/woodpeckerci/plugin-docker-buildx
+	"docker.io/woodpeckerci/plugin-docker-buildx:4.2.0",
```
lafriks, line 26: It will be crazy to maintain and you will have to either make this very long or you will be forced to update all pipelines to new versions, or am I missing something? 🤔

pipeline/frontend/yaml/linter/linter.go
lafriks: I think this will be crazy to maintain and real pain for users who will want to use latest version (because renovate will update it for example) but will have to ask admin all the time to add it to the list (that would require server restart that might be not easy in working time)

6543: the `latest` tag and `next` tag are also valid and allowd by default

> So, won't work, right? But it's just new version of official plugin.
I dont get it ... it woll work if you use the ''official plugin" in any case ...

shared/constant/constant.go
6543: yes at some point we would force to update the plugin version if a user did pin it to an old one ...
... or he has to explicit allow it

pipeline/frontend/yaml/linter/linter.go
zc-devs: I've just tested this:
```go

{
	from: "{steps: { build: { image: golang, settings: { test: 'true' } } }, when: { branch: main, event: push }, clone: { git: { image: woodpeckerci/plugin-git:2.5.3 } } }",
	want: "Specified clone image does not match allow list, netrc will not be injected",
},
```
and it passed.

Suppose, I use woodpecker version `3.0`. New versions of clone plugin released: `2.5.3` and `2.6.0`. I want to use `2.5.3` in my pipeline.
`2.5.2` (in WP `3.0`) - > `2.5.3` (will work?) -> `2.6.0` (aka `latest`)
In other words, would clone plugin work, if I use versions between the last hard-coded in WP (`2.5.2`) and the last real version (`latest=2.6.0`)?

6543: yes 

thats the list: https://github.com/woodpecker-ci/woodpecker/pull/4056/files#diff-0f5e5e5da97281ce5b634780f49f2b7783809e6780f517311aa5a7d1b420af3dR50

6543:
broke out the least controversial part:  #4069

6543:
it should be enforced in critical stuff ... else users just shoot themself in the foot

docs/docs/30-administration/10-server-config.md
```diff
@@ -351,6 +351,7 @@ a user can log into Woodpecker, without re-authentication.
 > Defaults are defined in [shared/constant/constant.go](https://github.com/woodpecker-ci/woodpecker/blob/main/shared/constant/constant.go)
 
 Docker images to run in privileged mode. Only change if you are sure what you do!
+You should specify the tag as images are compared via exact matches.
```
6543, line 354:
```suggestion
You should specify the tag of your images too, as this enforces exact matches.
```

pipeline/frontend/yaml/compiler/convert.go
```diff
@@ -138,7 +138,7 @@ func (c *Compiler) createProcess(container *yaml_types.Container, stepType backe
 		environment[requested] = secretValue
 	}
 
-	if utils.MatchImage(container.Image, c.escalated...) && container.IsPlugin() {
+	if utils.MatchImageExact(container.Image, c.escalated...) && container.IsPlugin() {
```
6543, line 141:
```suggestion
	if utils.MatchImageDynamic(container.Image, c.escalated...) && container.IsPlugin() {
```

pipeline/frontend/yaml/linter/linter.go
```diff
@@ -96,6 +101,23 @@ func (l *Linter) lintFile(config *WorkflowConfig) error {
 	return linterErr
 }
 
+func (l *Linter) lintCloneSteps(config *WorkflowConfig) error {
+	if len(config.Workflow.Clone.ContainerList) == 0 {
+		return nil
+	}
+	var linterErr error
+	for _, container := range config.Workflow.Clone.ContainerList {
+		if !utils.MatchImageExact(container.Image, constant.TrustedCloneImages...) {
```
6543, line 110:
```suggestion
		if !utils.MatchImageDynamic(container.Image, constant.TrustedCloneImages...) {
```

pipeline/frontend/yaml/types/container.go
```diff
@@ -126,5 +126,5 @@ func (c *Container) IsPlugin() bool {
 }
 
 func (c *Container) IsTrustedCloneImage() bool {
-	return c.IsPlugin() && utils.MatchImage(c.Image, constant.TrustedCloneImages...)
+	return c.IsPlugin() && utils.MatchImageExact(c.Image, constant.TrustedCloneImages...)
```
6543, line 129:
```suggestion
	return c.IsPlugin() && utils.MatchImageDynamic(c.Image, constant.TrustedCloneImages...)
```

lafriks:
If image is specified without tag in allow list will it permit all tags?

6543:
yes now it does - changed docs acordingly

lafriks (approved):

lafriks (changes requested):
I was thinking about this and I don't really see how this will improve security but for sure it will add administration burden.

If you trust the owner of the plugin it really does not matter for versions/tags because unless you tag it to specific docker image hash tag actually does not mean much. If the owner goes rogue or it's credentials are stolen, attacker can easily push over the malicious plugin version with the same tag that is already in the trusted list and it does not really help to have specific versions listed

lafriks:
Same actually goes to privileged plugins

lafriks:
Then this is not the right way to approach this. How many vulnerable versions do we have for the git plugin?

Setting trusted plugins is about that plugin is allowed to access secrets and we trust that plugin won't reveal them and won't do anything malicious. Vulnerabilities must be dealt with other tools or methods

6543:
This is here to prevent posible downgrade atacks

6543:
How would you else prevent them?

lafriks:
> This is here to prevent posible downgrade atacks

Can you elaborate on what you mean by downgrade attacks in this context (as usually these mean when an attacker forces the user to use a less secure communication protocol to intercept data)

6543:
well lets say you are an admin and do allow `plugin/docker` because you need it. but you know that version <= 1.0.0 is vounerable to exec code based on wrong settings, now anybody (with push or pull-creation access) can just specify `plugin/docker:1.0.0` and the admin has no option at all to only allow `plugin/docker:2.0.0` or ``plugin/docker:lateset` etc ...

so for the admin it's a binary joice atm either allow all or non

same goes for our clone plugin.

we can only disable clone and let all admins decide if they wana have clone feature and dont have checks or let the user define the clone step and inject the secrets (and also specify good filters etc ...)

but at this point i guess woodpecker is just broken...

the other option is to have a "save" repo where we only have trusted images and delete them asap we know otherwhise.
but deleting old images was also abhorrent.

so as for our project we now have three choises: 1. remove all and tell all admins good luck now you it's your vault (with no solution how to solve it properly); 2. delete old images we know are vournerable; 3. allow filtering by tags (and do so for our plugins providing core features)

6543:
@lafriks if we want the one or the other as **DEFAULT** ... we can discust that here.

for the general concept to be able to filter based on tags, i split out a new pull  #4074

witch also make it possible to admins to alter trustedClonePlugins as they wish

6543:
@lafriks I split out ... again the parts that are non breaking and realy usefull:  #4074 +  #4075

shared/constant/constant.go
```diff
@@ -35,8 +35,42 @@ const (
 )
 
 // TrustedClonePlugins can be changed by 'WOODPECKER_PLUGINS_TRUSTED_CLONE' at runtime.
-var TrustedClonePlugins = []string{
+var TrustedCloneImages = []string{
```
6543, line 38:
```suggestion
var TrustedClonePlugins = []string{
```

lafriks:
Personally I think that allow-list is not the way to go, they are takes a lot of resources and are hard to maintain for every instance.

The best solution for this imho would be to have centrally managed vulnerable plugin list that server would download at specific intervals and would trigger warnings at pipelines and allow woodpecker instance admins to block specific versions. This could be done not only for clone/trusted plugins but other plugins as well.

6543:
that's an option we could go for ...

lafriks:
Also we could have option that would automatically block vulnerable plugin versions if admin chooses to go this way

6543:
as to my knowledge our clone plugin has no issues ... we can take our time to create a proper solution ...

... the best case would be to be able to have something like a CVE list just for containers ... but I dont know of none

6543:
or also if you could specify container images if you file CVEs ... I wonder if I could propose some additional flag to the CVE list to specify ''artifacts'' or ''deply methodes'' so tools can check against the usge of them ...

in this way we could then just file normal CVEs for plugins and use the official list to automatically create our own list, else it's also quite some maintinance burden on our side anyway ...

lafriks:
I don't think the CVE list is the way to go and I don't think maintaining a list for us would be that hard as currently there haven't even been situations where that is the case and we could add them on issue/PR bases

6543:
ok we can close this in fafour of https://github.com/woodpecker-ci/woodpecker/issues/4080

PR #4066, opened 2024-08-29
Fix schema with detached steps

xtexChooser:
`commands` is optional, when `detach` is set.

Also fixes #4065, which made `commands` required (it can be unset when `entrypoint` is available) and therefore some schema tests fail.

For a step like:
```yaml
  detached:
    image: redis
    detach: true
```

It matches `commands_step` and `plugin_step` at the same time, so change oneOf to anyOf.


anbraten:
Could you add some tests.

xtexChooser:
There are already tests in `test-step.yaml` and `test-dag.yaml`, which are enough for these fixes.

PR #4071, opened 2024-08-31
Shortened status messages

zc-devs:
```
WOODPECKER_STATUS_CONTEXT: ci
```

new-status


server/forge/common/status.go
```diff
@@ -58,21 +59,21 @@ func GetPipelineStatusContext(repo *model.Repo, pipeline *model.Pipeline, workfl
 func GetPipelineStatusDescription(status model.StatusValue) string {
 	switch status {
 	case model.StatusPending:
-		return "Pipeline is pending"
```
6543, line 61: this is about the description please revert it back!

if it is used somewhere, and that is to long, the StatusValue should be used directly

server/forge/common/status.go
```diff
@@ -38,12 +38,13 @@ func GetPipelineStatusContext(repo *model.Repo, pipeline *model.Pipeline, workfl
...
-		"context":  server.Config.Server.StatusContext,
-		"event":    event,
-		"workflow": workflow.Name,
-		"owner":    repo.Owner,
-		"repo":     repo.Name,
-		"axis_id":  workflow.AxisID,
+		"context":        server.Config.Server.StatusContext,
+		"event":          event,
+		"workflow":       workflow.Name,
+		"workflow_count": len(pipeline.Workflows),
```
6543, line 44: this is the parsed pipeline ... we must make sure that even if workflows are filtered out later, they still count

server/forge/common/status.go
zc-devs: What else could be here?
[Screenshot 2024-08-31-1](https://github.com/user-attachments/assets/c097de93-6f2a-4c0f-878d-070168a8a4d5)
[Screenshot 2024-08-31-2](https://github.com/user-attachments/assets/36790461-33b7-4b68-911e-eaab6a4f31f8)
[Screenshot 2024-08-31-3](https://github.com/user-attachments/assets/72d97e09-163b-4ff5-ad2f-aa41fb1199c0)

server/forge/common/status.go
zc-devs: If this is the parsed pipeline and if workflows are filtered out _later_, then they still count here, no?

server/forge/common/status.go
zc-devs: Pipeline or workflow, BTW? :)

6543: ah well yes it should be renamed to workflow :D

server/forge/common/status.go
6543: i dont remember that's what to find out ...

zc-devs: Is there a simple solution to get that info?

zc-devs: I think it's not worth it

zc-devs:
I think, there is no much value without shortened status (the first requested change) and it's not worth to go further with the workflow count (second requested change).
So, closing it for now.

PR #4087, opened 2024-09-05
Add indices to repo table

6543:
make some query's faster

### DIFF
--- a/server/model/repo.go
+++ b/server/model/repo.go
@@ -23,13 +23,13 @@ import (
 // Repo represents a repository.
 type Repo struct {
 	ID      int64 `json:"id,omitempty"                    xorm:"pk autoincr 'id'"`
-	UserID  int64 `json:"-"                               xorm:"user_id"`
+	UserID  int64 `json:"-"                               xorm:"INDEX 'user_id'"`
 	ForgeID int64 `json:"forge_id,omitempty"              xorm:"forge_id"`
 	// ForgeRemoteID is the unique identifier for the repository on the forge.
 	ForgeRemoteID                ForgeRemoteID  `json:"forge_remote_id"                 xorm:"forge_remote_id"`
-	OrgID                        int64          `json:"org_id"                          xorm:"org_id"`
-	Owner                        string         `json:"owner"                           xorm:"UNIQUE(name) 'owner'"`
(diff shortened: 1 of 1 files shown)


server/model/repo.go
```diff
@@ -23,13 +23,13 @@ import (
...
-	UserID  int64 `json:"-"                               xorm:"user_id"`
+	UserID  int64 `json:"-"                               xorm:"INDEX 'user_id'"`
 	ForgeID int64 `json:"forge_id,omitempty"              xorm:"forge_id"`
 	// ForgeRemoteID is the unique identifier for the repository on the forge.
 	ForgeRemoteID                ForgeRemoteID  `json:"forge_remote_id"                 xorm:"forge_remote_id"`
-	OrgID                        int64          `json:"org_id"                          xorm:"org_id"`
-	Owner                        string         `json:"owner"                           xorm:"UNIQUE(name) 'owner'"`
-	Name                         string         `json:"name"                            xorm:"UNIQUE(name) 'name'"`
+	OrgID                        int64          `json:"org_id"                          xorm:"INDEX 'org_id'"`
+	Owner                        string         `json:"owner"                           xorm:"INDEX UNIQUE(name) 'owner'"`
```
lafriks, line 31: Unique already means index, no need to create separate index for it

lafriks (changes requested):

server/model/repo.go
6543: yes but it creates an index in combination with name and owner ...

through that's the least usefull indices here ...

lafriks: Even in combination at least when filtering just by the first column of the combined index it will use it anyway. If filtering by only the second column it will not, so at least one of these is still duplicate. Are there cases where it's needed?

server/model/repo.go
```diff
@@ -23,13 +23,13 @@ import (
...
+	Name                         string         `json:"name"                            xorm:"INDEX UNIQUE(name) 'name'"`
```
6543, line 32:
```suggestion
	Owner                        string         `json:"owner"                           xorm:"UNIQUE(name) 'owner'"`
	Name                         string         `json:"name"                            xorm:"UNIQUE(name) 'name'"`
```

lafriks (approved):

PR #4092, opened 2024-09-07
Implement registries for Kubernetes backend

### LINKED ISSUE #2987: Kubernetes backend references non-existing pull secret: regcred
### Component

agent

### Describe the bug

Repository has registry secret for myregistry.tld
Step in workflow references myregistry.tld/path/my-plugin:latest
Agent creates build pod, but references non-existing pull-secret "regcred":

```yaml
apiVersion: v1
kind: Pod
#...
spec:
  containers:
  - #...
    image: myregistry.tld/path/my-plugin:latest
    imagePullPolicy: Always
    name: wp-...
  #...
  imagePullSecrets:
  - name: regcred
```

### System Info

```shell
docker.io/woodpeckerci/woodpecker-server:next (sha256:391eb21a843bf7e38bb4bfd5a529452e31733d90a494c6d3e783f8f1ac3630b3)
docker.io/woodpeckerci/woodpecker-agent:next (sha256:60a04e5b59bbec66bf5a61fc15fcbd8f4b44693602893d10e6f74270f21d2915)
kubernetes 1.28.2
```

### Additional context

_No response_

### Validations

meln5674:
According to [the documentation](https://woodpecker-ci.org/docs/administration/backends/kubernetes#images-from-private-registries), per-organization and per-pipeline registries are currently unsupported for the Kubernetes backend.

This patch implements this missing functionality by creating and deleting a matching secret for each pod with a matched registry, using the same name, labels, and annotations as the pod, and appending it to its `imagePullSecrets` list.

This patch adds tests for the new functionality, and has been manually end-to-end-tested in KinD by using a private image hosted in the matching gitea instance.

This will require updating the matching helm charts to add the create/delete permissions to the agent role, which **is already done**.

close  #2987


6543:
Please also update the docs :)

pipeline/backend/kubernetes/image_pull_secret.go
```diff
@@ -0,0 +1,57 @@
...
+	"go.woodpecker-ci.org/woodpecker/v2/pipeline/frontend/yaml/utils"
+	v1 "k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func needsImagePullSecret(step *types.Step) bool {
+	return step.AuthConfig.Username != "" && step.AuthConfig.Password != ""
+}
+
+func mkImagePullSecret(step *types.Step, config *config, podName, goos string, options BackendOptions) (*v1.Secret, error) {
```
zc-devs, line 19: Unused `goos`.

pipeline/backend/kubernetes/image_pull_secret.go
```diff
@@ -0,0 +1,57 @@
...
+	labels, err := podLabels(step, config, options)
+	if err != nil {
+		return nil, err
+	}
+	annotations := podAnnotations(config, options)
```
zc-devs, line 24: Why do you use _Pod_ annotations on secret?

pipeline/backend/kubernetes/image_pull_secret_test.go
```diff
@@ -0,0 +1,97 @@
...
+		"metadata": {
+			"name": "wp-01he8bebctabr3kgk0qj36d2me-0",
+			"namespace": "woodpecker",
+			"creationTimestamp": null,
+			"labels": {
+				"app": "test",
+				"part-of": "woodpecker-ci",
+				"step": "go-test"
+			},
+			"annotations": {
```
zc-devs, line 45: That's what I'm talking about. They are Pod specific.

pipeline/backend/kubernetes/image_pull_secret_test.go
```diff
@@ -0,0 +1,97 @@
...
+				"kubernetes.io/limit-ranger": "LimitRanger plugin set: cpu, memory request and limit for container"
+			}
+		},
+  	"type": "kubernetes.io/dockerconfigjson",
+	  "data": {
+			".dockerconfigjson": "eyJhdXRocyI6eyJkb2NrZXIuaW8iOnsidXNlcm5hbWUiOiJmb28iLCJwYXNzd29yZCI6ImJhciJ9fX0="
+	  }
+	}`
+
+	pullSecret, err := mkImagePullSecret(&types.Step{
```
zc-devs, line 56: You don't need a half of options to test secrets.

pipeline/backend/kubernetes/pod.go
```diff
@@ -505,6 +514,18 @@ func startPod(ctx context.Context, engine *kube, step *types.Step, options Backe
 		return nil, err
 	}
 
+	if needsImagePullSecret(step) {
+		pullSecret, err := mkImagePullSecret(step, engineConfig, podName, engine.goos, options)
+		if err != nil {
+			return nil, err
+		}
+		log.Trace().Msgf("creating secret: %s", pullSecret.Name)
+		_, err = engine.client.CoreV1().Secrets(engineConfig.Namespace).Create(ctx, pullSecret, meta_v1.CreateOptions{})
```
zc-devs, line 523: Make a function in `image_pull_secrets`.

pipeline/backend/kubernetes/pod.go
```diff
@@ -505,6 +514,18 @@ func startPod(ctx context.Context, engine *kube, step *types.Step, options Backe
 		return nil, err
 	}
 
+	if needsImagePullSecret(step) {
```
zc-devs, line 517: And probably call it in `kubernetes.go`. See how volumes work.

pipeline/backend/kubernetes/pod.go
```diff
@@ -514,12 +535,23 @@ func stopPod(ctx context.Context, engine *kube, step *types.Step, deleteOpts met
 	if err != nil {
 		return err
 	}
+	errs := make([]error, 0, 2)
+	if needsImagePullSecret(step) {
```
zc-devs, line 539: The same as two above.

pipeline/backend/kubernetes/image_pull_secret.go
zc-devs: There is `secrets.go`. Secrets stuff should be there, I think.

zc-devs (commented):
1. Am I correct, that only one image pull secret is supported?
2. There was `registry` field in UI, as I remember. You don't use it, right?

https://woodpecker-ci.org/docs/usage/registries#images-from-private-registries

> Woodpecker matches the registry hostname to each image in your YAML. If the hostnames match, the registry credentials are used to authenticate to your registry and pull the image.

zc-devs:
#2987

pipeline/backend/kubernetes/image_pull_secret.go
zc-devs: Probably, applied to labels also.

meln5674:
> Please also update the docs :)

Can do, I wasn't sure if the docs site was live at HEAD or if there was some separate release process for them.

> 1. Am I correct, that only one image pull secret is supported?
> 2. There was `registry` field in UI, as I remember. You don't use it, right?
> 
> https://woodpecker-ci.org/docs/usage/registries#images-from-private-registries
> 
> > Woodpecker matches the registry hostname to each image in your YAML. If the hostnames match, the registry credentials are used to authenticate to your registry and pull the image.
> 
> Seems, registries are checked on the Server. As you create/delete secret each time, it should work as intended and comments above are invalid. However, I still believe, that creating one secret with all available registries/auths and then "attach" it (via `imagePullSecret`) to needed steps/pods is a better approach.

In Kubernetes? No, any number of secrets is supported, with any number of auths in each one . In woodpecker, I saw that the AuthConfig for a step is a single user:pass pair, which is matched against the list of repo, org, and global registries, so in that sense, a single extra secret is supported. That said, in the future, if sidecars are also supported, it may make sense for this one secret to contain multiple auths. Because you can have as many imagePullSecrets as you want in a pod, it makes much more sense to list the user-managed ones and then the agent-managed ones, instead of reading them from the API server and writing them back out, as the kubelet will do that for you, and retry against each one in sequence in the event of authentication errors.

pipeline/backend/kubernetes/image_pull_secret.go
meln5674: I assumed the user would want to label and annotate the secrets as well, but if you prefer a second pair of env vars for the secrets, I can do that.

pipeline/backend/kubernetes/image_pull_secret.go
meln5674: Removed goos, moved all secrets-related func's to secrets.go

pipeline/backend/kubernetes/image_pull_secret.go
meln5674: Added 4 new env's for labels, annotations, and allow for both for registry secrets. Added separate func's to return labels and annotations specifically for registry secrets.

pipeline/backend/kubernetes/image_pull_secret_test.go
meln5674: Removed all non-relevant fields

pipeline/backend/kubernetes/pod.go
meln5674: Done

pipeline/backend/kubernetes/pod.go
meln5674: Created startRegistrySecret and stopRegistrySecret to mirror startPod/Volume and stopPod/Volume

pipeline/backend/kubernetes/pod.go
meln5674: Done

zc-devs (commented):
Overall, looks very well to me.

---

> Am I correct, that only one image pull secret is supported?
> There was registry field in UI, as I remember. You don't use it, right?

I was confused by flashbacks from #3122. I thought, if registry isn't passed within AuthConfig, then how Kubernetes would choose right secret? Or the only one secret/registry would be supported.
Then I reread through my PR
> Made the secrets pipeline-wide instead of step-wide as in Docker backend

and understood everything :) ⬇️ 

> registries are checked on the Server. As you create/delete secret each time, it should work as intended

 --

> if sidecars are also supported, it may make sense for this one secret to contain multiple auths

Good catch.

> instead of reading them from the API server and writing them back out

Don't get this. Who reads regcreds from API? Kubernetes API? What is the "back"?

> kubelet will do that for you, and retry against each one

That is exactly, what I had in mind, implementing #3122. ⬇️ 

> However, I still believe, that creating one secret with all available registries/auths and then "attach" it (via imagePullSecret) to needed steps/pods is a better approach.

Yeah, maybe even not to "needed" pods, but all of them. And let Kubernetes cope with that.

pipeline/backend/kubernetes/secrets.go
```diff
@@ -189,3 +200,129 @@ func secretReference(name string) v1.LocalObjectReference {
...
+	authConfig := config_file.ConfigFile{
+		AuthConfigs: map[string]config_file_types.AuthConfig{
+			reference.Domain(named): {
+				Username: step.AuthConfig.Username,
+				Password: step.AuthConfig.Password,
+			},
+		},
+	}
+
+	configFileJSON, err := json.Marshal(authConfig)
```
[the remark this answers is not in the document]
meln5674: Yes. If you're thinking of needing to use base64 in secret YAML files, that is accomplished by [the encoding/json package](https://pkg.go.dev/encoding/json#Marshal) "Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string". The serialized JSON form contains a base64 string, but the k8s data structure contains raw bytes.

I've tested prior to each push with a full install into a k8s cluster, both testing that a private image fails without the registry, and succeeds with it, so if it didn't work without base64, that would have caught it. I'm happy to share or commit this test setup script if you'd like.

pipeline/backend/kubernetes/flags.go
```diff
@@ -89,4 +89,28 @@ var Flags = []cli.Flag{
...
+		Name:    "backend-k8s-registry-secret-annotations",
+		Usage:   "backend k8s additional Agent-wide registry image pull secret annotations",
+		Value:   "",
+	},
+	&cli.BoolFlag{
+		Sources: cli.EnvVars("WOODPECKER_BACKEND_K8S_REGISTRY_SECRET_ANNOTATIONS_ALLOW_FROM_STEP"),
+		Name:    "backend-k8s-registry-secret-annotations-allow-from-step",
+		Usage:   "whether to allow using annotations from step's backend options for registry image pull secret",
+		Value:   false,
+	},
```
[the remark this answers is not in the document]
meln5674: Fair enough. I tend to err on the side of flexibility with these sorts of things, but I guess it is in the commit history if we want it back.

Done.

pipeline/backend/kubernetes/secrets.go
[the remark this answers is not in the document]
zc-devs: I meant more `auth` field in the `auths`, than entire `.dockerconfigjson` in the secret.
https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
```json
{"auths":{"your.private.registry.example.com":{"username":"janedoe","password":"xxxxxxxxxxx","email":"jdoe@example.com","auth":"c3R...zE2"}}}
```
> To understand what is in the auth field, convert the base64-encoded data to a readable format:

```
$ echo "c3R...zE2" | base64 --decode
janedoe:xxxxxxxxxxx
```

zc-devs (approved):
As it is tested and works (https://github.com/woodpecker-ci/woodpecker/pull/4092#discussion_r1749313677), no more comments from me.
Thank you for the contribution 🎉

pipeline/backend/kubernetes/secrets.go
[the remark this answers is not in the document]
meln5674: Ah, I see. In k8s, as far as I know, that field isn't actually required, and empirically, it works without it.

6543 (approved):

PR #3967, opened 2024-07-24
Refactor `server/.../step_builder` into `pipeline/.../builder`

anbraten:
Extract the `step_builder` from the server to the pipeline package.
This cleans the interfaces / structure and will allow us to re-use it in the cli to correctly support pipeline execution (things like `depends_on` support).

PR #4102, opened 2024-09-11
Make local backend work with `cli exec`

6543:
Currently if you try to exec a pipeline via the cli tool and use the local backend,  all files are missing.
This will detect this specific case and use the new added option in the local backend to specify a execution directory upfront.

### DIFF
--- a/cli/exec/exec.go
+++ b/cli/exec/exec.go
@@ -111,6 +111,11 @@ func runExec(ctx context.Context, c *cli.Command, file, repoPath string) error {
 		return err
 	}
 
+	// if we use the local backend we should signal to run at $repoPath
+	if c.String("backend-engine") == "local" && !c.IsSet("backend-local-exec-dir") {
+		c.Set("backend-local-exec-dir", repoPath)
+	}
+
 	axes, err := matrix.ParseString(string(dat))
 	if err != nil {
 		return fmt.Errorf("parse matrix fail")
--- a/docs/docs/30-administration/22-backends/20-local.md
+++ b/docs/docs/30-administration/22-backends/20-local.md
@@ -65,3 +65,11 @@ In the context of the local backend, plugins are simply executable binaries, whi
 > Default: default temp directory
 
 Directory to create folders for workflows.
+
+#### `WOODPECKER_BACKEND_LOCAL_EXEC_DIR`
+
+> Default: *empty*
+
+:::warning
+Only set it if you know why
+:::
--- a/pipeline/backend/local/flags.go
+++ b/pipeline/backend/local/flags.go
@@ -24,7 +24,12 @@ var Flags = []cli.Flag{
 	&cli.StringFlag{
 		Name:    "backend-local-temp-dir",
 		Sources: cli.EnvVars("WOODPECKER_BACKEND_LOCAL_TEMP_DIR"),
-		Usage:   "set a different temp dir to clone workflows into",
+		Usage:   "Set a different temp dir to clone workflows into.",
 		Value:   os.TempDir(),
+	},
+	&cli.StringFlag{
+		Name:    "backend-local-exec-dir",
+		Sources: cli.EnvVars("WOODPECKER_BACKEND_LOCAL_EXEC_DIR"),
+		Usage:   "Instead of using an tmp dir, we use this directory to exec the workflow in (mostly used by 'woodpecker-cli exec ...').",
 	},
 }
--- a/pipeline/backend/local/local.go
+++ b/pipeline/backend/local/local.go
@@ -44,6 +44,7 @@ type workflowState struct {
 
 type local struct {
 	tempDir         string
+	execDir         string
 	workflows       sync.Map
 	output          io.ReadCloser
 	pluginGitBinary string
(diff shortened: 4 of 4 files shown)


docs/docs/30-administration/22-backends/20-local.md
```diff
@@ -65,3 +65,11 @@ In the context of the local backend, plugins are simply executable binaries, whi
 > Default: default temp directory
 
 Directory to create folders for workflows.
+
+#### `WOODPECKER_BACKEND_LOCAL_EXEC_DIR`
```
anbraten, line 69: This PR seems pretty redundant to `WOODPECKER_LOCAL`.

anbraten (changes requested):

docs/docs/30-administration/22-backends/20-local.md
6543: Well yes and no ...
... this is the most generic solution i did come up with to fix this.

The other var is not an engine option and only acept bool

6543: An other option would be to alter the backend config specific for this usecase ... which i found to be not as generic as this one

6543:
@anbraten if you know an better way to do:

```
woodpecker-cli.exe exec --backend-engine local --pipeline-event push --commit-branch main
```

to test local backend via cli in windows / linux please tell :)

anbraten:
Will do some testing. Fyi #3967 might be quite interesting for you.

6543:
Yes thus might allow more :)

Should i hide the flag and undocument it so we have it working now and can improve it later?

anbraten:
The usage of the `local` property was somehow not compatible with the local backend. This fixes it for me so I was able to run:

`go run ../cmd/cli/ exec --pipeline-event push --commit-branch main --backend-engine local --local`

https://github.com/woodpecker-ci/woodpecker/compare/main...anbraten:local-path?expand=1

6543:
I looked at your solution, this has 2 problems:
1. the 'repo-path' is now (miss-)used at the local backend even if it never was registered there (the lib we have do handle that but its outside of our convention)
2. artifacts that should not be there like temporary exec files are now stored inside the current repo, witch differs completly on how the backend is executed with server&agent combi
3. the bug mentioned in 2. affects docker backend if exec via cli now too (because of the removed volume)

the only thing i could addopt in my patch from your suggestion would be 1. witch breaks the convention ...
... not sure if this makes maintaining this harder...

6543:
what i could do and did, is to hide that option from the user: 7101cdba3bcfc5fd1c4bf2ab45761752f96efb47

6543:
@anbraten I can not find side effects you had concerns when we talked about this ...

... anyway I renamed the flag so it is clear that it is for internal usage only ...

cli/exec/exec.go
```diff
@@ -130,6 +130,13 @@ func runExec(ctx context.Context, c *cli.Command, file, repoPath string, singleE
 		return err
 	}
 
+	// if we use the local backend we should signal to run at $repoPath
+	if c.String("backend-engine") == "local" {
+		if err := c.Set("internal-backend-local-exec-dir", repoPath); err != nil {
```
xoxys: I don't like such `internal-` flag naming its ugly and pseudo-hidden. I know there were discussions about that but can you summarize whats the reason?

6543: for `woodpecker-cli exec` to be able to run the local backend localy (docker works because it can just mount current dir as volume)

6543: why we need that flag: because i found no better way to let the cli signal the engine backend what to do without breaking the hirrarchy of our architekture ...

... I might come up with another way to do this but until now i did not ...

PS: I also dont like this fix fully but I did not see something better _jet_

xoxys: That we need the flag is clear to me but why the `internal-` prefix? Thats what Im wondering about.

xoxys: Besides that Im fine with this change and would approve it as is.

6543: because `@`anbraten did not like to expose it as flag to users ... even if it is hidden ... so i added internal so it is clear that this is nothing a user should use

xoxys: mhm... Then let's leave it to @anbraten I don't like such pseudo-hidden flag names...

6543: me neither ... ideas please :D

6543:
ok @xoxys found a better way that does not expose stuff

6543:
## Test usecase
1. create new directory with .woodpecker.yaml config and echo.txt file:
```yaml file=.woodpecker.yaml
steps:
  echo:
    image: bash
    commands: cat echo.txt
```
```txt file=echo.txt
hello devs
```

2. exec with cli via docker backend -> should work
image
3. exec with cli via local backend

3.1 without patch shuld fail
image
3.2 with patch shuld work
image

6543:
@anbraten i still want your opinion and dont overule you ...

6543:
@anbraten will merge after ci pased

anbraten:
As this does not introduce new config settings, I am also fine with it. Will give it a test at some point and we can check now if we can further streamline the use-cases of the cli exec command here.

PR #4108, opened 2024-09-13
Don't expose task data via api

6543:
Currently an instance admin can just querny all secrets of active or pending workflows:

```sh
curl 'http://1.2.3.4:8000/api/queue/info' -H 'Cookie: user_sess=some.jwt.token' | jq .pending[].data | tr -d '"' | base64 -d | jq .config.secrets
```

this is not an serious issue but we never use this info in the webui anyway so we should just not make it available

### DIFF
--- a/server/api/agent.go
+++ b/server/api/agent.go
@@ -97,7 +97,10 @@ func GetAgentTasks(c *gin.Context) {
 	info := server.Config.Services.Queue.Info(c)
 	for _, task := range info.Running {
 		if task.AgentID == agent.ID {
-			tasks = append(tasks, task)
+			t := *task
+			// we copy the task and remove sensitive data
+			t.Data = nil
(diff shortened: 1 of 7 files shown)


server/api/agent.go
```diff
@@ -97,7 +97,10 @@ func GetAgentTasks(c *gin.Context) {
 	info := server.Config.Services.Queue.Info(c)
 	for _, task := range info.Running {
 		if task.AgentID == agent.ID {
-			tasks = append(tasks, task)
+			t := *task
+			// we copy the task and remove sensitive data
+			t.Data = nil
```
anbraten, line 102: Isn't this done by the Json `-` already?!

6543: It was at first but thi `-` it's not needed yes ...

anbraten (approved):

PR #4112, opened 2024-09-14
Require upgrade from 2.x

qwerty287:
Remove old DB migrations and require users to upgrade to 2.x first.

I don't think this has a big impact as there won't be many still using 1.x, but not sure - what do you think about this?


server/store/datastore/migration/001_add_org_id.go
```diff
@@ -26,12 +26,12 @@ import (
 var addOrgID = xormigrate.Migration{
 	ID: "add-org-id",
 	MigrateSession: func(sess *xorm.Session) error {
-		if err := sess.Sync(new(userV031)); err != nil {
+		if err := sess.Sync(new(userV009)); err != nil {
```
zc-devs, line 29: 1. Why `userV009`, not `userV001`?
2. Shouldn't the old records from 'migration' table be dropped also?
3. Maybe just drop old migration code without renaming?

zc-devs: 1. I thought it should be the same as filename: `019_drop_old_cols.go` - `oldPipeline019`, `022_add_orgs.go` - `repo022`. But it doesn't matter a lot :)

6543:
What are the benefits of removing it?

Else i just would let the upgrade path as wide as maintainable ...

zc-devs:
Sounds like [Nextcloud Approaching Upgrades](https://docs.nextcloud.com/server/latest/admin_manual/maintenance/upgrade.html#approaching-upgrades).

6543:
How did you generate the new test database?

server/store/datastore/migration/014_remove_old_migrations.go
```diff
@@ -0,0 +1,54 @@
...
+
+package migration
+
+import (
+	"src.techknowlogick.com/xormigrate"
+	"xorm.io/xorm"
+)
+
+var removeOldMigrations = xormigrate.Migration{
+	ID: "remove-old-migrations",
```
6543, line 23:
```suggestion
	ID: "remove-old-migrations-of-v1",
```

or so ...

6543 (approved):

PR #4114, opened 2024-09-15
Restart tasks on dead agents sooner

6543:
And run the steady queue process as background job


server/queue/fifo.go
```diff
@@ -51,26 +53,30 @@ type fifo struct {
 	paused        bool
 }
 
+const processTimeInterval = time.Second
+
 // New returns a new fifo queue.
 //
 //nolint:mnd
```
6543, line 60:
```suggestion
```

server/queue/fifo.go
```diff
@@ -254,28 +255,36 @@ func (q *fifo) KickAgentWorkers(agentID int64) {
...
 
 // helper function that loops through the queue and attempts to
-// match the item to a single subscriber.
+// match the item to a single subscriber until context got cancel.
 func (q *fifo) process() {
-	q.Lock()
-	defer q.Unlock()
+	for {
+		q.Lock()
+		if q.paused {
```
lafriks, line 262: Imho this will use a lot of CPU cycles of paused as it will loop and lock/unlock cpnstantly

6543: ah dammed :/

I had the time wait at the start first but moved it back for the tests

6543:
backport the queue fix: #4115

lafriks (approved):

PR #4115, opened 2024-09-16
Run queue.process() in background

6543:
Backported fix from  #4114

before this it could happen that worklows that depend on other got stuck untill a new queue task was added


6543:
~~could not reproduce that strange bug where tasks depending on other tasks get stuck in the queu, so I can not check if this here do solve it ...~~

If you set the agent to only query one and then exec it afterwards you have a stuck task that a new agent can not query.

lafriks (approved):

PR #4120, opened 2024-09-16
Remove unused struct from metadata

6543:
it is a bad idea anyway to call secrets metadata ... good that it is not in use!!!


6543:
was added at 442e05a4e129b7c669efde51080e5d03704ba717 ...

lafriks (approved):

PR #4122, opened 2024-09-16
WIP: Make agent labels use doublestar glob expressions

6543:
- [x] code
- [ ] docu
- [ ] tests

### DIFF
--- a/server/grpc/filter.go
+++ b/server/grpc/filter.go
@@ -15,6 +15,9 @@
 package grpc
 
 import (
+	"github.com/bmatcuk/doublestar/v4"
+	"github.com/rs/zerolog/log"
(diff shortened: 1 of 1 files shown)


pat-s:
@6543 If you want to get this in 3.0, this needs substantial work within the next days.

Personally, I think we should remove this from the 3.0 milestone, but it is up to you.

qwerty287:
Do we still need this? There is the wildcard `*` label already and I don't think there are many usecases that want to extend this to full globbing. Besides that doublestar is quite specific for filepaths.

6543:
well this would cover edgecases we currently dont support ... but I'm fine with the current feature set

Issue #571, opened 2021-12-07
Feature: let repo admin select the user a status is send back 


6543:
see https://github.com/woodpecker-ci/woodpecker/issues/558#issuecomment-987456959

It could happen that the original user who activated a repo reject there oauth2 access or left the project, dont have the rights anymore ...
Or you want to set it to a "bot" account ...

in all cases you at the moment have to change that in the database manually :/ - change that!

Gusted:
What would be a good place in the UI to offer this? New section in the 'actions' tab? Do Woodpecker even know which users have write access to the organisation, or would this need to be fetched from the forge?

qwerty287:
Maybe a general "access" overview or something like this would be good. But without actual permission management as this is taken from the forge.

> Do we even know which users have write access to the organisation, or would this need to be fetched from the forge?

This should be available.

PR #4128, opened 2024-09-20
Set repo user to repairing user when old user is missing

lukashass:
When the user that activated a repository is deleted, the repository is broken. Also described in #571.

This PR allows to fix the situation by running repair, which is assigning the repository to the repairing user.

### DIFF
--- a/server/api/repo.go
+++ b/server/api/repo.go
@@ -622,7 +622,7 @@ func repairRepo(c *gin.Context, repo *model.Repo, withPerms, skipOnErr bool) {
 			if err != nil {
 				_ = c.AbortWithError(http.StatusInternalServerError, err)
 			}
-			log.Info().Msgf("Could not find repo user with ID %d during repo repair, set to repair request user with ID %d", oldUserID, user.ID)
+			log.Debug().Msgf("Could not find repo user with ID %d during repo repair, set to repair request user with ID %d", oldUserID, user.ID)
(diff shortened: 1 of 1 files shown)


server/api/repo.go
```diff
@@ -615,10 +615,14 @@ func repairRepo(c *gin.Context, repo *model.Repo, withPerms, skipOnErr bool) {
...
-				c.AbortWithStatus(http.StatusNotFound)
+			oldUserID := repo.UserID
+			user = session.User(c)
+			repo.UserID = user.ID
+			err = _store.UpdateRepo(repo)
+			if err != nil {
+				_ = c.AbortWithError(http.StatusInternalServerError, err)
 			}
-			log.Error().Err(err).Msg("could not get user on repo repair")
+			log.Info().Msgf("Could not find repo user with ID %d during repo repair, set to repair request user with ID %d", oldUserID, user.ID)
```
anbraten, line 625: Probably enough to use .Debug now as server admins probably don't care about this

anbraten (approved):

server/api/repo.go
6543, line 625:
```suggestion
			log.Debug().Msgf("Could not find repo user with ID %d during repo repair, set to repair request user with ID %d", oldUserID, user.ID)
```

6543 (approved):

PR #4129, opened 2024-09-20
cli exec: fix flag type and remove uneffective ones

6543:
no idea how the things got there in the first place but the step specific envs are set by the engine.

And proper document the flags.


cli/exec/flags.go
```diff
@@ -125,229 +125,277 @@ var flags = []cli.Flag{
...
 	&cli.StringFlag{
 		Sources: cli.EnvVars("CI_PIPELINE_DEPLOY_TASK"),
 		Name:    "pipeline-deploy-task",
+		Usage:   "Set the metadata environment variable \"CI_PIPELINE_DEPLOY_TASK\".",
 	},
 	&cli.StringFlag{
 		Sources: cli.EnvVars("CI_PIPELINE_FILES"),
-		Usage:   "either json formatted list of strings, or comma separated string list",
+		Usage:   "Set the metadata environment variable \"CI_PIPELINE_FILES\", either json formatted list of strings, or comma separated string list.",
 		Name:    "pipeline-files",
```
anbraten, line 238:
```suggestion
		Name:    "changed-files",
```

6543: well it is an breaking change then :/

and it would break the name convention. what could work is:
```diff
-		Name:    "pipeline-files",
+
```		Name:    "pipeline-changed-files",

6543: well at least it is non breaking as not jet released ... https://github.com/woodpecker-ci/woodpecker/pull/4119

anbraten: `pipeline-changed-files` seems good

6543: ->  #4142

PR #4132, opened 2024-09-20
Ensure cli exec has by default not the same prefix

6543:
this prohibits collisions

this make it similar to the server:

https://github.com/woodpecker-ci/woodpecker/blob/eebaa10d104cbc3fa7ce4c0e344b0b7978405135/server/pipeline/stepbuilder/stepBuilder.go#L289-L295


6543:
-> it did randomize the wordir too :/

anbraten:
I would suggest to simply hard-code it in the cli to the same value we use in the agent. `wp_%s_%d`.

6543:
> I would suggest to simply hard-code it in the cli to the same value we use in the agent. `wp_%s_%d`.

no hardcode will make colisins as we have right now ... but I could change the format to have a wp_ prefix. the problem is that we dont have a number that is even near the truth for each workflow

cli/exec/flags.go
```diff
@@ -51,9 +52,10 @@ var flags = []cli.Flag{
 	&cli.StringFlag{
 		Sources: cli.EnvVars("WOODPECKER_PREFIX"),
 		Name:    "prefix",
-		Value:   "woodpecker",
-		Usage:   "prefix used for containers, volumes, networks, ... created by woodpecker",
-		Hidden:  true,
+		// emulate server behavior https://github.com/woodpecker-ci/woodpecker/blob/eebaa10d104cbc3fa7ce4c0e344b0b7978405135/server/pipeline/stepbuilder/stepBuilder.go#L289-L295
+		Value:  "woodpecker_" + ulid.Make().String(),
```
6543, line 56:
```suggestion
		Value:  "wp_" + ulid.Make().String(),
```

anbraten:
> no hardcode will make colisins as we have right now ... but I could change the format to have a wp_ prefix. the problem is that we dont have a number that is even near the truth for each workflow

`ulid.Make().String()` is fine. I meant we don't need a cli option for it at all as users can't change it for the agent as well.

6543:
well you normally dont need it but why remove it ... if we should not present this to the user we can just hide that flag

anbraten:
> *it's alrady hidden so why remove it etc ?!?

Why do we need to support changing it? It's just extra code ...

6543:
@anbraten done

anbraten (approved):

PR #4154, opened 2024-10-02
Add `.lycheeignore`

pat-s:
To avoid rate limiting for stackoverflow.

SO links should be *quite* stable and I don't think we need to check them all the time.


6543 (approved):

6543:
well from time to time we should but for the current pipeline yes

PR #4155, opened 2024-10-02
Print agent labels in debug mode

6543:
because right now there is no way to figure this out in any way


6543:
backported:  #4156

PR #4156, opened 2024-10-02
Print agent labels in debug mode (#4155)

6543:
Backport  #4155

because right now there is no way to figure this out in any way

### DIFF
--- a/cmd/agent/core/agent.go
+++ b/cmd/agent/core/agent.go
@@ -238,6 +238,8 @@ func run(ctx context.Context, c *cli.Command, backends []types.Backend) error {
 		return err
 	}
 
+	log.Debug().Msgf("agent configured with labels: %#v", labels)
+
 	filter := rpc.Filter{
 		Labels: labels,
 	}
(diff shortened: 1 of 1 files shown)


pat-s (commented):
This is the same as #4155 ?

cmd/agent/core/agent.go
```diff
@@ -238,6 +238,8 @@ func run(ctx context.Context, c *cli.Command, backends []types.Backend) error {
 		return err
 	}
 
+	log.Debug().Any("labels", labels).Msgf("agent configured with labels")
```
zc-devs, line 241: Maybe better to make it `info` and merge with line 285
> starting Woodpecker agent with version ... labels

6543: Normaly you as admin have set it so you should know this is for debugging porpuse

6543:
> This is the same as #4155 ?

Yes 1:1 backport

pat-s (approved):
Ah, missed the target branch...

PR #4160, opened 2024-10-04
Bump trivy

pat-s:
To avoid https://ci.woodpecker-ci.org/repos/3780/pipeline/21074/18

Should be fixed in 0.56 according to some discussions


6543:
-> #4188

PR #4162, opened 2024-10-04
renovate: set `automergetype=pr`

pat-s:
So that PRs are raised. Otherwise, the updates are getting forgotten in the branches due to the approval requirements of WP.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -15,6 +15,7 @@
     "apimachinery",
     "Archlinux",
     "autoincr",
+    "automerge",
     "autoscaler",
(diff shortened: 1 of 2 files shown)


6543 (approved):

pat-s:
I think the overall default should stay with `branch` as not all repos require an explicit approval for CI jobs. Besides, the idea is still to somehow add a whitelist support so that bots like renovate don't require approval.

6543:
> Shouldn't this be added in the global config?

We should make it default too yep!

PR #4201, opened 2024-10-06
Assign workflows to agents with the best label matches

6543:
This make sure if there are org and global agents, that workflows get fetched by the org agent they are in first and only if not available will the global one fetch it.

This also let you control agent priority etc ...

before the selection of an agent was kinda at random ...


6543:
or is it a feature?

6543:
https://app.codecov.io/gh/woodpecker-ci/woodpecker/pull/4201/blob/server/grpc/filter.go

100% test coverage :D

server/queue/fifo_test.go
```diff
@@ -504,3 +516,129 @@ func TestShouldRun(t *testing.T) {
...
+			assert.Equal(t, int64(3), task.AgentID, "Task with platform windows should be assigned to worker with filter3")
+		}
+	}
+
+	// Check if task with org-id "*" is assigned to worker with filter4
+	for _, task := range receivedTasks {
+		if task.ID == "5" {
+			assert.Equal(t, int64(5), task.AgentID, "Task with org-id * should be assigned to worker with filter4")
+		}
+	}
```
anbraten, line 643: Could be a bit shorter with a map (taskId => agentId) by using the above for loop again.

server/queue/fifo.go
```diff
@@ -325,17 +325,25 @@ func (q *fifo) filterWaiting() {
...
 	var next *list.Element
+	var bestWorker *worker
+	var bestScore int
+
 	for e := q.pending.Front(); e != nil; e = next {
 		next = e.Next()
 		task, _ := e.Value.(*model.Task)
 		log.Debug().Msgf("queue: trying to assign task: %v with deps %v", task.ID, task.Dependencies)
 
 		for w := range q.workers {
```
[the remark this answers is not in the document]
6543: done

server/queue/fifo_test.go
6543: done

anbraten (approved):

PR #4208, opened 2024-10-07
Use middleware to load org

### LINKED ISSUE #4207: Refactor: create middleware which set the org for the apis dedicated to orgs
currently each api endpoint function has the org-id parsing and loading from store defined on the start.

similar to how there is a middleware for repos witch use `session.Repo(c)` we should do the same for organisazions ...

6543:
close #4207


server/router/middleware/session/user.go
```diff
@@ -122,24 +122,17 @@ func MustUser() gin.HandlerFunc {
...
-			c.String(http.StatusBadRequest, "Error parsing org id. %s", err)
-			return
-		}
-
-		org, err := _store.OrgGet(orgID)
-		if err != nil {
-			c.String(http.StatusNotFound, "Organization not found")
+		org := Org(c)
+		if org == nil {
+			c.String(http.StatusBadRequest, "Org not loaded")
```
anbraten, line 134:
```suggestion
			c.String(http.StatusBadRequest, "Organization not loaded")
```

server/router/middleware/session/org.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		c.Set("org", org)
+		c.Next()
+	}
+}
+
+func MustOrg() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		SetOrg()(c)
```
anbraten, line 76: This seems to be rather implicit. Probably better to explicitly load / set the org in the router as we do it with `SetUser`.

6543: ?

Must will fail if org could not load and set want fail ... if desired i can remove set for now .... as we have only cades where it must load

anbraten: Works like this as well. I actually wanted to say, that `SetOrg` is kind of a side-effect in a middleware that gets the org and thereby should be explicitly used in the router same as for user. We `SetUser` first to later on be able to get the user through `User()` or `MustUser()`

server/router/middleware/session/org.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/gin-gonic/gin"
+
+	"go.woodpecker-ci.org/woodpecker/v2/server/model"
+	"go.woodpecker-ci.org/woodpecker/v2/server/store"
+	"go.woodpecker-ci.org/woodpecker/v2/server/store/types"
+)
+
+func Org(c *gin.Context) *model.Org {
+	v, ok := c.Get("org")
```
anbraten, line 30: This one only works now if `MustOrg` was called previously.

6543: yes as with other middleware ...

6543: it's  a feature not a bug ... I understand that it could be frigned if you just read the functions behind endpints ... but we do the dame to ensure permissions are set and respected etc ...

... this decuples the endpoint function a bit and also allow tests to be more easy to write agains it ...

anbraten: > but we do the dame to ensure permissions are set and respected etc ...

Actually not, we always call `SetUser`, `SetPerm`, `SetRepo` in the router to add sth to the context and later on, load it using `MustUser` or `User`. Can I push the change to adjust it this way?

6543: i dont get it :/ so ... Sure push it ...

anbraten (approved):

PR #4212, opened 2024-10-09
Add options to controll the database connections (open,idle,timeout)

6543:
this allow more fine control and make sure woodpecker do not crash in high use scenarios where the engine try to create more connection as the database do allow ... (mariadb=100)


6543:
the defaults i took form gitea

cmd/server/flags.go
```diff
@@ -36,6 +36,24 @@ var flags = append([]cli.Flag{
 		Name:    "log-xorm-sql",
 		Usage:   "enable xorm sql command logging",
 	},
+	&cli.IntFlag{
+		Sources: cli.EnvVars("WOODPECKER_XORM_MAX_CONNECTIONS"),
```
anbraten, line 40: Should get expert prefix and description hint.

6543: well I dont thing so ... if you config more connections for your db you can increase it here ...

and that's just an normal admin task

6543: @qwerty287 @zc-devs done

PR #4213, opened 2024-10-09
Only allow docker-buildx plugin >= 5.0.0 to be priviledged by default

6543:
bump buildx plugin to only be privileged if version is pinned and greater than v5.0.0 or latest.

https://codeberg.org/woodpecker-plugins/docker-buildx/releases/tag/v5.0.0

so by default only this images will get privildges:
- `woodpeckerci/plugin-docker-buildx` (latest)
- `woodpeckerci/plugin-docker-buildx:latest`
- `woodpeckerci/plugin-docker-buildx:5`
- `woodpeckerci/plugin-docker-buildx:5.0`
- `woodpeckerci/plugin-docker-buildx:5.0.0`

### DIFF
--- a/shared/constant/constant.go
+++ b/shared/constant/constant.go
@@ -19,8 +19,10 @@ var PrivilegedPlugins = []string{
 	"plugins/docker",
 	"plugins/gcr",
 	"plugins/ecr",
-	"woodpeckerci/plugin-docker-buildx",
-	"codeberg.org/woodpecker-plugins/docker-buildx",
+	"woodpeckerci/plugin-docker-buildx:5",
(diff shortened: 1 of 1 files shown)


6543:
well that's the as much non breaking as posible way

we don't have a better way to make users secure by default!!!

and if it breaks they where insecure and can decide to either stay so or fix it!

6543:
and to "FIX" it ... you can just read the old images you use ... but then it's up to you to secure it - I'll write that as heading into the release note

6543:
@qwerty287  see https://github.com/woodpecker-ci/woodpecker/pull/4214

Issue #3960, opened 2024-07-22
Major release


anbraten:
We should define a release cycle for major versions. The idea to release `3.0.0` next #3957 is currently somehow ignoring that users will have to prepare for some braking changes. For larger deployments with multiple users & repositories (There are deployments with >1000 repos 😉) this requires some time. 

Especially changes to pipeline configs take some time. To streamline this process we started adding some deprecation notices to the pipeline compiler some time ago. Therefore I propose defining a major release cycle of **> 12 months** (would result in a release date of earliest Nov 23, 2024 for 3.0.0). In addition every change that requires interaction by the Woodpecker users (not only admins) should be announced / marked as deprecated latest some time (proposing  **3 months**) before removal to provide enough time for admins to update to latest minor versions and users to notice / be notified by this change.

6543:
I remember that from v0.15.x to v1.0.0 we took quite a while and I often had to say ... you should use next as it is not jet released ... so a more rapid release schedule as we have now is a good. As the bump from 1.0.0 to 2.0.0 was a bit short I think 7months are quite OK.

So I would not specify strict dates as nobody works on WP full time btw ... but rather ranges.

So I'm OK to have a block majour versions to be made before 1/2year to the last one but it could be longer if no more than a few (2...3) breaking things are scheduled. blocking it for 1year is to much in my opinion.

What we could do is to have a release branch in times where a majour version is in development, to be able to still release new versions, if needed ... If we do it the other way around and hold breaking changes till we wana do a majour version we get a lot of conflicts etc ...

lafriks:
Hopefully we come to the phase to stabilize that at least pipeline definitions are stable to not have breaking changes

6543:
I think https://github.com/woodpecker-ci/woodpecker/issues/2780 shows quite some list

zc-devs:
Sorry for "spam" here and there, I just want to understand the rules and a flow.
Follow up of https://github.com/woodpecker-ci/woodpecker/pull/3918#issuecomment-2263739788 and https://github.com/woodpecker-ci/woodpecker/issues/3997#issuecomment-2263646174.

Should #3397 was opened before `2.7.0` to be eligible for `3.0`? Should PR was opened also or Issue is enough?
Suppose, I waited for next major to be announced, because I didn't want to waste time to make a PR into branch, which will become far outdated before actual merge. How should I know, that after `2.7` next major is coming? What is the flow? 

Besides, reading #2780:

[...]

PR #4218, opened 2024-10-10
Warn user when using lowercase secrets

zc-devs:
Follow up https://github.com/woodpecker-ci/woodpecker/issues/3960#issuecomment-2395470285

Lints
> Removed uppercasing all secret env vars, instead, the value of the secrets property is used


zc-devs:
it might not work properly = you have to access it with the lower case name or change secret name in pipeline

Fixed message in [b5cf0d0](https://github.com/woodpecker-ci/woodpecker/pull/4218/commits/b5cf0d0dcbe7019fb61904a74df785cc8541a0a0)

6543:
?!? did we not move to case sensitive iand depricate the secret key as a whole ?!?

zc-devs:
> move to case sensitive

Yes, therefore the PR.

> depricate the secret key as a whole

No, there was empty deprecation section in the code on `main` and of course no warnings in UI (when I tested and made the PR).

lafriks:
I don't think this is needed as we already will have warning in 3.0 about deprecated secrets section as whole

PR #4235, opened 2024-10-20
Deprecate `secrets`

qwerty287:
Deprecate the `secrets` directive completely.

Closes https://github.com/woodpecker-ci/woodpecker/discussions/4015

### DIFF
--- a/.woodpecker/binaries.yaml
+++ b/.woodpecker/binaries.yaml
@@ -6,7 +6,7 @@ when:
 
 variables:
   - &golang_image 'docker.io/golang:1.23'
-  - &node_image 'docker.io/node:22-alpine'
+  - &node_image 'docker.io/node:23-alpine'
   - &xgo_image 'docker.io/techknowlogick/xgo:go-1.22.x'
 
 # cspell:words bindata netgo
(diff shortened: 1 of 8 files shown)


zc-devs:
> replaces and closes https://github.com/woodpecker-ci/woodpecker/pull/4218

Are you going to backport this PR?

Anyway, it is not a replacement, because it says "in 3.0 you can still use `secrets`, but in 4.0 it will be removed".
And in my PR I warn people "you use lower-cased secret, which currently upper-cased; it won't be in 3.0, therefore those steps will fail (very?) likely". Which technically is not `changed pipeline syntax`, but result is the same from user's point of view.

zc-devs:
> Which technically is not changed pipeline syntax, but result is the same from user's point of view.

So, you could just remove it in 3.0
https://github.com/woodpecker-ci/woodpecker/issues/3960#issuecomment-2395481872

pat-s:
> So, you could just remove secrets in 3.0 and warn in 2.7.

If you mean 2.7.2 etc: I think many users are sticking with 2.7.0 due to the breaking changes in 2.7.1+, so they won't ever see warnings in 2.7.2.

zc-devs:
> If you mean 2.7.2

Yep

> many users are sticking with 2.7.0 due to the breaking changes in 2.7.1+, so they won't ever see warnings in 2.7.2

Sad

> I'd also rather deprecate in v3 and remove in v4

Sure, it was just thing to consider.

lafriks (approved):

PR #4245, opened 2024-10-23
Use descriptive pod names

pat-s:


6543:
the engine "frontend" referse to all the user input parsing and compiling into an actual pipeline that then will create backend types that the different backend types do implement. so we should not use the metadata in the backend. Also submiting more stuff witch should be already present via rpc is not good.

So from what i understand the container names are to random to debug stuff?

pat-s:
In k8s, I often have >5 builds running from different instances (which represent different forges). It is not possible to easily see which pod is which build as all are named `wp-<hash>`.

Using the pipeline names makes this more easy and descriptive without a downside.

PR #4259, opened 2024-10-26
Delete GitLab webhooks with partial URL match

### LINKED ISSUE #4248: GitLab: disabling repo doesn't remove webhook
### Component

server

### Describe the bug

Gitlab.org, GL 17.6

### Steps to reproduce

1. Enable repo 
2. Disable repo (red error appears in the bottom right)
3. Repo is disabled in WP dashboard but the Webhook still exists 

### Expected behavior

Webhook gets removed without error

### System Info

```shell
WP next-9dd796ad65 (Oct 13th 2024)
GL 17.6
```

### Additional context

_No response_

### Validations

pat-s:
fix #4248 

- [x] Ensure to process all matching webhooks and don't exit after the first one
- [x] Use partial string match as `webURL` != `hook.URL` (the latter always has the `api/token` suffix, the former doesn't)

Tested on gitlab.com.

### DIFF
--- a/server/forge/gitlab/gitlab.go
+++ b/server/forge/gitlab/gitlab.go
@@ -559,7 +559,6 @@ func (g *GitLab) Deactivate(ctx context.Context, user *model.User, repo *model.R
 				if err != nil {
 					return err
 				}
-				break
 			}
 		}
 
--- a/server/forge/gitlab/gitlab.go
+++ b/server/forge/gitlab/gitlab.go
@@ -555,12 +555,16 @@ func (g *GitLab) Deactivate(ctx context.Context, user *model.User, repo *model.R
 		for _, hook := range hooks {
 			if strings.Contains(hook.URL, webURL) {
 				hookID = hook.ID
+				_, err = client.Projects.DeleteProjectHook(_repo.ID, hookID, gitlab.WithContext(ctx))
+				if err != nil {
+					return err
+				}
 				break
 			}
 		}
 
 		// Exit the loop when we've seen all pages
-		if resp.CurrentPage >= resp.TotalPages {
+		if hookID != -1 || resp.CurrentPage >= resp.TotalPages {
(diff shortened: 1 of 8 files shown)


server/forge/gitlab/gitlab.go
```diff
@@ -553,28 +553,26 @@ func (g *GitLab) Deactivate(ctx context.Context, user *model.User, repo *model.R
...
-				break
+				_, err = client.Projects.DeleteProjectHook(_repo.ID, hookID, gitlab.WithContext(ctx))
+				log.Info().Msg(fmt.Sprintf("successfully deleted hook with ID %d for repo %s", hookID, repo.FullName))
+				if err != nil {
+					return err
+				}
 			}
 		}
 
 		// Exit the loop when we've seen all pages
```
anbraten, line 566: The comment doesn't fit anymore:
```suggestion
```

server/forge/gitlab/gitlab.go
```diff
@@ -553,28 +553,26 @@ func (g *GitLab) Deactivate(ctx context.Context, user *model.User, repo *model.R
...
-		if resp.CurrentPage >= resp.TotalPages {
+		if hookID != -1 || resp.CurrentPage >= resp.TotalPages {
```
anbraten, line 567: Looks like it would still stop after the first page with a match `hookID != -1`? 😉 

```suggestion
		if resp.CurrentPage >= resp.TotalPages {
```

pat-s: OK yes, in this new format we can just skim through all pages and don't need a custom exit logic.

6543 (approved):

6543:
```
server/forge/gitlab/gitlab.go:544:2: ineffectual assignment to hookID (ineffassign)
	hookID := -1
	^
```

PR #4262, opened 2024-10-27
Increase `WOODPECKER_FORGE_TIMEOUT` to fix config fetching for GitLab

### LINKED ISSUE #4260: GitLab: push triggers error with `[generic]could not load config from forge`
### Component

server

### Describe the bug

Only experiencing this in 1/6 repos right now. 
Ported the repo over from another forge and everything was fine there - so the config isn't the issue.

Need to investigate what causes this.

Manual triggers work fine 🤔 

<details>
<summary>Investigation</summary>

Logs don't even paste which config is being searched for

```
{"level":"debug","repo":"devxy/images/r","error":"context deadline exceeded","time":"2024-10-27T08:42:58Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/pipeline/create.go:90"," message":"error while fetching config '' in 'refs/heads/main' with user: 'pat-s'"}
```

</details>

### Steps to reproduce

.

### Expected behavior

_No response_

### System Info

```shell
d37c13873840d83a280a166f6dbbe3b129441e30
```

### Additional context

_No response_

### Validations

pat-s:
fix #4260 

Config fetching on GL takes a bit longer due to the subgroup fetching and permissions checks.
In my case, I had 1/6 repos failing with the default of 3s. The repo only had two workflow files. Both were somewhat complex.

I assume that the workflow parsing combined with the subgroup checks exceeded the 3s here.
After bumping `WOODPECKER_FORGE_TIMEOUT` to 5s, it worked.
Getting the initial config fetched seems to speed up future config pull as afterwards I was able to reduce `WOODPECKER_FORGE_TIMEOUT` again to 3s.

However, to avoid such issues for users I think the default value should be bumped to (at least) 5s.

Additionally, I improved some of the log/trace outputs which helped me a bit along the way.

### DIFF
--- a/cmd/server/flags.go
+++ b/cmd/server/flags.go
@@ -287,7 +287,7 @@ var flags = append([]cli.Flag{
 		Sources: cli.EnvVars("WOODPECKER_FORGE_TIMEOUT"),
 		Name:    "forge-timeout",
 		Usage:   "how many seconds before timeout when fetching the Woodpecker configuration from a Forge",
-		Value:   time.Second * 3,
+		Value:   time.Second * 5,
 	},
 	&cli.UintFlag{
 		Sources: cli.EnvVars("WOODPECKER_FORGE_RETRY"),
--- a/docs/docs/30-administration/10-server-config.md
+++ b/docs/docs/30-administration/10-server-config.md
@@ -484,7 +484,7 @@ Specify a configuration service endpoint, see [Configuration Extension](./40-adv
 
 ### `WOODPECKER_FORGE_TIMEOUT`
 
-> Default: 3s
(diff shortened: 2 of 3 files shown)


6543 (approved):

pat-s:
Hmm while the build now starts, the webhook in this repo still fails with

image

and GL disables the webhook after a few executions which resulted in a timeout.

6543:
still we can increase it a bit more ...

pat-s:
image

I don't think the timeout is the issue here. Even with 20s the GL webhook quits after ~ 10s.
I assume something is broken with the response?

PR #4263, opened 2024-10-27
WIP: Add option to build project with nix

6543:
play around more with nix stuff ...
- [ ] server does not compile webui
- [ ] how do dynamicly get the go vendor hash?


6543:
nix flake would need sha of vendored deps to change each time renovare does change this -> that's just not worth it

PR #4271, opened 2024-10-28
Do not require agent polling to make workflows that should be skiped done

6543:
currently we have in the task poll logic an check that clean up the queue of tasks that should be skipped.
so it requires agents to poll to actually do stuff.

but the agent poll logic should ignore all tasks that are not ready to run and the queue maintenance should happen in an dedicated place.

this pull starts to refactor this by at least move it in it's own function and go routine in the background.

### DIFF
--- a/cmd/server/grpc_server.go
+++ b/cmd/server/grpc_server.go
@@ -49,6 +49,7 @@ func runGrpcServer(ctx context.Context, c *cli.Command, _store store.Store) erro
 	)
 
 	woodpeckerServer := woodpeckerGrpcServer.NewWoodpeckerServer(
+		ctx,
 		server.Config.Services.Queue,
 		server.Config.Services.Logs,
 		server.Config.Services.Pubsub,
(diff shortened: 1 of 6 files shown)


server/queue/fifo.go
```diff
@@ -27,6 +27,9 @@ import (
 	"go.woodpecker-ci.org/woodpecker/v2/shared/constant"
 )
 
+// InternalWorkerID is pseudo agent ID for internal routines using the queue.
+const InternalWorkerID = -2
```
6543, line 31: -2 because -1 is used in the agentID context already ... so just any other negative integer is sufficient

6543:
todo: check that "virtual" agent is neither in database nor in webui/api visible

6543:
PS: once we refactored more of the queue related code we might have a chance to simplify it :)

6543:
currently i work on refactoring the queue stuff:
- https://github.com/woodpecker-ci/woodpecker/pull/6011
- https://github.com/woodpecker-ci/woodpecker/pull/6017
- https://github.com/woodpecker-ci/woodpecker/pull/6056
- ...

6543:
should be mostly fixed... rest will be done once refactoring progressed

PR #4272, opened 2024-10-28
Don't parse forge config files multiple times if no error occured

### LINKED ISSUE #4264: GitLab: push hook event reponses are not sent
### Component

server

### Describe the bug

Push event hooks don't get replied to properly. Builds still start as they are parsed internally and started but there's an issue with the response to the incoming hook.

This leads to GL deactivating the hook after a few hook iterations that didn't get a response.

image

### Steps to reproduce

- trigger a push event hook

### Expected behavior

next (9a7623359ecbfb65e9f9e4ad9bbb0aa169e03531) (Oct 27th)

### System Info

```shell
.
```

### Additional context

_No response_

### Validations

pat-s:
fix #4264 
introduced in #3699

This applies to all forges. Currently, all config files are fetched three times from the forge.
As this takes a while for gitlab.com, this resulted in an overall timeout error of > 10s for the webhook processing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ CGO_ENABLED ?= 1 # only used to compile server
 
 HAS_GO = $(shell hash go > /dev/null 2>&1 && echo "GO" || echo "NOGO" )
 ifeq ($(HAS_GO),GO)
-	XGO_VERSION ?= go-1.20.x
+	XGO_VERSION ?= go-1.22.x
 	CGO_CFLAGS ?= $(shell go env CGO_CFLAGS)
 endif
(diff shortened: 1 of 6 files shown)


anbraten:
Nice catch. Seems to be introduced by https://github.com/woodpecker-ci/woodpecker/pull/371 from +3years ago 🙈
Could you try to add a test checking this on `f.On("File")` with sth like `.Times(1)` or by changing it to `.Once()`

6543:
tested with and without tests -> it's covered now :tada:

6543:
-> #4273

PR #4276, opened 2024-10-29
Fix renovate support for `xgo`

pat-s:
- Use semver versions instead of `.x` (not detectable by renovate)
- Include xgo version in Makefile
- Clean `Dockerfile.make`: don't pin alpine versions, simplify node tags


.cspell.json
```diff
@@ -129,6 +129,7 @@
     "nosniff",
     "ntfy",
     "octocat",
+    "ontainer",
```
6543, line 132: why not just ignore .github/renovate.json

Makefile
```diff
@@ -39,7 +39,7 @@ CGO_ENABLED ?= 1 # only used to compile server
 
 HAS_GO = $(shell hash go > /dev/null 2>&1 && echo "GO" || echo "NOGO" )
 ifeq ($(HAS_GO),GO)
-	XGO_VERSION ?= go-1.20.x
+	XGO_VERSION ?= 1.22.8
```
xoxys, line 42: Sure that this change still works?

xoxys (commented):
xgo support would have been possible with the `x` version as well https://github.com/thegeeklab/renovate-presets/blob/main/golang.json#L40-L45

Makefile
xoxys: I don't think so, see https://pkg.go.dev/src.techknowlogick.com/xgo#readme-go-releases

pat-s:
Ah, then we should change it to this 👍

pat-s:
Switched to @xoxys approach. Linked PR in preset: https://github.com/woodpecker-ci/renovate-config/pull/21

.github/renovate.json
```diff
@@ -30,7 +31,7 @@
     },
     {
       "groupName": "golang-lang",
-      "matchPackagePatterns": ["^golang$", "xgo"],
```
xoxys, line 33: Why have you removed "xgo" from the group here? IMO xgo and golang updates should still be grouped.

pat-s: Ah, not on purpose. Thanks for catching!

xoxys (approved):

PR #4283, opened 2024-11-02
Fix deployment event note

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/discussions/4275


pat-s (changes requested):
Thanks for taking a stab at this.

I think the whole deployment topic needs to be distinguished in more detail:

- the only forge which issues a `deployment` event is AFAIK GH
- `deployment`s can be triggered from within WP itself

I agree on removing this the highlighted part here but I think all pages referencing it (including the forge matrix listing the part about "deployment") need a footnote or some other way of explaining the difference between the forge event and the WP event in detail. Otherwise such questions will continue to pop up.

pat-s (approved):
Thanks, definitely better than before 👍 Let's see if more issues related to this come up in the future.

Issue #3827, opened 2024-06-24
Separate admin settings tabs into separate pages


anbraten:
The tabs used in the admin settings are currently loading a single page and therefore when entering the admin settings the complete data like secrets, registries, agents, users, ... is loaded at one. Instead I

PR #4285, opened 2024-11-02
Use separate routes instead of anchors

qwerty287:
first part of https://github.com/woodpecker-ci/woodpecker/issues/3827


web/src/router.ts
```diff
@@ -104,9 +104,47 @@ const routes: RouteRecordRaw[] = [
           {
             path: 'settings',
             name: 'repo-settings',
-            component: (): Component => import('~/views/repo/RepoSettings.vue'),
+            component: (): Component => import('~/views/repo/settings/RepoSettings.vue'),
```
anbraten, line 107: Component can not be used with redirect simultaneously? Otherwise we could let `repo-settings` redirect to `repo-settings-general`.

web/src/views/admin/AdminSettingsWrapper.vue
```diff
@@ -0,0 +1,92 @@
...
+
+  if (!user?.admin) {
+    notifications.notify({ type: 'error', title: i18n.t('admin.settings.not_allowed') });
+    await router.replace({ name: 'home' });
+  }
+});
+
+const activeTab = computed({
+  get() {
+    if (route.name === 'admin-settings-secrets') {
```
anbraten, line 49: Wouldn't it be easier to use the route name as the tab id directly? Or we change the tab component to take a `:to=` prop, that checks if the route is active and routes on click?

web/src/views/admin/AdminSettingsWrapper.vue
```diff
@@ -0,0 +1,45 @@
...
+onMounted(async () => {
+  if (!user?.admin) {
+    notifications.notify({ type: 'error', title: i18n.t('admin.settings.not_allowed') });
+    await router.replace({ name: 'home' });
+  }
+
+  if (!user?.admin) {
+    notifications.notify({ type: 'error', title: i18n.t('admin.settings.not_allowed') });
+    await router.replace({ name: 'home' });
+  }
```
anbraten, line 43:
```suggestion
```

web/src/views/org/settings/OrgSettingsWrapper.vue
```diff
@@ -11,17 +11,11 @@
       </span>
     </template>
 
-    <Tab id="secrets" :title="$t('secrets.secrets')">
-      <OrgSecretsTab />
-    </Tab>
+    <Tab to="org-settings-secrets" :title="$t('secrets.secrets')" />
```
anbraten, line 14: Would suggest to use Vue's `:to="{ name: 'org-settings-secrets'}` syntax instead.

web/src/components/layout/scaffold/Tab.vue
```diff
@@ -11,7 +11,8 @@ import type { IconNames } from '~/components/atomic/Icon.vue';
 import { useTabsClient, type Tab } from '~/compositions/useTabs';
 
 const props = defineProps<{
-  id?: string;
+  to?: string;
+  alternativeRoute?: string;
```
anbraten, line 15: `$route.matched.some(({ name }) => name === 'customer_users')"` is probably what you were looking for?

web/src/router.ts
anbraten: `name: 'repo-settings',` is basically not needed anymore, right? If it is, you could rename 'repo-settings-general' to it.

web/src/components/layout/scaffold/Tab.vue
anbraten: Just copied that sample over. You could check if a route is matching (should work for your question above as well) using `const route = useRoute()` With `route.matched`.

web/src/views/admin/AdminSettingsWrapper.vue
anbraten: It's twice in the code.

web/src/views/repo/RepoWrapper.vue
```diff
@@ -134,6 +133,8 @@ watch([repositoryId], () => {
 });
 
 const badgeUrl = computed(() => repo.value && `${config.rootPath}/api/badges/${repo.value.id}/status.svg`);
+<<<<<<< HEAD
```
6543, line 136: Conflict relicz

web/src/router.ts
anbraten: The wrapper does not need a name anymore as it has a child with a default route:

```
path: '',
   name: 'repo-settings-general',
```.

You can rename that default child to `repo-settings` or change all router links to point to the new route ` repo-settings-general`

web/src/views/org/settings/OrgSettingsWrapper.vue
anbraten: That has to be possible somehow. When using `<router-link :to="...">` it also detects if the route is [currently active](https://router.vuejs.org/guide/essentials/active-links#Active-links) and sets `router-link-active` and `router-link-exact-active` respectively. That would make the syntax a bit cleaner.

anbraten:
> I just changed it now to use router-link instead of button to solve all the issues you had

That's smart as we don't have to extract the route matching logic and in addition get open to new tab with proper links

anbraten (dismissed):

anbraten (approved):
Tested and everything seems to work now
@qwerty287 Thanks for the nice work. Feel free to merge or adjust

6543 (approved):

PR #4297, opened 2024-11-04
Add a release process

anbraten:
This PR adds a checklist to prepare new releases.


xoxys:
I would recommend a checklist in an issue template then. This way, every release can easily get a tracking issue and its visible for everyone which parts are still missing for the new release.

Issue #3328, opened 2024-02-05
An unknown error occurred on User Secrets page (migration v1 -> v2)


zc-devs:
### Component

server

### Describe the bug

Some users face `An unknown error occurred` popup when follow to their profile Secrets page.

---

On that old instance (the first deployed version was 1.0, as I remember) we have:
```
woodpecker=# select user_id, user_login, user_org_id from users order by user_id;
 user_id |  user_login   | user_org_id
---------+---------------+-------------
       1 | admin         |           3
       2 | maria         |           3
       3 | kevin         |           3
       4 | piter         |           3
       5 | john          |           5
(5 rows)

woodpecker=# select * from orgs;
 id |     name      | is_user | private
----+---------------+---------+---------
  1 | maria         | t       | f
  2 | piter         | t       | f
  3 | kevin         | t       | f
  4 | admin         | t       | f
  5 | john          | t       | f
(5 rows)
```

As we can see, majority of users are members of an organization `3, kevin`. `john` is a pretty new user.
When `piter` goes to `https://woodpecker.example.com/user#secrets`, he sees error in browser console
```
GET
https://https://woodpecker.example.com/api/orgs/3/secrets?page=1
[HTTP/2 403  138ms]
```

On test instance (the first deployment was around version 2, as I remember) we have:
```
woodpecker_test=# select user_id, user_login, user_org_id from users;
 user_id | user_login | user_org_id
---------+------------+-------------
       3 | usr2       |           3
       2 | usr        |           2
       1 | admin      |           1
(3 rows)

woodpecker_test=# select * from orgs;
 id | name  | is_user | private
----+-------+---------+---------
  1 | admin | t       | f
  2 | usr   | t       | f
  3 | usr2  | t       | f
(3 rows)
```

All users have their own orgs. So there are no errors.

### System Info

```shell
Woodpecker 2.3.0, Gitea 1.21.5, PostgreSQL 16
```

### Additional context

https://github.com/woodpecker-ci/woodpecker/pull/3319

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

zc-devs:
`1.0.5`

```
[...]

PR #4307, opened 2024-11-05
Add migration to autofix corrupted users.org_id entrys in db

6543:
No idea why the corrupt entrys did exist ... we might should dig into the reason.

anyway it only happens on very old accounts and I did the migrations mannually at codeberg and now at kithara.

so we clearly should run it to catch potential hick-ups ...


6543:
might be an regression of #2126

zc-devs:
Perhaps, related #3328

6543:
symptoms: you get just weard behaviour and some functions 'randomly' dont work ..

deleting all of that ward user related stuff (in db) and recreate the user by log in newly do fix it.

pat-s:
> it only happens on very old accounts 

Define "very old" - if it happened on v1 -> v2, I'd not use that adjective here.

Also: #3319 -> can likely be closed then. Great that there is a fix now for this, a bit sad that it took almost a year. However I am not sure whether you discovered this "just recently" or were actually aware of it the whole time but it simply wasn't important enough (?).

> anyway it only happens on very old accounts and I did the migrations mannually at codeberg and now at kithara.

Hmm, just for the protocol, I did 20-30 manual corrections at CB throughout this time and was also solely commenting to user requests. Not sure which one you're referring to specifically but unless you did some in the recent days, this sounds a bit unfortunate.

pat-s (approved):
Assuming you tested/applied these commands on active DBs multiple times already, as stated.

6543:
> but unless you did some in the recent days,

On monday to be exact

6543:
-> #4327

PR #4319, opened 2024-11-06
Respect `directory` option for steps again

### LINKED ISSUE #4314: step `directory:` is ignored
after updating woodpecker to the latest image

```
  build-web:
    image: *node_image
    directory: web/
    commands:
      - corepack enable
      - pnpm install --frozen-lockfile
      - pnpm build
```

the directory is ignored

I suspect it is because of #4151  ... future investigation needed :/

cc @pat-s

6543:
close #4314


6543:
-> not tested jet but should fix it

pipeline/backend/kubernetes/pod.go
```diff
@@ -183,7 +183,7 @@ func podContainer(step *types.Step, podName, goos string, options BackendOptions
 	container := v1.Container{
 		Name:            podName,
 		Image:           step.Image,
-		WorkingDir:      step.WorkspaceBase,
+		WorkingDir:      step.WorkspaceBase, // step.WorkingDir will be respected by the generated script // TODO: plugins wont work with this at all!!!
```
6543, line 186: @pat-s what should we do with plugins who explizite wana run with there own entrypoint and as such we can not `cd` via an self generated script

?

lafriks (approved):

pat-s (changes requested):
@lafriks did you test if this works? I am skeptical for k8s as IIRC there was a reason I didn't include `workingDir` in the first place. 

I think the important part was that the mount point was a single dir which then can be adjusted by `fsGroup`. tl;dr, needs testing with PR images.

pipeline/backend/common/script_posix.go
```diff
@@ -53,10 +59,12 @@ fi
 unset CI_NETRC_USERNAME
 unset CI_NETRC_PASSWORD
 unset CI_SCRIPT
-mkdir -p "$CI_WORKSPACE"
-cd "$CI_WORKSPACE"
+mkdir -p "{{.WorkDir}}"
```
lafriks, line 62: this could probably be left as it was

pipeline/backend/common/script_win.go
```diff
@@ -31,16 +40,13 @@ func generateScriptWindows(commands []string) string {
...
-	return script
+
+	return buf.String()
 }
 
-const setupScriptWin = `
+const setupScriptWinProto = `
 $ErrorActionPreference = 'Stop';
-if ([Environment]::GetEnvironmentVariable('CI_WORKSPACE')) { if (-not (Test-Path "$env:CI_WORKSPACE")) { New-Item -Path "$env:CI_WORKSPACE" -ItemType Directory -Force }};
+if (-not (Test-Path "{{.WorkDir}}")) { New-Item -Path "{{.WorkDir}}" -ItemType Directory -Force };
```
lafriks, line 49: same here

lafriks:
I did not test but I don't see why working directory be anyhow related to workspace or permissions (it could be of course relative but not limited to)

pipeline/backend/common/script_posix.go
lafriks: or could be both just in case they do not match

pipeline/backend/kubernetes/pod.go
lafriks: this should have `step.WorkingDir` imho

pat-s:
On a related note: as this is all just for `directory:`: arguably this is a niche param which almost not used anywhere and not doing much else than `cd` to a subdir (for often no reason). If this happens to cause general issues for the whole rootless approach, I'd be in strong favor of removing it.

lafriks:
> On a related note: as this is all just because of `directory:`: arguably this is a niche param which almost not used anywhere and not doing much else than `cd` to a subdir (for often no reason). If this happens to cause general issues for the whole rootless approach, I'd be in strong favor of removing it.

I strongly disagree on this, we use it quite often, especially since for plugins you can't specify commands option to `cd subdir` otherwise plugin itself won't run and would not be secure anymore but there are many cases where for example reviewdog plugins where you need to run eslint only for subdir where frontend code is located and golangci-lint on directory where backend code is, it's also way more clean this way

lafriks:
> As said in my previous comment, it is about the mount point and its permissions which is different for subdirs on k8s. I wouldn't have changed it (i.e. get rid of `workDir`) otherwise in the previous/initial PR.
> 
> While I am not saying that this PR won't work per se (before I tested it), I am not so happy about an (early) approval here without any (practical) tests (given the impact and size of the previous PR) without waiting for at least an reply from the former PR creator (which in this case happens to be me but that doesn't matter overall). I.e, if I wouldn't have put a "please wait" just shortly after, this one would have likely already been merged.

I'm sorry for a bit rushed approval but initially code looked good code wise and as this is something that is pretty much regression for me I approved it

6543:
The direcotory is important for plugins yes :)

I'd like to see both working (rootles & dir) ... I tested docker backend witch works just fine with workdir and an non-root user.

For k8s I need to test things, is pull here is first and formost to discusst getting a good solution.

pat-s (approved):
Tested on k8s. As expected, the ownership doesn't match. Yet somehow also no error are thrown for operations like `mkdir` etc, so it might (luckily) work in practice.

https://ci-dev.devxy.io/repos/1/pipeline/63/4

The linked tests both use a rootless image and allow (at least) an `mkdir` to succeed within `$CI_WORKSPACE`.

TBH I am not too strict on getting the owner:group correct as long as rootless builds work and the user can operate within $CI_WORKSPACE. If this solves the issue and fixes the regression for now, I am fine with it. 

It should be tested in `main` for a while anyhow to find more potential issues and allow refinement.

pipeline/backend/common/script_posix.go
6543: no that's totaly unrelated - and normaly the clone plugin is in charge of makeing it exist

PR #4329, opened 2024-11-06
Bump release plugin

6543:


anbraten (approved):

anbraten:
The manual filter could be deleted as it's not supported

PR #4331, opened 2024-11-06
Cleanup openapi generation

anbraten:
- rename swagger to openapi (as we convert to openapi v3 afterwards & swagger is openapi v2)
- run `swag fmt` to format openapi annotations of routes
- fix `agent_id` param was named `agent`
- fix `version` was using object & string => only string
- use `tools.go` and `go run` to install correct version of `swag`


Makefile
```diff
@@ -109,15 +109,15 @@ clean: ## Clean build artifacts
...
+	rm -rf docs/docs/40-cli.md docs/openapi.json
 
 .PHONY: generate
 generate: install-tools generate-swagger ## Run all code generations
 	CGO_ENABLED=0 go generate ./...
 
-generate-swagger: install-tools ## Run swagger code generation
-	swag init -g server/api/ -g cmd/server/swagger.go --outputTypes go -output cmd/server/docs
-	CGO_ENABLED=0 go generate cmd/server/swagger.go
+generate-swagger: install-tools ## Run openapi code generation and format it
```
6543, line 118: if we rename it all ... I think we can rename the make targets too ...

anbraten: renamed it

cmd/server/tools.go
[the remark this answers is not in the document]
anbraten: moved to `tools/tools.go`, would add the other tools in a separate pr.

6543 (dismissed):

anbraten:
I had the dismiss on new commits option active unintentinally while changing settings and thereby it was dismissed. Code should be ready though.

PR #4336, opened 2024-11-07
Use settings for service

6543:
if service settings work , this pull also pass :)


6543:
works now with https://github.com/woodpecker-ci/woodpecker/pull/4336 applyed & deployed

PR #4340, opened 2024-11-08
Fix avatar column type

### LINKED ISSUE #4278: avatar url too long for pipeline 'avatar' column
### Component

server

### Describe the bug

For some unknown reason (still trying to reproduce) one of our users had their github avatar url generated with a token. For example, a common GitHub avatar url might be:

https://avatars.githubusercontent.com/u/7269710?v=4

While for this user it was showing as:

https://private-avatars.githubusercontent.com/u/2729053?jwt=<token>&v=4

This url had a total of 312 characters. The avatar displays correctly in GitHub, but the full URL with the jwt token is sent by the Github webhook to Woodpecker, which tries to store it in the database. The `users` table uses `character varying(500)` for avatar but the `pipelines` table uses `character varying(255)`.

```
\d+ pipelines;
                                                                   Table "public.pipelines"
        Column        |          Type          | Collation | Nullable |                    Default                     | Storage  | Stats target | Description
----------------------+------------------------+-----------+----------+------------------------------------------------+----------+--------------+-------------
 ...
 avatar               | character varying(255) |           |          |                                                | extended |              |
```

Then we get:

```
pq: value too long for type character varying(255)
```

### Steps to reproduce

Unfortunately we're not sure how to generate a GitHub avatar url with a jwt again. Presumably the user simply changed their avatar normally, by going to their user settings page at https://github.com/<settings> and uploading a new image. It could be a GitHub update or bug, but I couldn't find any information about it.

### Expected behavior

Wether we can reproduce or not, it seems reasonable to use `character varying(500)` for the avatar field in both tables.

### System Info

```shell
2.7.0, Kubernetes
```

### Additional context

_No response_

### Validations

qwerty287:
closes #4278

### DIFF
--- a/server/api/repo.go
+++ b/server/api/repo.go
@@ -362,7 +362,6 @@ func GetRepoPermissions(c *gin.Context) {
 func GetRepoBranches(c *gin.Context) {
 	_store := store.FromContext(c)
 	repo := session.Repo(c)
-	user := session.User(c)
 	_forge, err := server.Config.Services.Manager.ForgeFromRepo(repo)
 	if err != nil {
 		log.Error().Err(err).Msg("Cannot get forge from repo")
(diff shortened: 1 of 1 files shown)


zc-devs:
I know, the issue is about `avatar`, but there is `email` with the same "bug".

xoxys:
Does it require a migration? Or what happens with existing tables?

PR #4346, opened 2024-11-10
Support Git as only VCS

qwerty287:
Only support git as vcs and remove the corresponding env var `CI_REPO_SCM`.


docs/src/pages/migrations.md
```diff
@@ -16,6 +16,7 @@ This will be the next version of Woodpecker.
   - `CI_PIPELINE_FINISHED` as empty during execution
   - `CI_PIPELINE_STATUS` was always `success`
   - `CI_STEP_STATUS` was always `success`
+  - `CI_REPO_SCM`
```
anbraten, line 19: Would be nice if you could add some more details here.

anbraten (dismissed):

6543 (dismissed):
e.g. https://github.com/woodpecker-ci/woodpecker/discussions/2355 integration

and more could come, so it does not add much maintinance to keep it as generic as it is now

6543:
why was this initially proposed at all?!? e.g. what's the reason behind?

6543:
I see the current architecture that we **can** support more than git even as far as an "selling" over other cicd systems

6543:
PS: still thanks for thinking how to slim down and remove legacy code :)

6543:
> we can't build the "egg-laying woolly milk sow"

agree but this is realy not much code and losing it is a shame ...

6543:
**I** as an owner wana veto this change

pat-s:
Without any opinion: Should this PR be closed then?

xoxys:
If a single owner has the power to block PRs at any time, yes.

6543:
well the argument stands: either merge it, in this case I will also merge things, that got denyed from me, because an owner is for it ... or block it untill veto is removed  in this case untill I'm no owner anymore. next election is in 1 months btw. ...

so yes our community gudelines dont have a clear statement of when the owners dissagree - wich would not be a problem if we would have `$owner.mod(2) = 0`, .. witch we currently do.

so I personally interpret it as veto. ?

xoxys:
I just dont understand your point. Right now there is not a single none-git forge supported in woodpecker core right? So this code is unused and useless. To keep it just because we might or might not support none-git forges doesnt make sense to me. Even if we would like to support such forges we could also do it by an official extension in the same way we do it for plugins.

pat-s:
As this is now more a political issue than a code-one, it would be great if you @6543 @anbraten could discuss the general governance part how to proceed in such situations. 

@qwerty287 I think there's not much value in further branch updated until this is resolved?

6543:
another inportant pease of note: I personally wana add mercry support for a long time but it's  just low priority as it requieres to:
 - [ ] create an official mrcury clone plugin
 - [ ] add support for sourcehut ...
 
 and this is not as important as #3411 or  #2254 (I already have some non ready code moving towards it) ...

... so if we merge this pull ... we would soon have to reverte it and remigrate to it, so I dont see why we should do so

6543:
> This is not even necessary. The information that woodpecker stores here is used nowhere. You can use any plugin/container for cloning.

well
https://github.com/woodpecker-ci/woodpecker/blob/ebf9f9ccbb2784b23428032a42ff61dfda7374ef/pipeline/frontend/yaml/compiler/compiler.go#L159-L182

should use it to decide based on the default plugins what to use ...

6543:
well you could argue with altering  #4352 to not have a slice but store the DefaultClonePlugin that can be changed ... and is set on enabling an repo (this logic could adopt to check what vcs is used) you could remove it ...

... but still in this case e.g. plugins like ready-release-go will just fail with no clear error to point users to 

checking against `CI_REPO_SCM==git` can display a nice error msg ... same as missuse of the wrong clone plugin

6543:
and extensions are we... like DKMS modules to kernels ... they get the job done but fail over and over and just shift the burden to the sysadmins - I wana have first class support for different VCS ? what argument is against that?

xoxys:
Can we just do a poll? Sorry, but everyone already wanted someone in Woodpecker that was rejected, and it seems this discussion is going nowhere.

pat-s:
Mercurial or any other VCS are super niche and maybe have a user base of <5%, maybe under 2%?

We never had active requests for other VCS AFAIK. Most maintainers have likely never used anything else than `git`. Hence I don't understand why we would want to keep/extend this part.

Besides, I don't understand why you (@6543) are defending this PR so tremendously? It feels there must be something we don't see/know, also because you haven't acted similarly in the past for other PRs which seemed potentially more important.

Last, only having one person being responsible for such an important cornerstone of a codebase feels risky and might incur more complicated discussions in the future, for both users and maintainers.

Overall, I don't have a strong opinion on this personally but I can't see good (enough) arguments so far why it should be kept.

zc-devs:
Wait, who's here responsible for others? ;)
Having no one responsible person is even worse.

pat-s:
@qwerty287 I am not really sure what this PR is waiting on. I see you rebase frequently. So, we are stuck here somehow and nobody wants to act I guess?

pat-s:
The tendency seems to be in clear favor or removal judging from the last comments and overall discussion. To move on here and get this into 3.0 as it is breaking, I'll merge this now.

pat-s (approved):

xoxys:
Well then dismiss it?

pat-s:
I guess only owners can?

xoxys:
No

image

xoxys:
Done now.

6543:
well I'm still against this cange but It looks like i have to creat a pull and rebase it untill all others see it worth adding it again ... :/

PR #4351, opened 2024-11-11
Improve linter messages in UI

xoxys:
- Render linter errors as markdown string and convert it to HTML in frontend
- Unify wording a bit
- Restructure grid to improve mobile view
- Use existing `DocsLink` component

---

Before:

image

After:

image


6543:
tests fail :sweat_smile:

xoxys:
The error messages will also be used in CLI IIRC. So while the backticks also look ok on the CLI HTML does not.

xoxys:
I would also avoid the possibility to inject complex HTML (classes etc.) from the backend in the frontend, so I think markdown supports basic formatting while we don't need to worry about injected HTML breaks the frontend formatting. But that might be my personal impression only.

xoxys:
Agree with that. What should we do in this particular case? IMO, properly formatting the linter messages in the UI is valuable and makes it more readable.

xoxys:
The remaining test failure looks unrelated to my code changed? :thinking:

web/src/components/atomic/RenderMarkdown.vue
```diff
@@ -0,0 +1,11 @@
+<template>
+  <span v-html="marked(props.content)" />
+</template>
+
+<script setup lang="ts">
+import { marked } from 'marked';
+
+const props = defineProps<{
+  content: string;
```
6543, line 9: about sanitizing ... if there is a script embeded as html coverd as markdown, is it sanitized?

xoxys: No, its not sanitized by default. I have not added sanitizing yet, as it's not really exposed to end-users and only handles code that we control. If you prefer to add sanitizing, I can add it of course.

xoxys: Works for me, will take care later today.

6543 (approved):

Issue #2601, opened 2023-10-17
Replace `NetrcOnlyTrusted` with list of trusted plugins for netrc


anbraten:
As noticed in several PRs the current option `NetrcOnlyTrusted` is limiting users, but on the other hand not really protecting them as well as netrc credentials could be stolen by a custom clone step with custom commands. 

- https://github.com/woodpecker-ci/woodpecker/pull/2585#issuecomment-1765756023
- https://github.com/woodpecker-ci/woodpecker/pull/2214

To prevent this a `pass-netrc-to-plugins` option should replace the `NetrcOnlyTrusted` option. This option would contain list of images which will receive the netrc credentials if they are used as plugins. If the image however uses custom commands it wont get the credentials as those commands could be changed by others than the admin.

This will allow all steps and the clone step to use netrc credentials. It will however be breaking as it wont be possible anymore to use clone steps with custom commands.

pat-s:
>  If the image however uses custom commands it wont get the credentials

Then we're still missing an option that allows the use of credentials in custom steps/images. If the repo admin wants to allow it (e.g. in private repos).
Currently this is only possible on the instance level by setting the repo to "trusted" but not on the repo level.

So why not structure it like this:

1. Option to allow injecting netrc into plugins/trusted images
2. Option to allow injecting netrc into any image

The use case I had was a step executing arbitrary git commands, among them `git push`, to push back to the repo. Many people like this style instead of having to use a plugin as it "feels like freedom" ;)

6543:
is this not solved by simply make

https://github.com/woodpecker-ci/woodpecker/blob/26b77ed677bb6121f35cdba2731692d26bd5b3de/shared/constant/constant.go#L39-L42

as an configuration option?

6543:
an option only instance admins should change in any case

6543:
well that would allow any one to steal other users credentials at codeberg:
1. propose for ci acces
2. create own repo
3. let other users creat pulls
4. you got all there oauth2 tokens :tada:

anbraten:
> you got all there oauth2 tokens 🎉

I am also not for exposing those secrets to easily (as commented before), but your point seems to be wrong.

> Is cloning using the tokens of the pipeline authors? Shouldn't it use the repo users/owners?

It uses the oauth token of the first repo admin (might be updated to another repo admin later not 100% sure).

>  Maybe we should limit this to plugins?

[...]

Issue #4232, opened 2024-10-17
Gerrit Forge


pelme:
### Clear and concise description of the problem

As a user of [Gerrit Code Review](https://www.gerritcodereview.com/), I would like to use Woodpecker for CI. Opening this issue to see if there would be any interest in having this in core Woodpecker as a supported forge or if it would be a better approach to add it as an addon forge.

# Guidelines for adding new forges
As documented (https://woodpecker-ci.org/docs/development/core-ideas#guidelines), a to add a new forge to Woodpecker, it must support webhooks and oauth2: 

## Webhooks
Gerrit supports webooks via the builtin core plugin `webhooks`: https://gerrit-review.googlesource.com/Documentation/config-plugins.html#webhooks.

## OAuth2
Gerrit is not an OAuth2 provider itself, does that disqualify it from being added as a core?There is a popular Gerrit plugin to make Gerrit use OAuth: https://gerrit.googlesource.com/plugins/oauth/  In our setup, we use Google SSO OAuth to authenticate with Gerrit. So we would use Google SSO with Woodpecker too. 

Would it be possible to reconsider the OAuth2 requirement? Or make it possible to have another OAuth2 provider while still having Gerrit support in core Woodpecker?

### Suggested solution

Add Gerrit as a builtin forge. 

### Alternative

Build a Gerrit integration as an addon forge.

### Additional context

My company would potentially be happy to open source/contribute code and/or money towards getting Woodpecker/Gerrit working somehow. Of course we could start building it as an addon forge.

Gerrit is a great tool for git+code review and a lot of people like that flow as an alternative to the typical PR workflow Github/Gitlab/Gitea etc. Gerrit is open source and has been around for a long time and is actively developed. Using Gerrit and Woodpecker together seems like it would be an amazing combination for a lot of people! 🔥 

Both Gerrit and Woodpecker uses birds as their logos/mascots! 🐦‍⬛ 

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

zc-devs:
Even though I do not need Gerrit support, you are rising interested questions, one of them I also thought.

> it must support webhooks and OAuth2

[...]

PR #4352, opened 2024-11-11
Allow to set custom trusted clone plugins

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/2601


pat-s:
How is that different to `WOODPECKER_PLUGINS_TRUSTED_CLONE`?

xoxys:
You can now set it per repo as well.

6543:
> You can now set it per repo as well.

did not checked the code jet - **but ony an instance admin** should be able to change it

6543:
one question that comes into my mind: why not add the config into TrustedConfiguration ?

server/model/repo.go
```diff
@@ -50,7 +50,7 @@ type Repo struct {
 	Hash                         string               `json:"-"                               xorm:"varchar(500) 'hash'"`
 	Perm                         *Perm                `json:"-"                               xorm:"-"`
 	CancelPreviousPipelineEvents []WebhookEvent       `json:"cancel_previous_pipeline_events" xorm:"json 'cancel_previous_pipeline_events'"`
-	NetrcOnlyTrusted             bool                 `json:"netrc_only_trusted"              xorm:"NOT NULL DEFAULT true 'netrc_only_trusted'"`
+	NetrcTrusted                 []string             `json:"netrc_trusted" xorm:"json 'netrc_trusted"`
```
zc-devs, line 53:
```suggestion
	NetrcTrusted                 []string             `json:"netrc_trusted" xorm:"json 'netrc_trusted'"`
```

xoxys:
> > You can now set it per repo as well.
> 
> did not checked the code jet - **but ony an instance admin** should be able to change it

Repeating it doesnt help. Can we keep the discussion in the issue? You never responded to https://github.com/woodpecker-ci/woodpecker/issues/2601#issuecomment-1826871274

server/store/datastore/migration/migration.go
```diff
@@ -47,7 +47,11 @@ var migrationTasks = []*xormigrate.Migration{
 	&addCustomLabelsToAgent,
 	&splitTrusted,
 	&correctPotentialCorruptOrgsUsersRelation,
+<<<<<<< HEAD
```
xoxys, line 50: Rebase leftovers

6543:
> Why the instance admin? This is about per-repo/per-user credentials, so the repo admins should decide how they are used.

on public instances like e.g. codeberg:
1. an malicious aktor creates a new repo (-> repo admin (same is valid for org))
2. change the trusted clone plugin to one that extracts netrc and send it tho him
3. motivate other users to create pull requests to e.g. fix a type or what not
4. other users got there account hijacked

xoxys:
But this would need a private repo then. And if the repo is private you can not easily convince people to contribute.

6543:
limited repos are enouth ...

xoxys:
What are limited repos? Netrc credentials are only required for cloning private repos and at least for gh there is only private/public.

pat-s:
Limited -> logged-in users only
Private -> Fully-private / only visible to explicit org/repo members

---

How about adding a global option which allows changing `WOODPECKER_PLUGINS_TRUSTED_CLONE` on a repo level in the first place?

xoxys:
Ok this only applies to gitea/forgejo then? Global option to toggle it sounds good to me.

anbraten:
> I still don't get why. Cloning should always happen with the credentials from the "repo-user", i.e. the user that activeated the repo. If you now have a malicious repo and somebody creates a PR to this, how can this expose credentials except the repo-user ones?

I think so as well. The person activating a repo (or repairing later on) will be used for cloning. You could only try to ask someone to create an org repo for you and then steal those credentials (which is possible already).

pat-s:
> That's actually impossible because wp can't have the credentials if the user who opens a pr is not registered at the wp instance (for example any non-maintainer on our repos).

Makes sense. Hence, all security concerns don't apply?

zc-devs:
> You could only try to ask someone to create an org repo for you and then steal those credentials

Not necessary to ask. There could be a couple of repo admins in the org. One admin adds a repo, another adds custom image and steals creds of the first. But this is matter of trust between admins of an org. It's unlikely that they would do this in their right minds.

> which is possible already
 
How?

---

This magic should be in the docs (#4232). Seems, Bitbucket Data center works [differently](https://woodpecker-ci.org/docs/administration/forges/bitbucket_datacenter#woodpecker_bitbucket_dc_git_username).

6543:
> > That's actually impossible because wp can't have the credentials if the user who opens a pr is not registered at the wp instance (for example any non-maintainer on our repos).
> 
> Makes sense. Hence, all security concerns don't apply?

uh nice - that's one of the things I was going to have to lookup :)

in this case I'm ok with as is :)

pat-s (approved):
Fine for you given the discussion above.

docs/docs/20-usage/75-project-settings.md
```diff
@@ -39,16 +39,11 @@ Only server admins can set this option. If you are not a server admin this optio
...
 
 The clone step may require Git credentials (e.g. for private repos) which are injected via `netrc`.
 
-By default, they are only injected into trusted clone plugins listed in the env var `WOODPECKER_PLUGINS_TRUSTED_CLONE`.
-If this option is disabled, the Git credentials are injected into every clone plugin, regardless of whether it is trusted or not.
-
-:::note
-This option has no effect on steps other than the clone step.
-:::
+They are only injected into trusted clone plugins listed in the env var `WOODPECKER_PLUGINS_TRUSTED_CLONE` or in this repo setting.
```
anbraten, line 46: Some example what the user can do with this would be quite helpful I guess:

```suggestion
They are only injected into trusted plugins listed in the env var `WOODPECKER_PLUGINS_TRUSTED_CLONE` or in this repo setting.

This allows you to use a trusted plugin for in the clone section or as a step to pull or push using your git credentials.  
```

server/model/repo.go
```diff
@@ -71,7 +71,7 @@ type Repo struct {
 	Hash                         string               `json:"-"                               xorm:"varchar(500) 'hash'"`
 	Perm                         *Perm                `json:"-"                               xorm:"-"`
 	CancelPreviousPipelineEvents []WebhookEvent       `json:"cancel_previous_pipeline_events" xorm:"json 'cancel_previous_pipeline_events'"`
-	NetrcOnlyTrusted             bool                 `json:"netrc_only_trusted"              xorm:"NOT NULL DEFAULT true 'netrc_only_trusted'"`
+	NetrcTrusted                 []string             `json:"netrc_trusted" xorm:"json 'netrc_trusted'"`
```
anbraten, line 74: Calling it sth like `NetrcTrustedPlugins` could make it a bit more descriptive.

web/src/assets/locales/en.json
```diff
@@ -104,8 +104,8 @@
           "desc": "Permit 'deployment' runs for successful pipelines. All users with with push permissions can trigger these, so use with caution."
         },
         "netrc_only_trusted": {
-          "netrc_only_trusted": "Only inject git credentials into trusted clone plugins",
-          "desc": "When enabled, git credentials are accessible only to trusted clone plugins specified in WOODPECKER_PLUGINS_TRUSTED_CLONE. Otherwise, custom clone plugins can use git credentials. This setting has no affect on non-clone steps."
+          "netrc_only_trusted": "Custom trusted clone plugins",
+          "desc": "Plugins listed here will get access to netrc credentials that can be used to clone repositories from the forge."
```
anbraten, line 108:
```suggestion
          "desc": "Plugins listed here will get access to netrc credentials that can be used to clone repositories from the forge or push to it."
```

anbraten (approved):

pat-s (approved):

PR #4354, opened 2024-11-11
address review lint issues

6543:
address lint errors ( https://ci.woodpecker-ci.org/repos/3780/pipeline/22530/31 )


xoxys:
While its nice that we skip some steps/workflows depending on the changed files, we somehow miss some breaking changes that should have been covered by the introduced PR CI run already.

xoxys:
Ah didnt know, thanks. Is there a reason to use latest instead of pinning + automerge?

Issue #4366, opened 2024-11-12
Unnecessary errors in the log output of Woodpecker agent with the Docker backend using a Podman socket


IvoWingelaar:
### Component

agent

### Describe the bug

The logs produced by the Woodpecker agent misleadingly contain unnecessary error entries when running the Docker backend with Podman.

Similar to: https://github.com/woodpecker-ci/woodpecker/issues/85#issuecomment-1836750811

### Steps to reproduce

1. Install the latest Woodpecker server and agent (v2.7.3, both running in containers).
2. Configure the agent to use a rootless Podman socket with the Docker backend. Make sure that the minimum log level is "error" on the agent.
3. Run any pipeline that creates some containers with that agent. Verify that the pipeline finishes without errors in the Woodpecker server.
4. Verify that the environment managed by the Woodpecker agent is cleaned up.
5. View system log output of Woodpecker agent.

### Expected behavior

I expect to see no error messages in the Woodpecker agent logs if the pipeline succeeds fully and the environment managed by the agent is cleaned up afterwards.  However, during cleanup, the following output is produced:

```
Nov 12 14:50:56 test-vm woodpecker-agent[457598]: {"level":"error","error":"Error response from daemon: no container with name or ID \"wp_01JCGDP7ZXN06SS9FE0PMAND10\" found: no such container","time":"2024-11-12T14:50:56Z","caller":"/src/pipeline/backend/docker/docker.go:330","message":"could not remove container 'test'"}
Nov 12 14:50:56 test-vm woodpecker-agent[457598]: {"level":"error","error":"Error response from daemon: no container with name or ID \"wp_01JCGDP7ZXN06SS9FE0PMAND10\" found: no such container","time":"2024-11-12T14:50:56Z","caller":"/src/pipeline/backend/docker/docker.go:327","message":"could not kill container 'test'"}
Nov 12 14:50:56 test-vm woodpecker-agent[457598]: {"level":"error","error":"Error response from daemon: no container with name or ID \"wp_01JCGDP7ZXN06SS9FE0MP2RT54\" found: no such container","time":"2024-11-12T14:50:56Z","caller":"/src/pipeline/backend/docker/docker.go:330","message":"could not remove container 'build'"}
[...]

PR #4367, opened 2024-11-12
Podman doesnt supported

zc-devs:
Addresses #4366

### DIFF
--- a/docs/docs/30-administration/22-backends/10-docker.md
+++ b/docs/docs/30-administration/22-backends/10-docker.md
@@ -18,10 +18,6 @@ FROM woodpeckerci/woodpecker-server:latest-alpine
 RUN apk add -U --no-cache docker-credential-ecr-login
 ```
 
-## Podman support
-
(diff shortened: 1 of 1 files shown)


IvoWingelaar:
I want to strongly share my opposition to this change; two out of the five entries on Woodpecker's [blog](https://woodpecker-ci.org/blog) mention Podman explicitly, and suddenly it becomes not-supported?

zc-devs:
They are user's posts. You can write and share whatever you want, tips and tricks. It is not the documentation.

Also a maintainer said the same
> there isn't any official support for it

No support - no docs :) It's simple to me.

---

But it's not my "war" anymore, let's maintainers decide 😶

6543:
#305 .... etc - so yes I would call the support "experimental" ... but yes it would be nice to have it natively so we can address podman quirks etc ...

6543:
@zc-devs could alter it and add a warning that this is experimental instead of removing it?

zc-devs:
@6543, sorry for not following your request.
Experimental support used to be in Kubernetes backend. At the time there already was dedicated backend.
Experimental support used to be in Forgejo forge. At the time there already was dedicated forge-code and library.
There is no Podman backend, even in experimental stage. Therefore, I've placed it in tips and tricks.

6543 (approved):

PR #4371, opened 2024-11-13
Add EditorConfig-Checker Plugin to docs

6543:
REPO: https://codeberg.org/woodpecker-plugins/editorconfig-checker

DOCS: https://woodpecker-ci-woodpecker-pr-4371.surge.sh/plugins/EditorConfig%20Checker

### DIFF
--- a/.woodpecker/static.yaml
+++ b/.woodpecker/static.yaml
@@ -5,7 +5,7 @@ when:
 
 steps:
   - name: lint-editorconfig
-    image: docker.io/mstruebing/editorconfig-checker:v3.0.3
+    image: docker.io/woodpeckerci/plugin-editorconfig-checker:0.1.0
     depends_on: []
     when:
       - event: pull_request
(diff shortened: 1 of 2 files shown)


xoxys:
Why? Im still not a fan of this wrapping approach as it adds a lot of overhead

xoxys:
Ah this time its a real plugin, nevermind

pat-s:
While I also don't like the idea of shell wrappers as plugins without any additional beef behind it, I think the overall idea is to have a large suite of "plug and play" plugins that are perceived as "secure" compared to normal steps that would execute the same job.

6543:
Plungins are not only for security ... they are also for beginners to not write scripts.

In this case the main benefit is feedback on sucess and the option to only lint changed files (via simple bool option)

As told the target audience of such plugins (prettier, pm, editorconfig, ...) are beginners not advanced users.

And if it exist we cam dogfood ...

I think I need to make an actual survay about this to support my argument with hard numbers 🤔.

xoxys (approved):

Issue #4378, opened 2024-11-14
Refactor: move pipeline compilers getSecretValue() into an compiler Option


6543:
https://github.com/woodpecker-ci/woodpecker/blob/1c7728fae3b5f2f6ce99580066bed6b2f98167d2/pipeline/frontend/yaml/compiler/convert.go#L101
https://github.com/woodpecker-ci/woodpecker/blob/1c7728fae3b5f2f6ce99580066bed6b2f98167d2/pipeline/frontend/yaml/compiler/compiler.go#L97

instead of definding on the fly func and passing al secrets onto the compiler, we can inject an function that can directly query the database ... and for cli the args etc...

6543:
NOTE: if #3723 merged should als be done for it

PR #4380, opened 2024-11-14
Define and implement  secret querier via func interface to let pipeline compiler secrets

6543:
Address #4378

- this only sends secrets used to the agent
- allow for future improvement to inject a secret-querier that query an secret backend directly


cli/exec/exec.go
```diff
@@ -146,13 +146,17 @@ func execWithAxis(ctx context.Context, c *cli.Command, file, repoPath string, ax
...
-	for key, val := range metadata.Workflow.Matrix {
-		environ[key] = val
-		secrets = append(secrets, compiler.Secret{
-			Name:  key,
-			Value: val,
-		})
+
+	secretMap := metadata.Workflow.Matrix
+	secretQuerier := func(name string) (*compiler.Secret, error) {
+		secret, found := secretMap[strings.ToLower(name)]
```
6543, line 152: currently we use case-insensitive for name matiching ... If ok I want to make a breaking change afterwards that make it truly case sensitive ...

pipeline/frontend/yaml/compiler/compiler.go
```diff
@@ -139,14 +141,6 @@ func (c *Compiler) Compile(conf *yaml_types.Workflow) (*backend_types.Config, er
 		Name: fmt.Sprintf("%s_default", c.prefix),
 	})
 
-	// create secrets for mask
-	for _, sec := range c.secrets {
-		config.Secrets = append(config.Secrets, &backend_types.Secret{
```
6543, line 144: this is finally gone :tada: so the agent only has to know the secrets that are actually used

cli/exec/exec.go
6543: PS: I would do that in an dedicated pull

cli/exec/exec.go
```diff
@@ -146,13 +146,17 @@ func execWithAxis(ctx context.Context, c *cli.Command, file, repoPath string, ax
...
 	environ := metadata.Environ()
-	var secrets []compiler.Secret
-	for key, val := range metadata.Workflow.Matrix {
-		environ[key] = val
-		secrets = append(secrets, compiler.Secret{
-			Name:  key,
-			Value: val,
-		})
+
+	secretMap := metadata.Workflow.Matrix
```
6543, line 150: I have no idea why `metadata.Workflow.Matrix` is used in cli for secrets in the first place ?!? ... I think I'll add an dedicated option afterwards in an dedicated pull ...

6543:
-> this destryes the usage of secrets to only maks the log of certain output ... 

I'll worke on an other method to not have to send all secrets to the agent

6543:
-> https://github.com/woodpecker-ci/woodpecker/pull/4384

PR #2552, opened 2023-10-08
Improve repo list

anbraten:
- [x] sort repos by last access
- [x] show repo icon, if public 
- [ ] add endpoint to get a list of last pipelines for a list of #747
  - [ ] user repos closes #1872
  - [ ] repo pull-requests
  - [ ] repo branches
  - [ ] org repos
- [ ] show fallback if no pipeline was executed yet
- [ ] use same list for 
  - [ ] org repos page
  - [ ] repo branches tab
  - [ ] repo pull-request tab

image

PR #4386, opened 2024-11-16
Redesign repo list and include last pipeline

xoxys:
Supersedes: https://github.com/woodpecker-ci/woodpecker/pull/2552

- Use the same list for Repos/OrgRepos
- Show the last four visited Repos
- List all pipelines sorted by last pipeline creation

I tried out many variants of what information should be displayed in the repo list with different layouts and decided on the absolute minimum. Everything else seemed kind of overloaded and cluttered  (especially as it has to work consistently on mobile). However, if you have any ideas, please let me know, but I would like to keep this PR small and focus on the repo list for now.

Repo list (desktop):

image

Repo list (mobile):

OrgRepo list (mobile):


anbraten:
I would suggest to place the status icon in front of the last pipeline message to group all info that belongs to the pipeline. The time feels a bit far away on the desktop screen, however I've no nice idea were to place it instead. To avoid two icons we could remove the pipeline type icon (push, pr, ...). 

Some similar UIs I've found:

image

image

<https://v0.dev/chat/fWkCn1IFMjk?b=b_jUv2KjQbRRj>

image

web/src/lib/api/types/repo.ts
```diff
@@ -65,7 +67,7 @@ export interface Repo {
 
   visibility: RepoVisibility;
 
-  last_pipeline: number;
+  last_pipeline: Pipeline;
```
anbraten, line 70: Isn't this a number coming from the db. If so would rather suggest to use an additional `last_pipeline_item?: Pipeline` to avoid conflicts. Might also be an issue that the last pipeline gets unset when we are receiving updates via the sse stream.

web/src/components/repo/RepoItems.vue
```diff
@@ -0,0 +1,82 @@
...
+      class="flex flex-col border rounded-md bg-wp-background-100 overflow-hidden p-4 border-wp-background-400 dark:bg-wp-background-200 cursor-pointer hover:shadow-md hover:bg-wp-background-300 dark:hover:bg-wp-background-300"
+    >
+      <div class="grid grid-cols-[1.5rem,auto,1fr] gap-2 items-center">
+        <PipelineStatusIcon v-if="pipeline" :status="pipeline.status" />
+        <Icon v-else name="new" class="text-wp-text-100" />
+        <span class="text-wp-text-100 text-lg">{{ `${repo.owner} / ${repo.name}` }}</span>
+        <span class="ml-auto">
+          <Badge
+            v-if="repo.visibility === RepoVisibility.Public"
+            :label="$t('repo.settings.general.visibility.public.public')"
```
anbraten, line 13: Could you bring those translations a bit to the top level so sth like: `visibility.public`.

web/src/store/repos.ts
```diff
@@ -32,9 +32,13 @@ export const useRepoStore = defineStore('repos', () => {
 
   async function loadRepos() {
     const _ownedRepos = await apiClient.getRepoList();
-    _ownedRepos.forEach((repo) => {
-      repos.set(repo.id, repo);
-    });
+    await Promise.all(
+      _ownedRepos.map(async (repo) => {
+        const latestPipeline = await apiClient.getPipelineList(repo.id, { page: 1, perPage: 1 });
```
anbraten, line 37: Nice and easy solution.

web/src/views/Repos.vue
```diff
@@ -8,30 +8,42 @@
...
 const search = ref('');
 
 const { searchedRepos } = useRepoSearch(repos, search);
+const { sortReposByLastAccess, sortReposByLastActivity } = useRepos();
+
+const repoListAccess = computed(() => sortReposByLastAccess(repos.value || []));
+const repoListActivity = computed(() => sortReposByLastActivity(searchedRepos.value || []));
 
-onMounted(async () => {
+router.beforeEach(async () => {
```
anbraten, line 46: This would load the list again each time you navigate to a route (not only this one).

xoxys: Yeah, this was more or less a hack anyway where I wanted to ask for help :see_no_evil: `onMount` behaves weird when moving back and forward in the browser history e.g. the last duration is missing  (see video). Do you have an idea why?

https://github.com/user-attachments/assets/8a8a4b8e-d895-46c4-b693-460fcd58bee2

web/src/components/repo/RepoItems.vue
xoxys: Not sure what you mean. I have just used the existing translation from `en.json` and not added a new one.

web/src/lib/api/types/repo.ts
xoxys: At least, I have not found any ref on the server for this field.

web/src/views/Repos.vue
anbraten: I think when opening the repo it loads the specific repo again and overwrites `last_pipeline` with the number and therefore you loose the `last_pipeline` data (see my other comment).

web/src/components/repo/RepoItems.vue
anbraten: Yep. You are using `repo.settings.general.visibility.public.public` which was for the settings page as you can see in the name. As duplicating such entries makes no sense IMO, I started to move things like `repo.settings.general.visibility.public.public` to a rather flat property like `visibility.public` away from the place it's used to rather match just the content.

xoxys: Ah ok got it :+1: will take care

web/src/views/Repos.vue
xoxys: Ok even if I dont understand where `last_pipeline` (the number) is coming from I have added a new property `last_pipeline_item` as you suggested and this seems to fix the problem. Thanks!

xoxys:
> I would suggest to place the status icon in front of the last pipeline message to group all info that belongs to the pipeline. The time feels a bit far away on the desktop screen, however I've no nice idea were to place it instead. To avoid two icons we could remove the pipeline type icon (push, pr, ...).


Did some rework:

image

web/src/components/repo/RepoItems.vue
xoxys: Done

web/src/views/Repos.vue
xoxys: @anbraten looks like its still not fixed -.- I still see missing information while using the browser navigation. Any other ideas?

pat-s:
I like the new layout. Thanks for putting the time and effort in!

There's still room for UX improvement, but we can go from there and iterate on it. E.g. I think we have too much free space in the middle now where the repo list starts.

Additionally, I don't like the coloring of the "public" badge so much. I know it is already in use like this in the "add repository" list. Personally, I'd prefer no background but only a rounded border with a solid colored text. The gray color doesn't really fit into the light theme well, IMO. But this can also be done in a follow-up PR as this one is mainly about the layout.

Another idea would be to add possible repo avatars, now that we have so much space. And deciding for a visually pleasing default image.

pat-s (approved):

xoxys:
> E.g. I think we have too much free space in the middle now where the repo list starts.

Spacing is always a bit difficult. Too less and it looks weird, too much is wasting space.

> Additionally, I don't like the coloring of the "public" badge so much.

I would like to rework the badges globally, I also don't like them.

> Another idea would be to add possible repo avatars

We can try again, but for me, it was too much and looked overloaded.

web/src/views/Repos.vue
anbraten: Did some adjustments to setting the repos in store as it was overriding the entries again when opening a repo page.

web/src/components/repo/RepoCard.vue
```diff
@@ -0,0 +1,93 @@
...
+          <PipelineStatusIcon v-if="lastPipeline" :status="lastPipeline.status" />
+          <span class="whitespace-nowrap overflow-hidden overflow-ellipsis">{{ shortMessage }}</span>
+        </div>
+
+        <div v-else class="flex gap-x-2">
+          <span>{{ $t('repo.pipeline.no_pipelines') }}</span>
+        </div>
+      </div>
+
+      <div class="col-span-2 text-wp-text-100 text-sm min-h-5 mt-3">
```
anbraten, line 24:
```suggestion
      <div class="col-span-2 text-wp-text-100 text-sm min-h-4 mt-4">
```

Spacings / heights etc should try to be a multiple of 2.

web/src/views/Repos.vue
```diff
@@ -8,10 +8,14 @@
       <Button :to="{ name: 'repo-add' }" start-icon="plus" :text="$t('repo.add')" />
     </template>
 
-    <div class="space-y-4">
-      <ListItem v-for="repo in searchedRepos" :key="repo.id" :to="{ name: 'repo', params: { repoId: repo.id } }">
-        <span class="text-wp-text-100">{{ `${repo.owner} / ${repo.name}` }}</span>
-      </ListItem>
+    <div class="flex flex-col gap-4">
+      <div class="gap-4 grid grid-cols-1 md:grid-cols-2 lg:grid-cols-2">
+        <RepoCard v-for="repo in repoListAccess" :key="repo.id" :repo="repo" />
```
anbraten, line 13: While testing this it was quite confusing to me as repos where duplicated in both lists and I was not able to search in the upper one with the searchable items being quite far down:

image

We should probably decide for one list that might be sorted by both

xoxys:
image

You have removed the spacing at the top block completely now (its not equal to all other card spaces). But IMO this is confusing, the top 4 items a re sorted differently and not reflected in search filter etc. That are two different repo lists and they should be separated visibly.

web/src/views/Repos.vue
anbraten: Just tried the spacing gap-4 spacing for testing wasn't intended to be pushed.

web/src/components/repo/RepoCard.vue
xoxys: I don't agree in general but fine for me.

web/src/views/Repos.vue
xoxys: Personally, I also don't like the two columns layout if we use it for a huge list. Its hard to read and cant be caught easily by eye. It can be used for a short list e.g. pinned repos on GitHub or similar but not for such long repo lists.

anbraten: > The idea to keep both lists was:

That makes sense.

> the last four accessed repos

Did I removed that, as I was getting a longer list?

> don't filter them by search

It feels a bit strange seeing them when searching.

Maybe small titles like: `Last used` and `All repos` might help 🤔

xoxys: > Did I removed that, as I was getting a longer list?

Also possible I have never implemented it :D at least that was what I wanted to do

> Maybe small titles like: Last used and All repos might help 🤔

Yes, good point, that should help 👍 

Do you want to take care of finishing this PR, or should I?

anbraten: I am experimenting a bit right now. My current favorite is sth like the following. I would actually favorite to show only the pipeline message without any further context. One other thing I have to check is if we can update the `lastPipelineItem` when receiving updates to its state (or even showing the new latest pipeline in case we receive such an event).

image

xoxys: I liked the context information but fine for me as well.

> I am experimenting a bit right now.

Ok then I'll leave this PR to you. Ping me if you want me to review again.

xoxys:
The `RepoCard` renaming is IMO a unicorn, while we use `...Item` for almost all other list item templates.

anbraten:
> The RepoCard renaming is IMO a unicorn, while we use ...Item for almost all other list item templates.

Good point. Was `Items` before which confused me.

xoxys:
👍 that was my fault then

anbraten:
Finally got the updating of the last_pipeline to work, so new pipelines will replace the existing last pipelines and it will always show the current pipeline state. 

I've adjusted the lists now to show 4 highlighted repos and the list (without the highlighted repos) by default. When starting to search it only shows the search results. What do you think of that?

xoxys:
For me it doesnt work:

image

xoxys:
And there are now a lot of conflicts.

xoxys:
> Finally got the updating of the last_pipeline to work, so new pipelines will replace the existing last pipelines and it will always show the current pipeline state.

Nice!

> I've adjusted the lists now to show 4 highlighted repos and the list (without the highlighted repos) by default. When starting to search it only shows the search results. 

For me, its too dynamic and too much UI layout changes on interaction, which is IMO a bad UX. At least from your description, I could not test it yet due to the error.

anbraten:
While searching with the highlights staying at the top it was quite tough to notice the results being shown especially when using a mobile device.

xoxys:
Ok, and if we move the search field don't to the "All Repos" list?

xoxys:
That's how it looks right now for me on Firefox:

https://github.com/user-attachments/assets/4d6b6d9f-3ce3-4f11-bd39-6c01a17a2bfe

6543:
just based on the video, an idea poped up: could the transition from hiding and showing be sothed out (e.g. transition animation) - of course you are free to adopt the idea or not

xoxys:
My opinion stands, changing the layout (hiding/moving/transforming parts of the overall page layout) on keyboard inputs is a bad UX and should be avoided in any case.

xoxys:
Ok, after playing around with it, I can't find any better solution. The issue on mobile described by @anbraten is valid, and I agree on that. So to get forward, I would say we live with the tradeoff for now. I just added some changes:

- I still expect the "All repos" list to contain all repos (search is applied o this list)
- Smooth in-out transition while using search
- Layout adjustments to show feedback on empty search result and empty repo lists

I added all changes to a single commit so they can be easily reverted.


https://github.com/user-attachments/assets/292d75e0-9232-4d45-8872-6d42bf8fc79a

anbraten (approved):

pat-s:
Most users don't use mobile anyway I guess? If so, there will be future efforts to improve it. 

Thanks to both of you for finding a consensus and getting this improvement shipped 💙

xoxys:
Pretty confusing post for me. I try to sort it:

> Is this per user info then, right?

If the "last access" list is not user based, we should look into that.

> Can I adjust the number of recently visited repos to be displayed? Can I hide it or disable completely?

No, atm it can't be disabled you can of course hide it by custom CSS.

> But the issue is about repositories list. What is this about?

I don't understand this point. What is your problem/question? 

> What is the order?

Its 1)

> What visibility is it supposed to be?

It's supposed to be not "Public" so "Private" or "Internal" but we can improve this part if that's what you are asking for.

> Is there the N+1 problem? It is event worse: 2*N+1.

Also, something we should investigate.

zc-devs:
> we should look into that
> we should investigate

Thanks in advance.

> No, atm it can't be disabled you can of course hide it by custom CSS
> Its 1

Thanks for answers.

> I planned to rework the badges anyway, so I'll keep it in mind.

👍 

> I don't understand this point. What is your problem/question?

Just tried to understand what is the "pipeline list" you are talking about? Where it is and how I can test it?

xoxys:
> Just tried to understand what is the "pipeline list" you are talking about? Where it is and how I can test it?

Oh, now I got it. That's just a typo in the PR description, sorry for the confusion. It was meant to be:

- List all repos sorted by last pipeline creation

Ill fix it in the PR description

xoxys:
> So, it is Wodpecker's visibility, not the Forge.

Yes it shows the woodpecker state of the repo, whatever you have configured for it in the WP repos settings will apply here.

anbraten:
> Is this per user info then, right?
> 
> Under User 1 I've added some repos. I did not visited that repos under User 2, but they are on top:

It's the last 4 visited repos. If you never visited a repo it shows the first 4 repos from the "general" list (probably sorted by last pipeline time).


Just noticed the repo list for an org doesn't seem to be sorted by pipeline time.

zc-devs:
> List all repos sorted by last pipeline creation

Then I have another feedback.

Repos list used to be sorted by name, right? So, it was kinda static. One could remember at a subconscious level where some repo is. Like it is somewhere "one scroll from the top" and then "on the middle of the page". So, after "Repositories" clicked, I do one scroll and move pointer to the center of the screen. Then find the repo visually, adjust position of the pointer and click.

Now, this list is dynamic, it always changes. I even can do nothing, just hit F5 and there will be completely reordered list, just because my teammates push changes in other repos or some cron jobs. Reminds me of https://www.youtube.com/watch?v=uRGljemfwUE&t=507s 🤣 

It would be great to have sorting setting like in GitHub and default is by name.

xoxys:
Sure, there will always be different use cases and preferences. But IMO, the repo list was quite useless. But I agree, sorting options would be useful. Besides that, I would recommend creating separate issues for all the missing features you discovered.

zc-devs:
> GitHub repo list is sorted by last activity by default 
> in my opinion that should be the default for WP as well even if we add other sort options.

Maybe. At least it is in Gitea.
It would be great if this setting will be saved somewhere.

> same for Issues PR's, etc.

Not for me

[Issues Default](https://github.com/user-attachments/assets/96cdd93b-15c1-4ad7-9757-bfbba8ba8312)
[Issues Recently updated](https://github.com/user-attachments/assets/6fc488c8-61d8-485e-a729-753523bc996d)

[PRs Default](https://github.com/user-attachments/assets/4aec0c1b-d997-408a-8188-e23db2f3868b)
[PRs Recently updated](https://github.com/user-attachments/assets/1c0aa137-dbb3-4368-84ac-533f431f58ad)

PR #4387, opened 2024-11-16
Remove enableManagers from renovate config

xoxys:
Issue introduced in https://github.com/woodpecker-ci/woodpecker/pull/4276

Not sure why this config was added, but all none-experimental managers are enabled by default. Besides that, the correct way is:

```json
  "some-new-manager": {
    "enabled": true
  }
```

`enabledManagers` will automatically disable all managers and only use managers explicitly enabled by this parameter.


anbraten (approved):

pat-s:
Maybe a c/p mistake but strange that it doesn't connect with the other changes. 

Sorry and thanks for the fix!

PR #4393, opened 2024-11-16
Add back button to repo pipeline list

xoxys:
image

Was this button always missing? Or was it removed (by accident) at some point?


6543 (approved):

xoxys:
Ah I see, so we should revert this.

PR #4403, opened 2024-11-18
Untangle settings / header slots

anbraten:
- extract warning from settings component (use slot instead)
- grow search box width
- move docs link after title instead of description (was previously not visible when no description was set)
- rename desc => description and some slot names
- tested for desktop & mobile

Screenshot from 2024-11-18 12-36-23
Screenshot from 2024-11-18 12-36-12
Screenshot from 2024-11-18 12-36-00

Screenshot from 2024-11-18 12-35-00


6543 (approved):

6543:
file conflict :/

PR #4406, opened 2024-11-18
Dont run pipeline on push events to renovate branches

6543:
As we require renovate to create pull requests, the ci coverage will happen at the pull-request event


pat-s (commented):
The initial idea of this was to let renovate automerge if CI checks succeeds for branch runs (which are happening in the background). This is now possible with #3348. This would reduce the renovate noise by a lot.

If checks fail, renovate will automatically create a PR anyway.

The downside is that no labels would be applied to these merge commits, which means it wouldn't play well with the release plugin.
This was one reasoning why I proposed to not only rely on labels but also on conv commits a few weeks ago.

But as this was not agreed on, branch automerge will likely never happen. In this case, removing the branch triggers makes sense.

6543:
thanks for writing **why** it was added in the first place ...

It's better to remove stuff if you know why it existed in the first place :)

I'm personally for let the renovate auto merge, but I do wana se that through pulls ...

pat-s (approved):

6543:
> docs fail unrelated ... need to dig into that ...

PR #4416, opened 2024-11-19
Add filter to list repository pipelines API

6543:
we currently filter thing on the client side:

https://github.com/woodpecker-ci/woodpecker/blob/ce8cd029b4cbc9035bd4572c2b30af5bd128a079/web/src/views/repo/RepoPullRequest.vue#L27-L40

this makes it possible to filter server side in an upcomming pull addressing the frontend


anbraten:
We still need to filter on the client though to get reactive queries updating on changes coming from the sse stream.

6543:
:point_up: well yes but we also need it from the server as getting a random count of entry's even if there should exist more ... is bad

e.g: https://ci.woodpecker-ci.org/repos/3780/branches/main

-> somethimes it just displays 2 ... or 3 ... or some other random number

xoxys:
> but we have https://ci.woodpecker-ci.org/repos/3780/pipeline/15602 ....

Can you explain how server-side filtering help in this case?

6543:
well you currently can not discover it if you dont have the link or are on the specific pull to retrive the link

xoxys:
That was clear already, I still don't understand how server-side filtering will solve it.

6543:
you have to query the server again and use that filter if you switch the pipeline list view to a different one

6543:
If I have time I'll also update the frontend within this pull at some point

server/api/pipeline.go
```diff
@@ -116,14 +117,32 @@ func createTmpPipeline(event model.WebhookEvent, commit *model.Commit, user *mod
 //	@Param			perPage			query	int		false	"for response pagination, max items per page"	default(50)
 //	@Param			before			query	string	false	"only return pipelines before this RFC3339 date"
 //	@Param			after			query	string	false	"only return pipelines after this RFC3339 date"
+//	@Param			branch			query	string	false	"filter pipelines by branch"
+//	@Param			event			query	string	false	"filter pipelines by webhook events (comma separated)"
+//	@Param			ref				query	string	false	"filter pipelines by strings contained in ref"
```
xoxys, line 122: What's the use case for the "ref contains" filter?

xoxys (approved):

PR #4421, opened 2024-11-20
Add icons to all tabs

xoxys:
I would like to add icons to all tabs in pipeline view (and later on also in other views). In my opinion it looks better and results in a consistent look, especially on mobile where tabs are displayed in vertically.

@woodpecker-ci/maintainers as this might be another controversial design decision, please vote on the initial post of this PR with :+1: and :-1: 

---

### Pipelines

Desktop:
image

Mobile:

---

### Repos

image

---

### User settings

image

---

### Admin settings

image
---

### Repo settings

image

### Org settings

image


pat-s:
I like the idea but don't like the icon choice in the screenshot too much.

Thanks for starting with a poll right away. It might be that we need a separate one for each icon 😅️

pat-s:
Here are some possible alternatives up for discussion:

- Tasks: https://pictogrammers.com/library/mdi/icon/list-status/
- Warnings: https://pictogrammers.com/library/mdi/icon/alert-circle-outline/
- Config: https://pictogrammers.com/library/mdi/icon/cog-outline/
- Debug: https://pictogrammers.com/library/mdi/icon/bug-outline/

xoxys:
Thanks for the feedback.

I had something in mind while choosing exactly these items. I was looking for iconst that somehow fit together semantically. I started with a stack of sheets aka workflows/tasks. A sheet from this stack symbolizes a part of the config. Thats why I chose the paper sheet icons for the config and the config parsing issues. The last icon for debugging was the closest I found to inspect a sheet (aka workflow/task). The last but not least all icons have a similar shape to avoid a too chaotic and distracting look. 

Sure there might be "better" icons for each tab but I dont know if that should be the only criteria but Im also not a professional ui designer :)

Nevertheless Im not interested in months of discussions for single icons we have too much of them. It that is the outcome thats totally fine but then Ill leave this task to someone else :)

pat-s:
> I had something in mind while choosing exactly these icons. I was looking for icons that somehow fit together semantically.

OK, didn't know that :)

Let's see what others think!

> m also not a professional ui designer :)

None of us is, so no worries. I guess most people just follow their gut feeling which might be more or less matching with others.

xoxys:
> OK, didn't know that :)

How could you :) Don't want to blame anyone. If you don't like the icons, that perfectly fine.

xoxys:
As written above just wanted to get a consensus at all. But yes I planned to add icons to all tabs (if we can find an agreememt for particular icons)

zc-devs:
Do you use the same icon for Errors tab as for Warnings?
There is no colors (yellow, red) in those icons now, right?

pat-s:
@xoxys Just in case you want to continue here at some point: adding icons to all tabs in general is fine for me. While I don't like most of the icons per se, I would only have a hard veto for the warnings icon (due to already mentioned points).

xoxys:
Yes I know. As it looks like you have a stronger opinion for the UI these days and want to be more involved, feel free to take over.

pat-s:
Don't wanna take over 🙂️ Appreciate that others have opinions regarding UI and want to change something. 
Stronger is relative. If we'd ask 100 people, we likely get 90 different UIs.

So far most people here agreed that the warnings icon should be changed, the others are accepted by all, even if (maybe) not preferred. 🙂️

xoxys:
> [!CAUTION]
> I have also removed the actions tab and moved the 3 buttons to the general settings tab.

image

To me, this makes more sense instead of having an extra tab for only 3 buttons.

web/src/views/repo/settings/Actions.vue
```diff
@@ -1,109 +0,0 @@
-<template>
```
anbraten, line 1: Would put this change in a separate PR.

anbraten: Just leave it without an icon then. I am fine with removing the tab in general, found the save button to be in the middle of the page a bit wrong on the other hand.

xoxys: Fair. Ok as this might need discussion I'll add an icon for now so we can merge this PR in a consistent state.

anbraten (approved):

PR #4429, opened 2024-11-21
Extend approval options (#3348)

anbraten:
Backport of #3348


server/store/datastore/migration/017_gated_to_require_approval.go
```diff
@@ -0,0 +1,72 @@
...
+		if _, err := sess.Exec(
+			builder.Update(builder.Eq{"require_approval": RequireApprovalAllEvents}).
+				From("repos").
+				Where(builder.Eq{"gated": true})); err != nil {
+			return err
+		}
+
+		// migrate public repos to new default require approval
+		if _, err := sess.Exec(
+			builder.Update(builder.Eq{"require_approval": RequireApprovalForks}).
```
anbraten, line 56: Should this stay as no approval at all to not risk breaking any setups?

docs/docs/20-usage/75-project-settings.md
```diff
@@ -25,10 +25,9 @@ Only activate this option if you trust all users who have push access to your re
 Otherwise, these users will be able to steal secrets that are only available for `deploy` events.
 :::
 
-## Protected
+## Require approval for
```
anbraten, line 28: This can stay as we don't deploy docs from 2.8 anyways, right?!

server/api/repo.go
```diff
@@ -236,8 +237,20 @@ func PatchRepo(c *gin.Context) {
...
+			repo.RequireApproval = mode
+		} else {
+			c.String(http.StatusBadRequest, "Invalid require-approval setting")
+			return
+		}
+	} else if in.IsGated != nil { // TODO: remove isGated in next major release
+		if *in.IsGated {
+			repo.RequireApproval = model.RequireApprovalAllEvents
+		} else {
+			repo.RequireApproval = model.RequireApprovalForks
```
anbraten, line 252: Should this stay at no approval to not break api calls?

server/store/datastore/migration/017_gated_to_require_approval.go
xoxys: IMO it should. But if we decide to change the behavior in 2.8 already (which is also valid due to security reasons) we should take care of a clear note in the release notes and upgrade docs.

server/api/repo.go
xoxys: Same here I would say.

server/store/datastore/migration/017_gated_to_require_approval.go
pat-s: I think having a changed default only in 3.0 makes things easier. So far, people also survived without this option and having in 3.0 is better WRT to semver.

6543:
Time to address things?

server/store/datastore/migration/017_gated_to_require_approval.go
anbraten: Changed it to be not breaking. Don't know however, how we would do the migration in 3.0 without overwriting explicit changes made in between.

6543: Add new type `NotSet = ""` and add smal new migration that set the new default in main ... + add smal if check for it and handle it acordingly ...

6543: We can also add a changelog note that gated field is depricated as of v2.8 so we can already remove it in v3.0.0 ...

6543:
with my last commit and https://github.com/woodpecker-ci/woodpecker/pull/4456 this is non breaking anymore :)

server/store/datastore/migration/017_gated_to_require_approval.go
6543: done with https://github.com/woodpecker-ci/woodpecker/pull/4429/commits/a7db4fe94e4d96a49e21f96327feee96dd1a1f4c
and https://github.com/woodpecker-ci/woodpecker/pull/4456

server/api/repo.go
```diff
@@ -236,8 +237,21 @@ func PatchRepo(c *gin.Context) {
...
+
+	if in.RequireApproval != nil {
+		if mode := model.ApprovalMode(*in.RequireApproval); mode.Valid() {
+			repo.RequireApproval = mode
+		} else {
+			c.String(http.StatusBadRequest, "Invalid require-approval setting")
+			return
+		}
+	} else if in.IsGated != nil { // TODO: remove isGated in next major release
+		log.Warn().Msgf("Deprecated 'gated' option was used, use 'require_approval' in the future to update an repo settings.")
```
6543, line 249: might have a better message or is this ok?

docs/docs/20-usage/75-project-settings.md
```diff
@@ -25,10 +25,9 @@ Only activate this option if you trust all users who have push access to your re
...
 
-Every pipeline initiated by an webhook event needs to be approved by a project members with push permissions before being executed.
-The protected option can be used as an additional review process before running potentially harmful pipelines. Especially if pipelines can be executed by third-parties through pull-requests.
+To prevent malicious pipelines from extracting secrets or running harmful commands or to prevent accidental pipeline runs, you can require approval for an additional review process. Depending on the enabled option, a pipeline will be put on hold after creation and will only continue after approval. The default restrictive setting is `Approvals for forked repositories`.
```
6543, line 30:
```suggestion
To prevent malicious pipelines from extracting secrets or running harmful commands or to prevent accidental pipeline runs, you can require approval for an additional review process. Depending on the enabled option, a pipeline will be put on hold after creation and will only continue after approval. The default restrictive setting is `None`.
```

docs/docs/20-usage/75-project-settings.md
```diff
@@ -25,10 +25,9 @@ Only activate this option if you trust all users who have push access to your re
 Otherwise, these users will be able to steal secrets that are only available for `deploy` events.
 :::
 
-## Protected
+## Require approval for
 
-Every pipeline initiated by an webhook event needs to be approved by a project members with push permissions before being executed.
-The protected option can be used as an additional review process before running potentially harmful pipelines. Especially if pipelines can be executed by third-parties through pull-requests.
+To prevent malicious pipelines from extracting secrets or running harmful commands or to prevent accidental pipeline runs, you can require approval for an additional review process. Depending on the enabled option, a pipeline will be put on hold after creation and will only continue after approval. The default restrictive setting is `None`.
```
6543, line 30: I think I'll checkout the release branch to create versioned docs and use that for main then ... once v2.8.0 was released

cli/repo/repo_update.go
```diff
@@ -87,8 +92,29 @@ func repoUpdate(ctx context.Context, c *cli.Command) error {
 	if c.IsSet("trusted") {
 		patch.IsTrusted = &trusted
 	}
+	// TODO: remove isGated in next major release
 	if c.IsSet("gated") {
-		patch.IsGated = &gated
+		if gated {
```
6543, line 97:
```suggestion
		log.Warn().Msgf("Deprecated 'gated' option was used, use 'require-approval' in the future to update an repo settings.")
		if gated {
```

cli/repo/repo_update.go
```diff
@@ -87,8 +92,30 @@ func repoUpdate(ctx context.Context, c *cli.Command) error {
 	if c.IsSet("trusted") {
 		patch.IsTrusted = &trusted
 	}
+	// TODO: remove isGated in next major release
 	if c.IsSet("gated") {
-		patch.IsGated = &gated
+		log.Warn().Msgf("Deprecated 'gated' option was used, use 'require-approval' in the future to update an repo settings.")
```
6543, line 97: same: 

might have a better message or is this ok?

anbraten:
```suggestion
		log.Warn().Msgf("The 'gated' flag was deprecated, use 'require-approval' instead.")
```

server/api/repo.go
anbraten: The server admin does not need to be the repo admin, therefore this information would never reach the actual user and I would remove it.

docs/docs/20-usage/75-project-settings.md
anbraten: For new repos this is actually not required and could stay as is.

server/pipeline/gated.go
```diff
@@ -31,6 +31,7 @@ func needsApproval(repo *model.Repo, pipeline *model.Pipeline) bool {
 		return false
 	}
 
+	// to be non breaking we have this case till v3.0.0
```
anbraten, line 34: Should be a todo or similar

cli/repo/repo_update.go
6543: b142c677d21ac4192542c9590326412defdb14ba

server/pipeline/gated.go
6543:
```suggestion
	// TODO: to be non breaking we have this case till v3.0.0
```

6543:
https://github.com/woodpecker-ci/woodpecker/pull/4429#discussion_r1857164519

> For new repos this is actually not required and could stay as is.

Well I still would say this is a _soft_ breaking change if we add it back ... and the new migration at https://github.com/woodpecker-ci/woodpecker/pull/4456 do take it into account already

6543 (approved):

PR #4447, opened 2024-11-23
Improve code highlighting in web UI

pat-s:
@woodpecker-ci/maintainers as this might be another controversial design decision, please vote on the initial post of this PR with 👍 and 👎

---

This introduces a custom color. However, I found no color of the current palette to be matching here and/or creating a good contrast.
In addition, I slightly tweaked the padding and rounding values.

## Before

image

image

## After

image

image

image


xoxys:
Im not against changing the code block color but not this way. If we want a new one we should create a new color theme with `--wp-code-{100,200,300}`.

xoxys:
Keep in mind its used in various other places as well.

pat-s:
I don't think we need a full palette for a single light/dark code highlighting color.

I agree the naming is not optimal and doesn't fit into the current schema. Open for suggestions.

xoxys:
We dont need a full pallete right now we only have 100 and 200 just change the color of these.

xoxys:
We are getting lost in discussions again. At the end, I don't really care, and `wp-code` works for me as well.

pat-s:
Hm, I think this was just a normal discussion? You brought up some reasons, I replied and shared my thoughts. It is not even about subjective UI but code syntax, so I'd say such discussions are normal (and needed and good).

xoxys:
Sorry this was not meant to be offensive or rude :) The color is only used in one location. While I have some personal preferences, I don't care much enough to discuss further how it is done in detail. `wp-code` works for me we can still change it later on if there is any need for a color palette. Sorry if my last message was misleading.

pat-s:
Consolidated all mixed wp-code definitions into `wp-code-bg` now, which aligns with the logic of `wp-code-text`.

@xoxys Is this acceptable in your view? Yes, the color is custom and not derived from the main colour palettes but I don't think that any value from these matches well here.

Need to run the PR image first to see in which places the changes trigger a change.

xoxys:
Fine for me.

pat-s:
OK, my recent changes also change the bg color of the main log container. More work needed.

xoxys:
Ah yeah, I remember. That's why it's a bit questionable if we don't to keep it equal, but I'm still fine if we want to change it only for real code blocks.

pat-s:
@xoxys 

I refactored to keep the old color definitions and added `var(--wp-code-300)` with a custom color definitions for both light and dark. These are needed as none of the existing palette colors do even closely match a good UI outcome for these inline code boxes as they need to distinguish in a clean way from the surrounding elements.
Is the new approach OK with you?

I also updated the screenshots.

xoxys:
Can we not use the same color as we use for the counter badge already?

image


This would be `colors.gray.200` and to be honest I dont see a big difference and no reason to introduce another custom color. What do you think?

pat-s:
@xoxys I did quite some refactoring. (Sorry for the messy diff due to the sorting.)
To summarize: I aligned the color palettes in their shades both in terms of absolute values as well as for light and dark having matching shades.
E.g. before light and dark did not have matching shades (e.g. light was only up to 300 while dark was up to 400) and the individual shades were also not matching the referenced color palette shades, making it even more confusing.

Now everything is aligned and it should be much less confusing when checking `style.css`.

You primarily want to check the changes in `style.css` and `windi` to spot these changes. For all the color related changes in the UI, deploying the PR image and inspecting visually might be the solution (I guess you already do so anyhow).

Thanks for all the work 💙️

xoxys (changes requested):
Sorry to say, but I don't really like the current state. It lacks a lot of contrast in many areas. On example:

image

For me also, the separator on the top is really missing now:

image

Maybe it was not a good idea to change the overall look and feel in a PR that was nearly ready to merge with the improved code blocks? :see_no_evil:

pat-s:
Weird I didn't catch the log background. Too many places to check I guess.

I'd say having a somewhat consistent color mapping between tailwind and the palette is worth it. 
I don't mind iterating a bit more. The state before was somewhat "wild west", I think you agree.

xoxys:
As I have created this wild west, I see it a bit different :D We need a very granular color pallette of gray tones to get the different layers and overlays working in terms of contrast including over effects. That's why I have picked exactly these colors. While it might look a bit random IMO it was a valid palette where 100 was the lightest value that gets darker with each value.

If we find a better pallette fine for me but I had a hard time to get contrasts working without getting too dark in the light theme and vice versa.

pat-s:
I'd didn't apply color changes to the palette, I mainly aligned the tailwind - palette mappings. 

This caused some tailwind classes to be different. I already updated some to recreate the old contrasts but haven't caught all yet. You're welcome to help to update the ones I missed but I can also do it myself. 

We should be able recreate the old contrasts easily. 
I also used your suggestion to use the existing gray color for the code box backgrounds.

pat-s:
@xoxys You can take another look. I think I was able to keep all previous colors while achieving a much more consistent palette - tailwind mapping that should ease future UI changes.

xoxys:
Uhm the overall look and feel is still completely different and I dont like it:

image

image

xoxys:
image

On this view, the hover/non-hover color is inverted compared to the one in the screenshot above. Non-hover is white hover, is gray. This is IMO the correct behavior but should be used everywhere. The borders are way too dark now.

pat-s:
> Uhm the overall look and feel is still completely different and I dont like it:

I am actually confused why it looks like this now. The last time I checked it looked completely different. 
Not sure, why maybe I did too many PRs in parallel.
Just to state explicitly, that this is not the look I indented.

web/src/components/layout/Settings.vue
```diff
@@ -1,15 +1,15 @@
 <template>
   <Panel>
-    <div class="flex flex-col border-b mb-4 pb-4 justify-center dark:border-wp-background-100">
-      <div class="flex items-center justify-between">
-        <h1 class="text-xl text-wp-text-100 flex items-center gap-1">
+    <div class="flex flex-col justify-center dark:border-wp-background-200 mb-4 pb-4 border-b">
```
xoxys, line 3:
```suggestion
    <div class="flex flex-col justify-center dark:border-wp-background-100 mb-4 pb-4 border-b">
```

web/src/components/admin/settings/queue/AdminQueueStats.vue
```diff
@@ -1,22 +1,22 @@
 <template>
   <div v-if="stats" class="flex justify-center">
     <div
-      class="bg-wp-background-200 border border-wp-background-300 dark:bg-wp-background-100 text-wp-text-100 rounded-md py-5 px-5 w-full"
+      class="border-wp-background-400 bg-wp-background-200 dark:bg-wp-background-100 px-5 py-5 border rounded-md w-full text-wp-text-100"
```
xoxys, line 4:
```suggestion
      class="border-wp-background-400 bg-wp-background-200 dark:bg-wp-background-300 px-5 py-5 border rounded-md w-full text-wp-text-100"
```

xoxys (commented):
`border-wp-background-400` should be changed to `border-wp-background-500` to match the old color.

web/src/style.css
```diff
@@ -49,10 +59,12 @@
 }
 
 :root[data-theme='dark'] {
-  --wp-background-100: theme('colors.int-wp-secondary.300');
-  --wp-background-200: theme('colors.int-wp-secondary.400');
-  --wp-background-300: theme('colors.int-wp-secondary.500');
-  --wp-background-400: theme('colors.int-wp-secondary.600');
+  --wp-background-100: theme('colors.black');
```
xoxys, line 62: This seems to be wrong. 100 should not be the darkest variant.

web/src/components/repo/pipeline/PipelineStepList.vue
```diff
@@ -102,14 +102,14 @@
               :key="step.pid"
               type="button"
               :title="step.name"
-              class="flex p-2 gap-2 border-2 border-transparent rounded-md items-center hover-effect hover:bg-wp-background-300 dark:hover:bg-wp-background-400 w-full"
+              class="flex items-center gap-2 border-2 hover:bg-wp-background-300 dark:hover:bg-wp-background-600 p-2 border-transparent rounded-md w-full hover-effect"
               :class="{
-                'bg-wp-background-300 dark:bg-wp-background-400': selectedStepId && selectedStepId === step.pid,
+                'bg-wp-background-400 dark:bg-wp-background-500': selectedStepId && selectedStepId === step.pid,
```
xoxys, line 107:
```suggestion
                'bg-wp-background-400 dark:bg-wp-background-600': selectedStepId && selectedStepId === step.pid,
```

xoxys:
image

Code bocks in cli view are not covered.

web/src/components/admin/settings/queue/AdminQueueStats.vue
pat-s: Verified.

pat-s:
> Code bocks in cli view are not covered.

This was actually on purpose. The change was only intended for inline code, not code blocks. But alright, lets apply it consistently to all code related boxes then.

xoxys:
Why would we use different styles for inline code blocks and multiline code blocks?

pat-s:
> Why would we use different styles for inline code blocks and multiline code blocks?

My main concern was the hard contrast on the inline code parts. The larger code blocks were actually "OK".

After changing the CLI code blocks, the pipeline log is now also again affected and needs to be decoupled. Unless the light one should also become light in general? (with matching line number colors etc. of course). It wouldn't look too bad with a matching lighter grey tone.

xoxys:
Yeah thats why I said its questionable to change it at all... however changing inlines only is weird too... However the pipeline log should stay as it if we want to change code blocks and Im ok to live with this tradeoff.

pat-s:
> Yeah thats why I said its questionable to change it at all.

It means work but I'd say that these classes should have been decoupled right from the start so they can more easily be changed.

> If you want to propose a light variant of the pipeline log view, feel free to do so as long as we can get enough contrast for all parts.

Maybe in a different PR.

pat-s:
> @pat-s looks like your last commit is not related to this PR?

Actually I just merged `main` locally. Weird it is no merge commit. 

I'll fix it.

web/src/components/agent/AgentList.vue
```diff
@@ -3,11 +3,11 @@
     <ListItem
       v-for="agent in props.agents"
       :key="agent.id"
-      class="items-center !bg-wp-background-200 !dark:bg-wp-background-100"
+      class="items-center !bg-wp-background-200 !dark:bg-wp-background-300"
```
xoxys, line 6: This change seems to be unrelated to this PR?

web/src/components/atomic/ListItem.vue
```diff
@@ -2,9 +2,9 @@
   <component
     :is="to ? 'router-link' : clickable ? 'button' : 'div'"
     :to="to"
-    class="w-full flex border rounded-md bg-wp-background-100 overflow-hidden p-4 border-wp-background-400 dark:bg-wp-background-200"
+    class="flex border-wp-background-400 bg-wp-background-100 dark:bg-wp-background-400 p-4 border rounded-md w-full overflow-hidden"
     :class="{
-      'cursor-pointer hover:shadow-md hover:bg-wp-background-300 dark:hover:bg-wp-background-300': clickable || to,
+      'cursor-pointer hover:shadow-md hover:bg-wp-background-300 dark:hover:bg-wp-background-500': clickable || to,
```
xoxys, line 7: Same here.

web/src/components/atomic/ListItem.vue
```diff
@@ -2,9 +2,9 @@
   <component
     :is="to ? 'router-link' : clickable ? 'button' : 'div'"
     :to="to"
-    class="w-full flex border rounded-md bg-wp-background-100 overflow-hidden p-4 border-wp-background-400 dark:bg-wp-background-200"
+    class="flex border-wp-background-400 bg-wp-background-100 dark:bg-wp-background-400 p-4 border rounded-md w-full overflow-hidden"
```
xoxys, line 5: Same here for ` dark:bg-wp-background-400`

web/src/components/form/TextField.vue
```diff
@@ -2,7 +2,7 @@
   <input
     v-if="lines === 1"
     v-model="innerValue"
-    class="w-full border border-wp-control-neutral-200 py-1 px-2 rounded-md bg-wp-background-100 focus-visible:outline-none focus-visible:border-wp-control-neutral-300"
+    class="border-wp-control-neutral-200 focus-visible:border-wp-control-neutral-300 bg-wp-background-200 px-2 py-1 border rounded-md w-full focus-visible:outline-none"
```
xoxys, line 5: Same here for border and bg color

web/src/components/form/TextField.vue
```diff
@@ -11,7 +11,7 @@
   <textarea
     v-else
     v-model="innerValue"
-    class="w-full border border-wp-control-neutral-200 py-1 px-2 rounded-md bg-wp-background-100 focus-visible:outline-none focus-visible:border-wp-control-neutral-300"
+    class="border-wp-control-neutral-200 focus-visible:border-wp-control-neutral-300 bg-wp-background-200 px-2 py-1 border rounded-md w-full focus-visible:outline-none"
```
xoxys, line 14: Same

web/src/components/layout/Panel.vue
```diff
@@ -1,18 +1,18 @@
 <template>
   <div
-    class="rounded-md w-full shadow overflow-hidden bg-wp-background-100 dark:bg-wp-background-200 border border-wp-background-400"
+    class="border-wp-background-400 bg-wp-background-100 dark:bg-wp-background-400 shadow border rounded-md w-full overflow-hidden"
```
xoxys, line 3: Ill stop here :smile: please revert all border and background color changes.

pat-s:
@xoxys I've adjusted the color scheme ordering, which is why most bg classes are changed. 

But yeah, as I don't need these colors for the code-box adjustments, it might be easier to revert them again to keep the changes smaller.

web/src/style.css
```diff
@@ -39,8 +39,18 @@
   --wp-hint-warn-100: theme('colors.int-wp-hint-warn.100');
   --wp-hint-warn-200: theme('colors.int-wp-hint-warn.200');
 
+  --wp-code-inline-100: theme('colors.gray.200');
+  --wp-code-inline-200: theme('colors.gray.200');
+
+  --wp-code-inline-text-100: theme('colors.gray.600');
+
+  /* --wp-code-300: theme('colors.int-wp-secondary.300');
+  --wp-code-600: theme('colors.int-wp-secondary.600');  */
```
xoxys, line 48:
```suggestion
```

pat-s:
Hm, something is still off with the prettier tailwind sorting in `main` and here even after merging main.

pat-s:
> Hm, something is still off with the prettier tailwind sorting in main and here even after merging main.

prettier tailwind sorting seems to be quite forgiving, e.g. it doesn't enforce within group-sorting and some other instances with overrides were also not accounted for in terms of being at the beginning or end of the class definition.

web/src/components/form/InputField.vue
```diff
@@ -6,7 +6,7 @@
       <slot v-else-if="$slots.titleActions" name="titleActions" />
     </div>
     <slot :id="id" />
-    <div v-if="$slots.description" class="text-wp-text-alt-100 ml-1">
+    <div v-if="$slots.description" class="text-wp-text-alt-100 ml-1 mt-1">
```
pat-s, line 9: This ensures enough margin between the text and the form.

image

lafriks (approved):

pat-s:
@xoxys FYI you're still on block here.

xoxys:
Yes would like to give it a final test but will unblock.

PR #4448, opened 2024-11-23
Various fixes for `awesome.md`

pat-s:
- git.rustybever cannot be reached. Might be temporary 
- wording updates
-

### DIFF
--- a/docs/src/pages/awesome.md
+++ b/docs/src/pages/awesome.md
@@ -1,8 +1,8 @@
 # Awesome Woodpecker
 
-A curated list of awesome things related to Woodpecker CI.
+A curated list of assets (tools, projects, blog posts) related to Woodpecker CI.
 
-If you have some missing resources, please feel free to [open a pull-request](https://github.com/woodpecker-ci/woodpecker/edit/main/docs/docs/92-awesome.md) and add them.
+If you want to add a new entry, open a pull-request against [awesome.md](https://github.com/woodpecker-ci/woodpecker/edit/main/docs/docs/92-awesome.md).
 
 ## Official Resources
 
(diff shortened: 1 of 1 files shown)


docs/src/pages/awesome.md
```diff
@@ -1,8 +1,8 @@
 # Awesome Woodpecker
 
-A curated list of awesome things related to Woodpecker CI.
+A curated list of assets (tools, projects, blog posts) related to Woodpecker CI.
 
-If you have some missing resources, please feel free to [open a pull-request](https://github.com/woodpecker-ci/woodpecker/edit/main/docs/docs/92-awesome.md) and add them.
+If you want to add a new entry, open a pull-request against [awesome.md](https://github.com/woodpecker-ci/woodpecker/edit/main/docs/docs/92-awesome.md).
```
xoxys, line 5: IMO too much details. I would prefer


```suggestion
If you want to add a new entry, open a [pull-request](https://github.com/woodpecker-ci/woodpecker/edit/main/docs/docs/92-awesome.md).
```

xoxys (approved):

xoxys:
Not sure how we want to handle (temp) unavailable resources.

pat-s:
Maybe having a waiting period for ~ 3 days makes sense? It surely differs for the domain at hand, e.g. I'd expect a git.something to be back earlier than a random personal website domain.

pat-s:
The problem is that within that week all CI checks will fail without any action. 

And this causes interruption for automated processed and also takes developer time looking at runs and causes unneeded discussions about the failure per se.

xoxys:
Move linkchecking to a dedicated workflow as we do for security scanning and make it not required.

pat-s:
Good ideas overall.

The check is already optional.

To not block this PR due to this, I've added the resources again, so this PR is not blocked by this discussion.

xoxys (approved):

6543 (approved):

PR #4458, opened 2024-11-25
Change default icon size to 20

xoxys:
In most places, we use icons right beside text. This PR changes the default icon size to `20` as this fits better with the default text size. 

Before:

image

After:

image


pat-s (approved):
Agree that it fits better.

pat-s:
image

This is mad. So when the reviewer merges main as the last person, the review isn't counted 👀

xoxys:
Yeah @qwerty287 was reporting this in our internal chat already. Maybe @woodpecker-ci/owner can check if we can disable it.

xoxys:
Lol... it was enough to add another commit to ensure the last commit was not done by the approver? GitHub you are drunk.

PR #4465, opened 2024-11-26
Trim space from agent token

6543:
so e.g. newlines don't mather if you load the secret from an file


xoxys (approved):

6543:
skip-changelog as bigger followup is at  #4468

PR #4468, opened 2024-11-27
Trim space to all config flags that allow to read value from file

6543:
files often have a newline at the end but we expect simple values, so we should trim space for these options.

based on https://github.com/urfave/cli/issues/2013#issuecomment-2503855416

followup to https://github.com/woodpecker-ci/woodpecker/pull/4465

### DIFF
--- a/cmd/server/flags.go
+++ b/cmd/server/flags.go
@@ -116,6 +116,9 @@ var flags = append([]cli.Flag{
 		Name:  "grpc-secret",
 		Usage: "grpc jwt secret",
 		Value: "secret",
+		Config: cli.StringConfig{
+			TrimSpace: true,
+		},
 	},
 	&cli.StringFlag{
 		Sources: cli.EnvVars("WOODPECKER_METRICS_SERVER_ADDR"),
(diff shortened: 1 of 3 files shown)


cmd/server/flags.go
```diff
@@ -500,11 +524,17 @@ var flags = append([]cli.Flag{
...
 		Usage: "Raw encryption key",
+		Config: cli.StringConfig{
+			TrimSpace: true,
+		},
 	},
 	&cli.StringFlag{
 		Sources: cli.EnvVars("WOODPECKER_ENCRYPTION_TINK_KEYSET_FILE"),
 		Name:    "encryption-tink-keyset",
 		Usage:   "Google tink AEAD-compatible keyset file to encrypt secrets in DB",
+		Config: cli.StringConfig{
```
xoxys, line 535: This one seems to be wrong. It does not load the value from a file.

xoxys (changes requested):

cmd/server/flags.go
6543: hehe right that sliped in

cmd/server/flags.go
```diff
@@ -500,11 +524,17 @@ var flags = append([]cli.Flag{
...
+			TrimSpace: true,
+		},
```
6543, line 537:
```suggestion
```

PR #4475, opened 2024-11-28
Add option to limit the resultset returned by paginate helper

xoxys:
In some cases, it might be helpful to get a fixed number of results from the server. This PR adds a `limit` parameter to the paginate helper to achieve this.


shared/utils/paginate.go
```diff
@@ -15,16 +15,31 @@
...
 // Paginate iterates over a func call until it does not return new items and return it as list.
-func Paginate[T any](get func(page int) ([]T, error)) ([]T, error) {
+func Paginate[T any](get func(page int) ([]T, error), limit int) ([]T, error) {
 	items := make([]T, 0, 10)
 	page := 1
 	lenFirstBatch := -1
 
 	for {
+		// limit < 0 means get all results
+		remaining := -1
```
6543, line 25: i would say 0 should also be all ...  if you ise it and want a list with nothing ... dont use it ...

6543 (approved):

PR #4478, opened 2024-11-29
Use pagination helper to list pipelines in cli

xoxys:
This PR fixes the `woodecker-cli pipeline ls` command. Before this change it was no possible to list more than 50 pipelines as this is the hardcoded limit of the API for pipelines per page.

### DIFF
--- a/woodpecker-go/woodpecker/repo.go
+++ b/woodpecker-go/woodpecker/repo.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 )
 
@@ -31,8 +32,12 @@ const (
 
 type PipelineListOptions struct {
 	ListOptions
-	Before time.Time
-	After  time.Time
+	Before      time.Time
+	After       time.Time
+	Branch      string
+	Events      []string
+	RefContains string
+	Status      string
 }
 
 type CronListOptions struct {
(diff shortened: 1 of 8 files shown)


xoxys:
Yes I know but this was already the case before thos change. 

Yes there is now server side filtering but not for all three options and I would now look into the open autogen client PR instead of copying everything manually to woodpecker-go again.

So can we merge this PR as is?

xoxys:
You are right. Then lets keep it open for now and I will look into to clientgen first.

6543:
> ... not for all three options ...

the woodpecker-go currently does not have support for the new filters yes ... I'm also looking forward to generate it

xoxys:
Depends-on: https://github.com/woodpecker-ci/woodpecker/pull/4494

woodpecker-go/woodpecker/repo.go
```diff
@@ -77,6 +82,18 @@ func (opt *PipelineListOptions) QueryEncode() string {
...
 		query.Add("after", opt.After.Format(time.RFC3339))
 	}
+	if opt.Branch != "" {
+		query.Add("branch", opt.Branch)
+	}
+	if len(opt.Events) > 0 {
+		query.Add("event", strings.Join(opt.Events, ","))
+	}
+	if opt.RefContains != "" {
+		query.Add("ref_contains", opt.RefContains)
```
6543, line 92:
```suggestion
		query.Add("ref", opt.RefContains)
```

6543 (approved):

pat-s (approved):

PR #4480, opened 2024-11-29
Move pipeline logs command

xoxys:
After restructuring the cli we had `woodpecker-cli pipeline log purge` and `woodpecker-cli pipeline logs` command. This PR moves `woodpecker-cli pipeline logs` to `woodpecker-cli pipeline log show` command.


xoxys:
We can if you prefer it that way but we dont do it for any other command.

PR #4490, opened 2024-11-30
 Generate go-client from swagger

xoxys:
Supersedes: https://github.com/woodpecker-ci/woodpecker/pull/2691

- go tidy
- fix tests
- add interface and mocks
- drop stream.go: this functionality does not exist in the old client anyway and can be implemented in a dedicated PR


6543:
https://ci.woodpecker-ci.org/repos/3780/pipeline/23454/42

xoxys:
@anbraten I had to drop the entire history due to the pretty old merge commits, and I was not able to do a proper rebase to main. Sorry about this, if you are not ok with it please let me know!

xoxys:
> https://ci.woodpecker-ci.org/repos/3780/pipeline/23454/42

Yes, I'm aware of it. I would like to silence it for now and do the migration to the new go client in a dedicated PR? What do you think?

6543:
sure just add a rule to the golint file ;)

.mockery.yaml
```diff
@@ -1,2 +1,11 @@
 ---
 disable-version-string: true
+outpkg: 'mocks'
+dir: '{{.InterfaceDir}}/mocks'
+packages:
+  go.woodpecker-ci.org/woodpecker/v2/woodpecker-go/client:
```
6543, line 6: do we need this or can we do this in `woodpecker-go/client` ?

xoxys: We need it. The inline mockery annotation is deprecated.

xoxys:
@qwerty287 This PR is not really breaking, is it? It does not remove the legacy client.

xoxys:
I start to share the opinion from @lafriks https://github.com/woodpecker-ci/woodpecker/issues/3053#issuecomment-1884839660

The autogenerated sdk is really pain to use... No idea if we can improve this somehow, otherwise I also tend to stick to a handcrafted sdk.

xoxys:
Yes the types are nice but also have some issues e.g.

```
	StepId *int          `json:"step_id,omitempty"`
```

should be `StepID` or `stepID` and I dont see a way to control this behavior. So do we just accept it?

xoxys:
Looks like it can even lead to clashed... https://github.com/oapi-codegen/oapi-codegen/issues/452 All in all Im not convinced. I think Ill stick to our maintained sdk.

6543:
This is not the only lib to gen code ... If it has to many downsides, we might look for another lib that does the same ...
... in tje worst case we should automate waht we do manually now ourselfe

pat-s:
So, #2691 can be closed then and removed from the 3.0 milestone?

xoxys:
At least in my opinion its unlikely to be part of v3 and the release should not be blocked by it.

PR #4492, opened 2024-11-30
Update kin-openapi to v0.128.1-0.20241126180935-793b28df2a1e

6543:


xoxys (approved):

6543:
```

# github.com/getkin/kin-openapi/openapi3
../../vendor/github.com/getkin/kin-openapi/openapi3/marsh.go:28:20: undefined: yaml.UnmarshalWithOrigin
cmd/server/openapi.go:24: running "go": exit status 1
```

this I have tested against latest `github.com/invopop/yaml@57832ada4901f7c9054875d2c4bd4c0194178aee ` but that does not have such an API ... so it's probably that they use an outdated invopop/yaml ... 

I guess the best bet is to resolve the todo on there side :/

pat-s:
Seems like this about to get fixed upstream soonish, so we don't need to update to a digest build and can stick with semver?

xoxys:
This PR doesnt even work, so yes we can close it and wait for the upstream fix.

PR #4500, opened 2024-12-01
fix: change default for `CI_COMMIT_BRANCH` and fix default-branch triggers in PR events

### LINKED ISSUE #3560: Change value of "CI_COMMIT_BRANCH" for pulls to source branch
### Describe the bug

Somewhat related to " PR triggers the CI twice #2888 ".

To reproduce:

* Create a simple yaml in `main` branch:

```
steps:
  test-when:
    when: 
      branch: main
    image: ....
    commands: 
      - echo "running on CI_COMMIT_BRANCH=$CI_COMMIT_BRANCH"
```

* Create a new branch `new-branch` based on main and open a PR on Gitea
* Make some changes in `new-branch` branch
* When woodpecker receives the `pull_request_sync` event for the PR branch `new-branch`, it still runs it and considers it as `main` branch
* Expected behavior: such event should be skipped

image

### System Info

```shell
server & client: 2.4.1
```

### Additional context

_No response_

### Validations

pat-s:
fix #3560

`Commit.branch` has been set to the `BaseRef` value in PRs for the linked forges, which equals the target branch. This then lead to a successful match with the default branch name in the `when.Match` logic, causing the behaviour described in the PR.

Changing `Commit.branch` to `HeadRef`, i.e. the branch name of the incoming PR, solves this. 
AFAICS, no other logic makes active use of this, so I think we should be fine.

Tested this in a test repo of mine following the example of the linked issue.

---

Adjusting the value of `CI_COMMIT_BRANCH` is actually a side-effect here, but doesn't relate to the actual issue at hand. As said, the matching is happening between `Commit.branch` and the value of the `when: branch` condition, `CI_COMMIT_BRANCH` is not involved in this.

NB: For BB and BB datacenter, the source ref has already been set.


docs/src/pages/migrations.md
```diff
@@ -17,6 +17,7 @@ This will be the next version of Woodpecker.
   - `CI_PIPELINE_FINISHED` as empty during execution
   - `CI_PIPELINE_STATUS` was always `success`
   - `CI_STEP_STATUS` was always `success`
+- `CI_COMMIT_BRANCH` now always returns the source branch, even for pull requests (previously it returned the target branch)
```
xoxys, line 20: As discussed in the chat yesterday. Please add more context. Why, impact and what do I have to change as a user.

xoxys (changes requested):
Docs adjustmens needed https://github.com/woodpecker-ci/woodpecker/blob/main/docs/docs/20-usage/20-workflow-syntax.md?plain=1#L251

docs/src/pages/migrations.md
pat-s: I think this is the wrong place. This is (currently) a list which briefly lists all migrations without going into much detail.

All points listed there should be explained in more detail in the release notes or in a different place.
Or, if this should be the place, then the whole section should be rewritten in one take rather than doing it for one item here.

xoxys: Ok, but IMO every breaking PRs should already contain the required details notes. Otherwise, someone will have to do a lot of work in release preparation.

xoxys: Why does it not help? At least for this change, we still need it for the v3.0 release or not?

pat-s: Yes, but it would not fit with all the other bullet-list only descriptions that exist right now. 

I committed to do the release notes for 3.0 and hence I have to do all the translation work anyhow. 

It doesn't matter for the *very last* issue of 3.0 now; i.e. forcing a proper migration description within the PR won't change a thing for anyone at this stage.

I don't think this is worth a huge discussion now given all the circumstances 🙃️

xoxys: > Yes, but it would not fit with all the other bullet-list only descriptions that exist right now.

That's why I proposed to create a folder and place a per-pr md file there.

> It doesn't matter for the very last issue of 3.0 now; i.e. forcing a proper migration description within the PR won't change a thing for anyone at this stage.

Well, then we will never change it, and we can skip such governance decisions in the future 🤷

pat-s: I still strongly believe this shouldn't be part of this PR given all the *current* standings but I am repeating myself.

> Well, then we will never change it, and we can skip such governance discussions in the future 

This is a completely different point. I am glad you raised it in the first place. Still I think this is/should be a different discussion/PR. This PR just adopted the scheme for all PRs so far that included some migration notes. I don't think we need a general discussion about a workflow in the (hopefully) very last instance of such a PR. 

> That's why I proposed to create a folder and place a per-pr md file there.

Again, out-of-scope for *this* PR IMO but agreed on.

xoxys:
👍 that's what I already tried to say with https://github.com/woodpecker-ci/woodpecker/pull/4500#pullrequestreview-2484713565

pat-s:
@xoxys Sorry I missed that, just saw your comment on migrations.md back then.

@qwerty287 Yes, agreed.

pat-s:
1. See the migrations PR and previous discussions in this PR 
2. I don't see why this should be documented. After this PR this is the expected behavior and `main` is just one normal branch name then. Why is that a special behavior?

pat-s:
Isn't that the intuitive behavior? A branch is a branch, whether or not a PR is associated with it. If one adds a filter to skip something on branch `X`, one expects that this is honored no matter if the initial trigger came from a `push` or `pull_request` event. 

This is also exactly what has been reported in the linked issue what has now been fixed. 

*Because* this is the expected/intuitive action, I am a bit confused why this needs to be highlighted. If it gets highlighted, it implies that it is *not* the intuitive behavior and needs documentation therefore.

Before this PR, branch filters did not apply on PR branches as the env vars which sets the `branch` value had been set to the target branch, not the source branch. This effectively ignored all branch filters on branches when being triggered via `pull_request` event, even though the branch name matched. (which also brings me back to the point that this PR *is* a bug fix IMO).

zc-devs:
> Isn't that the intuitive behavior?

No.

> env var should be the source
>  the filter should be the target

👍 

[Question](https://github.com/woodpecker-ci/woodpecker/issues/3560#issuecomment-2022662619) has not been answered.

[Are you shooting yourself in the foot](https://github.com/woodpecker-ci/woodpecker/blob/2e490f3c7b442798c39129f1b92999c3d49658b6/.woodpecker/binaries.yaml#L4)?

PR #4515, opened 2024-12-06
Move link checks into cron-curated issue dashboard

pat-s:
Link checks often indicate failed CI checks. While the `static` run is not required, it takes time to inspect the failure and move on.
This is irritating and not helpful, as failing link checks often do not have any relation to the PR at hand.

This change moves the link checks out of the PR checks into a static issue dashboard which is updated daily by a CRON job.

This way:

- `static.yaml` could be added as a required CI check
- False-positive CI check noise gets reduced
- A static issue dashboard shows the up-to-date state of all links

### DIFF
--- a/.woodpecker/links.yaml
+++ b/.woodpecker/links.yaml
@@ -1,0 +1,30 @@
+when:
+  - event: cron
+    cron: links
+
+steps:
+  - name: links
+    image: docker.io/lycheeverse/lychee:0.15.1
+    depends_on: []
+    commands:
+      - lychee pipeline/frontend/yaml/linter/schema/schema.json > links.md
+      - lychee --exclude localhost docs/docs/ >> links.md
+      - lychee --exclude localhost docs/src/pages/ >> links.md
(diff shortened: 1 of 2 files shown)


pat-s:
It shouldn't get closed. Link checks will never stop. Same with dependency checks.

pat-s:
Hm, while I understand the motivation, I don't think this will ever apply. Actually I also like closing issues and keep things tidy, but there is no other way to pin such dashboard-like information in today's git hosters than within "issues".

I still think moving out the link checker from the always-running CI checks is an improvement.

jolheiser:
What about using a "discussion" item instead?

xoxys:
IMO it doesnt harm if there are a hand full pinned tracking issues. I like the idea.

pat-s:
@woodpecker-ci/maintainers Any reviews?

pat-s:
Sure, the question is only what the purpose of this issue would be? 
Simply for visual grouping? Feel free to create one 👍

xoxys:
A dashboard label would make sense and helps with filtering 👍

Issue #3131, opened 2024-01-08
Docker Tag vX.X not working


dR3b:
### Component

other

### Describe the bug

The docker tag `vX.X` (v2.1) (details: https://woodpecker-ci.org/docs/administration/deployment/docker-compose#docker-images) does not work.

### Additional context
```
docker pull woodpeckerci/woodpecker-server:v2.1-alpne
Error response from daemon: manifest for woodpeckerci/woodpecker-server:v2.1-alpne not found: manifest unknown: manifest unknown
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

dR3b:
I don't think that's a good idea. If there is a release 3.0 later, I'll get this version straight away without a warning!

pat-s:
@qwerty287 Using `auto_tag` of the buildx plugin should take care of this automatically - or a slight modification of the manifest template. It doesn't cost anything and the usefulness by @dR3b is well described.

dR3b:
Is there any status update?

pat-s:
Not yet, somebody would need to add the `auto_tag` to all docker release builds. PR welcome :)

PR #4532, opened 2024-12-08
Auto tag docker images

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/3131


pat-s (commented):
Did you test this? `auto_tag` is kinda broken for a long time in the plugin, at least for complex tags.

pat-s:
That's great overall, I forgot about the issue.

I would need to lookup the exact issues, but already the `-alpine` suffix might create problems. 
As this runs only on `tag`, it is kinda dangerous to use it without proper testing. Personally I moved all my repos away from `auto_tag` some time ago. Imo it needs a proper refactor in the plugin.

PR #4535, opened 2024-12-08
disable cron event for securityscan job

pat-s:
- because there is no cron job configured for it and it should not run when running the link checker
- replace curl image with alpine as the former is rootless and hence cannot install `jq` interactively

### DIFF
--- a/.woodpecker/links.yaml
+++ b/.woodpecker/links.yaml
@@ -14,13 +14,13 @@ steps:
       - echo -e "\nLast checked:$(date)" >> links.md
 
   - name: Update issue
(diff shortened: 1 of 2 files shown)


xoxys (approved):
It might make sense to run the security scan frequently via cron. 

> replace curl image with alpine as the former is rootless and hence cannot install jq interactively

You could now also use the exposed `user` docker backend option to switch to the root user.

pat-s:
Yes, the securityscan could also be outsourced to a dashboard. This would also lower the resource pressure on trivvy.

PR #4547, opened 2024-12-10
Add Bitbucket fixes to Release/2.8

j04n-f:
## Description

We currently use a fork of the original code because the following fixes prevent us from using the upstream release. Add the Bitbucket fixes to release 2.8.1.

You are doing an amazing work! Let us know if we can help you.

### DIFF
--- a/server/forge/bitbucket/bitbucket.go
+++ b/server/forge/bitbucket/bitbucket.go
@@ -290,11 +290,11 @@ func (c *config) Dir(ctx context.Context, u *model.User, r *model.Repo, p *model
 }
 
 // Status creates a pipeline status for the Bitbucket commit.
-func (c *config) Status(ctx context.Context, user *model.User, repo *model.Repo, pipeline *model.Pipeline, _ *model.Workflow) error {
+func (c *config) Status(ctx context.Context, user *model.User, repo *model.Repo, pipeline *model.Pipeline, workflow *model.Workflow) error {
 	status := internal.PipelineStatus{
 		State: convertStatus(pipeline.Status),
 		Desc:  common.GetPipelineStatusDescription(pipeline.Status),
-		Key:   "Woodpecker",
+		Key:   common.GetPipelineStatusContext(repo, pipeline, workflow),
 		URL:   common.GetPipelineStatusURL(repo, pipeline, nil),
 	}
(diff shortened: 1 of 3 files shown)


pat-s:
Thanks. Can you please summarize the changes in OP and add the BB version which introduced these changes?
I guess that changing this will break WP for some users on older BB versions?

j04n-f:
As far as I know, WP does not have BB users because the WP - BB integration was broken until we introduced the following changes:

- https://github.com/woodpecker-ci/plugin-git/commit/4dfe5e1aa9a55bcdff84fd0cd68a4673a386ae4f
- https://github.com/woodpecker-ci/woodpecker/commit/3a4ffe951e403548a8829f00627e52e72248f69a
- https://github.com/woodpecker-ci/woodpecker/commit/047eb19d4226a9a84684daf55ea5c7a45c3b34f0

These changes fixed the PR workflow, which was broken. The last changes to the BB code were done almost one and a half years ago. BB did not change anything, this issue regarding the SHA length has been a recurring issue since 2021: https://jira.atlassian.com/browse/BCLOUD-21201

But, as I said, this is a hypothesis. No one complained about BB not working with WP, no one is using the BB integration.

Updated the PR description. What do you mean by "summarize the changes in OP"?

pat-s:
Thanks for the heads-up. Might very well be that there are no BB users. 
Just saw some API changes that could have originated from a major BB release (I guess they must have been at some point) and if it was somewhat recent, they might break stuff.
But you're right, there are probably no users then if it didn't work at all for you until now.

> What do you mean by "summarize the changes in OP"?

Updating the PR description in the [O]riginal [P]ost. Like you now did 👍️

We usually apply such patches first to the `main` branch and then do a backport. While the order doesn't really matter, do you also want to add a PR for `main` after this got accepted? 3.x is round the corner.

zc-devs:
> We usually apply such patches first to the main branch

Screenshot 2024-12-10 1

>  and then do a backport

This is a backport.

pat-s:
Ah, got it. Thanks for clarifying.

pat-s (approved):
LGTM

pat-s:
I would also be fine issuing a patch release right away given that this fix is critical to use WP in the first place for BB users. @woodpecker-ci/maintainers opinions?

PR #4549, opened 2024-12-10
Add bluesky post plugin

anbraten:

### DIFF
--- a/docs/plugins/woodpecker-plugins/plugins.json
+++ b/docs/plugins/woodpecker-plugins/plugins.json
@@ -186,6 +186,11 @@
       "verified": true
     },
     {
+      "name": "Bluesky Post",
+      "docs": "https://codeberg.org/woodpecker-plugins/bluesky-post/raw/branch/main/docs.md",
+      "verified": false
+    },
+    {
       "name": "Discord",
       "docs": "https://raw.githubusercontent.com/appleboy/drone-discord/master/DOCS.md",
       "verified": false


pat-s (approved):

anbraten:
> Why is it not verified?

 Probably as someone said there are too many official plugins.

pat-s:
If anbraten wants to maintain it and puts his name into CODEOWNERS and possibly the README, I think it can live there.

We still don't have a clear rule for where plugins should live WRT to ownership and maintenance. 
One approach could be to say that only maitainers can host plugins in the CB org. But then again, if a maintainer drops out, the plugin stays within the org and someone has to maintain it. Guess there will never be a "perfect" solution.

pat-s (commented):
I think we have two options here:

- "verfiedi" + CB org
- "unverified" + personal username

I'd let @anbraten decide what he wants to do?

xoxys:
This is pending for weeks now, @anbraten @qwerty287 what to do now?

xoxys:
Yeah, but moving the repo owner requires participation by anbraten and there was no response since a month now. So we can either close this PR or change it to `verified` and merge it. To be honest, I would just close this PR for now.

xoxys:
Alright... Then I'll change it to verified now, even if I don't like it and already see no activity on the plugin.

xoxys (approved):
Thanks

xoxys (approved):

Issue #4503, opened 2024-12-02
NGINX 504's on the `/proto.Woodpecker/Next` route between agent and server when using `grpc_pass`


IvoWingelaar:
### Component

server, agent

### Describe the bug

I configured NGINX as a gRPC reverse proxy with TLS offloading between the agents and the server. However, the NGINX logs quickly are filled with lines complaining about timeouts:
```
upstream timed out (110: Connection timed out) while reading response header from upstream, client: x.x.x.x, server: example.org, request: "POST /proto.Woodpecker/Next HTTP/2.0", upstream: "grpc://127.0.0.1:3002", host: "example.org"
```
Which means the access log of NGINX will log 504's being sent back to the clients:
```
[02/Dec/2024:13:27:57 +0000] "POST /proto.Woodpecker/Next HTTP/2.0" 504 167 "-" "grpc-go/1.65.0"
```
From the debug logs of NGINX it can be determined that the timeouts happen after 60s.
Setting `WOODPECKER_KEEPALIVE_TIME=10s` on the agent to try and keep the connection open does nothing.

### Steps to reproduce

1. Install the latest Woodpecker server. Let it listen to `gRPC` on port `3002`.
2. Configure an NGINX reverse proxy to perform `gRPC` TLS offloading as follows:
```
upstream wp {
  server 127.0.0.1:3002;
}

server {
  listen 443 ssl http2;
  listen [::]:443 ssl http2;

  server_name example.org;

  ssl_certificate /etc/example.org/fullchain.pem;
  ssl_certificate_key /etc/example.org/privkey.pem;

  location / {
    grpc_pass grpc://wp;
  }
}
```
3. Configure a Woodpecker agent with `WOODPECKER_SERVER=example.org`, and `WOODPECKER_GRPC_SECURE=true`.
4. Notice the Woodpecker agent can connect to the server and take tasks from the queue successfully, but there are frequent (~1 minute interval) 504's in the `access.log` and upstream timeouts in the `error.log` of NGINX.

### Expected behavior

The agent appears to be polling for new tasks on the `/proto.Woodpecker/Next` route to the server. The implementation of this form of long-polling is fragile, as an intermediary infra-component like NGINX in the path of the request can terminate connections if they are too long lived. The `WOODPECKER_KEEPALIVE_TIME` argument on the agent does not keep the connection alive.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.7.3"}
```

[...]

PR #4551, opened 2024-12-10
Do not log forge tokens

qwerty287:


server/forge/setup/setup.go
```diff
@@ -64,7 +64,7 @@ func setupGitea(forge *model.Forge) (forge.Forge, error) {
 	if len(opts.URL) == 0 {
 		return nil, fmt.Errorf("WOODPECKER_GITEA_URL must be set")
 	}
-	log.Trace().Msgf("Forge (gitea) opts: %#v", opts)
+	log.Trace().Msg("Setting up gitea")
```
zc-devs, line 67: 1. But it was a `trace` log, not even `debug`.
2. What if I want to log/debug forge parameters like URL, SkipVerify and OAuthHost?

zc-devs: 1. Yeah, sure.
2. No :)
2.1. They are go through one transformation [at least](https://github.com/woodpecker-ci/woodpecker/blob/355ebcd508afe107cd5641c4eeb2e4e0c9474795/cmd/server/flags.go#L336). Which [might not work](https://github.com/woodpecker-ci/woodpecker/issues/4503).
2.2. The second transformation might be in some [setup logic](https://github.com/woodpecker-ci/woodpecker/blob/main/server/services/setup.go#L124-L159).
2.3. Also some settings can be changed via API, it appears
Screenshot 2024-12-10 1

6543:
trace log is fine to have secrets it's ment for debugging in development ...

zc-devs:
JFYI, https://github.com/showa-93/go-mask?tab=readme-ov-file#field-name--map-key solves initial intention as well as my comment number 2.
Another solution could be copying the structure, deleting value of `Secret` filed and logging this redacted structure.

pat-s:
Tokens should never be logged. At most you might want to care that they are not empty to ensure they didn't get lost in between. In this case you could print an abbreviated subset of the token.

pat-s:
@qwerty287 This is currently the last issue in the milestone. Do you want to finish it or postpone? I don't think it is that critical that it has to go into 3.0 necessarily.

6543 (approved):
as this specific part works fine I'm ok to limit it's verbosity to harden security ...

... still trace log is for devs or hard to catch bugs ...

server/forge/setup/setup.go
```diff
@@ -64,7 +64,7 @@ func setupGitea(forge *model.Forge) (forge.Forge, error) {
 	if len(opts.URL) == 0 {
 		return nil, fmt.Errorf("WOODPECKER_GITEA_URL must be set")
 	}
-	log.Trace().Msgf("Forge (gitea) opts: %#v", opts)
+	log.Trace().Str("url", opts.URL).Str("oauth-host", opts.OAuthHost).Bool("skip-verify", opts.SkipVerify).Msg("Setting up gitea")
```
zc-devs, line 67: Maybe `info` then? Like
```

{"level":"info","time":"2024-12-14T08:45:52Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/cmd/server/server.go:112","message":"starting Woodpecker server with version 'next-2e490f3c7b'"}
{"level":"info","time":"2024-12-14T08:45:52Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/cmd/server/server.go:117","message":"starting cron service ..."}
{"level":"info","time":"2024-12-14T08:45:52Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/cmd/server/server.go:128","message":"starting grpc server ..."}
{"level":"info","time":"2024-12-14T08:45:52Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/cmd/server/server.go:253","message":"starting http server ..."}
```

xoxys:
Sorry, but why Info log level now? Why should we put that many details into info log?

zc-devs:
IDK, if I may speak or I'm not who was asked. Hope, it won't be an offtopic.

> why Info log level now?

Because it would be aligned with [already existed logs](https://github.com/woodpecker-ci/woodpecker/pull/4551#discussion_r1884984561).

> Why should we put that many details into info log?

It's not necessary, though. I just suggested `info` first.
1. It could be two log lines: `info` with `starting...` and `debug` with details. 
2. Or else just one `debug` line.

xoxys:
I would prefer debug for the additional logs.

pat-s:
My vote goes for `debug` or `trace` log level incl. an abbreviated output of the token to ensure it is not empty (maybe 3 or 4 chars).

xoxys:
What do you think about adding the masked token as well to at least signalize it is set? Something like:

```
logEvent := log.Debug().
    Str("url", opts.URL).
    Str("oauth-host", opts.OAuthHost).
    Bool("skip-verify", opts.SkipVerify)

if opts.Token != "" {
    logEvent = logEvent.Str("token", "********")
}

logEvent.Msg("Setting up gitea")
```

xoxys:
Thats fine for me as well :) I have also seen strings like "_masked_" in logs of other tools but I dont have a strong opinion in how we signalize it.

xoxys (approved):

xoxys:
> It's btw a similar problem for database connections, there we print the full data source as well which contains the password… I don't see an easy way to redact that though without completely omitting the data source

Good point 👍 lets look into it if we can find a solution, otherwise I agree we should drop it for now.

server/forge/setup/setup.go
```diff
@@ -142,7 +142,7 @@ func setupBitbucketDatacenter(forge *model.Forge) (forge.Forge, error) {
 		Password:     gitPassword,
 		OAuthHost:    forge.OAuthHost,
 	}
-	log.Info().Str("url", opts.URL).Str("oauth-host", opts.OAuthHost).Msg("Setting up bitbucketdatacenter")
+	log.Debug().Str("url", opts.URL).Str("oauth-host", opts.OAuthHost).Msg("Setting up bitbucketdatacenter")
```
zc-devs:
```suggestion
	log.Debug().Str("url", opts.URL).Str("oauth-host", opts.OAuthHost).Msg("Setting up Bitbucket Datacenter")
```

xoxys:
Your suggestions can't be applied as you have done them based on an old commit. Instead of hard-coding the names, we should IMO reuse https://github.com/woodpecker-ci/woodpecker/blob/main/server/model/forge.go#L19-L27

server/forge/setup/setup.go
```diff
@@ -44,7 +44,10 @@ func setupBitbucket(forge *model.Forge) (forge.Forge, error) {
 		Client: forge.Client,
 		Secret: forge.ClientSecret,
 	}
-	log.Trace().Msgf("Forge (bitbucket) opts: %#v", opts)
+	log.Debug().
+		Bool("client-set", opts.Client != "").
+		Bool("secret-set", opts.Secret != "").
+		Msg("Setting up bitbucket")
```
xoxys, line 50:
```suggestion
		Msgf("Setting up forge %s", forge.Type)
```

server/forge/setup/setup.go
```diff
@@ -64,7 +67,13 @@ func setupGitea(forge *model.Forge) (forge.Forge, error) {
...
 		return nil, fmt.Errorf("WOODPECKER_GITEA_URL must be set")
 	}
-	log.Trace().Msgf("Forge (gitea) opts: %#v", opts)
+	log.Debug().
+		Str("url", opts.URL).
+		Str("oauth-host", opts.OAuthHost).
+		Bool("skip-verify", opts.SkipVerify).
+		Bool("client-set", opts.Client != "").
+		Bool("secret-set", opts.Secret != "").
+		Msg("Setting up gitea")
```
xoxys, line 76:
```suggestion
		Msgf("Setting up forge %s", forge.Type)
```

server/forge/setup/setup.go
```diff
@@ -84,18 +93,32 @@ func setupForgejo(forge *model.Forge) (forge.Forge, error) {
...
 		return nil, fmt.Errorf("WOODPECKER_FORGEJO_URL must be set")
 	}
-	log.Trace().Msgf("Forge (forgejo) opts: %#v", opts)
+	log.Debug().
+		Str("url", opts.URL).
+		Str("oauth2-url", opts.OAuth2URL).
+		Bool("skip-verify", opts.SkipVerify).
+		Bool("client-set", opts.Client != "").
+		Bool("secret-set", opts.Secret != "").
+		Msg("Setting up forgejo")
```
xoxys, line 102:
```suggestion
		Msgf("Setting up forge %s", forge.Type)
```

server/forge/setup/setup.go
```diff
@@ -84,18 +93,32 @@ func setupForgejo(forge *model.Forge) (forge.Forge, error) {
...
 		OAuthHost:    forge.OAuthHost,
-	})
+	}
+	log.Debug().
+		Str("url", opts.URL).
+		Str("oauth-host", opts.OAuthHost).
+		Bool("skip-verify", opts.SkipVerify).
+		Bool("client-id-set", opts.ClientID != "").
+		Bool("client-secret-set", opts.ClientSecret != "").
+		Msg("Setting up gitlab")
```
xoxys, line 120:
```suggestion
		Msgf("Setting up forge %s", forge.Type)
```

server/forge/setup/setup.go
```diff
@@ -118,7 +141,15 @@ func setupGitHub(forge *model.Forge) (forge.Forge, error) {
...
-	log.Trace().Msgf("Forge (github) opts: %#v", opts)
+	log.Debug().
+		Str("url", opts.URL).
+		Str("oauth-host", opts.OAuthHost).
+		Bool("merge-ref", opts.MergeRef).
+		Bool("only-public", opts.OnlyPublic).
+		Bool("skip-verify", opts.SkipVerify).
+		Bool("client-set", opts.Client != "").
+		Bool("secret-set", opts.Secret != "").
+		Msg("Setting up github")
```
xoxys, line 152:
```suggestion
		Msgf("Setting up forge %s", forge.Type)
```

server/forge/setup/setup.go
```diff
@@ -140,16 +171,21 @@ func setupBitbucketDatacenter(forge *model.Forge) (forge.Forge, error) {
 		Password:     gitPassword,
 		OAuthHost:    forge.OAuthHost,
 	}
-	log.Trace().Msgf("Forge (bitbucketdatacenter) opts: %#v", opts)
+	log.Debug().
+		Str("url", opts.URL).
+		Str("oauth-host", opts.OAuthHost).
+		Bool("client-id-set", opts.ClientID != "").
+		Bool("client-secret-set", opts.ClientSecret != "").
+		Msg("Setting up bitbucketdatacenter")
```
xoxys, line 179:
```suggestion
		Msgf("Setting up forge %s", forge.Type)
```

server/forge/setup/setup.go
```diff
@@ -140,16 +171,21 @@ func setupBitbucketDatacenter(forge *model.Forge) (forge.Forge, error) {
...
 
 func setupAddon(forge *model.Forge) (forge.Forge, error) {
 	executable, ok := forge.AdditionalOptions["executable"].(string)
 	if !ok {
-		return nil, fmt.Errorf("missing git-username")
+		return nil, fmt.Errorf("missing addon executable")
 	}
 
-	log.Trace().Msgf("Forge (addon) executable: %#v", executable)
+	log.Debug().Msgf("Forge (addon) executable: %s", executable)
```
xoxys, line 189:
```suggestion
	log.Debug().Str("executable", executable).Msgf("Setting up forge %s", forge.Type)
```

xoxys:
Yes but it ensures we use the exact same spelling and name everywhere e.g. bitbucket-dc vs bitbucketdatacenter. However, If you prefer to keep the current approach fine for me as well.

zc-devs:
I prefer it be human readable as I suggested.
Or else
```
.Str("type", forge.Type).Msg("Setting up forge")
```

xoxys:
Adjusted the suggestions to use `.Str("type", forge.Type).Msg("Setting up forge")`

xoxys (approved):

xoxys (approved):

PR #4552, opened 2024-12-10
Allow renovate security updates for backport branches

pat-s:
Not fully sure it will work, might need more tweaks, but let's see.

- `dependencyDashboardApproval`: this should restrict renovate to not apply dependency updates but only security updates. Non-security updates should still show in the dependency dashboard and can be triggered manually there.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -2,7 +2,7 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["github>woodpecker-ci/renovate-config"],
   "automergeType": "pr",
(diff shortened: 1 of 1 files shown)


xoxys:
Not sure if I like it. It will create a lot of noise in the dashboard. Why not stick to backports for this process as well?

pat-s:
Because this is tedious, (unneeded) manual work and renovate is perfect for handling that. 

To me, "noise" in the dependency dashhoard is not a strong arguments against renovate doing silent security updates.

pat-s:
Closing then, I guess.

PR #4556, opened 2024-12-11
Set `CI_COMMIT_BRANCH` as source branch for PRs

pat-s:
fix https://github.com/woodpecker-ci/woodpecker/issues/3560

ref #4500

### DIFF
--- a/docs/src/pages/migrations.md
+++ b/docs/src/pages/migrations.md
@@ -17,6 +17,7 @@ This will be the next version of Woodpecker.
   - `CI_PIPELINE_FINISHED` as empty during execution
   - `CI_PIPELINE_STATUS` was always `success`
   - `CI_STEP_STATUS` was always `success`
+- `CI_COMMIT_BRANCH` now always returns the source branch, even for pull requests (previously it returned the target branch)
 - Set `/woodpecker` as default workdir for the **woodpecker-cli** container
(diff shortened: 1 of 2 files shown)


lafriks (approved):

pat-s:
> In the previous PR you wrote that the bitbucket drivers set the branch value to the source branch which would be wrong in this case. Or did I misunderstood that?

https://github.com/woodpecker-ci/woodpecker/blob/defe6c418f6c6daa6af632fe1466528baf4bb8e4/server/forge/bitbucket/convert.go#L180

Even if this is possibly wrong (idk what is behind `from.PullRequest.Source.Branch.Name` for BB exactly), it should not be content of this PR.

> And refspec is also set for pushes so this doesn't break anything, right?

No refspec is touched here, only the content of the env var is adjusted which isn't involved in any filtering processes. It is still breaking as it changes the value of an env var and hence users that make use of it now will see a different behavior.

pat-s:
Argh, good point. It doesn't. I only checked for PRs but yeah, refspec isn't defined for pushes and hence the env var is empty.

Alright, at this point I'll give up. I think the only way to solve this for PRs is to split `branch` into `source_branch` and `target_branch` at some point. 

Changing `CI_COMMIT_BRANCH` is only cosmetic anyhow and won't actually fix the linked issue as the branch filter still applies.

PR #4562, opened 2024-12-12
Use `bun` for docs instead of pnpm

pat-s:
Superseedes #3676 #3440 

Discussion ref #2769 

Building and serving locally worked.


pat-s:
@anbraten Not sure if you have more experience here, but `bun` is working just fine locally (`bun i && bun run build`) and is 10x faster then `pnpm`.

I don't see why CI complains about `docusaurus/faster` being missing, it is listed in the config file and properly added in the lockfiles.

anbraten:
Bun is locally sometimes using just node underneath could you try `bun --bun`

docs/package.json
```diff
@@ -4,8 +4,8 @@
   "private": true,
   "scripts": {
     "start": "cd ../ && make docs && cd docs && docusaurus start",
-    "build": "pnpm build:woodpecker-plugins && docusaurus build",
-    "build:woodpecker-plugins": "cd plugins/woodpecker-plugins && pnpm i && pnpm build",
+    "build": "bun build:woodpecker-plugins && docusaurus build",
+    "build:woodpecker-plugins": "cd plugins/woodpecker-plugins && bun install && bun pm trust --all && bun run build",
```
anbraten, line 8: Should probably stick to workspaces instead

pat-s:
Can't find a solution for the CI errors, locally everything works 🤷️

PR #4526, opened 2024-12-07
Add user as docker backend_option

xoxys:
As Woodpecker is now able to use non-root images https://github.com/woodpecker-ci/woodpecker/pull/4151 it would be helpful to be able to set the run user per step. This PR exposes `user` as backend option for the docker backend.

PR #4573, opened 2024-12-15
Replace discontinued mitchellh/mapstructure by maintained fork

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/pull/4526#issuecomment-2525649383

Replace the discontinued package `github.com/mitchellh/mapstructure` by the maintained fork `github.com/go-viper/mapstructure/v2`. There were some discussions about this fork in https://gist.github.com/mitchellh/90029601268e59a29e64e55bab1c5bdc but as the maintainer of `go-viper/mapstructure` is also the core maintainer `spf13/viper` it is trustworthy enough for me.

### DIFF
--- a/cli/output/table.go
+++ b/cli/output/table.go
@@ -9,7 +9,7 @@ import (
 	"text/tabwriter"
 	"unicode"
 
-	"github.com/mitchellh/mapstructure"
+	"github.com/go-viper/mapstructure/v2"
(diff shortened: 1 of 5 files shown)


pat-s (approved):
And if it doesn't work, we can also switch to another solution or fork.

PR #4574, opened 2024-12-15
Add client error to sdk and fix purge cli

xoxys:
The PR introduces a `ClientError` type to the sdk. This way, the HTTP error returned by the API can be handled more easily.  The PR also fixes a bug in the cli purge command where the purge action of a repo was interrupted on the first error. This behavior was changed to not fail on HTTP 422 errors but log the error and continue with the next pipeline instead.

### DIFF
--- a/woodpecker-go/woodpecker/client.go
+++ b/woodpecker-go/woodpecker/client.go
@@ -34,6 +34,15 @@ const (
 	// pathVersion        = "%s/version"
 )
 
+type ClientError struct {
+	StatusCode int
+	Message    string
+}
+
+func (e *ClientError) Error() string {
+	return fmt.Sprintf("client error %d: %s", e.StatusCode, e.Message)
+}
+
 type client struct {
 	client *http.Client
 	addr   string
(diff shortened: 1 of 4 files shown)


pat-s (approved):
Cool!

Would it be hard to add a basic test regarding the continuation after the first error? I guess if there was one in the first place, you would not have encountered the error in the first place. 

Not a requirement for this PR though.

xoxys:
Test case added.

pat-s (approved):

woodpecker-go/woodpecker/client.go
```diff
@@ -116,16 +125,25 @@ func (c *client) do(rawURL, method string, in, out any) error {
 func (c *client) open(rawURL, method string, in any) (io.ReadCloser, error) {
 	uri, err := url.Parse(rawURL)
 	if err != nil {
-		return nil, err
+		return nil, &ClientError{
+			StatusCode: http.StatusInternalServerError,
+			Message:    err.Error(),
+		}
```
anbraten, line 131: Such an url error is not really related to a http status. Would keep the original errors here instead of using `http.StatusInternalServerError`.

xoxys:
Please dont merge, need to push a cleanup.

woodpecker-go/woodpecker/client.go
xoxys: But doesn't internal server error still fit? I mean an API call failed without a user error.

anbraten: A http call could deliver a status code. Why would `url.Parse` deliver one or is there a reason to do so? The other errors are most of the time already custom types like https://pkg.go.dev/net/url#Error with additional details. Using `ClientError` would drop those details.

woodpecker-go/woodpecker/client.go
```diff
@@ -120,7 +129,10 @@ func (c *client) open(rawURL, method string, in any) (io.ReadCloser, error) {
 	}
 	req, err := http.NewRequest(method, uri.String(), nil)
 	if err != nil {
-		return nil, err
+		return nil, &ClientError{
```
xoxys, line 132: @anbraten Would you revert it here as well?

anbraten: For a similar reason I would avoid extracting / wrapping existing errors there as well.

anbraten (approved):

PR #4575, opened 2024-12-16
Use icon sizes relative to font size

pat-s:
fix https://github.com/woodpecker-ci/woodpecker/discussions/2162#discussioncomment-11579022

Icon are now relative to the font-size in the browser and the height is aligned with the text height.
New classes were added for the badge and forge/settings icons in the repo view, they didn't had any size limits applied before.

Tested with different font-size settings in FF, from 12 to 16. 

Example with size 12 in FF:

image

image

image

## Changes

- Removed unneeded extra padding around status bar icons and aligned them in height
- Aligned the pipeline info icons with the text size next to it
- Pipeline status and avatar icons have the same height now and are a substantially bigger than the text size of the commit message but a little bit smaller than before
- Forge icon and repo settings icon do now scale with the font size instead of being hard-coded in size 
- Fix checkbox and radio button margins (and make them relative to font size)


docs/static/img/feat-opensource.svg
```diff
@@ -1,2 +1,2 @@
...
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" aria-hidden="true" role="img" class="iconify iconify--el" width="1.5em" height="1.5em" preserveAspectRatio="xMidYMid meet" viewBox="0 0 1200 1200"><path d="M600.073 17.426C268.728 17.426 0 286.008 0 617.353c0 260.49 166.117 482.172 398.146 565.076L546.04 791.313c-74.269-23.013-128.273-92.136-128.273-173.96c0-100.646 81.661-182.307 182.308-182.307c100.646 0 182.16 81.66 182.16 182.307c0 81.888-53.952 151.147-128.273 174.106l147.896 391.115C1033.938 1099.72 1200 877.909 1200 617.353c0-331.345-268.581-599.927-599.927-599.927z" fill="currentColor"></path></svg>
```
anbraten, line 2: Using rem instead of em might be even better to avoid issues with parents changing font-size and would follow windi / tailwind and other custom usages of sizes in WP.

xoxys:
Why are there such much diffs related to a changed property order? Makes it a bit hard to review.

pat-s:
I think this comes from some vscode tailwind plugin I am using which automatically sorts the TW classes. I didn't configure this, it seems to be the default. I think it happens when auto-format is applied on file-save.

xoxys (changes requested):
Some parts look weird to me now:

After:
image


Before:
image

pat-s:
ok, let's discuss UI again 🙃️

To me they finally look "not broken" now 😄️ Here are my arguments:

Before, line-height of icons didn't match with the text and it always looked non-harmonic to me.
Now, `icon-height` = `line-height`, making the whole UI much more smooth to me.
Besides, I so far had the assumption that icon size was chosen somewhat randomly as we had different sizes all over the place (20px, 24px, 32px), also in places where IMO they should be identical.

Last, the forge and settings icons are way too big IMO and didn't had any size restrictions set at all. They now match in height with the line height of the owner/repo name to the left.

The same goes for the badge showing the pipeline status. Besides the fact that to me it isn't adding value at this place and could be removed entirely, it is way too large.

Overall my main point is though that **all icons should have a relative size so the chosen font-size** that "matches", i.e. either equals the font-size in line height or has a visible substantial difference (i.e. at least 1.5x).
Right now the difference is like 1.2x for font-size 16 (which is IMO way to small by default) and increases to 2-3x when reducing the font-size, as the icons don't scale alongside.

pat-s:
Your putting my comment somewhat out of context.

I didn't say "all icons should be identical" but some that match context-wise, being it with text attach right next to it (btw you even did #4458 because of that). But it doesn't make any sense from a UI PoV to have icons exceed neighboring elements by a fraction in height. There should be a logic behind sizing and no wild-west approach. Without such a logic, UI is "human bullshit" to me.

There are rules for building UIs (just to use the contrast ratio rules as an example). Similar ones apply to sizing and margins.

I am not bound to these sizes even though I consider them good but I'd like to see meaningful alternative proposals then that don't just follow a gut feeling.

xoxys:
Ok if we want to strictly follow some ui design guidelines now fine but which one? If you are looking at the before/after screenshots, do you really think it looks good? If we want to change the icon sizes, fine, but then the surrounding space is now way too much. The icons for the pipeline status in particular now just look lost. And I still think that especially icon buttons should have the same size regardless of the used icon. Now, every icon button as a completely different size.

pat-s:
> Ok if we want to strictly follow some ui design guidelines now fine but which one?

I am not a UI expert but if you check the icons and line height of this PR on GH, you'll also see that the icons on the right align with the line-height of the attached text.

Not everything needs to go into (this) one PR WRT to related changes (spacing, margins, etc). 

I agree on the point of the surrounding space being to large somewhat. Yet again, I'd like to focus on the icon sizings here and if you don't like a 1:1 that's fine, but please come up with some other numbers than that are not 1.1x line-height like it has been before. Such relative sizes are confusing for the mind as they are so close to "match" but then they don't and your mind realizes this when looking at it and gives you the feeling that "something doesn't add up". If there is a clear difference (like e.g. 2x) then it is clear and non-confusing.

> If you are looking at the before/after screenshots, do you really think it looks good?

Yes, much better than before WRT to sizings.

> And I still think that especially icon buttons should have the same size regardless of the used icon.

I don't know yet what the "icon buttons" are in this context. Do you mean the forge icon? They should all have the same relative size now and the same height as the line-height of the owner/repo button on the left. If not, there's an issue. 
In your screenshot I see that the Gitea icon looks good.

pat-s:
> No the gitea icon doesnt look good. As I said these icons should be squares now the hight is bigger than the width and the width of the forge icon is different than the width of the settings icon button...

Agree on that, height wasn't enforced by mistake.

pat-s:
> but making them way smaller without adjusting the spacing looks ugly and unpolished to me and thats why I will not approve it.

Fair enough. In this case I'll update everything and this PR will get bigger than initially thought, to have the possibility to potentially convince you with respect to styling.

xoxys:
No need to convince me if others are fine with the current approach.

pat-s:
I have increased the icon size to 1.5x now which is quite close to the previous appearance, yet "clearer" than before. Additionally I reduced margins and the most recent commit increases the overall container width a bit. 

> No need to convince me if others are fine with the current approach. What also bothers me is that the status icon is now smaller than the avatar image right beside it. All in all I dont think this path will convince me in general. Maybe I just need to build my own UI 😆

Ideally we find a matching middle-ground that suits most people.

xoxys:
The icon buttons (I mean [IconButton.vue](https://github.com/woodpecker-ci/woodpecker/pull/4575/files#diff-56e6ebfb2e2fc5c30c0a6bb517d063320e80f1a1e63d40be693973974b9c026b)) are still broken. Not even the height matches:

image

silverwind:
I can't recommend rem-based icon sizes. I find it better to fine-tune icon size based on the place they appear in. The vast majority of icons should all be one size (16px), very rarely one needs to divert from that.

pat-s:
Can you please add some justification to your statements? 
I can't see why they shouldn't scale with the font-size, as deviating from the default font-size makes the UI look odd. 16px only makes sense if you go with the browsers default font-size, which is 16px these days. If you change that, the UI looks odd if icons stay static.

Also I just checked GH and GL. Both scale their sidebar icons using `rem`. For GL even the navbar icons adjust.

Besides, I asked ChatGPT and here's the answer

> 	•	For inline, content-related icons (like text links or buttons): Scale with em or rem relative to the font-size.
	•	For standalone or UI icons (like a navigation bar or logo): Keep them static using fixed sizes in px.


Which makes sense to me. I also agree with keeping the navbar icons static, but all others should scale IMO.

Did you try the PR image and changing fonts? It is *much* better than in the overall behavior when changing font size (unrelated to the other topic of what the best default icon sizes are).

xoxys:
@pat-s Have you seen my comment https://github.com/woodpecker-ci/woodpecker/pull/4575#issuecomment-2546547611?

> I have increased the icon size to 1.5x now which is quite close to the previous appearance, yet "clearer" than before. 

Thanks!

> Additionally I reduced margins and the most recent commit increases the overall container width a bit.

Why? IIRC there are also rules available for such things. If you compare it with e.g. the view size of this PR it's now way to broad to be easily catchable for the eyes.

pat-s:
> Why? IIRC there are also rules available for max content-width. If you compare it with e.g. max content-width of this PR view, it's now way too broad to be easily catchable for the eyes.

I thought having a bit more space could help with long commit message. But I agree `max-w-6` is too wide.

I'll tweak things a bit and will ping again once I think a new reviewable version is ready. I need to find a way to restrict the icon size in the navbar natively as right now there is no size argument set and the tailwind classes surrounding it do not always "win" in terms of sizing.

web/src/components/layout/header/Navbar.vue
```diff
@@ -1,39 +1,39 @@
...
-        class="navbar-icon relative"
+        class="relative navbar-icon"
         :title="$t('settings')"
         :to="{ name: 'admin-settings' }"
       >
         <Icon name="settings" />
         <div
           v-if="version?.needsUpdate"
-          class="absolute top-2 right-2 bg-int-wp-state-error-100 rounded-full w-3 h-3"
+          class="top-[0.15rem] right-[0.4rem] absolute bg-int-wp-state-error-100 rounded-full w-3 h-3"
```
xoxys, line 31: I know It's still a draft (let me know if you don't want to have feedback during in draft state) but I'm against such micro-changes. On the one side you would like to generalize things and in this place you switch to custom values to micro-optimize things.

xoxys: Not sure if I like the overall direction 🙈

xoxys:
Mobile nav bar broken:

image

web/src/components/layout/header/Navbar.vue
pat-s: You said before you don't want half-baked solutions for UI, i.e. only changing the relative icon size 🙃️

Hence I am now trying to achieve matching overall picture that works OOB, including aligning icons in height that had a random height before (e.g. in the navbar, the admin settings icon is not aligned with the two ones to its right).

Remember that I also adjusted the icons in size again to match your opinion, not mine ;)

Please have some patience, I'll ping you again once I have something ready "to show".

xoxys:
The icon size beside the text is IMO wrong:

image

As you can see the bottom part from the `g` is out of the lines (while it should be included in the overall line height.

image

pat-s:
@xoxys You can try again with the most recent version. I also added changes and new screenshots in OP.

web/src/components/layout/header/Navbar.vue
xoxys: > You said before you don't want half-baked solutions for UI, i.e. only changing the relative icon size 🙃️

:D ok, point for you, however let's try to work together (I know I can also still improve on this point)

Regarding the Icon, I don't think we need to change it at all. It now looks pretty big. Why do you try to make the icon of the icon button the same size as the avatar? The entire button should match the size, but not the icon:

grafik

xoxys:
https://github.com/woodpecker-ci/woodpecker/pull/4575#discussion_r1890148222

This was not answered/addressed yet, right?

web/src/components/layout/header/Navbar.vue
pat-s: Because there is no adjacent component next to it matching the size. And it that case it looks non-harmonic not having a match with the other two components right next to it. 

Yes, it matches with the line height on the other side. But that is "the other side". One option could also be to move it to the left then. 

I somewhat agree with the overall size though I almost got used to it already in the instance I am running this image on. If this would have been the standard until now, you likely wouldn't have noticed 🙃

xoxys: Ok, for me it's way too big, but I'm ok to keep it that way.

> One option could also be to move it to the left then.

What do you mean with this?

pat-s: Move the admin settings icon to the left of the navbar so it matches visually with the line height and the existing text.

xoxys: Ok, definitely don't want to move it to the left. If that's the only alternative, just keep the current approach.

pat-s:
@xoxys I reverted the size change of the settings icon.

pat-s:
@xoxys Is there still anything missing/blocking from your side?

xoxys:
Need to review it again.

web/src/components/atomic/Icon.vue
```diff
@@ -1,59 +1,59 @@
...
-  <SvgIcon v-else-if="name === 'since'" :path="mdiClockTimeEightOutline" size="20" />
-  <SvgIcon v-else-if="name === 'push'" :path="mdiSourceBranch" size="20" />
-  <SvgIcon v-else-if="name === 'pull-request'" :path="mdiSourcePull" size="20" />
-  <SvgIcon v-else-if="name === 'pull-request-closed'" :path="mdiSourceMerge" size="20" />
-  <SvgIcon v-else-if="name === 'manual-pipeline'" :path="mdiGestureTap" size="20" />
-  <SvgIcon v-else-if="name === 'tag'" :path="mdiTagOutline" size="20" />
-  <SvgIcon v-else-if="name === 'deployment'" :path="mdiPackageVariant" size="20" />
-  <SvgIcon v-else-if="name === 'commit'" :path="mdiSourceCommit" size="20" />
-  <SvgIcon v-else-if="name === 'back'" :path="mdiArrowLeft" size="20" />
+  <SvgIcon v-if="name === 'duration'" :path="mdiTimelapse" size="1.5rem" />
```
xoxys, line 3: The default size for icons is still too big. The line height for text with the default font-size is 20px or ~ 1.3rem not 1.5

pat-s:
> Can we have a bit more inner spacing?

Inner spacing towards? Trying to understand what you mean, also WRT to the old look. 

image

xoxys:
> (Also is there a reason you always highlight the hidden padding div of the element in screenshots?)

Not sure what you mean with this. 

> Inner spacing towards? Trying to understand what you mean, also WRT to the old look.

I thought the space around the icon related to the hover background was bigger, but this has not changed, so fine for this PR.

web/src/components/atomic/Icon.vue
pat-s: Agreed, well spotted!

web/src/components/atomic/Icon.vue
```diff
@@ -66,7 +66,7 @@
     />
     <path
       fill="currentColor"
-      d="M12,4a8,8,0,0,1,7.89,6.7A1.53,1.53,0,0,0,21.38,12h0a1.5,1.5,0,0,0,1.48-1.75,11,11,0,0,0-21.72,0A1.5,1.5,0,0,0,2.62,12h0a1.53,1.53,0,0,0,1.49-1.3A8,8,0,0,1,12,4Z"
+      d="M12,4a8,8,0,0,1,7.89,6.7A1.33,1.33,0,0,0,21.38,12h0a1.3,1.3,0,0,0,1.48-1.75,11,11,0,0,0-21.72,0A1.3,1.3,0,0,0,2.62,12h0a1.33,1.33,0,0,0,1.49-1.3A8,8,0,0,1,12,4Z"
```
xoxys, line 69: Why was this changed?

web/src/components/form/Checkbox.vue
```diff
@@ -40,6 +40,8 @@ const id = (Math.random() + 1).toString(36).substring(7);
 
 <style scoped>
 .checkbox {
+  width: 1.5rem;
+  height: 1.5rem;
```
xoxys, line 44: Should be 1.3rem as well.

web/src/components/form/RadioField.vue
```diff
@@ -42,6 +42,8 @@ const id = (Math.random() + 1).toString(36).substring(7);
 
 <style scoped>
 .radio {
+  width: 1rem;
+  height: 1rem;
```
xoxys, line 46: Should be 1.3rem as well.

web/src/components/form/RadioField.vue
```diff
@@ -54,8 +56,8 @@ const id = (Math.random() + 1).toString(36).substring(7);
   display: block;
   top: 50%;
   left: 50%;
-  width: 8px;
-  height: 8px;
+  width: 0.3rem;
+  height: 0.3rem;
```
xoxys, line 60: After changing the overall size this needs to be adjusted as well to 0.6rem.

web/src/components/atomic/Icon.vue
pat-s: Can't remember I did this. Can revert it.

pat-s:
@xoxys I think the inner part of the radio button is too big with `0.6rem`. What do you think about using `0.4rem`?

0.6:

image


0.4:

image

xoxys:
IMO 0.6 looks better and should be the same ratio we had before. But no strong opinion here.

pat-s:
OK, then let's keep it. Are we good here finally? 🙂️

xoxys (approved):
Thanks for your work.

xoxys:
I guess thats a rendering issue in some browsers while using relative sizes like rem 😔

pat-s:
@qwerty287 Which browser? Checked with FF, Brave and Safari and couldn't reproduce when zooming in/out.

xoxys:
We can check if e.g. 0.5rem has the same issue or 0.65 might fix it.

pat-s:
My first check used a FF fork (Zen). I can reproduce with FF vanilla 👍️. However, given that it fails only for FF vanilla and works for others, I'd argue that this is a FF bug then? (and hence shouldn't block this PR).

xoxys:
Have you tried the mentioned approaches?

pat-s:
Not yet but now. Indeed `0.5rem` seems to work. Did you have this issue before to know this potential workaround?
Still a strange bug IMO.

Setting to 0.5 then.

xoxys (approved):
Yeah, I had this sort of issue already. Good that this hack still works :rofl: Yes, really weird rendering bug.

Tested again, everything else looks good to me.

pat-s:
Great we were able to finish this ✨

PR #4578, opened 2024-12-16
Remove CLI docs from web docs

pat-s:
IMO having the CLI docs just causes noise in the web docs. They are long, not well formatted and I don't think they add much value.

In addition, building docs requires building the CLI first, which takes about a minute in every CI run.

CLI people usually look at the docs in the CLI itself and `grep` for options etc.  

The new `build-docs` makefile rule aligns with the existing `build-ui` rule.


xoxys:
> CLI people usually look at the docs in the CLI itself and grep for options etc. 

This doesn't help for sub-commands or flags and the reason why man pages still exists and used :) I don't see a reason to remove it even if the formatting can be improved for better readability.

pat-s:
> This doesn't help for sub-commands or flags 

Why not?

```
woodpecker-cli org -h
NAME:
   woodpecker-cli org - manage organizations

USAGE:
   woodpecker-cli org [command [command options]] [arguments...]

COMMANDS:
   registry  manage organization registries
   help, h   Shows a list of commands or help for one command

OPTIONS:
   --help, -h  show help
```

> I don't see a reason

It heavily increases/blocks building the docs website and is badly formatted.

xoxys:
Because I need to know that a sub-command `woodpecker-cli pipeline purte` exists to look it up. For me cli docs are valuable, as I said I also use man pages quiet frequently.

pat-s:
> Because I need to know that a sub-command woodpecker-cli pipeline purte exists to look it up

Hm I usually do `<cli> -h` to see subcommands and then `<cli> <subcommand>` to get more info and so on. This works well with CLIs from `ufarve/cli` and most others.

Fair enough, I'll close here then even though I am not happy with the (heavy) build dep for building docs.

Issue #2337, opened 2023-08-28
Skip pipelines waiting for approvals from badge


anbraten:
### Component

server, web-ui

### Describe the bug

![Screenshot from 2023-08-28 08-15-36](https://github.com/woodpecker-ci/woodpecker/assets/6918444/05185001-61f1-4bd1-b609-93ad1cbee5a5)

If the last pipelines are waiting for approvals we currently show a `none` badge. Instead we should skip them and show the status of the last executed / denied pipeline.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-c1731524"}
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

PR #4582, opened 2024-12-17
Ignore blocked pipelines for badge rendering

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/2337

Use the existing pagination helper to find the last pipeline that has a status != blocked. While it works, I'm not sure if that's the best solution. If you have any other suggestion, please let me know. 

image


xoxys:
Sure, sounds like a better idea 😅

xoxys:
Thanks for your review!

PR #4583, opened 2024-12-18
Don't log DB passwords

qwerty287:
as of https://github.com/woodpecker-ci/woodpecker/pull/4551#issuecomment-2550669324


xoxys (approved):

xoxys:
We don't even parse the data source later on... So yeah, you are right, no easy way to get just particular information.

xoxys (approved):

zc-devs:
URL could be split into 5 parts: host, port, database name, user, password.
This way we can not only log parameters and do not log password, but also store host, port, database name in Kubernetes ConfigMap and user, password in Secret.

PR #4414, opened 2024-11-19
Unify pipeline status icons

xoxys:
I know this might be controversial and there were already a lot of back and forth by the pipeline status icons. But right now it is kind of inconsistent, e.g. status `success`, `running` and `skipped` are circled icons while others are not. I would really like to harmonize it and use `circleOutline` icons for all pipeline status icons.

image

PR #4584, opened 2024-12-18
Refactor icons: harmonize contextually, use filled status icons

pat-s:
Follow-up to #4414 which started to use the outline icon-set for the pipeline status icons.

- Harmonized status icons to use `circleFilled` icons
- Harmonized other contextual sub-groups to use `outline` icons
- Squash `alert/error/warning` icons to a single icons and distinguish by color


pat-s:
Ah, thanks for referencing, wasn't aware of it.
The main goal of this PR is to use `circle-outline` icons consistently. We still use many icons from `mdi` which are from another subtype.

A side-component (which I will likely outsource again) is to use colored forge icons from `devicon` instead of the current ones from `simple-icons`.

I had a hard time importing `devicon` without `iconify`. I didn't use the `unplugin` approach but the "web component" one, which seems to be the recommended approach. I am not sure if this one also imports all icons by default or just on demand - do you know more?

And do you see a simple solution for importing `devicon` if not using iconify?

pat-s:
> What's the advantage of devicons over simple-icons here? Just that it's coloured?

Yes, the idea was to propose coloured forge icons but as said, this should actually be a separate PR.

The iconify discussion is only related to this possible change though as the other changes to use `circle-outline` icons doesn't need the `devicon` icon set.

pat-s:
> Ah OK, so we can move this whole discussion about devicons/simple-icons/unplugin/iconify/coloured icons to its own PR? And you revert the changes you made for this?

Yes, this was the initial plan. Was just playing around and misused the PR to test the forge icon change.

xoxys:
This is another all or nothing discussion:

- all icons should be used from the same library 
- all semantically related icons should use the same style
- NOT all used icons need to use the same style

To me, there is no need to use ALL outlined icons whenever possible. But this time I'll leave the decision to someone else 😃

xoxys:
@pat-s what do you think about `alert-circle-outline` for the linter tab instead of the octagon icons?

pat-s:
Fine with me. Overall I am wondering if there should be different icons for error and warning.

pat-s:
Maybe `mdi:stop-circle-outline` in red for errors and `mdi:alert-circle-outline` in yellow for warnings

xoxys:
I dont think the stop icon is really meaningful in this context. Lets stick to alter and different colors I would say.

web/src/components/atomic/Icon.vue
```diff
@@ -31,26 +31,25 @@
...
-  <SvgIcon v-else-if="name === 'error'" :path="mdiAlertCircle" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'gitlab'" :path="mdiGitlab" size="32" />
+  <SvgIcon v-else-if="name === 'error'" :path="mdiAlertOutline" size="1.3rem" />
   <SvgIcon v-else-if="name === 'bitbucket' || name === 'bitbucket-dc'" :path="mdiBitbucket" size="32" />
   <SvgIcon v-else-if="name === 'question'" :path="mdiHelpCircleOutline" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'plus'" :path="mdiPlus" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'plus'" :path="mdiPlusCircleOutline" size="1.3rem" />
   <SvgIcon v-else-if="name === 'list'" :path="mdiFormatListBulleted" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'heal'" :path="mdiBandage" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'heal'" :path="mdiWrenchCogOutline" size="1.3rem" />
```
xoxys, line 41: This icon doesnt look nice IMO

xoxys (changes requested):
I still have some concerns:

- the outline cog ion looks not good at all, it is even harder to see in smaller sizes e.g. if used as service icon in pipelines
image

- the circel-outlined icons look weird if used for buttons, especially for icon buttons
image

I really dont see a reason why we need to use outlines icons everywhere at all costs.

pat-s:
I am not sure how to proceed. I feel there are many UI-related points that can be improved by a lot and don't wanna push all of them into a fork. But I see that finding a common ground is hard and we also don't have a clear governance idea how to approach these discussions. Additionally, these discussions take a lot of time and energy.

To me, all the icon changes of this PR are *clear* visual improvements as they use a consistent icon set all over the place. But I don't wanna force anything and since there is no additional response and governance rule, I guess I need to close this PR?

xoxys:
I dont think we need to close it but if you think the discussion is too time-consuming then I dont see how to get forward. To me the mentioned two points are UX issues and cant be ignored.

- The outlined cog works in most places but have you checked it in the pipeline list where it is used for services? The screenshot does not really reflect it as it is zoomed by GitHub but in reality it is so tiny that the symbol itself is not even really recognizable anymore. Do you disagree?

- The pencil. Why do we really need the circle around it? The pencil itself is not even outlined at all. Its the same pencil symbol we had before, just surrounded by a circle. It looks really weird in this particular use case (buttons) and as the pencil is exactly the same for the outlined and non-outlined icon I really dont understand why you want to use the outlined one here.

pat-s:
> I dont think we need to close it but if you think the discussion is too time-consuming then I dont see how to get forward.

I mean, I kinda feel sorry for both you and me. This costs a lot of time and energy, doesn't it?

> The outlined cog works in most places but have you checked it in the pipeline list where it is used for services?

Not in services, no. On the screenshot it looks fine.
I can see the argument with the icon being too small so that the outlines are not clearly visible anymore. Need to check it again.

> The pencil. Why do we really need the circle around it? 

My argument would to to have a consistent line in the icons, as said before. I don't have strong feelings about the pencil icon in either way (outlined or not). The main point is just that it feels somewhat "random" to me to sometimes use icons from other subsets for no apparent reason (pencil, plus, pause).

xoxys:
Ok to me its not really random we use a consistent icon set and I dont see why we can just use outlined or not outlined ones. But thats just my opinion of course if we really want to go that route then my vote is going the other way around and drop all outlines icons as they have mentioned issues at some sizes.

pat-s:
> Well last time you argued discussions in PRs are a regular process and required to find a consensus and I would agree. So no I dont think its worthless.

Yes. I guess I am just doing a bit too much and opened too many controversial PRs so I ran out of energy myself 🙃️

I checked the cog-outline one in services now. Using font-size 12 on my Lap. 
I gotta say it actually looks fine to me WRT to the icon shape and a possible line squash/overlap. Even at this font-size/zoom size I can see it just fine 🤷. I'd even say it is "better" than before as the non-outlined icon looks like an outlier compared to the other outlined ones surrounding it. 

image

Here, it sticks out much more and looks kinda "unharmonic" to me - not sure how else I should express it.

image

In this version it would pair better with non-outlined status icons.

So, I am not sure how to proceed. I am not bound to the pencil or others per se and I see your point that some are not a great fit even though they "fit" into the icon set subcategory. But some kind of subjective taste will always stick within these discussions (btw I also would prefer non-outlined status icons but this is another discussion again).

---

So, what kind of middle-way can we find, given that nobody else seems to want to comment here? From the current state I think you have a hard opinion on reverting `cog-outline` and `pencil-outline`? I am fine with this then, as we still have a lot of icons that are more consistent than before.

pat-s:
I just found by chance that MS uses a different outlined gear icon that could actually fit well. It is rotated a bit and the borders are a bit thinner, making it easier to look at smaller scales. Couldn't find out yet what the icon set is though. 

image

xoxys:
But if you prefer non-outlined status icons why not drop outlined icons at all?

pat-s:
😄️ Because this would be an even bigger change and you introduced the new outlined ones recently for the pipeline status icons and everyone else agreed on this change. So I assumed this is fixed and I didn't wanted to propose a full changeover for these (and others to stay somewhat consistent).

xoxys:
Ok. I just think it makes more sense as we use more non-outlined icons than outlined.

pat-s:
Wait: you mean change the status icons (back) to non-outlined ones (in a different PR) and use more non-outlined ones overall?

xoxys:
Not really back to the old ones but Im ok to use e.g. check-circle instead of check-circle-outline.

pat-s:
Paired with `cross-circle`, `play-circle`?

xoxys:
No... if you now start to argue that we can just use circled icons Im finally out 😆

pat-s:
I am just trying to make some sense between consistency and your personal prefs 😄️ Gotta admit I am completely confused now. 

So you propose to go back to `check-circle` paired with` cross-circle-outline` and `play-circle-outline`?

I mean I understand your desire and the general point to not only use one sub-iconset but icons belonging to a specific context group like all status icons (running, stop, play, etc.) should IMO *really* be from one subiconset at least, no?

xoxys:
No. I want to have the regular icons we had for play pause etc. just before this PR. You argued then well but we use outlined icons somewhere else. I can agree on that (even if I dont think thats a real problem). To fix this inconsistency I made the approach to replace the outlined icons we had already in use with non-outlined ones.

pat-s:
@xoxys In addition to the points discussed in chat I opted for `cogOutline` for the admin list panels as the adjacent icon (e.g. trash) is also outlined. Hope that is OK?

xoxys:
Wasnt the idea to harmonize the icons? Why mixing outlined and regular ones again?

pat-s:
> Wasnt the idea to harmonize the icons? Why mixing outlined and regular ones again?

Harmonizing on a contextual level. Some were outlined before, e.g. trashcan. The regular trashcan looks bulky, I think you agree?

Here's a preview of what I mean:

image

-> Both outlined

All status icons are now filled. The other standalone ones (e.g. cog in the navbar) are regular.

xoxys:
Sure some were outlined before nevertheless we wanted to harmonize it.

IMO trash-can looks good and not too bulky, if we want to avoid individual preferences (as much as possible) this looks a bit random to me and could cause discussions in the future again.

pat-s:
Hm. To me, the combined outlined icons from my previous comment look way more harmonic together than the current state:

image

If you agree with the rest, I'd draw my "virtual joker" and ask for this combination (both outlined as in my previous comment) to get accepted 😅️

xoxys:
Alright lets keep it this way.

web/src/components/atomic/Icon.vue
```diff
@@ -15,7 +15,7 @@
   <SvgIcon v-else-if="name === 'settings'" :path="mdiCog" size="1.3rem" />
   <SvgIcon v-else-if="name === 'settings-outline'" :path="mdiCogOutline" size="1.3rem" />
   <SvgIcon v-else-if="name === 'trash'" :path="mdiTrashCanOutline" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'status-blocked'" :path="mdiPlayCircle" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'status-blocked'" :path="mdiPlayOutline" size="1.3rem" />
```
xoxys, line 18: This change is wrong. Status icons should now all be the filled circle.

web/src/components/atomic/Icon.vue
```diff
@@ -29,34 +30,33 @@
...
+  <SvgIcon v-else-if="name === 'edit'" :path="mdiPencilOutline" size="1.3rem" />
   <SvgIcon v-else-if="name === 'download'" :path="mdiDownloadCircle" size="1.3rem" />
   <SvgIcon v-else-if="name === 'stopwatch'" :path="mdiAlarm" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'auto-scroll'" :path="mdiDownload" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'auto-scroll'" :path="mdiDownloadCircle" size="1.3rem" />
   <SvgIcon v-else-if="name === 'auto-scroll-off'" :path="mdiDownloadOff" size="1.3rem" />
   <SvgIcon v-else-if="name === 'pause'" :path="mdiPause" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'play'" :path="mdiPlay" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'remove'" :path="mdiClose" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'play'" :path="mdiPlayOutline" size="1.3rem" />
```
xoxys, line 52: This looks weird

image

I would like to keep those simple symbols (play, pause, plus etc) as regular varian.

pat-s: Ah I changed it for the CRON action buttons. There it matches with the other two.

image

Then we need both, I'd say. I'll change it again for the status icon and queue one to non-outlined.

xoxys: Ah I see... pain. Yeah I guess then we need both for now.

pat-s: Used a filled icon now without color.

xoxys: Why not keep the color?

anbraten:
> Changed icon button size to w-9 (2.25 rem) from w-8 as this aligns better with the line height the icons are in.

For consistency I would suggest again to stick to a 4-point / 8-point UI system. Those spacing issues often result from other parts having a wrong margin, padding or border not being aligned to it.

web/src/components/atomic/Icon.vue
pat-s: It is somewhat out of place, there is no other color nearby. It is not clickable. Idk, having a color there felt "out of place"?

pat-s:
> For consistency I would suggest to stick to a 4-point / 8-point UI system. Those spacing issues often result from other parts having a wrong margin, padding or border not being aligned to it.

So you mean reduce the line-height from 1.3rem to something like 1.2? This would then likely match with h-8.

anbraten:
That would be probably the easiest here, right?

pat-s:
To match with h-8, yes. But I am not sure 1.2 is maybe a bit to narrow. Need to check.

web/src/components/atomic/Icon.vue
pat-s: (Sorry, I know I am continuously making unrelated changes while "at it")

xoxys:
But how exactly does 1.2rem fit into an 8 point grid system that suggests using multiples of 8 and 4?

xoxys:
While it might make sense to use w-8 h-8 instead of -9 how is this related to the default size of the icons? Changing them to 1.2rem will not make them fit with the default line height of 24px.

pat-s:
Can we outsource this into a discussion/chat?

web/src/components/atomic/Icon.vue
```diff
@@ -15,12 +15,12 @@
   <SvgIcon v-else-if="name === 'settings'" :path="mdiCog" size="1.3rem" />
   <SvgIcon v-else-if="name === 'settings-outline'" :path="mdiCogOutline" size="1.3rem" />
   <SvgIcon v-else-if="name === 'trash'" :path="mdiTrashCanOutline" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'status-blocked'" :path="mdiPlayOutline" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'status-declined'" :path="mdiStopCircleOutline" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'status-blocked'" :path="mdiPlay" size="1.3rem" />
```
xoxys, line 18: Why do you go in this direction now? Please use the `mdiStopCircle` and `mdiPlayCircle` icons here as we agreed to use filled circle icons for **all** status icons.

pat-s: My mistake, wasn't on purpose.

web/src/components/atomic/Icon.vue
```diff
@@ -10,16 +10,17 @@
...
   <SvgIcon v-else-if="name === 'trash'" :path="mdiTrashCanOutline" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'status-blocked'" :path="mdiPlayCircleOutline" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'status-declined'" :path="mdiStopCircleOutline" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'status-blocked'" :path="mdiPlayCircle" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'status-declined'" :path="mdiStopCircle" size="1.3rem" />
   <SvgIcon
     v-else-if="name === 'status-failure' || name === 'status-error' || name === 'status-killed'"
     type="mdi"
-    :path="mdiCloseCircleOutline"
+    :path="mdiClose"
```
xoxys, line 23:
```suggestion
    :path="mdiCloseCircle"
```

web/src/components/atomic/Icon.vue
```diff
@@ -29,34 +30,34 @@
...
-  <SvgIcon v-else-if="name === 'status-success'" :path="mdiCheckCircleOutline" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'attention'" :path="mdiAlert" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'warning'" :path="mdiAlertOutline" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'error'" :path="mdiAlertCircle" size="1.3rem" />
-  <SvgIcon v-else-if="name === 'gitlab'" :path="mdiGitlab" size="32" />
+  <SvgIcon v-else-if="name === 'status-skipped'" :path="mdiMinusCircle" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'status-success'" :path="mdiCheckCircle" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'attention'" :path="mdiAlertCircle" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'warning'" :path="mdiAlert" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'error'" :path="mdiAlert" size="1.3rem" />
```
xoxys, line 37: Having two icons under different name with the same target `mdiAlert` doesnt make sense. Before that change we had three different icons for attention, warning and error. Why was this changed? How to we differentiate between these 3 now?

pat-s: With color. See OP. We also discussed this earlier in this PR already.

xoxys: Ok sorry. Got confused by the different names.

xoxys:
@pat-s I did the remaining cleanups:

- remove duplicate icons with the same target but different name
- harmonize pipeline icons
image
- dont re-use the download icons for scroll/no-scroll and replace it with the eye icon varians

I hope it was ok to push to your branch. If not please feel free to revert. Please re-check, if you dont have anything to add we are good to go now.

xoxys (approved):

pat-s:
Thanks cleaning up and all the feedback! Much appreciated.

xoxys (approved):

xoxys:
Now we have merge conflicts again 😐

xoxys (approved):

xoxys:
Finally 😁 thanks again!

PR #4595, opened 2024-12-20
Propagate version updates after artifacts are released

xoxys:
Wait until container images and binaries were built and uploaded successfully before publishing the `version.json` update.


pat-s (approved):

xoxys:
Mhm... that doesn't work due to the heavy use of conditions. Need to think about it again.

PR #4605, opened 2024-12-21
Update Go imports paths

qwerty287:


xoxys:
Why was this not covered by renovate?

xoxys:
For the go imports that shouldnt really matter, but unrelated to the PR.

pat-s (approved):
Might be good if renovate could catch it, but as we only need to to it once for every major, manual is also okayish.

PR #4608, opened 2024-12-22
pre-commit prettier: switch to maintained fork

pat-s:
The current one is archived and pre-commit itself doesn't maintain it anymore. It also fails with some strange errors which don't occur in the referenced fork.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,8 +27,8 @@ repos:
     rev: v2.13.1-beta
     hooks:
       - id: hadolint
-  - repo: https://github.com/pre-commit/mirrors-prettier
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -27,8 +27,8 @@ repos:
     rev: v2.13.1-beta
     hooks:
       - id: hadolint
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v4.0.0-alpha.8
+  - repo: https://github.com//rbubley/mirrors-prettier
```
xoxys, line 30:
```suggestion
  - repo: https://github.com/rbubley/mirrors-prettier
```

xoxys (approved):

PR #4612, opened 2024-12-23
Typo fix: instace -> instance

jensenbox:
instace -> instance


pat-s:
Already fixed in latest docs.

PR #4613, opened 2024-12-23
Fix misleading example in Workflow syntax/Privileged mode

plafue:


plafue:
Hi @qwerty287 and thanks for the quick reply. I changed the documentation accordingly under `docs/docs`.

pat-s (approved):
Thanks!

PR #4617, opened 2024-12-24
Add rootless (alpine) images

pat-s:
## Background

Currently, two image variants exist:

- scratch-based images
- alpine-based

The former is a few MB smaller than the alpine one but lacks the ability to exec into it as it has no shell.
The latter has a shell but as both run as `root` user, using the alpine image is somewhat risky.

## Possible solutions

There are two ways to approach this:

1. Add a new `rootless` variant for the alpine variant (-> allows to exec into the pod but without being `root`)
2. Replace the `scratch` variant by a rootless alpine variant and make it the default 

With (1), there would be three image variants that need to be build and maintained. Having this many, possible confusion among users exists. Also sticking with the `scratch` image as the default is not super convenient as adminds always have to switch to the alpine one first for interactive actions

With (2), there would be only two image variants (`alpine-rootless` and `alpine-rootful`). Making `alpine-rootless` the default (i.e. tagging with only semver versions) provides the same security as before because running rootless = same as using `scratch`, one cannot do anything destructive within the container.)

It would additionally provide the ability to exec into the container/pod. A possible downside would be that the image is a few MB bigger (e.g. 17mb instead of 13mb for the server image).

I'd like to get a maintainers vote here how to proceed. Based on that, I would like to a add a new docs page which explains the existing tags and their indented usage, similar to what has been added recently to the repository descriptions on DockerHub.

@woodpecker-ci/maintainers 
Vote with 🚀 for keeping `scratch` and having a new `alpine-rootless` variant.
Vote with 👀 for dropping `scratch` and replacing it with the new `alpine-rootless` variant.

PS: I tested the server and agent images in a k8s instance. The CLI image was tested locally.

### DIFF
--- a/docker/Dockerfile.agent.alpine.multiarch.rootless
+++ b/docker/Dockerfile.agent.alpine.multiarch.rootless
@@ -1,0 +1,26 @@
+FROM --platform=$BUILDPLATFORM docker.io/golang:1.23 AS build
+
+WORKDIR /src
+COPY . .
+ARG TARGETOS TARGETARCH CI_COMMIT_SHA CI_COMMIT_TAG CI_COMMIT_BRANCH
+RUN --mount=type=cache,target=/root/.cache/go-build \
+    --mount=type=cache,target=/go/pkg \
+    make build-agent
+
+FROM docker.io/alpine:3.21
+RUN apk add -U --no-cache ca-certificates
+ENV GODEBUG=netdns=go
+# Internal setting do NOT change! Signals that woodpecker is running inside a container
+ENV WOODPECKER_IN_CONTAINER=true
+EXPOSE 3000
+
+COPY --from=build /src/dist/woodpecker-agent /bin/
+RUN mkdir -p /etc/woodpecker
+
+RUN addgroup -S woodpecker && adduser -S woodpecker -G woodpecker
+RUN mkdir -p /var/lib/woodpecker && chown -R woodpecker:woodpecker /etc/woodpecker
+
+USER woodpecker
+
+HEALTHCHECK CMD ["/bin/woodpecker-agent", "ping"]
+ENTRYPOINT ["/bin/woodpecker-agent"]
--- a/docker/Dockerfile.agent.alpine.multiarch.rootless
+++ b/docker/Dockerfile.agent.alpine.multiarch.rootless
@@ -18,7 +18,7 @@ COPY --from=build /src/dist/woodpecker-agent /bin/
 RUN mkdir -p /etc/woodpecker
 
 RUN addgroup -S woodpecker && adduser -S woodpecker -G woodpecker
-RUN mkdir -p /var/lib/woodpecker && chown -R woodpecker:woodpecker /etc/woodpecker
+RUN mkdir -p /etc/woodpecker && chown -R woodpecker:woodpecker /etc/woodpecker
 
 USER woodpecker
 
--- a/docker/Dockerfile.cli.alpine.multiarch.rootless
+++ b/docker/Dockerfile.cli.alpine.multiarch.rootless
@@ -1,0 +1,25 @@
+FROM --platform=$BUILDPLATFORM docker.io/golang:1.23 AS build
+
+WORKDIR /src
+COPY . .
+ARG TARGETOS TARGETARCH CI_COMMIT_SHA CI_COMMIT_TAG CI_COMMIT_BRANCH
+RUN --mount=type=cache,target=/root/.cache/go-build \
+    --mount=type=cache,target=/go/pkg \
+    make build-cli
+
+FROM docker.io/alpine:3.21
+WORKDIR /woodpecker
+
+RUN apk add -U --no-cache ca-certificates
+
+ENV GODEBUG=netdns=go
+ENV WOODPECKER_DISABLE_UPDATE_CHECK=true
+
+COPY --from=build /src/dist/woodpecker-cli /bin/
+
+RUN addgroup -S woodpecker && adduser -S woodpecker -G woodpecker
+
+USER woodpecker
+
+HEALTHCHECK CMD ["/bin/woodpecker-cli", "ping"]
+ENTRYPOINT ["/bin/woodpecker-cli"]
--- a/docs/docs/30-administration/04-image-variants.md
+++ b/docs/docs/30-administration/04-image-variants.md
@@ -1,0 +1,3 @@
+# Image variants
+
+FIXME
--- a/.woodpecker/docker.yaml
+++ b/.woodpecker/docker.yaml
@@ -214,6 +214,23 @@ steps:
       logins: *publish_logins
     when: *when-release
 
+  release-server-alpine-rootless:
+    depends_on:
+      - cross-compile-server
+    image: *buildx_plugin
+    settings:
+      repo: *publish_repos_server
+      dockerfile: docker/Dockerfile.server.alpine.multiarch.rootless
+      platforms: *platforms_alpine
+      tag:
+        [
+          '${CI_COMMIT_TAG%%.*}-alpine-rootless',
+          '${CI_COMMIT_TAG%.*}-alpine-rootless',
+          '${CI_COMMIT_TAG}-alpine-rootless',
+        ]
+      logins: *publish_logins
+    when: *when-release
+
   #############
   # A g e n t #
   #############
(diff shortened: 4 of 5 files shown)


xoxys:
> The latter has a shell but as both run as root user, using the alpine image is somewhat risky.

Im a bit confused. The scratch image is not running as root. So in production just use the scratch one and for debugging if you need a shell use the alpine one?

pat-s:
> Im a bit confused. The scratch image is not running as root.

It is? Unless a different `USER` is set, `root` is being used. Also the directories are owned by `root`. It is only "secure" because there is no shell, yet all operations are done as `root`.

> So in production just use the scratch one and for debugging if you need a shell use the alpine one? For everyday operation I dont see any reason to have a shell.

Exactly this is the problem. Switching the image just for this is tedious. Scratch images are fine for plugins and other "read-only" approaches IMO but the `agent` and `server` images should have a shell.

Hence, a better default would be rootless image (i.e. running as an unprivileged non-root user) but having the ability to exec in. This way, one can inspect logs and other things in a limited way. 

> For everyday operation I don't see any reason to have a shell.

I'd argue that there should always be a shell paired with an unprivileged user by default for applications like WP.

---

If the scratch image would be replaced by the rootless-alpine one, the alpine one per se would have to be renamed as well then. So then

- `vx.y.z` -> alpine-rootless with shell
- `vx.y.z-rootful` -> alpine-rootful with shell

xoxys:
Could you explain why you need a shell for the server container?

pat-s:
For all sorts of debugging activities? Inspecting connection issues to the agent, checking ports, grpc, DB connections.
The runtime env (k8s, docker) doesn't matter here. A shell in a container is *always* helpful.
Yet it should be non-privileged by default.

> It is a very common practice to switch images for debugging.

To me, using scratch images is a predated approach to securing containers before the rootless idea became usable/accepted. One should always have the option to undertake the above mentioned checks in a limited and non-destructive environment without having to switch images upfront.

xoxys:
How often do you need to debug your apps? So yes I do this every time I have to debug something.

I disagree that rootless inages fix every problem that could be fixed by distroless base containers.

pat-s:
> How often do you need to debug your apps?

I am managing close to 10 WP instances, so it sums up. Especially in highly-restricted environments, it is a quite regular tasks, as changes to the networking layer often happens without prior information.

>  rootless alpine should just replace the existing alpine variant.

Alright, as it seems this is the only agreement we have here, I'll then proceed this way. This at least enables the option to use the alpine variants by default without running a shell container as root.

pat-s:
Update for silent readers: We had some private chat discussions and eventually agreed on using filled-icons for the status ones and a mix of regular and outlined ones for all others.

xoxys:
@pat-s Your last comment seems to be on the wrong PR? :sweat_smile:

pat-s:
Ah, fck, I am really operating on too many PRs 🙈️

lafriks:
Rootless image can be also easily implemented for scratch images

docker/Dockerfile.agent.alpine.multiarch.rootless
```diff
@@ -17,5 +17,10 @@ EXPOSE 3000
 COPY --from=build /src/dist/woodpecker-agent /bin/
 RUN mkdir -p /etc/woodpecker
 
+RUN addgroup -S woodpecker && adduser -S woodpecker -G woodpecker
+RUN mkdir -p /etc/woodpecker && chown -R woodpecker:woodpecker /etc/woodpecker
```
lafriks, line 21: RUN's should be merged as much as possible to reduce layer count

xoxys:
@pat-s do you plan to finish this one?

lafriks:
If I understand correctly that alpine version will be rootless only?

Imho best solution if we do not provide both rootfull and rootless images would be to switch both (scratch and alpine) images as rootless so that there are no permission problems when switching between them etc

pat-s:
Yes, switching between both variants is then not easily possible anymore as the directory permissions are different.

To make this change smoother, there should ideally be an init-container for the helm chart which adjust permissions, if needed. 

But yes, this change will break all existing instances and manual permission adjustments are required.

lafriks:
Ideally we should probably switch both types then to the same user id and release notes should warn about permission changes required when upgrading to 3.0

docker/Dockerfile.agent.multiarch.rootless
```diff
@@ -1,4 +1,8 @@
-FROM --platform=$BUILDPLATFORM docker.io/golang:1.23 AS build
+FROM --platform=$BUILDPLATFORM docker.io/golang:1.23 AS builder
```
xoxys, line 1: Why do you call it builder now while it was called build in most places already and and still is named that way in some places?

docker/Dockerfile.agent.multiarch.rootless
```diff
@@ -1,4 +1,8 @@
...
+
+RUN adduser woodpecker && \
+  mkdir -p /var/lib/woodpecker && \
```
xoxys, line 4: Why is this directory created here? Looks like its never used and also doesnt exist in the alpine image.

docs/docs/30-administration/04-image-variants.md
```diff
@@ -0,0 +1,30 @@
+# Image variants
+
+:::info
+The `latest` tag has been deprecated as of v3.0 and will be completely removed in the future.
+This was done to prevent accidental major version upgrades.
+:::
+
+- `vX.Y.Z`: Sem-ver tags for specific releases, no entrypoint shell (scratch image)
```
xoxys, line 8:
```suggestion
- `vX.Y.Z`: SemVer tags for specific releases, no entrypoint shell (scratch image)
```

docs/docs/30-administration/04-image-variants.md
```diff
@@ -0,0 +1,30 @@
...
+  - `vX.Y`
+  - `vX`
+- `vX.Y.Z-alpine`: Sem-ver tags for specific releases, based on Alpine, rootless (as of v3.0).
```
xoxys, line 11:
```suggestion
- `vX.Y.Z-alpine`: SemVer tags for specific releases, based on Alpine, rootless (as of v3.0).
```

docs/src/pages/migrations.md
```diff
@@ -172,6 +172,11 @@ The following restructuring was done to achieve a more consistent grouping:
 
 - Webhook signatures now use the `rfc9421` protocol
 
+#### Rootless images
+
+From 3.0 onward, all Woodpecker images use a non-privileged user (`woodpecker`) by default.
```
xoxys, line 177: Does it make sense to put the version in here? Docs are versioned and the migration notes always reflect required changes of the next major anyway?

xoxys (changes requested):

xoxys:
The created group and user should be created with a static uid/gid. A common choice is `1000`.

docker/Dockerfile.cli.alpine.multiarch.rootless
```diff
@@ -17,5 +17,9 @@ ENV WOODPECKER_DISABLE_UPDATE_CHECK=true
 
 COPY --from=build /src/dist/woodpecker-cli /bin/
```
xoxys, line 19: Can we group the RUN commands together?

docker/Dockerfile.agent.alpine.multiarch.rootless
```diff
@@ -15,7 +15,12 @@ ENV WOODPECKER_IN_CONTAINER=true
 EXPOSE 3000
 
 COPY --from=build /src/dist/woodpecker-agent /bin/
-RUN mkdir -p /etc/woodpecker
+
+RUN adduser -u 1000 -g 1000 woodpecker && \
```
xoxys, line 19: There is already a RUN apk line can we group this?

docker/Dockerfile.agent.multiarch.rootless
```diff
@@ -1,12 +1,15 @@
 FROM --platform=$BUILDPLATFORM docker.io/golang:1.23 AS build
 
+RUN adduser -u 1000 -g 1000 woodpecker && \
```
xoxys, line 3: Sam here. We dont need to chain but at least position the RUN commands together?

xoxys:
Have you tested the new images?

xoxys (approved):

PR #4620, opened 2024-12-26
Add docker in docker example to advanced usage in docs

plafue:


docs/docs/20-usage/90-advanced-usage.md
```diff
@@ -127,3 +127,40 @@ WOODPECKER_ENVIRONMENT=first_var:value1,second_var:value2
...
+
+:::warning
+This set up will only work on trusted repositories and for security reasons should only be used in private environments. See [project settings](./75-project-settings.md#trusted) to enable trusted mode.
+:::
+
+The snippet below shows how a step can communicate with the docker deamon via a `docker:dind` service.
+
+The most important bit ist to allow the dind service to write/read certificates on a mounted volume that is also shared with the step that will contact the daemon (it needs the service's CA to do so successfully).
+
+The `DOCKER_*` environment variables in the step are generic settings that should work with your framework of choice (e.g. TestContainers or similar):
```
pat-s, line 141: What does "TestContainers or similar" mean here exactly? Can't make much sense out of that part.

pat-s (commented):
Thanks! I have been thinking about adding this in the past as well, so surely good to have. And the place matches as well.

I would say that there should be an additional note at the top which states something like "use the buildx plugins by default, this is only a fallback for advanced/custom use of docker build".

In regards to that (and this might answer the second question): what has motivated you to go this route instead of using the buildx plugin? (this would be a good use case example to add as well).

docs/docs/20-usage/90-advanced-usage.md
plafue: Thanks for the incredibly quick reply!

My usecase is not to build images, but to use a framework like [testcontainers](https://testcontainers.com/) to start/stop containers during test, so that close-to-production infrastructure can be started on demand. One could compare it to the services feature of woodpecker but allows more advanced scenarios.

I'd be happy to remove that comment if it is a bit too specific / out of scope.

I will gladly add that for normal oci building scenarios, the buildx plugin is the way to go

pat-s: Ah, didn't know that project before. Makes sense then 👍️ I think when you add a link, it is fine and even a great example of when to use this approach. Bundled with the note that buildx should be used for building, this looks like a great addition then.

plafue: Thanks for the feedback. I added two examples of frameworks in the end, with a couple of links.

If there's anything I should add/modify, please let me know!

docs/docs/20-usage/90-advanced-usage.md
```diff
@@ -127,3 +127,44 @@ WOODPECKER_ENVIRONMENT=first_var:value1,second_var:value2
...
+This set up will only work on trusted repositories and for security reasons should only be used in private environments. See [project settings](./75-project-settings.md#trusted) to enable trusted mode.
+:::
+
+The snippet below shows how a step can communicate with the docker daemon via a `docker:dind` service.
+
+:::note
+If your aim ist to build/publish OCI images, consider using the [Docker Buildx Plugin](https://woodpecker-ci.org/plugins/Docker%20Buildx) instead.
+:::
+
+The most important bit ist to allow the dind service to write/read certificates on a mounted volume that is also shared with the step that will contact the daemon (it needs the service's CA to do so successfully).
```
pat-s, line 143:
```suggestion
The most important bit is to allow the `dind` service to write/read certificates on a mounted volume that is also shared with the step that will contact the daemon (it needs the service's CA to do so successfully).
```

TLS is enforced since docker v27 AFAIR. Maybe a short link to that can help to explain this requirement a bit better with less complicated/technical wording?

docs/docs/20-usage/90-advanced-usage.md
```diff
@@ -127,3 +127,44 @@ WOODPECKER_ENVIRONMENT=first_var:value1,second_var:value2
...
+
+The `DOCKER_*` environment variables in the step are generic settings that should work with the docker client used by your framework of choice (e.g. [TestContainers](https://testcontainers.com/), [Spring Boot Docker Compose](https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-docker-compose) or similar):
+
+```yaml
+steps:
+  - name: test
+    image: docker:latest
```
pat-s, line 150: I'd suggest to add a comment like "in production, you usually want to use (at least) a fixed major version like `:27`". Same for the `dind` service.

docs/docs/20-usage/90-advanced-usage.md
```diff
@@ -127,3 +127,44 @@ WOODPECKER_ENVIRONMENT=first_var:value1,second_var:value2
 ```
 
 Note that this tightly couples the server and app configurations (where the app is a completely separate application). But this is a good option for truly global variables which should apply to all steps in all pipelines for all apps.
+
+## Docker in docker (dind) setup
+
+:::warning
+This set up will only work on trusted repositories and for security reasons should only be used in private environments. See [project settings](./75-project-settings.md#trusted) to enable trusted mode.
```
pat-s, line 134:
```suggestion
This set up will only work on trusted repositories and for security reasons should only be used in private environments.
See [project settings](./75-project-settings.md#trusted) to enable trusted mode.
```

docs/docs/20-usage/90-advanced-usage.md
plafue: Fair point! i added a note about it in [212f29e](https://github.com/woodpecker-ci/woodpecker/pull/4620/commits/212f29e2045c04ded666be1497d0a3098b344b9a)

docs/docs/20-usage/90-advanced-usage.md
plafue: Good catch, i just changed both to the latest `27` tags in [212f29e](https://github.com/woodpecker-ci/woodpecker/pull/4620/commits/212f29e2045c04ded666be1497d0a3098b344b9a)

docs/docs/20-usage/90-advanced-usage.md
```diff
@@ -141,17 +141,65 @@ The snippet below shows how a step can communicate with the docker daemon via a
...
+```
+
+This step should output version information of the client and the server if everything has been set correctly.
+
+Complete example:
 
 ```yaml
 steps:
   - name: test
-    image: docker:latest
```
pat-s: I meant to just use a comment here for the major version hint and keep `latest`. Otherwise the docs don't age well.

plafue:
@pat-s i rewrote the section quite a bit, guiding the reader through the changes and a bit of reasoning behind them. Became a bit wordy. i hope it's ok. Otherwise, feel free to make changes at your own discretion.

pat-s:
Thanks, I'll check again later and apply suggestions as needed then 👍

pat-s (approved):

xoxys (approved):

plafue:
Thanks for the support, @pat-s !

PR #4622, opened 2024-12-26
Add vscode tailwind extensions & prettier tailwind linter, sort tailwind classes

pat-s:
These are the ones that cause the auto-sorting in all my recent PRs. If we would all use these (and sort all classes once), we'd have a predictable and consistent order of tailwind classes.

- Added prettier checks to `docs/` and `web/` enforcing the sorting. As the release process in `plugin-prettier` has some issues currently, I've used `:next` to move on


xoxys:
Can we add linting as well then? Not everybody uses VSCode.

pat-s:
blocked by https://codeberg.org/woodpecker-plugins/prettier/pulls/11

.gitpod.yml
```diff
@@ -117,4 +117,6 @@ vscode:
     - 'redhat.vscode-yaml'
     - 'davidanson.vscode-markdownlint'
     - 'streetsidesoftware.code-spell-checker'
+    - 'dejmedus.tailwind-sorter'
```
anbraten, line 120: Seems to be doing the same as the prettier sorting?

.prettierrc.json
```diff
@@ -4,5 +4,6 @@
   "singleQuote": true,
   "printWidth": 120,
   "tabWidth": 2,
-  "endOfLine": "lf"
+  "endOfLine": "lf",
+  "plugins": ["prettier-plugin-tailwindcss"]
```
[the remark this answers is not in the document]
pat-s: web/ and docs/ refer to the top-level config currently. See also #4628 

Maybe it can be done in a different way, then one does not need to install this in their global npm package stack.

.gitpod.yml
pat-s: Yes, that's the idea?

web/.prettierrc.js
```diff
@@ -5,12 +5,5 @@ const config = JSON.parse(await readFile(new URL('../.prettierrc.json', import.m
...
   ...config,
-  plugins: ['@ianvs/prettier-plugin-sort-imports'],
-  importOrder: [
-    '<THIRD_PARTY_MODULES>', // Imports not matched by other special words or groups.
-    '', // Empty string will match any import not matched by other special words or groups.
-    '^(#|@|~|\\$)(/.*)$',
-    '',
-    '^[./]',
-  ],
+  plugins: ['prettier-plugin-tailwindcss'],
```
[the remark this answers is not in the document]
pat-s: Ah, good catch. I mistakenly thought this is for something else. Good to re-add it.

.gitpod.yml
anbraten: I am auto formatting with prettier on save which would now already sort the classes. Isn't using that plugin redundant?

pat-s: Yeah they might be indeed doing the same. The vscode one can use custom sorting but if we don't need that, we can stick with the prettier plugin.

image

.prettierrc.json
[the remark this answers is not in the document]
pat-s: Ah forget what I said, I think your proposed change should be just fine.

xoxys (approved):

.gitpod.yml
pat-s: I dropped it already.

PR #4625, opened 2024-12-27
Two row layout for title and context of pipeline list

qwerty287:
supersedes https://github.com/woodpecker-ci/woodpecker/pull/3976

Before:
image

After:
image


pat-s:
Can you add a screenshot?

xoxys:
Added one to the PR description.

xoxys (commented):
Im really missing the `#<pipeline number>` information. Can we add it back?

pat-s:
Would it be better to have it in the second line? This line already holds other static metadata while the first one holds the title information.

xoxys:
Fine for me as well even if I like it more in the title. If we add it to the second line I would add as the first element and we need to add an `pound` icon instead of the text `#`.

xoxys:
Not sure if I can follow. But IMO this is getting way too complicated now. Either keep it in the title as it was before or leave it out and improve it in a dedicated PR.

xoxys:
Ok got it. I can look into this in another PR if you like.

xoxys:
I was looking it and Im sure this will not work, at least not with pure css.

xoxys:
@pat-s @anbraten Are you ok to keep it in the title for now until we find a way to add it to the second line? Otherwise we might want to remove it again from this PR to get forward.

pat-s:
Temporary OK, if you commit to look into this in more detail. The question is just what happens if this is "not so easy" or gets stuck - then we have a complicated discussion.

Maybe the motivation would be higher for a follow-up PR if the workaround wouldn't be included? 🙈️

Your choice, I'm either way.

xoxys:
Just dont like that we loose a functionality here. To be honest I dont know how to solve the issue with the second line. This is not only an issue with the pipeline number. With the current approach we will get this problem with every other information that can have a dynamic length.

xoxys:
The other alternative it to accept that we will get a similar look like this with the second line:

image

@qwerty287 could we live with that?

pat-s (approved):

PR #3421, opened 2024-02-22
Use actual title of event subjects if possible

6543:
At the moment we did mostly ignore the title of things and used the first line of the object descriptions.

Now we try to retrieve the actual title if an object can have one.

depends on #3420

PR #4626, opened 2024-12-27
Rework pipeline data structure

### LINKED ISSUE #2971: Environment Variable CI_COMMIT_AUTHOR_EMAIL is wrong
### Component

server, agent

### Describe the bug

I am running Gitea with Woodpecker CI. I have noticed that for some time now, the built in `CI_COMMIT_AUTHOR_EMAIL` environment variable is wrong. I have setup a sample project, with nothing more than a simple pipeline.

Here is the git log from the CLI:

```
commit 3d6d4085e41d1a044a67c15ddb99a7ea4746ded7 (HEAD -> main, origin/main, origin/HEAD)
Author: Philip Kozeny <philip.kozeny@mydomain.com>
Date:   Tue Dec 19 10:32:26 2023 +0000

    Update .woodpecker.yml
```

Here is the woodpecker.yml

```
steps:
  build:
    image: alpine:edge
    commands:
      - echo $CI_COMMIT_AUTHOR
      - echo $CI_COMMIT_AUTHOR_EMAIL
```

Here is the output after running the pipeline:

```
+ echo $CI_COMMIT_AUTHOR
philip.kozeny
+ echo $CI_COMMIT_AUTHOR_EMAIL
philip.kozeny@anotherdomain.com
```

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.0.0"}
Powered by Gitea Version: 1.21.2
```

### Additional context

_No response_

### Validations

### LINKED ISSUE #4195: Wrong value of CI_PREV_COMMIT_MESSAGE
### Component

server

### Describe the bug

`CI_PREV_COMMIT_MESSAGE` sets to `MANUAL PIPELINE @ 01-hello-world`, for example.

### Steps to reproduce

1. Manually run pipeline
```
skip_clone: true
steps:
  hello:
    image: alpine
    commands:
      - echo Hello world
```
2. Find `CI_PREV_COMMIT_MESSAGE` among Pod's environment variables
3. Get wrong value like `MANUAL PIPELINE @ 01-hello-world`

### Expected behavior

`MANUAL PIPELINE @ 01-hello-world` - is pipeline message, not commit.
So, variable should be named like `CI_PREV_PIPELINE_MESSAGE` or value should be as actual _commit_ message.

### System Info

```shell
`next-8e7cace29a`, Kubernetes
```

### Additional context

_No response_

### Validations

### LINKED ISSUE #4677: `tag` pipelines are missing avatar and commit title
### Component

web-ui

### Describe the bug

At least the avatar should be there. The commit title should be the commit message, (which was set for this tag to the tag name via `git tag -m "v1.0.0" v1.0.0`).

image

### Steps to reproduce

.

### Expected behavior

_No response_

### System Info

```shell
"next-d126cf9e35"
```

### Additional context

_No response_

### Validations

qwerty287:
supersedes https://github.com/woodpecker-ci/woodpecker/pull/3421
closes https://github.com/woodpecker-ci/woodpecker/issues/2971
closes https://github.com/woodpecker-ci/woodpecker/issues/4677
closes https://github.com/woodpecker-ci/woodpecker/issues/4195

Currently, our pipeline model has fields like the message, title etc., but every field is used differently, dependending on the event.

This PR changes this and splits these reused fields to have a more clear pipeline model.

The model now has fields like:
- commit (containing sha, author, forge url and message)
- pull request (index, title, from fork, labels)
- release title

and similar.

Advantages:
1. cleaner model and less differently used fields
2. fix the problem that there is a difference between commit and pipeline author (pipeline author = pusher) #2971 
3. store and show the commit message for every pipeline
4. store and show the context of a pipeline (PR title, deployment target, release title etc.) and show it

Especially 3 and 4 and are the enhancement parts here, as we can now store both context and commit message and not only one.


pat-s:
@qwerty287 Why is that breaking? Because of the env var change?

qwerty287:
Yes, env var change and it also changes the json for the pipeline model

xoxys:
Can you add a bit more context about what has changed and how it fixes the linked issue?

xoxys:
Thank you! ❤️

xoxys:
Will finally take a look this evening.

xoxys:
After switching to this branch I get 
image

xoxys:
image

I would stick to a two line layout, even if the second line would be empty. The pipeline id is not show for me.

xoxys:
> I'm not sure if I'd do this.

True. But I would prefer to not change the layout depending on the event. Ill try to find another approach if not Im also fine with the current way.

server/forge/bitbucketdatacenter/convert.go
```diff
@@ -111,18 +112,19 @@ func convertRepositoryPushEvent(ev *bb.RepositoryPushEvent, baseURL string) *mod
...
+			SHA:      ev.PullRequest.Source.Latest,
+			ForgeURL: fmt.Sprintf("%s/projects/%s/repos/%s/commits/%s", baseURL, ev.PullRequest.Source.Repository.Project.Key, ev.PullRequest.Source.Repository.Slug, ev.PullRequest.Source.Latest),
+		},
+		Branch:      ev.PullRequest.Source.DisplayID,
+		Avatar:      bitbucketAvatarURL(baseURL, ev.Actor.Slug),
+		Author:      ev.Actor.Name,
+		Ref:         fmt.Sprintf("refs/pull-requests/%d/from", ev.PullRequest.ID),
+		PullRequest: convertPullRequest(&ev.PullRequest),
+		// TODO should link to the Pr
+		ForgeURL: fmt.Sprintf("%s/projects/%s/repos/%s/commits/%s", baseURL, ev.PullRequest.Source.Repository.Project.Key, ev.PullRequest.Source.Repository.Slug, ev.PullRequest.Source.Latest),
```
langecode, line 126: The link to the PR in the Bitbucket UI would be something like:

```
fmt.Sprintf("%s/projects/%s/repos/%s/pull-requests/%d", baseURL, ev.PullRequest.Source.Repository.Project.Key, ev.PullRequest.Source.Repository.Slug, ev.PullRequest.ID)
```

server/forge/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -364,22 +364,28 @@ func (c *client) BranchHead(ctx context.Context, u *model.User, r *model.Repo, b
...
-				ForgeURL: fmt.Sprintf("%s/commits/%s", r.ForgeURL, branch.LatestCommit),
-			}, nil
-		}
-	}
-	return nil, fmt.Errorf("no matching branches found")
+
+	cm := commits[0]
+
+	return &model.Commit{
+		SHA:      cm.ID, // TODO check id or displayid?
```
langecode, line 381: `cm.ID` is the full sha - the `displayID` contains a shortened sha.

server/forge/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -364,22 +364,28 @@ func (c *client) BranchHead(ctx context.Context, u *model.User, r *model.Repo, b
 	if err != nil {
 		return nil, fmt.Errorf("unable to create bitbucket client: %w", err)
 	}
-	branches, _, err := bc.Projects.SearchBranches(ctx, r.Owner, r.Name, &bb.BranchSearchOptions{Filter: b})
+	commits, _, err := bc.Projects.SearchCommits(ctx, r.Owner, r.Name, &bb.CommitSearchOptions{
```
langecode, line 367: I'll see if I can do this API call to Bitbucket to verify that the `Until` property actually returns what we expect. The `b` variable will contain the branch name not the ref?

langecode (commented):
added a few comments to the bitbucket datacenter implementation.

anbraten:
I like the change especially as it will help is in the future with adding PRs like #5174. 
Two suggestions:
- Extract the UI changes to a separate PR as they often involve some more discussions regarding the layout, etc
- Keep the current model structure for the api endpoints so this one could be merged soonishly without being breaking (therefore we could create a model that extends the new one and map the old properties accordingly)

anbraten:
> I can remove the ui changes, but why can't we just do 4.0 and merge this as is with the breaking changes?

Breaking it without deprecating it in a previous release makes it extremely difficult for larger instances / teams to adopt to those api changes. Especially the pipeline apis are used by some external tools.

server/store/datastore/migration/026_update_pipeline_messages.go
```diff
@@ -0,0 +1,159 @@
+// Copyright 2024 Woodpecker Authors
```
anbraten:
```suggestion
// Copyright 2025 Woodpecker Authors
```

pipeline/frontend/metadata/environment.go
```diff
@@ -143,13 +142,16 @@ func (m *Metadata) Environ() map[string]string {
...
 	setNonEmptyEnvVar(params, "CI_PREV_COMMIT_AUTHOR_EMAIL", prevCommit.Author.Email)
-	setNonEmptyEnvVar(params, "CI_PREV_COMMIT_AUTHOR_AVATAR", prevCommit.Author.Avatar)
 	if prevPipeline.Event == EventPull || prevPipeline.Event == EventPullClosed {
 		prevSourceBranch, prevTargetBranch := getSourceTargetBranches(prevCommit.Refspec)
 		setNonEmptyEnvVar(params, "CI_PREV_COMMIT_SOURCE_BRANCH", prevSourceBranch)
 		setNonEmptyEnvVar(params, "CI_PREV_COMMIT_TARGET_BRANCH", prevTargetBranch)
 	}
 
+	// TODO Deprecated, remove in next major
+	setNonEmptyEnvVar(params, "CI_COMMIT_AUTHOR_AVATAR", pipeline.Avatar)
```
6543, line 152: I just wonder if CI_PIPELINE_AVATAR is interesting for people to be used?

server/api/pipeline.go
```diff
@@ -79,25 +79,21 @@ func CreatePipeline(c *gin.Context) {
...
 	return &model.Pipeline{
-		Event:     event,
-		Commit:    commit.SHA,
-		Branch:    opts.Branch,
-		Timestamp: time.Now().UTC().Unix(),
-
-		Avatar:  user.Avatar,
-		Message: "MANUAL PIPELINE @ " + opts.Branch,
+		Event:  event,
+		Commit: commit,
```
6543, line 89: timestamp is now missing ?!?

server/model/pipeline.go
```diff
@@ -20,46 +20,85 @@ import (
...
-	Finished            int64                  `json:"finished"                xorm:"finished"`
-	DeployTo            string                 `json:"deploy_to"               xorm:"deploy"`
-	DeployTask          string                 `json:"deploy_task"             xorm:"deploy_task"`
-	Commit              string                 `json:"commit"                  xorm:"commit"`
-	Branch              string                 `json:"branch"                  xorm:"branch"`
-	Ref                 string                 `json:"ref"                     xorm:"ref"`
-	Refspec             string                 `json:"refspec"                 xorm:"refspec"`
-	Title               string                 `json:"title"                   xorm:"title"`
-	Message             string                 `json:"message"                 xorm:"TEXT 'message'"`
-	Timestamp           int64                  `json:"timestamp"               xorm:"'timestamp'"`
```
6543, line 43: timestamp got removed why?

docs/src/pages/migrations.md
```diff
@@ -1,10 +1,19 @@
...
 ## `next`
 
+### User-facing changes
+
 - (Kubernetes) Deprecated `step` label on pod in favor of new namespaced label `woodpecker-ci.org/step`. The `step` label will be removed in a future update.
+- deprecated `CI_COMMIT_AUTHOR_AVATAR` and `CI_PREV_COMMIT_AUTHOR_AVATAR` env vars as commit authors don't have an avatar
+
+### API changes
+
+- Changed the pipeline model to have different objects for different event metadata (e.g. pull request title)
```
anbraten: Might be best to list the changes. Helped me with understanding them while reviewing as  well.

```suggestion
- Changed the pipeline model to use nested objects grouped based on the event (e.g. instead of a generic `title` it now uses `pr.title`)

`author` => `commit.author`
`deploy_to` => `deployment.target`
`deploy_task` => `deployment.task`
`commit` (sha) => `commit.sha`
`title` => `release.title` (for release events) & `pr.title` (for pull-request events)
`message` => `commit.message`
`timestamp` => `created` (somehow dropped the value, was this the commit timestamp before?)
`sender` => `author`
`avatar` => `author_avatar`
`author_email` => `commit.author.email`
`pr_labels` => `pr.labels`
`is_prerelease` => `is_prerelease` (not `release.is_prerelease` ?)
??? => `release_tag_title` (not `release.tag_title`? was this the normal `title` before?)
`from_fork` => `pr.from_fork`
```

docs/src/pages/migrations.md
```diff
@@ -83,6 +92,38 @@ The following syntax deprecations will now result in an error:
 - `platform:` ([#3916](https://github.com/woodpecker-ci/woodpecker/pull/3916))
 - `branches:` ([#3916](https://github.com/woodpecker-ci/woodpecker/pull/3916))
 
+#### Workflow syntax changes
```
anbraten: Where is this coming from? Should we add it in a separate PR?

pipeline/frontend/metadata/types.go
```diff
@@ -52,9 +52,13 @@ type (
 		ForgeURL   string `json:"forge_url,omitempty"`
 		DeployTo   string `json:"target,omitempty"`
 		DeployTask string `json:"task,omitempty"`
-		Commit     Commit `json:"commit,omitempty"`
+		Commit     Commit `json:"commit"`
```
anbraten: Why not `omitempty` here. Do we expect all pipelines to have a commit in the future (thinking of a pipeline being triggered on issue creation)?

server/api/pipeline.go
anbraten: Hasn't it been set to the timestamp of the related commit at some point?

server/api/pipeline.go
```diff
@@ -638,7 +641,7 @@ func PostPipeline(c *gin.Context) {
 			return
 		}
 
-		pl.DeployTo = c.DefaultQuery("deploy_to", pl.DeployTo)
+		pl.Deployment.Target = c.DefaultQuery("deploy_to", pl.Deployment.Target)
```
anbraten: This property is called `deployment.target` now while the query parameter is `deploy_to`. Should use one term to be consistent with a deprecation and fallback if necessary.

server/model/pipeline.go
```diff
@@ -20,46 +20,85 @@ import (
...
 // TableName return database table name for xorm.
 func (Pipeline) TableName() string {
 	return "pipelines"
 }
 
+func (p *Pipeline) ToAPIModel() *APIPipeline {
+	ap := &APIPipeline{
+		Pipeline:  p,
+		Commit:    p.Commit.SHA,
+		Title:     p.Commit.Message,
```
anbraten: This should be using sth like the `pr.title` or `release_tag_title` or `cron` based on the event to be backwards compatible I guess.

web/src/components/repo/pipeline/PipelineItem.vue
```diff
@@ -54,7 +54,7 @@
 
         <div class="flex min-w-0 items-center space-x-2">
           <Icon name="commit" />
-          <span class="truncate">{{ pipeline.commit.slice(0, 10) }}</span>
+          <span class="truncate">{{ pipeline.commit_pipeline.sha.slice(0, 10) }}</span>
```
anbraten: As the api is now backwards compatible, I would suggest to extract these changes to a followup PR, so we can check such fundamental changes one-by-one.

cli/exec/flags.go
```diff
@@ -270,11 +270,6 @@ var flags = []cli.Flag{
 		Name:    "commit-author-name",
 		Usage:   "Set the metadata environment variable \"CI_COMMIT_AUTHOR\".",
 	},
-	&cli.StringFlag{
```
anbraten, line 302: I would suggest to extract this change to #5227 or a new PR as it will be breaking.

anbraten (commented):
Nice work. Code mainly lgtm (haven't check on the mappings from the forges to the model too much yet). As those changes are quite fundamental I would suggest to extract all possible parts like the UI and env var changes from this first PR, so we can easily test and find the change if an error occurs.

pipeline/frontend/metadata/types.go
qwerty287: My editor marked this and said that `omitempty` has no effect on struct fields

docs/src/pages/migrations.md
qwerty287: > `is_prerelease` => `is_prerelease` (not `release.is_prerelease` ?)

I didn't add a release field because it would only have the `is_prerelease` as the tag title exists on both tags and releases.

> ??? => `release_tag_title` (not `release.tag_title`? was this the normal `title` before?)

It was extracted from the ref.

docs/src/pages/migrations.md
qwerty287: merge issue

server/api/pipeline.go
qwerty287: Not that I would know. It was unused. If it was, I wouldn't store this information as we don't use it.

web/src/components/repo/pipeline/PipelineItem.vue
qwerty287: I can do, but I don't really think that's necessary because web and server and separated from each other. So you can just review them in one PR without having to think about possible integration issues (as long as the API is compatible).

server/api/pipeline.go
anbraten: In that case we can drop it internally without breaking from myside.

qwerty287: Not quite sure as the field was part of the api I think so removing also removes it from the json fields. If I remember correctly, the value is always the same as `created` so migration shouldn't be hard but still it has to be done if some 3rd-parties use this field

docs/src/pages/migrations.md
anbraten: > I didn't add a release field because it would only have the is_prerelease as the tag title exists on both tags and releases.

In case we later want to add new properties for a release it would be beneficial to have it as an object directly to avoid breaking changes I guess (this is something we should check for all properties now directly).

> It was extracted from the ref.

Perfect. Could you add the updated list to the migrations page.

web/src/components/repo/pipeline/PipelineItem.vue
anbraten: Would favor to do this in small steps as it would allow us to deploy this first and check if the api is still working as before and later on we could check the ui update using the new data fields.

qwerty287: #5255

server/forge/bitbucketdatacenter/bitbucketdatacenter.go
qwerty287: @langecode did you check this out in the meantime?

server/api/pipeline.go
```diff
@@ -593,7 +593,7 @@ func GetPipelineQueue(c *gin.Context) {
 //	@Param			repo_id			path	int		true	"the repository id"
 //	@Param			number			path	int		true	"the number of the pipeline"
 //	@Param			event			query	string	false	"override the event type"
-//	@Param			deploy_to		query	string	false	"override the target deploy value"
+//	@Param			deploy_target	query	string	false	"override the deployment target value"
```
anbraten, line 593:
```suggestion
//	@Param			deploy_to	query	string	false	"deprecated use deploy_target instead"
//	@Param			deploy_target	query	string	false	"override the deployment target value"
```

qwerty287:
Can we get this done so it is in `next` for enough time before the next release?

cmd/server/openapi/docs.go
```diff
@@ -3130,8 +3130,8 @@ const docTemplate = `{
                     },
                     {
                         "type": "string",
-                        "description": "override the target deploy value",
-                        "name": "deploy_to",
+                        "description": "override the deployment target value",
+                        "name": "deploy_target",
```
anbraten, line 3133: This still seems to be breaking change. Could you adjust the comment for openapi and add a todo?

docs/src/pages/migrations.md
```diff
@@ -1,10 +1,33 @@
...
 # Migrations
 
 To enhance the usability of Woodpecker and meet evolving security standards, occasional migrations are necessary. While we aim to minimize these changes, some are unavoidable. If you experience significant issues during a migration to a new version, please let us know so maintainers can reassess the updates.
 
 ## `next`
 
+### User-facing changes
+
 - (Kubernetes) Deprecated `step` label on pod in favor of new namespaced label `woodpecker-ci.org/step`. The `step` label will be removed in a future update.
+- deprecated `CI_COMMIT_AUTHOR_AVATAR` and `CI_PREV_COMMIT_AUTHOR_AVATAR` env vars as commit authors don't have an avatar
```
anbraten: should be adjusted in https://github.com/woodpecker-ci/woodpecker/pull/5227 directly afterwards so ppl can migrate accordingly.

docs/src/pages/migrations.md
```diff
@@ -1,10 +1,33 @@
...
+
+### API changes
+
+- Changed the pipeline model to use nested objects grouped based on the event (e.g. instead of a generic `title` it now uses `pr.title`)
```
anbraten:
```suggestion
- The pipeline model has been changed to use nested objects grouped based on the event (e.g. instead of a generic `title` it now uses `pr.title`). Following properties are deprecated and should be replaced by the their new counterparts:
```

server/forge/forgejo/fixtures/HookPush.json
```diff
@@ -18,6 +18,19 @@
       "modified": ["app/controller/application.rb"]
     }
   ],
+  "head_commit": {
```
anbraten, line 21: Has this been added to forgejo?

server/forge/forgejo/forgejo.go
```diff
@@ -643,34 +667,75 @@ func (c *Forgejo) getChangedFilesForPR(ctx context.Context, repo *model.Repo, in
 	}, -1)
 }
 
-func (c *Forgejo) getTagCommitSHA(ctx context.Context, repo *model.Repo, tagName string) (string, error) {
+func (c *Forgejo) getTagMessage(ctx context.Context, repo *model.Repo, tagName string) (*model.Commit, string, error) {
```
anbraten:
```suggestion
func (c *Forgejo) getTagCommitAndMessage(ctx context.Context, repo *model.Repo, tagName string) (*model.Commit, string, error) {
```

server/forge/forgejo/forgejo.go
```diff
@@ -489,13 +494,32 @@ func (c *Forgejo) Hook(ctx context.Context, r *http.Request) (*model.Repo, *mode
...
+			pipeline.Commit = sha
+		case pipeline.Event == model.EventPull || pipeline.Event == model.EventPullClosed:
+			sha, err := c.getCommitFromSHAStore(ctx, repo, pipeline.Commit.SHA)
+			if err != nil {
+				return nil, nil, err
+			}
+			pipeline.Commit = sha
+		case pipeline.Event == model.EventTag:
+			tagName := strings.Split(pipeline.Ref, "/")[2]
+			sha, msg, err := c.getTagMessage(ctx, repo, tagName)
```
anbraten:
```suggestion
			commit, tagMsg, err := c.getTagMessage(ctx, repo, tagName)
```

server/forge/forgejo/forgejo.go
```diff
@@ -489,13 +494,32 @@ func (c *Forgejo) Hook(ctx context.Context, r *http.Request) (*model.Repo, *mode
...
-	if pipeline != nil && pipeline.Event == model.EventRelease && pipeline.Commit == "" {
-		tagName := strings.Split(pipeline.Ref, "/")[2]
-		sha, err := c.getTagCommitSHA(ctx, repo, tagName)
-		if err != nil {
-			return nil, nil, err
+	if pipeline != nil {
+		switch {
+		case pipeline.Event == model.EventRelease && pipeline.Commit.SHA == "":
+			tagName := strings.Split(pipeline.Ref, "/")[2]
+			sha, _, err := c.getTagMessage(ctx, repo, tagName)
```
anbraten:
```suggestion
			commit, _, err := c.getTagMessage(ctx, repo, tagName)
```

server/forge/forgejo/forgejo.go
```diff
@@ -489,13 +494,32 @@ func (c *Forgejo) Hook(ctx context.Context, r *http.Request) (*model.Repo, *mode
...
+			}
+			pipeline.Commit = sha
+		case pipeline.Event == model.EventTag:
+			tagName := strings.Split(pipeline.Ref, "/")[2]
+			sha, msg, err := c.getTagMessage(ctx, repo, tagName)
+			if err != nil {
+				return nil, nil, err
+			}
+			pipeline.Commit = sha
+			if pipeline.Commit.Message != msg {
```
anbraten: Why should this only be set if its not the same?

server/forge/forgejo/forgejo.go
```diff
@@ -643,34 +667,75 @@ func (c *Forgejo) getChangedFilesForPR(ctx context.Context, repo *model.Repo, in
...
 	}
 
 	tag, _, err := client.GetTag(repo.Owner, repo.Name, tagName)
 	if err != nil {
-		return "", err
+		return nil, "", err
 	}
 
-	return tag.Commit.SHA, nil
+	commit, err := c.getCommitFromSHA(ctx, user, repo, tag.Commit.SHA)
```
anbraten, line 708: Might be cleaner to not include this call directly in here, but call it in the hook directly

server/forge/forgejo/forgejo.go
```diff
@@ -643,34 +667,75 @@ func (c *Forgejo) getChangedFilesForPR(ctx context.Context, repo *model.Repo, in
...
+	return commit, tag.Message, err
+}
+
+func (c *Forgejo) getCommitFromSHAStore(ctx context.Context, repo *model.Repo, sha string) (*model.Commit, error) {
```
anbraten: This name is kinda confusing as it does not get the sha from the store, but loads some other data from it instead to call the forge api underneath

server/forge/gitea/gitea.go
```diff
@@ -496,13 +497,32 @@ func (c *Gitea) Hook(ctx context.Context, r *http.Request) (*model.Repo, *model.
...
-	if pipeline != nil && pipeline.Event == model.EventRelease && pipeline.Commit == "" {
-		tagName := strings.Split(pipeline.Ref, "/")[2]
-		sha, err := c.getTagCommitSHA(ctx, repo, tagName)
-		if err != nil {
-			return nil, nil, err
+	if pipeline != nil {
+		switch {
+		case pipeline.Event == model.EventRelease && pipeline.Commit.SHA == "":
+			tagName := strings.Split(pipeline.Ref, "/")[2]
+			sha, _, err := c.getTagMessages(ctx, repo, tagName)
```
anbraten: same as for forgejo

server/forge/gitea/helper.go
```diff
@@ -267,6 +259,15 @@ func matchingHooks(hooks []*gitea.Hook, rawURL string) *gitea.Hook {
 	return nil
 }
 
+func convertPullRequests(from *gitea.PullRequest) *model.PullRequest {
```
anbraten:
```suggestion
func convertPullRequest(from *gitea.PullRequest) *model.PullRequest {
```

server/forge/gitea/parse_test.go
```diff
@@ -64,15 +64,21 @@ func TestGiteaParser(t *testing.T) {
...
-				Author:       "6543",
-				Event:        "push",
-				Commit:       "28c3613ae62640216bea5e7dc71aa65356e4298b",
-				Branch:       "fdsafdsa",
-				Ref:          "refs/heads/fdsafdsa",
-				Message:      "Delete '.woodpecker/.check.yml'\n",
-				Sender:       "6543",
-				Avatar:       "https://codeberg.org/avatars/09a234c768cb9bca78f6b2f82d6af173",
-				Email:        "6543@obermui.de",
+				Event: "push",
```
anbraten:
```suggestion
				Event: model.EventPush,
```

server/forge/github/convert.go
```diff
@@ -168,3 +178,15 @@ func convertLabels(from []*github.Label) []string {
...
 	return labels
 }
+
+func convertCommit(from *github.Commit) *model.Commit {
+	return &model.Commit{
+		SHA:      from.GetSHA(),
+		ForgeURL: from.GetHTMLURL(),
+		Message:  from.GetMessage(),
+		Author: model.CommitAuthor{
+			Author: from.GetAuthor().GetName(),
```
anbraten: This is actually a quite nice idea to use `author.name` instead of `author.author`

server/forge/github/github.go
```diff
@@ -610,13 +604,21 @@ func (c *client) Hook(ctx context.Context, r *http.Request) (*model.Repo, *model
...
+	if pipeline != nil {
+		if pipeline.Event == model.EventRelease && pipeline.Commit.SHA == "" {
+			tagName := strings.Split(pipeline.Ref, "/")[2]
+			sha, err := c.getTagCommitSHA(ctx, repo, tagName)
+			if err != nil {
+				return nil, nil, err
+			}
+			pipeline.Commit = sha
+		} else if pipeline.Event == model.EventDeploy || pipeline.Event == model.EventPull || pipeline.Event == model.EventPullClosed {
+			sha, err := c.getCommitFromSHA(ctx, repo, pipeline.Commit.SHA)
```
anbraten:
```suggestion
			commit, err := c.getCommitFromSHA(ctx, repo, pipeline.Commit.SHA)
```

server/forge/github/github.go
```diff
@@ -610,13 +604,21 @@ func (c *client) Hook(ctx context.Context, r *http.Request) (*model.Repo, *model
...
 
-	if pipeline != nil && pipeline.Event == model.EventRelease && pipeline.Commit == "" {
-		tagName := strings.Split(pipeline.Ref, "/")[2]
-		sha, err := c.getTagCommitSHA(ctx, repo, tagName)
-		if err != nil {
-			return nil, nil, err
+	if pipeline != nil {
+		if pipeline.Event == model.EventRelease && pipeline.Commit.SHA == "" {
+			tagName := strings.Split(pipeline.Ref, "/")[2]
+			sha, err := c.getTagCommitSHA(ctx, repo, tagName)
```
anbraten:
```suggestion
			commit, err := c.getTagCommitSHA(ctx, repo, tagName)
```

server/forge/github/github.go
```diff
@@ -667,21 +669,20 @@ func (c *client) loadChangedFilesFromPullRequest(ctx context.Context, pull *gith
 	return pipeline, err
 }
 
-func (c *client) getTagCommitSHA(ctx context.Context, repo *model.Repo, tagName string) (string, error) {
+func (c *client) getTagCommitSHA(ctx context.Context, repo *model.Repo, tagName string) (*model.Commit, error) {
```
anbraten: Similar to gitea / forgejo I would improve those two helpers to do less "magic" and rather one thing like one returning the sha and one the commit from a sha or similar.

```suggestion
func (c *client) getCommitAndMessageFromTag(ctx context.Context, repo *model.Repo, tagName string) (*model.Commit, error) {
```

server/forge/github/parse.go
```diff
@@ -85,32 +85,28 @@ func parsePushHook(hook *github.PushEvent) (*model.Repo, *model.Pipeline) {
...
 
-	if len(pipeline.Author) == 0 {
-		pipeline.Author = hook.GetHeadCommit().GetAuthor().GetLogin()
-	}
 	if strings.HasPrefix(pipeline.Ref, "refs/tags/") {
 		// just kidding, this is actually a tag event. Why did this come as a push
 		// event we'll never know!
 		pipeline.Event = model.EventTag
 		pipeline.ChangedFiles = nil
-		// For tags, if the base_ref (tag's base branch) is set, we're using it
```
anbraten, line 123: Shouldn't we keep this?

server/forge/github/github.go
```diff
@@ -610,13 +604,21 @@ func (c *client) Hook(ctx context.Context, r *http.Request) (*model.Repo, *model
 		return nil, nil, err
 	}
 
-	if pipeline != nil && pipeline.Event == model.EventRelease && pipeline.Commit == "" {
-		tagName := strings.Split(pipeline.Ref, "/")[2]
-		sha, err := c.getTagCommitSHA(ctx, repo, tagName)
-		if err != nil {
-			return nil, nil, err
+	if pipeline != nil {
+		if pipeline.Event == model.EventRelease && pipeline.Commit.SHA == "" {
```
anbraten: Maybe a check if message is empty but sha is set to load the commit details would be cleaner and could help loading the corresponding data always if possible. Cculd be especially helpful when adding new events later on.

server/model/feed.go
```diff
@@ -17,21 +17,42 @@ package model
...
+	Started     int64        `json:"started,omitempty"           xorm:"pipeline_started"`
+	Finished    int64        `json:"finished,omitempty"          xorm:"pipeline_finished"`
+	Branch      string       `json:"branch,omitempty"            xorm:"pipeline_branch"`
+	Ref         string       `json:"ref,omitempty"               xorm:"pipeline_ref"`
+	Refspec     string       `json:"refspec,omitempty"           xorm:"pipeline_refspec"`
+	Deployment  *Deployment  `json:"deployment"                  xorm:"json 'pipeline_deployment'"`
+	PullRequest *PullRequest `json:"pull_request,omitempty"      xorm:"json 'pipeline_pr'"`
+	Release     *Release     `json:"release,omitempty" xorm:"json 'pipeline_release'"`
+	// TODO change json to 'commit' in next major
+	Commit *Commit `json:"commit_pipeline,omitempty"   xorm:"json 'pipeline_commit'"`
```
anbraten, line 39: Am I right that we will drop `pipeline_commit` in favor of `commit` in the next major or would we rename again?

server/model/pipeline.go
```diff
@@ -20,46 +20,97 @@ import (
...
+	// event related
+
+	Event WebhookEvent `json:"event"                       xorm:"event"`
+	// TODO change json to 'commit' in next major
+	Commit       *Commit      `json:"commit_pipeline"             xorm:"json 'commit'"`
+	Branch       string       `json:"branch"                      xorm:"branch"`
+	Ref          string       `json:"ref"                         xorm:"ref"`
+	Refspec      string       `json:"refspec"                     xorm:"refspec"`
+	ForgeURL     string       `json:"forge_url"                   xorm:"forge_url"`
+	Author       string       `json:"author"                      xorm:"author"`
```
anbraten:
```suggestion
	Author       string       `json:"author"                      xorm:"author"` // The user sending the webhook data or triggering the pipeline event
```

server/model/pipeline.go
```diff
@@ -78,3 +129,14 @@ type PipelineOptions struct {
...
+
+type Deployment struct {
+	Target      string `json:"target"`
+	Task        string `json:"task"`
+	Description string `json:"description"`
+}
+
+type Release struct {
+	IsPrerelease bool   `json:"is_prerelease,omitempty"`
+	TagTitle     string `json:"release_tag_title,omitempty"`
```
anbraten:
```suggestion
	TagTitle     string `json:"tag_title,omitempty"`
```

Do we want to know the tag title or the release title here?

woodpecker-go/woodpecker/types.go
```diff
@@ -96,32 +96,54 @@ type (
...
+		Parent      int64            `json:"parent"`
+		Event       string           `json:"event"`
+		Status      string           `json:"status"`
+		Errors      []*PipelineError `json:"errors"`
+		Created     int64            `json:"created_at"`
+		Updated     int64            `json:"updated_at"`
+		Started     int64            `json:"started_at"`
+		Finished    int64            `json:"finished_at"`
+		Deployment  *Deployment      `json:"deployment"`
+		Commit      *Commit          `json:"commit"`
```
anbraten: `pipeline_commit`?

server/forge/forgejo/fixtures/HookPush.json
qwerty287: The fixtures we are using are just an excerpt of the real hooks. So I added this. So yes forgejo sends this data.

server/forge/github/parse.go
qwerty287: I don't think so. Git tags don't have a branch by themselves, they're just linked to the corresponding commit

server/model/feed.go
qwerty287: I would like to rename it to just `commit` in the next major when we drop `commit` being the SHA.

server/model/pipeline.go
qwerty287: It was actually mixed which I would avoid.

I would add a new tag type and field to store the tag's name and its message. Is that fine?

server/forge/github/github.go
qwerty287: This doesn't work afaik because the github release hook does not include the commit sha at all, just the tag name. So it has to fetch the data by the tag name.

woodpecker-go/woodpecker/types.go
qwerty287: changed it to a string so it uses the old value for now

server/forge/forgejo/forgejo.go
qwerty287: I just adapted to the style that was used before as well. Can we refactor that in another PR?

server/forge/forgejo/forgejo.go
qwerty287: I renamed it to `getCommitFromSHAWithUserFromStore` which is a bit long - you've got a different idea?

server/forge/forgejo/forgejo.go
anbraten: sure, we can clean those helpers up in a followup.

server/forge/forgejo/forgejo.go
anbraten: No nice idea, would change the helpers scope in a followup and rename in that moment than.

woodpecker-go/woodpecker/types.go
```diff
@@ -96,32 +96,54 @@ type (
...
+
+	Commit struct {
+		SHA      string        `json:"sha"`
+		Message  string        `json:"message"`
+		ForgeURL string        `json:"forge_url"`
+		Author   *CommitAuthor `json:"author"`
+	}
+
+	CommitAuthor struct {
+		Author string `json:"author"`
```
anbraten:
```suggestion
		Name string `json:"name"`
```

docs/docs/20-usage/50-environment.md
```diff
@@ -77,7 +77,6 @@ This is the reference list of all environment variables available to your pipeli
 | `CI_COMMIT_MESSAGE`              | commit message                                                                                                     | `Initial commit`                                                                           |
 | `CI_COMMIT_AUTHOR`               | commit author username                                                                                             | `john-doe`                                                                                 |
 | `CI_COMMIT_AUTHOR_EMAIL`         | commit author email address                                                                                        | `john-doe@example.com`                                                                     |
-| `CI_COMMIT_AUTHOR_AVATAR`        | commit author avatar                                                                                               | `https://git.example.com/avatars/5dcbcadbce6f87f8abef`                                     |
```
anbraten: Should be in the migration notes or reverted and be changed as part  of #5227?!

server/store/datastore/migration/026_update_pipeline_messages.go
```diff
@@ -0,0 +1,159 @@
...
+			FromFork bool                `json:"from_fork,omitempty"`
+		}
+
+		type deployment struct {
+			Target      string `json:"target"`
+			Task        string `json:"task"`
+			Description string `json:"description"`
+		}
+
+		type pipelines struct {
```
anbraten, line 70:
```suggestion
		type pipeline struct {
```

server/store/datastore/migration/026_update_pipeline_messages.go
```diff
@@ -0,0 +1,159 @@
...
+			DeployTask        string   `xorm:"deploy_task"`
+			PullRequestLabels []string `xorm:"json 'pr_labels'"`
+			FromFork          bool     `xorm:"from_fork"`
+
+			// new fields
+			CommitNew       *commit      `xorm:"json 'commit_new'"`
+			Deployment      *deployment  `xorm:"json 'deployment'"`
+			PullRequest     *pullRequest `xorm:"json 'pr'"`
+			Cron            string       `xorm:"cron"`
+			ReleaseTagTitle string       `xorm:"release_tag_title"`
```
anbraten: this needs to be adjusted?

anbraten: would it help to define separate types for old and new pipelines?

server/store/datastore/migration/026_update_pipeline_messages.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"go.woodpecker-ci.org/woodpecker/v3/server/model"
+)
+
+// perPage024 set the size of the slice to read per page.
+var perPage024 = 100
+
+var updatePipelineMessages = xormigrate.Migration{
+	ID: "update-pipeline-messages",
+	MigrateSession: func(sess *xorm.Session) error {
```
anbraten, line 33: Before merging we should do one final test using a copy of our production db.

anbraten (commented):
Hope I detected all remaining pieces. Should be almost rdy to go.

server/forge/forgejo/forgejo.go
qwerty287: What do you suggest for now?

server/store/datastore/migration/026_update_pipeline_messages.go
qwerty287: This should be plural because the table name is constructed from this and the table name is `pipelines`

server/store/datastore/migration/026_update_pipeline_messages.go
qwerty287: fixed. I wouldn't split

server/store/datastore/migration/026_update_pipeline_messages.go
qwerty287: Can you take care of this or should I do?

qwerty287:
What still needs to be done and tested is storing of the tag name. @anbraten I changed some parts there again, but it's not finished yet. Sorry btw that my code quality totally decreased with the reviews…

qwerty287:
I changed the tag storing to use its own field independent from release called `tag_title`

server/api/pipeline_test.go
```diff
@@ -37,6 +37,7 @@ var fakePipeline = &model.Pipeline{
 	ID:     2,
 	Number: 2,
 	Status: model.StatusSuccess,
+	Commit: &model.Commit{},
```
6543, line 49: why?

server/forge/bitbucket/internal/client.go
```diff
@@ -206,6 +207,13 @@ func (c *Client) GetBranchHead(owner, name, branch string) (*Commit, error) {
 	return out.Values[0], nil
 }
 
+func (c *Client) GetCommit(owner, name, sha string) (*Commit, error) {
```
6543, line 210: should this not be it's own pull?

server/model/pipeline.go
```diff
@@ -32,34 +32,92 @@ type Pipeline struct {
...
-	Reviewed            int64                  `json:"reviewed"                xorm:"reviewed"`
+	Reviewed            int64                  `json:"reviewed"                xorm:"reviewed"` // timestamp of the review
 	Workflows           []*Workflow            `json:"workflows,omitempty"     xorm:"-"`
 	ChangedFiles        []string               `json:"changed_files,omitempty" xorm:"LONGTEXT 'changed_files'"`
 	AdditionalVariables map[string]string      `json:"variables,omitempty"     xorm:"json 'additional_variables'"`
-	PullRequestLabels   []string               `json:"pr_labels,omitempty"     xorm:"json 'pr_labels'"`
-	IsPrerelease        bool                   `json:"is_prerelease,omitempty" xorm:"is_prerelease"`
-	FromFork            bool                   `json:"from_fork,omitempty"     xorm:"from_fork"`
+	Deployment          *Deployment            `json:"deployment,omitempty"    xorm:"json 'deployment'"`
+	PullRequest         *PullRequest           `json:"pull_request,omitempty"  xorm:"json 'pull_request'"`
```
6543: if we drop sqlite and stick to postgresql then we can do so ... else we have to normalize our database!

server/forge/bitbucket/internal/client.go
qwerty287: What do you mean by "pull"?

server/model/pipeline.go
qwerty287: What do you mean? Using the json feature works with sqlite as well. I guess xorm is doing this directly then.

6543: yes as data-blob you can store it ... but you can then no longer do standart sql querys e.g. for filtering purpose.

we should filter more stuff in the backend and not the frontend as we do now ...

6543: qostgresql is an exeption as it can understand json stored objects andquery inside that json stuff ... but that's non sql behaviour - its NoSQL e.g. an document database in this case

qwerty287: Yes, that's correct, but I don't think we ever have to filter using an item of the json object. Only for the PR index this might be helpful.

What would be your approach for this? Adding a new `pull_request` table? We can do this as well (would be probably the most clean, but also most complex solution) but I would like to not do this in this PR.

server/forge/bitbucket/internal/client.go
6543: I'm adding this addition into https://github.com/woodpecker-ci/woodpecker/pull/5410

6543:
conflicts because of me ... if ok, I'll try to resolve them and push on to your branch Y/N?

qwerty287:
Sure, go on. Thanks

qwerty287:
@woodpecker-ci/maintainers could I ask again for reviews here? This is open for almost a year now. And it simplifies the pipeline structure a lot. (Also feel free to directly modify the branch if there's something to fix)

server/api/pipeline_test.go
qwerty287: Tests fail otherwise with nil pointer. This way e.g. `Commit.SHA` is defined as empty string.

anbraten:
Thanks for updating to main again. I will try to have a look at it and will do some testing. Could you try to test with a copy of our ci.woodpecker-ci.org db as well. Last time it had some hickups with the migrations.

qwerty287:
@anbraten I just tested this with our DB, and it seems to be working. 

Check out an excerpt of the table:
[database-excerpt.csv](https://github.com/user-attachments/files/24788970/database-excerpt.csv)

The columns `commit_new` and `pr` contain the new stuff.

Migration takes a very long time however (ca. 15 min). 

It seems there are many weird unnecessary columns in our DB like `pipeline_signed`?

6543:
> Migration takes a very long time however (ca. 15 min).

it it takes 15min for us it will take days for big instances!!! that's not good at all :/

I'll benchmark it and ask the codeberg-ci maintainer to test a migration and give me the times

qwerty287:
@6543 any results? I could also change that to be an optional migration, and pipelines that are not migrated won't be visible until the migration ran. Similar to how we did that with the logs.

server/api/pipeline.go
```diff
@@ -78,26 +78,27 @@ func CreatePipeline(c *gin.Context) {
...
 	if err != nil {
 		handlePipelineErr(c, err)
+		return
+	}
+
+	if pl != nil {
+		c.JSON(http.StatusOK, pl.ToAPIModel())
 	} else {
-		c.JSON(http.StatusOK, pl)
+		c.String(http.StatusOK, "nothing to run")
```
6543: api expects json ... we should not change that

esp. in a non error paths

server/pipeline/create.go
```diff
@@ -42,13 +42,9 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline
 	}
 
 	if pipeline.Event == model.EventPush || pipeline.IsPullRequest() {
-		skipMatch := skipPipelineRegex.FindString(pipeline.Message)
+		skipMatch := skipPipelineRegex.FindString(pipeline.Commit.Message)
```
6543: we should this check as public func into the pipeline/ package ...

so for here we might add an TODO comment

server/pipeline/gated_test.go
```diff
@@ -79,8 +79,10 @@ func TestSetGatedState(t *testing.T) {
 				RequireApproval: model.RequireApprovalPullRequests,
 			},
 			pipeline: &model.Pipeline{
-				Event:    model.EventPullMetadata,
-				FromFork: false,
+				Event: model.EventPullMetadata,
+				PullRequest: &model.PullRequest{
+					FromFork: false,
+				},
```
6543:
```suggestion
				PullRequest: &model.PullRequest{FromFork: false},
```

server/store/datastore/feed_test.go
```diff
@@ -75,10 +79,11 @@ func TestGetPipelineQueue(t *testing.T) {
 	assert.Equal(t, pipeline1.Commit, feedItem.Commit)
 	assert.Equal(t, pipeline1.Branch, feedItem.Branch)
 	assert.Equal(t, pipeline1.Ref, feedItem.Ref)
-	assert.Equal(t, pipeline1.Title, feedItem.Title)
-	assert.Equal(t, pipeline1.Message, feedItem.Message)
+	assert.Equal(t, pipeline1.Commit.Message, feedItem.Commit.Message)
+	assert.Equal(t, pipeline1.Commit.SHA, feedItem.Commit.SHA)
```
6543, line 83: well this requires WebUI changes too...

6543:
It's a huge code diff... I would recommend to cherry-pick parts that can be done seperate into smaller pulls, so we move forward with this :)

qwerty287:
I'll try to check again in the next days.

6543:
@qwerty287 I like the approach, and if was in smaller chunks it would be mostly merged already :sweat_smile: 

I just have some concerns: server side filtering based on json -> only supported by postgress
so if we have metadata we wana filter based on that, we have to query DB and do filtering in our server code, that's inefficient. but the other option would be to not inline json but create additional tables in the DB, witch makes the DB stuff more complicated :/

also some changes are breaking changes (esp. the API stuff, so i would move that out in a dedicated pull to be blocked till we wana do majour update)

server/store/datastore/migration/028_update_pipeline_structure.go
```diff
@@ -0,0 +1,176 @@
...
+					}
+					newPipeline.TagTitle = strings.TrimPrefix(oldPipeline.Ref, "refs/tags/")
+				case model.EventTag:
+					newPipeline.TagTitle = strings.TrimPrefix(oldPipeline.Ref, "refs/tags/")
+				case model.EventCron:
+					newPipeline.Cron = oldPipeline.Sender
+				case model.EventPull, model.EventPullClosed:
+					newPipeline.PullRequest = &pullRequest{
+						Title: oldPipeline.Title,
+						Index: model.ForgeRemoteID(
```
6543: i would move that out into a helper func to have more code redabilety

qwerty287:
> I just have some concerns: server side filtering based on json -> only supported by postgress
> so if we have metadata we wana filter based on that, we have to query DB and do filtering in our server code, that's inefficient. but the other option would be to not inline json but create additional tables in the DB, witch makes the DB stuff more complicated :/

Yes I know. However the only thing that can be useful for filtering that is wrapped in json is the PR index. I think this should definitely get moved out. But of course you could extend that. We could add full-text-search for pull request title or commit message.

Is there a way to only let xorm split those nested fields into separate columns but keep them nested in Go? That would be the best way in my opinion.

> also some changes are breaking changes (esp. the API stuff, so i would move that out in a dedicated pull to be blocked till we wana do majour update)

Please tell me what's breaking. I tried to make it non-breaking already, what did I miss?

server/store/datastore/feed_test.go
qwerty287: No, it should not, as the commit field itself did not change for the API. That's why we have the `APIPipeline` struct.

6543:
efen with xorm you can write raw sql querys, and postgress allows json filters... so it's not a huge tradeoff, we could state, that for big installation we only recomed tjat database and for the rest we filter via go-code

we just should make that deccission in consensus

https://www.postgresql.org/docs/current/functions-json.html

the only problem might be schema update not supported by xorm (jet ... we could upstream that)

qwerty287:
@6543 what do you think about adding a second model that splits the nested fields and use that model for the database? So we can use xorm handling and add some conversion function to the regular model so we can use more clean code.

I.e. if we have a field like
```go
PullRequest struct{
    Index int `json:"index"`
    Title string `json:"title"`
}
```

then in the DB model do

```go
PullRequestIndex int `xorm:"pull_request_index"`
PullRequestTitle string `xorm:"pull_request_title"`
```

I know that this is basically the structure of how it was before, but it still allows us to write clean code in woodpecker but still use xorm without hacks.

server/pipeline/create.go
6543: -> https://github.com/woodpecker-ci/woodpecker/pull/6437

just did the refactoring

6543:
well about the DB normalisation ... i think we can still do that later ... so to move this forward let that as is and if needed we can always have a followup.

I would suggest to merge this avter the v3.14 release so we have time to pick up issues

server/api/pipeline.go
```diff
@@ -90,19 +90,15 @@ func CreatePipeline(c *gin.Context) {
 
 func createTmpPipeline(event model.WebhookEvent, commit *model.Commit, user *model.User, opts *model.PipelineOptions) *model.Pipeline {
 	return &model.Pipeline{
-		Event:     event,
-		Commit:    commit.SHA,
-		Branch:    opts.Branch,
-		Timestamp: time.Now().UTC().Unix(),
```
6543, line 96: the timestamp where did that info go?

server/model/pipeline.go
```diff
@@ -73,13 +80,46 @@ func (Pipeline) TableName() string {
...
+	case EventTag:
+		ap.Message = p.TagTitle
+	case EventRelease:
+		if p.Release != nil {
+			ap.IsPrerelease = p.Release.IsPrerelease
+			ap.Title = p.Release.Title
+		}
+		ap.Message = "created release " + p.TagTitle
+	case EventManual:
+		ap.Message = "MANUAL PIPELINE" + p.Branch
```
6543:
```suggestion
		ap.Message = "MANUAL PIPELINE @ " + p.Branch
```

server/api/pipeline.go
6543: cah created...

6543:
the only thing blocking is that i have to review the migration part ... but it looks costly ...

qwerty287:
We still need to discuss how to work with the migration time. As you said it will take days on instances like codeberg… https://github.com/woodpecker-ci/woodpecker/pull/4626#issuecomment-3808027801

Also I'd like to finish the discussion on the database model before merging, otherwise we basically revert that at DB level later… https://github.com/woodpecker-ci/woodpecker/pull/4626#issuecomment-4204265522

server/forge/gitlab/convert.go
```diff
@@ -275,22 +276,30 @@ func convertPushHook(hook *gitlab.PushEvent) (*model.Repo, *model.Pipeline, erro
...
+	pipeline.Author = hook.UserUsername
+	pipeline.AuthorAvatar = hook.UserAvatar
 
 	pipeline.Author = hook.UserUsername
 
 	// assume a capacity of 4 changed files per commit
 	files := make([]string, 0, len(hook.Commits)*4)
 	for _, cm := range hook.Commits {
 		if hook.After == cm.ID {
+<<<<<<< HEAD
```
6543: merge conflict relicts got in with the last merge

6543:
thx!

server/model/pipeline.go
```diff
@@ -21,50 +21,57 @@ import (
...
+	Parent              int64                   `json:"parent"                  xorm:"parent"`
+	Event               WebhookEvent            `json:"event"                   xorm:"event"`
+	EventReason         []string                `json:"event_reason"            xorm:"json 'event_reason'"`
+	Status              StatusValue             `json:"status"                  xorm:"INDEX 'status'"`
+	Errors              []*errors.PipelineError `json:"errors"                  xorm:"json 'errors'"`
+	Created             int64                   `json:"created"                 xorm:"'created' NOT NULL DEFAULT 0 created"`
+	Updated             int64                   `json:"updated"                 xorm:"'updated' NOT NULL DEFAULT 0 updated"`
+	Started             int64                   `json:"started"                 xorm:"started"`
+	Finished            int64                   `json:"finished"                xorm:"finished"`
+	Commit              *Commit                 `json:"commit_pipeline"         xorm:"json 'commit'"` // TODO change json to 'commit' in next major
```
6543, line 39: this colides

6543:
I found a lot of issues in the migration code ... this is now all fixed ... but I am by no means confident to have catched them all :/

-> I think we should migrate one data type at once ... this makes pull requests smaller ... I'll do so tomorrow

server/api/pipeline.go
6543: timestamp is now proper in commit ...

server/pipeline/create.go
```diff
@@ -40,12 +40,18 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline
...
-	if constraint.IsSkipCommitMessage(metadata.Event(pipeline.Event), pipeline.Message) {
-		ref := pipeline.Commit
+	var commitMessage, commitSHA string
+	if pipeline.Commit != nil {
+		commitMessage = pipeline.Commit.Message
+		commitSHA = pipeline.Commit.SHA
+	}
+
+	if constraint.IsSkipCommitMessage(metadata.Event(pipeline.Event), commitMessage) {
+		ref := commitSHA
```
6543:
```suggestion
```

server/pipeline/create.go
```diff
@@ -40,12 +40,18 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline
...
 		if len(ref) == 0 {
 			ref = pipeline.Ref
 		}
-		log.Debug().Str("repo", repo.FullName).Msgf("ignoring pipeline as skip-ci was found in the commit (%s) message '%s'", ref, pipeline.Message)
+		log.Debug().Str("repo", repo.FullName).Msgf("ignoring pipeline as skip-ci was found in the commit (%s) message '%s'", ref, commitMessage)
```
6543:
```suggestion
		log.Debug().Str("repo", repo.FullName).Msgf("ignoring pipeline as skip-ci was found in the commit (%s) message '%s'", commitSHA, commitMessage)
```

6543:
split out:
- tag & release related changes: https://github.com/woodpecker-ci/woodpecker/pull/6774
- e2e test: https://github.com/woodpecker-ci/woodpecker/pull/6775
- https://github.com/woodpecker-ci/woodpecker/pull/6772

6543:
commit: #6779 

but that one was extracted with this pull diff and claude opus4.8 ... so before i can give it free for review, it needs to be scrutinized by me first ;)

6543:
will rebase as #6774 was merged :tada: :rocket:

PR #4628, opened 2024-12-27
Add `prettier` to `docs/`

pat-s:
With same config as for `web/`


.woodpecker/docs.yaml
```diff
@@ -41,18 +41,28 @@ steps:
     image: *golang_image
     commands:
       - make docs
+    depends_on: prettier
     when:
       - path: *when_path
         event: [tag, pull_request, push]
       - event: manual
 
+  prettier:
```
xoxys, line 50: Can we put this step before the build-cli step if it depends on this one?

xoxys (approved):

Issue #4616, opened 2024-12-24
Inline styles


zc-devs:
### Component

web-ui

### Describe the bug

The code contains the inline styles.
Inline style, seems, was only in `web/src/components/admin/settings/queue/AdminQueueStats.vue:25` before.

### Expected behavior

Should not be inline styles.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/style
https://stackoverflow.com/questions/2612483/whats-so-bad-about-in-line-css
https://content-security-policy.com/unsafe-inline/

### System Info

`next`

### Additional context

I accidentally saw #4609 and wanted to comment there, but I'm not allowed to comment on `pat-s` issues and PRs.

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

xoxys:
> I accidentally saw https://github.com/woodpecker-ci/woodpecker/pull/4609 and wanted to comment there, but I'm not allowed to comment on pat-s issues and PRs.

Can you please stop with such stuff? Most of the time your input is helpful and is taken into account, but some comments seem as if they are only intended to disturb development without making any constructive contribution. Thus, they might get ignored. If you have constructive feedback to add, pretty sure nobody told you to stop on PRs of a particular author.

zc-devs:
> Can you please stop with such stuff?

Sorry, I do not understand, what are you asking to stop? But, sure, it won't be a problem from my side. Are you asking to stop filing the issues?

> I accidentally saw https://github.com/woodpecker-ci/woodpecker/pull/4609 and wanted to comment there, but I'm not allowed to comment on pat-s issues and PRs.

I just explained, why this issue appears. The same as with #4619.

> they are only intended to disturb development

OMG!

> without making any constructive contribution
> If you have constructive feedback to add

Testing and filing a bugs are not contribution, aren't? Or they are, but are not constructive? The are not constructive in general? Only my? Or only this one? How can I improve this one?

> they might get ignored

No problem. Feel free to close, if you find it non-constructive, disturbing or whatever.

[...]

PR #4629, opened 2024-12-27
Avoid usage of inline css style

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/4616


pat-s:
There are also some instances in `docs/` - do you want to fix them as well?

xoxys:
Not in this PR. We dont use a css framework for docs so not that easy as we use dynamic values from react vars in some places.

pat-s (approved):

web/src/components/admin/settings/queue/AdminQueueStats.vue
```diff
@@ -21,8 +21,7 @@
                 v-for="item in data"
                 :key="item.key"
                 class="h-full"
-                :class="`${item.color}`"
-                :style="{ width: `${item.percentage}%` }"
+                :class="[item.color, item.percentage <= 0 ? 'w-0' : `w-[${item.percentage}%]`]"
```
anbraten, line 24: Is this working in production builds? Might be that windi / tailwind would not be able to determine the necessary classes to bundle beforehand.

web/src/components/repo/pipeline/PipelineStepList.vue
```diff
@@ -90,12 +90,7 @@
           </div>
           <div
             class="transition-height duration-150 overflow-hidden"
-            :class="{
-              'max-h-0': workflowsCollapsed[workflow.id],
-            }"
-            :style="{
-              marginLeft: !singleConfig ? '1.6rem' : '',
-            }"
+            :class="[workflowsCollapsed[workflow.id] ? 'max-h-0' : '', !singleConfig ? `ml-[1.6rem]` : '']"
```
anbraten, line 93:
```suggestion
            :class="{ 'max-h-0' : workflowsCollapsed[workflow.id], `ml-[1.6rem]` : !singleConfig }"
```

web/src/components/admin/settings/queue/AdminQueueStats.vue
xoxys: Where is the difference between production builds/dev compose stack?

web/src/components/repo/pipeline/PipelineStepList.vue
xoxys: Can you explain this change?

web/src/components/admin/settings/queue/AdminQueueStats.vue
xoxys: Ok got it. Yes you are right this cant work.

xoxys: Reverted. While we could use workarounds like 

```
<script lang="ts" setup>
const getStatWidthClass = computed(() => {
  return (percentage: number) => {
    if (percentage <= 0) return 'w-0'
    if (percentage <= 5) return 'w-[5%]'
    if (percentage <= 10) return 'w-[10%]'
    if (percentage <= 15) return 'w-[15%]'
    // ... continue for each 5% increment
    if (percentage <= 95) return 'w-[95%]'
    return 'w-full'
  }
})
</script>
```

this seems to be way more complicated and adds a lot of maintenance overhead instead of using inline style for this functionality.

anbraten (approved):

PR #4636, opened 2024-12-28
Remove workflow-level volumes and networks

qwerty287:
Undocumented and unusable because the yaml values were not passed through.


pat-s:
Fine for me in general if unused, but there there ever an idea behind it? Maybe checking with the person who put it there (git blame) just to make sure we don't overlook anything?

xoxys (approved):

pat-s (approved):

pat-s (approved):

pat-s (approved):

PR #4637, opened 2024-12-28
fix `depends_on` for docs workflow

pat-s:
https://ci.woodpecker-ci.org/repos/3780/pipeline/24401/7


xoxys:
I feared such issues and I thought about it in the original PR already. Do we really need this complex dependency chain here? I know you tried to parallelize steps but its pretty hard to read what depends on what especially if also complex file-based filters are also involved...

xoxys (approved):

PR #4638, opened 2024-12-28
Drop dependency chain from docs workflow

xoxys:
Fixes: https://ci.woodpecker-ci.org/repos/3780/pipeline/24411

Introduced in https://github.com/woodpecker-ci/woodpecker/pull/4628 and still not fixed after https://github.com/woodpecker-ci/woodpecker/pull/4637 I think we should just drop the dependency chain, see my comment in https://github.com/woodpecker-ci/woodpecker/pull/4637#issuecomment-2564373206

### DIFF
--- a/.woodpecker/docs.yaml
+++ b/.woodpecker/docs.yaml
@@ -41,7 +41,9 @@ steps:
     image: docker.io/woodpeckerci/plugin-prettier:1.0.0
     settings:
       version: 3.3.3
-
+    when:
+      event:
+        pull_request
   build-cli:
     image: *golang_image
     commands:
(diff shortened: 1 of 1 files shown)


pat-s:
I did this as the workflow should fail early if there is a linter error. El

Without a dep chain, prettier starts alongside the others and it might take some secs until an error state is reached. 
Yes, it might only up to 10s but I still thought it might be worthwhile. But I also see the issues that come with it, especially as we can't really test them if we don't have proper dry runs. 

Won't block, leaving the decision to somebody else.

xoxys:
Without a dependency chain steps run in sequential order. As prettier is the first one nothing will run if prettier fails.

xoxys:
For the record: that we allow multiple event syntaxes is such a pain...

pat-s (approved):

PR #4639, opened 2024-12-29
Inline web helpers

qwerty287:
Functions used once just wrapping another func are just useless…


xoxys:
While I think using external lib methods directly makes sense why do you also moved our own helper functions from a central place? Whats the benefit?

PR #4641, opened 2024-12-29
Add priority menu to tabs

xoxys:
Supersedes: https://github.com/woodpecker-ci/woodpecker/pull/4610
Fixes: https://github.com/woodpecker-ci/woodpecker/discussions/2162#discussioncomment-11355337

Adds a priority plus menu to tabs that contains all tabs items that don't fix into the available screen size. The wordwrap issue seen in one of the demo videos was fixed already.

https://github.com/user-attachments/assets/0064550c-dbdf-49fe-bdd8-4e6d0eeea5b2

https://github.com/user-attachments/assets/f65d20ad-91df-4ae6-940b-538548b840a2

### DIFF
--- a/web/src/components/atomic/Icon.vue
+++ b/web/src/components/atomic/Icon.vue
@@ -48,6 +48,7 @@
   <SvgIcon v-else-if="name === 'pause'" :path="mdiPause" size="1.3rem" />
   <SvgIcon v-else-if="name === 'play'" :path="mdiPlay" size="1.3rem" />
   <SvgIcon v-else-if="name === 'play-outline'" :path="mdiPlayOutline" size="1.3rem" />
+  <SvgIcon v-else-if="name === 'dots'" :path="mdiDotsVertical" size="1.3rem" />
 
   <SvgIcon v-else-if="name === 'visibility-private'" :path="mdiLockOutline" size="1.3rem" />
   <SvgIcon v-else-if="name === 'visibility-internal'" :path="mdiLockOpenOutline" size="1.3rem" />
@@ -93,6 +94,7 @@ import {
   mdiCloseCircle,
   mdiCog,
   mdiCogOutline,
+  mdiDotsVertical,
   mdiDownloadOutline,
   mdiEyeOffOutline,
   mdiEyeOutline,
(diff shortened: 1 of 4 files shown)


pat-s:
I agree this is a better one, thanks for putting the effort in. 

Can we use horizontal dots instead of vertical ones? This should also be the common default for other sites.

xoxys:
Sure I can but without a border around the button and beside text this variant looks better IMO

xoxys:
@pat-s you would still prefer horizontal dots?

pat-s:
I can also live with vertical ones if you have a strong preference for it.

Yet this is what I see on iPhone 15 (safari) with default settings:

image

This is caused if the language is "german" as the "pipeline ausführen" button is then too large and collapses everything on the left. The dots then expand to the left and leave the screen.

Would it be complicated to enforce showing at least one item at all times? This would then avoid the menu issue going out of bounds and improve the overall feel.

xoxys:
Good catch, will fix that case.

xoxys:
@anbraten as your javascript knowledge is way better than mine your feedback would be very welcome.

pat-s:
@xoxys Do you want to wait here until @anbraten replied?

xoxys:
No we can proceed IMO. Can be improved later as well if there is any feedback.

pat-s:
@xoxys Leaving the conflicts to you.

pat-s:
The duration infos should be place above the row containing the menu? I am not sure if you wanted to do this on purpose to save space or if this is an undesired change?

image

xoxys:
Fixed it:

https://github.com/user-attachments/assets/374d3fa2-13a7-4a2a-afe0-fd6cec448a1a

xoxys:
I also adjusted the drop down:

image

The counter badge was removed intentionally due to the limited space.

pat-s:
Thanks! The only last nit I would have is that the menu doesn't vanish if one clicks on an element outside the menu that is within the overall tab list (e.g. 'tasks' or 'errors').
Seems like a niche/edge case, I am also fine with the current state!

pat-s (approved):

xoxys:
Lets fix it before merge. Should be easy enough :)

pat-s (approved):
Feels great now!

(and just a "little bit" of work in the end 😆️)

Issue #2112, opened 2023-08-04
woodpecker-cli misnamed in RPM and DEB releases


inflatador:
### Component

cli

### Describe the bug

woodpecker-cli's RPM and DEB packages install a binary called 'woodpecker' as opposed to 'woodpecker-cli'. The binary appears to work, but this is confusing as the examples (and tarball releases) use the name 'woodpecker-cli'.

### System Info

```shell
Fedora 38 (RPM), Debian 11 (DEB)
```

### Additional context

Fedora:
```
$sudo rpm -ivh https://github.com/woodpecker-ci/woodpecker/releases/download/v1.0.0/woodpecker-cli-1.0.0.x86_64.rpm
$rpm -ql woodpecker-cli
/usr/local/bin/woodpecker
```

Debian:
```
 $sudo dpkg -i woodpecker-cli_1.0.0_amd64.deb
Selecting previously unselected package woodpecker-cli.
$ dpkg -L woodpecker-cli
/usr
/usr/local
/usr/local/bin
/usr/local/bin/woodpecker
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

inflatador:
In that case, should I change the report to 'standardize woodpecker-cli binary name' or something like that?

6543:
binary currently should be:

```
woodpecker         # cli
woodpecker-server  # server
woodpecker-agent   # agent
```

as it's handy for autocompletion in shell, I would argue ... but if the package name also should just be called that way I dont know ...

PR #4643, opened 2024-12-29
fix cli exec statement in debug tab

pat-s:
At least for me on macOS, the cli is always `woodpecker-cli` and not `woodpecker`. Our releases indicate that this is also the case for all other OS, so I guess this is just a typo?

Besides, I couldn't get the debug execution to work even though I followed the instructions. Might have done something wrong but if I don't understand the issue, others won't either?

After skimming over the PR, I think there should be some information highlighting that secrets are not included (obviously) and must be injected manually. Instructions should also show how this is done.

I think it would be good if the section would contain a link to the docs explaining everything in more detail.


xoxys:
I would say lets fix the binary name for the cli then?

lafriks (approved):

pat-s:
Given that most people use the tarball or other means of installation, it should likely be `woodpecker-cli` instead of `woodpecker` unless this should be breaking change then.

pat-s:
So #2112 states that it should be `woodpecker`. So what now? A follow-up PR with a breaking change for the woodpecker CLI name of the tarball and macOS brew?

PR #4099, opened 2024-09-10
Clarify usage of secrets in docs

6543:
as per https://github.com/woodpecker-ci/woodpecker/issues/4096#issuecomment-2339769294

PR #4644, opened 2024-12-29
Refactor `secrets` page in docs

pat-s:
Superseedes #4099


docs/docs/20-usage/40-secrets.md
```diff
@@ -1,45 +1,60 @@
...
+Woodpecker can mask secrets from its native secret store, but it cannot apply the same protection to external secrets. As a result, these external secrets may be exposed in the pipeline logs.
+:::
 
 ## Usage
 
-You can set a setting or environment value from secrets using the `from_secret` syntax.
+You can set a setting or environment value from Woodpecker secrets using the `from_secret` syntax.
 
-The example below passes a secret called `secret_token` as an environment variable that will be called `TOKEN_ENV`:
+The example below passes a secret called `SECRET_TOKEN` which will be stored in an environment variable named `TOKEN_ENV`:
```
lafriks, line 28: Might be just my preference but imho in example it would be better to keep `secret_token` lowercase as it was

pat-s: IIRC we agreed on using uppercase secret refs everywhere some time ago. Can maybe somebody else verify this or was this a dream of mine?

lafriks: Have not seen such a discussion, imho it would be good if we would use documentation same as we use in woodpecker pipelines in repos (we also have secret names lowercase) as we should follow our own recommendations to show example. But to be honest I don't see any good reason to use secret names uppercase

lafriks: That should be somewhere in the description of secret naming, examples imho should have consistent naming to not make confusion

xoxys: As it is mixed in the docs already can we move this discussion/cleanup in a dedicated PR?

pat-s: It might have confused this discussion then with one that at least discussed the uppercase/lowercase mixup in general. At least there was something about uppercasing all secrets, maybe only the env vars in the end. Anyhow.

xoxys (approved):

PR #4650, opened 2024-12-31
Fix workflow volume and network config

anbraten:
Reverts the change from `Volumes` to `Volume` of the backend config from #4636. Apart from the nil-pointer error using arrays seems to be more consistent here. 

Changing the backend config would need a rpc bump. We might also think of how we can make sure agents would not pick up "old" jobs from the queue, so we would need to do sth like dropping the queue in such cases 🤔

### DIFF
--- a/pipeline/backend/types/config.go
+++ b/pipeline/backend/types/config.go
@@ -16,7 +16,7 @@ package types
 
 // Config defines the runtime configuration of a workflow.
 type Config struct {
-	Stages   []*Stage   `json:"pipeline"` // workflow stages4649
+	Stages   []*Stage   `json:"pipeline"` // workflow stages
 	Networks []*Network `json:"network"`  // network definitions
 	Volumes  []*Volume  `json:"volume"`   // volume definitions
 	Secrets  []*Secret  `json:"secrets"`  // secret definitions
--- a/pipeline/backend/types/config.go
+++ b/pipeline/backend/types/config.go
@@ -16,10 +16,10 @@ package types
 
 // Config defines the runtime configuration of a workflow.
 type Config struct {
-	Stages  []*Stage  `json:"pipeline"` // workflow stages
-	Network *Network  `json:"network"`  // network definitions
-	Volume  *Volume   `json:"volume"`   // volume definition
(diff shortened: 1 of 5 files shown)


pat-s (approved):

xoxys:
I also dont see why we need a list here.

6543:
because you should be able to bind other resources into woodpecker beside the current workspace: e.g. GPUs ... or other different volumes that you manually use for caching etc ... the pull that makes volums and networks form list to single option is breaking and limits woodpeckers ability ... In my opinion it should stay a list at least in the backend to not technically limit ourselve

pipeline/backend/types/config.go
```diff
@@ -16,10 +16,10 @@ package types
 
 // Config defines the runtime configuration of a workflow.
 type Config struct {
-	Stages  []*Stage  `json:"pipeline"` // workflow stages
-	Network *Network  `json:"network"`  // network definitions
-	Volume  *Volume   `json:"volume"`   // volume definition
-	Secrets []*Secret `json:"secrets"`  // secret definitions
+	Stages   []*Stage   `json:"pipeline"` // workflow stages4649
```
6543, line 19:
```suggestion
	Stages   []*Stage   `json:"pipeline"` // workflow stages
```

pipeline/backend/types/config.go
```diff
@@ -16,10 +16,10 @@ package types
 
 // Config defines the runtime configuration of a workflow.
 type Config struct {
-	Stages  []*Stage  `json:"pipeline"` // workflow stages
-	Network *Network  `json:"network"`  // network definitions
-	Volume  *Volume   `json:"volume"`   // volume definition
-	Secrets []*Secret `json:"secrets"`  // secret definitions
+	Stages   []*Stage   `json:"pipeline"` // workflow stages
+	Networks []*Network `json:"network"`  // network definitions
+	Volumes  []*Volume  `json:"volume"`   // volume definitions
```
6543, line 21:
```suggestion
	Volumes  []*Volume  `json:"volume"`   // volume definitions
```

6543 (approved):

PR #4654, opened 2025-01-01
Support non git vcs

6543:
reverts #4346


6543:
I do wana have this pull open ... so CI can run against it ...

PR #4674, opened 2025-01-06
Add hosts for detached steps

### LINKED ISSUE #4627: A detached container cannot be access (at least with Kubernetes backend)
### Component

agent

### Describe the bug

https://github.com/woodpecker-ci/woodpecker/pull/3411

### Steps to reproduce

1. Install Woodpecker and configure Kubernetes backend;
2. Run a detached step and access it in following steps with its name;
3. See "bad DNS name" or similar reports.

### Expected behavior

As documented, a `detached` step should behave like a service. If it cannot be accessed by DNS, it is not capable of replacing `service`.

### System Info

```shell
{
    "version": "2.8.0"
}
```

### Additional context

https://github.com/woodpecker-ci/woodpecker/pull/3411 should be favorable...

### Validations

qwerty287:
closes #4627 

@stevapple Once the CI finished, can you test this with the PR tag?

### DIFF
--- a/pipeline/backend/kubernetes/kubernetes.go
+++ b/pipeline/backend/kubernetes/kubernetes.go
@@ -199,7 +199,7 @@ func (e *kube) SetupWorkflow(ctx context.Context, conf *types.Config, taskUUID s
 	var extraHosts []types.HostAlias
 	for _, stage := range conf.Stages {
 		for _, step := range stage.Steps {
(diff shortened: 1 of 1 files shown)


anbraten (approved):

stevapple:
@qwerty287 While I cannot test this in production, it looks fairly correct to me👍

PR #4675, opened 2025-01-06
Move trusted config to agent

qwerty287:
#3758


lafriks:
Imho this is not a way to go, if you have multiple workers I can't see how you would keep these options in sync. If they are not if the correct worker gets a job it will run if not it will fail

6543:
if we wana do this, the linter can not make sure the user did missconfig something and we have to announce and filter agents based on the security capabilety

before:
linter did warn you if there is a security related function used but not enabled/configured

now:
pipelines just fail and on configuration drift efen wors: fail randomly

in genneral the idea to move that config to the agent is a good one as that agent should have the last say

**but** i would make it so that the server represents the default config to the agent and the agent can overwrite it.

this way we A. have no problem with lafriks usecase and B. the agent still has the last say.

also we should look into how we can make the linter aware of agents that have X capabilety e.g. via lable announced or what not and config needs Y capabilety, and produce a warnning if there is no agent registered witch has souch ... so to inform the user (autoscaler might add agents so we cant error out)

qwerty287:
I won't continue this approach for now. Feel free to take this as base, but it's quite complex and time-intensive to always fix the merge conflicts and there doesn't really seem to be interest in this… I still think however this is a very good enhancement esp. in combination with user/org agents. Because you can then use your own agents to use trusted features.

PR #4676, opened 2025-01-06
Remove deprecated API field

qwerty287:
I don't think we need that error.


pat-s (approved):

xoxys:
How do API user then know about this?

xoxys:
Oh ok I thought it will be removed after v3.

pat-s:
With v3 and this change is targeting v3, so it should not exist anymore there.

pat-s (approved):

PR #4687, opened 2025-01-08
Don't fail user creation if org exists

### LINKED ISSUE #4651: Adding users with the same name as an already existing organization fails
### Component

server

### Describe the bug

When adding a user, an organization with their name is stored. If an organization with the same name already exists (e.g. when one of the repos of the user-to-be has already been added to woodpecker) this operation fails. 

The [problematic code](https://github.com/woodpecker-ci/woodpecker/blob/5b0d7887373426628f61214ef59043caa48f8d9a/server/store/datastore/user.go#L62-L66) does not check if the organization already exists, it attempts to store the organization, which results in a unique constraint violation on the `orgs` table, which the user gets to see.

### Steps to reproduce

1. Install and configure fogejo, with an admin user and an additional user
2. Create a repo in the main user's account and in the additional user's account
3. Install woodpecker and configure the integration with forgejo
4. Enable the secondary user's repo in woodpeckers ui
5. Attempt now to add the additional user as a woodpecker user

### Expected behavior

The user can be added to woodpecker without problems

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.8.2"}


```yaml
services:
  woodpecker-server:
    image: woodpeckerci/woodpecker-server:v2.8.2
    ports:
      - 8000:8000
    volumes:
      - /woodpecker/data:/var/lib/woodpecker:rw
    environment:
      - WOODPECKER_AUTHENTICATE_PUBLIC_REPOS=true
      - WOODPECKER_OPEN=false
      - WOODPECKER_ADMIN=admin
      - WOODPECKER_FORGEJO_CLIENT=__________________________
      - WOODPECKER_FORGEJO=true
      - WOODPECKER_PLUGINS_PRIVILEGED=woodpeckerci/plugin-docker-buildx:5.0.0
      - WOODPECKER_HOST=https://woodpecker
      - WOODPECKER_FORGEJO_URL=https://forgejo
      - WOODPECKER_FORGEJO_SECRET=__________________________

  woodpecker-agent:
    image: woodpeckerci/woodpecker-agent:v2.8.2
    command: agent
    restart: always
    depends_on:
      - woodpecker-server
    volumes:
    environment:
      - WOODPECKER_SERVER=woodpecker-server:9000
      - WOODPECKER_AGENT_SECRET=___________
      - WOODPECKER_AUTHENTICATE_PUBLIC_REPOS=true

```
```


### Additional context

_No response_

### Validations

### PR DESCRIPTION
**Pull request by pat-s**
fix #4651

Also modified `OrgFindByName` to return `nil` if nothing is found instead of erroring with a SQL error.

### DIFF
--- a/server/store/datastore/org.go
+++ b/server/store/datastore/org.go
@@ -16,7 +16,6 @@ package datastore
 
 import (
 	"fmt"
-	"log"
 	"strings"
 
 	"xorm.io/xorm"
@@ -78,14 +77,12 @@ func (s storage) orgDelete(sess *xorm.Session, id int64) error {
 func (s storage) OrgFindByName(name string) (*model.Org, error) {
 	// sanitize
 	name = strings.ToLower(name)
-	// find
 	org := new(model.Org)
 	has, err := s.engine.Where("name = ?", name).Get(org)
 	if err != nil {
 		return nil, fmt.Errorf("failed to check if org exists: %w", err)
 	}
 	if !has {
-		log.Printf("Organization with name %s not found", name)
 		return nil, nil
 	}
 	return org, nil
--- a/server/store/datastore/org.go
+++ b/server/store/datastore/org.go
@@ -16,6 +16,7 @@ package datastore
 
 import (
 	"fmt"
+	"log"
 	"strings"
 
 	"xorm.io/xorm"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: server/store/datastore/org.go [hunk 1]
```diff
@@ -79,7 +80,15 @@ func (s storage) OrgFindByName(name string) (*model.Org, error) {
 	name = strings.ToLower(name)
 	// find
 	org := new(model.Org)
-	return org, wrapGet(s.engine.Where("name = ?", name).Get(org))
+	has, err := s.engine.Where("name = ?", name).Get(org)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if org exists: %w", err)
+	}
+	if !has {
+		log.Printf("Organization with name %s not found", name)
```
lafriks, line 88: Why would we need to log this?

lafriks (approved):

pat-s:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|release/v2.8|[](https://github.com/woodpecker-ci/woodpecker/pull/4692)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

PR #4689, opened 2025-01-09
Fix text overflow on login mask

pat-s:
regression of #4614

Before

image

After

image

### DIFF
--- a/web/src/components/atomic/Button.vue
+++ b/web/src/components/atomic/Button.vue
@@ -2,7 +2,7 @@
   <component
     :is="to === undefined ? 'button' : httpLink ? 'a' : 'router-link'"
     v-bind="btnAttrs"
-    class="relative flex flex-shrink-0 cursor-pointer items-center overflow-hidden whitespace-nowrap rounded-md border px-2 py-1 shadow-sm transition-all duration-150 disabled:cursor-not-allowed disabled:opacity-50 text-wrap"
+    class="relative flex flex-shrink-0 cursor-pointer items-center overflow-hidden whitespace-nowrap text-wrap rounded-md border px-2 py-1 shadow-sm transition-all duration-150 disabled:cursor-not-allowed disabled:opacity-50"
     :class="{
       'border-wp-control-neutral-300 bg-wp-control-neutral-100 text-wp-text-100 hover:bg-wp-control-neutral-200':
         color === 'gray',
--- a/web/src/components/atomic/Button.vue
+++ b/web/src/components/atomic/Button.vue
@@ -2,7 +2,7 @@
   <component
     :is="to === undefined ? 'button' : httpLink ? 'a' : 'router-link'"
     v-bind="btnAttrs"
-    class="relative flex flex-shrink-0 cursor-pointer items-center overflow-hidden whitespace-nowrap text-wrap rounded-md border px-2 py-1 shadow-sm transition-all duration-150 disabled:cursor-not-allowed disabled:opacity-50"
+    class="relative flex flex-shrink-0 cursor-pointer items-center overflow-hidden whitespace-nowrap rounded-md border px-2 py-1 shadow-sm transition-all duration-150 disabled:cursor-not-allowed disabled:opacity-50"
     :class="{
       'border-wp-control-neutral-300 bg-wp-control-neutral-100 text-wp-text-100 hover:bg-wp-control-neutral-200':
         color === 'gray',
--- a/web/src/components/atomic/Button.vue
+++ b/web/src/components/atomic/Button.vue
@@ -2,7 +2,7 @@
   <component
     :is="to === undefined ? 'button' : httpLink ? 'a' : 'router-link'"
     v-bind="btnAttrs"
-    class="relative flex flex-shrink-0 cursor-pointer items-center overflow-hidden whitespace-nowrap rounded-md border px-2 py-1 shadow-sm transition-all duration-150 disabled:cursor-not-allowed disabled:opacity-50"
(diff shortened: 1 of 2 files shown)


web/src/components/atomic/Button.vue
```diff
@@ -2,7 +2,7 @@
   <component
     :is="to === undefined ? 'button' : httpLink ? 'a' : 'router-link'"
     v-bind="btnAttrs"
-    class="relative flex flex-shrink-0 cursor-pointer items-center overflow-hidden whitespace-nowrap rounded-md border px-2 py-1 shadow-sm transition-all duration-150 disabled:cursor-not-allowed disabled:opacity-50"
+    class="relative flex flex-shrink-0 cursor-pointer items-center overflow-hidden whitespace-nowrap text-wrap rounded-md border px-2 py-1 shadow-sm transition-all duration-150 disabled:cursor-not-allowed disabled:opacity-50"
```
xoxys, line 5: `whitespace-nowrap text-wrap` should not be used together. I would also not break text in all buttons by default. Can we just allow text wrap on the login screen buttons?

xoxys: I would suggest to use `class="!whitespace-normal"` on the `Button` element in web/src/views/Login.vue instead.

web/src/views/Login.vue
```diff
@@ -19,7 +19,7 @@
       <div class="flex min-h-48 items-center justify-center bg-wp-primary-200 dark:bg-wp-primary-300 md:w-3/5">
         <WoodpeckerLogo preserveAspectRatio="xMinYMin slice" class="h-32 w-32 md:h-48 md:w-48" />
       </div>
-      <div class="flex min-h-48 flex-col items-center justify-center gap-4 text-center md:w-2/5">
+      <div class="flex min-h-48 flex-col items-center justify-center gap-4 px-2 text-center md:w-2/5">
```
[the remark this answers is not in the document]
pat-s: Why expand the vertical padding?

web/src/components/atomic/Button.vue
pat-s: Thanks, a good solution!

web/src/views/Login.vue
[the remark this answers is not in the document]
pat-s: Tried `p-2` and I don't think this should be changed.

lafriks (approved):

xoxys (changes requested):
Approving the current state doesnt make sense as the issue is not fixed.

pat-s:
@xoxys What is missing?

xoxys:
Ah you have now added everyrhing to the div? Why? I would also still suggest to use p-4 instead of px-4.

pat-s:
I am against `p-4` as said, it creates too much space between the "login" word and the button. Also the vertical space wasn't debated before, why change it now? The main goal/fix was to change the text overflow.

---

I didn't build an image to test this diff in detail, only applied your classes in devtools but I think it should be fine?

xoxys:
The padding is added to the outer div. While you have fixed the horizontal spacing when the button reaches the width of the container why not just fix the same potential issue with the vertical space if the content reaches the height?

xoxys:
Im still confused. Why the -mb-4 stuff now? Why not keep the p-4 on the outer div? There are now also two whitespace-wrap classes, why?

xoxys:
No I would just keep the padding on the outer div. This will ensure items just dont hit the container without adding extra space between container elements.

Yes on the button the whitespace-normal needs the important annotation.

web/src/views/Login.vue
```diff
@@ -19,13 +19,14 @@
       <div class="flex min-h-48 items-center justify-center bg-wp-primary-200 dark:bg-wp-primary-300 md:w-3/5">
         <WoodpeckerLogo preserveAspectRatio="xMinYMin slice" class="h-32 w-32 md:h-48 md:w-48" />
       </div>
-      <div class="flex min-h-48 flex-col items-center justify-center gap-4 text-center md:w-2/5">
-        <h1 class="text-xl text-wp-text-100">{{ $t('welcome') }}</h1>
+      <div class="flex min-h-48 flex-col items-center justify-center gap-4 p-4 text-center md:w-2/5">
+        <h1 class="-mb-4 text-xl text-wp-text-100">{{ $t('welcome') }}</h1>
```
xoxys, line 23: Now we just need to remove this `-mb-4` again 😁

pat-s: Still the think the margin is too big now but don't wanna argue anymore 😄️

xoxys: But why? We have not added any extra space here or what do I miss? The new padding is places on the outer div. The title and the buttons are both direct childs of the  div with the new padding. As such the parents padding does not add space between the childs. Or what do I miss?

pat-s: Sure, there was no vertical padding around the div before. Also, if I add this interactively, I see the space between the header and the button increasing.

xoxys: Alright, Ill give it a test again if Im back on a PC. If this change increases the space between the childs something else is wrong.

xoxys (approved):
I will check the discussed spacing issue later. Feel free to merge the current state.

xoxys:
I can't reproduce your described behavior:

With p-4:
image

Without p-4:
image

The spacing between the header and the button are the same in both cases. As expected, the padding is applied to the outer container:

image

pat-s:
Hm, alright. Maybe the issue was caused by placing `p-4` on the div surrounding the button?

In some scenario this definitely caused an issue. But great it doesn't now and even more great you checked in that much detail!

PR #4705, opened 2025-01-11
Add docs for 3.0

pat-s:


anbraten:
Would be nice to update `docs/src/pages/versions.md` here as well.

https://github.com/woodpecker-ci/woodpecker/pull/4297/files might be helpful with some additional hints.

pat-s:
> Would be nice to update docs/src/pages/versions.md here as well.

Going forward there really must be a slicker way than having to edit versions manually in like three or more places for every release.

pat-s:
Why are routes now missing? Can't spot what's the issue.

xoxys:
> Afaik you also have to update the docusuarus config. And please delete the oldest docs then (we keep only 3 versions)

@qwerty287 This PR deletes docs for v2.6 now, but I'm a bit confused. If we keep only the last 3 versions, does this mean the last 3 in total? If so we would also have to delete v1 now and only keep v2.7, v2.8 and v3.0 is this correct?

xoxys:
Mhm I disagree a bit. Keeping at least one older major might make sense but we dont even support v1 anymore or backport fixes to v1. Keeping the docs alive feels wrong somehow.

xoxys:
Ok as nobody else participates here, and we should finish the release, keep it or drop it? I vote for dropping it.

xoxys:
Ok, then let's keep it for now and proceed.

anbraten:
Would drop it some days after we've successfully released 3.0

anbraten:
I removed the 3.0 folder and created it again after merging main. Is next docs not up to date?

anbraten:
You are right. Now it should be fine,

xoxys:
Can you fix the spellcheck?

anbraten (approved):

xoxys (approved):

PR #4713, opened 2025-01-12
Fix admin badge alignment

qwerty287:
before:
grafik

(now it's directly next to the icon buttons)


xoxys:
Maybe its better this way according to https://github.com/woodpecker-ci/woodpecker/discussions/2162#discussioncomment-10497833?

xoxys (approved):

PR #4715, opened 2025-01-12
Fix broken z-index

xoxys:
Regression-of: https://github.com/woodpecker-ci/woodpecker/pull/4614

Tailwind has only a few pre-defined z-index levels.


xoxys:
Tested locally

image

PR #4719, opened 2025-01-12
Add docs-dependencies target to makefile

xoxys:
Analog to `ui-dependencies`.


xoxys:
You are right. I restructured it a bit more.

PR #4724, opened 2025-01-13
feat(bitbucketdatacenter.go): Add unique key to build , and add ref f…

Levy-Tal:
**Problem**
Previously, all build statuses sent to Bitbucket Datacenter shared the same key. This caused newer build statuses to overwrite earlier ones, resulting in only one build status being visible at any time.
Screenshot From 2025-01-13 12-45-28

**Solution**
Unique Keys: Updated the key used in the build status to include the pipeline number. This ensures each build has a unique key, preventing overwrites.
Git Reference Association: Added the Git ref field to the Bitbucket request. This explicitly associates each build with the correct branch or tag in Bitbucket, improving traceability.
Screenshot From 2025-01-13 13-39-10

Impact
Multiple build statuses for pipelines will now be displayed in Bitbucket, allowing users to see all relevant build information.
Enhanced association between builds and branches for better clarity and context in the development workflow.

### DIFF
--- a/server/forge/bitbucketdatacenter/bitbucketdatacenter.go
+++ b/server/forge/bitbucketdatacenter/bitbucketdatacenter.go
@@ -313,8 +313,9 @@ func (c *client) Status(ctx context.Context, u *model.User, repo *model.Repo, pi
 	status := &bb.BuildStatus{
 		State:       convertStatus(pipeline.Status),
 		URL:         common.GetPipelineStatusURL(repo, pipeline, workflow),
-		Key:         common.GetPipelineStatusContext(repo, pipeline, workflow),
+		Key:         fmt.Sprintf("%s-%d", common.GetPipelineStatusContext(repo, pipeline, workflow), pipeline.Number),
 		Description: common.GetPipelineStatusDescription(pipeline.Status),
+		Ref:         pipeline.Ref,
 	}
 	_, err = bc.Projects.CreateBuildStatus(ctx, repo.Owner, repo.Name, pipeline.Commit, status)
 	return err


langecode:
yeah, I'll take a look :) I am a little bit swamped the start of this week
so it might not be until Wednesday

Den man. 13. jan. 2025 kl. 15.22 skrev qwerty287 ***@***.***>:

Levy-Tal:
Hey @langecode , just a quick reminder to review my PR when you have a moment. Thanks!

langecode:
Makes perfect sense to add the `ref` property to make sure Bitbucket displays multiple active heads correctly. I think the status is actually already attached to a specific commit already as the REST API actually require the commit to be part of the URL path (that is why `pipeline.Commit` is part of the call in line 320). So if you look at the commit log in Bitbucket the build status is there on the specific commit - but it would seem that the view of the builds is not correct.

Makes perfect sense to add the `ref` - the documentation for the about says:
> *ref* (optional): The fully qualified git reference e.g. `refs/heads/master`

I am wondering whether the correct value is `pipeline.Ref` or `pipeline.Refspec` - I cannot see from the `Pipeline` struct what is in one or the other.

While changes are being made to the build status it could be considered whether to also add the build number to the status in the `buildNumber` property?
> *buildNumber* (optional): A unique identifier for this particular run of a plan

We actually changed the context template through the `WOODPECKER_STATUS_CONTEXT_FORMAT` environment variable in our local deployment of Woodpecker - I currently cannot remember what thet was with the default, I'll try to dig it out of my mind 🤔

Levy-Tal:
> **I think the status is actually already attached to a specific commit already.**

You’re correct — the build status is attached to the specific commit. Adding the Git ref will allow Bitbucket to associate multiple builds with the same branch and display them correctly in the UI.

---

> **I am wondering whether the correct value is `pipeline.Ref` or `pipeline.Refspec`.**

After debugging, I’ve confirmed that `pipeline.Ref` contains the fully qualified Git ref (e.g., `refs/heads/master`), whereas `pipeline.Refspec` is empty.  
Screenshot From 2025-01-20 23-14-16

---

> **While changes are being made to the build status, it could be considered whether to also add the build number to the status in the `buildNumber` property.**

From what I’ve observed, Bitbucket does not use `buildNumber` for any internal operations. It’s simply stored for users to retrieve build status information via the API. Including it would be optional — a "nice to have" — but it’s not strictly necessary. I’m happy to add it if you feel it would be beneficial.

---

**Regarding the context template and whether the `buildNumber` needs to be in the status key or not:**  

From the [Bitbucket documentation](https://developer.atlassian.com/server/bitbucket/how-tos/updating-build-status-for-commits/#updating-build-status-for-commits):  

> Bitbucket Data Center stores one build result for each key per commit, so you can update a previous build result by sending a request with:  
> - the same commit hash; and  
> - the same key attribute.  
> The new result will replace the previously submitted build result for that commit hash and key.  

This raises the question: Should rerunning a build on the same commit (e.g., restarting a build via the UI) overwrite the existing build status, or should it create a new build status (with a unique key, potentially including the build number)?  

Thank you for responding and for maintaining the Bitbucket Data Center codebase.

langecode:
Yeah, looks good to me then 👍

PR #4744, opened 2025-01-19
Add missing error message for org_access_denied

xoxys:


pat-s:
Besides the better msg "You are not allowed to access this organization", users facing this error often still don't know where it comes from, especially if they are new.

I'd argue that additional information related to the `WOODPECKER_ORGS` setting would help (i.e. that this one is blocking access as they are not within that org).
This setting is not obvious for new people and often overlooked, leading to initial setup frustration.

xoxys:
I'm not sure if merged PRs are the best place to discuss further improvements. Would you mind creating an issue instead?

pat-s:
I guess mentioning it after one was too slow to add it before a PR was merged should be OK? It would have likely become a review with "changes requested" 🙃️

Don't think this needs an issue, just wanted to bring it up as a potential point to think about which could improve user experience.

xoxys:
My main point against it is information disclosure. Especially for auth processes its not that easy to give detailed errors without providing helpful information to bad actors.

pat-s:
Mentioning the env var is all I had in mind. 

Anyhow, I think we at least partially misunderstand each other here. All good, let's move on.

Issue #4537, opened 2024-12-08
Race condition while rendering pipeline commit


xoxys:
### Component

web-ui

### Describe the bug

https://github.com/user-attachments/assets/ae56c58f-717b-4915-aad3-effdd9640fe2

Sometimes `pipeline.` is rendered instead of the commit hash. I guess that's a race condition during data loading from pinia. What I'm wondering about it, this only happens while a user is logged in. If the pipeline list is loaded as unauthenticated, everything seems to be working as intended.

### Steps to reproduce

Reload the pipeline list multiple times.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-d0e4a176f1"}
```

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

zc-devs:
Authenticated
![Screenshot 2024-12-09 200315](https://github.com/user-attachments/assets/07c0cfc9-6129-4948-9916-d5854e0b5f2b)

Unauthenticated
![Screenshot 2024-12-09 200853](https://github.com/user-attachments/assets/0ea8c1f5-23ca-446f-9e68-66a27acfe473)


1. Requests for pipelines per repo I have. Wasn't a case on `2.8`.
2. Request for current active repo.

xoxys:
Seems to be related to https://github.com/woodpecker-ci/woodpecker/pull/3806/files#r1921574171

xoxys:
https://github.com/woodpecker-ci/woodpecker/pull/3806/files#r1921575001

> AFAIK SQL this was a syntax error, but not sure.
> 
> Please double-check that with all DBs. I'm having a similar issue in my message rework PR, and there it is always working with some DBs, but not with the others.

xoxys:
I can reproduce the SQL error in sqlite

![Image](https://github.com/user-attachments/assets/208f4fea-b7c6-4a0f-889f-b67f0093eb32)

Will try to find a workaround while keeping the escaped syntax.

PR #4746, opened 2025-01-19
Fix sql identifier escaping in datastore feed

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/4537

Use the correct syntax to escape column identifiers. `"pipelines"."commit"` references the actual commit column in the pipelines table. This syntax should be part of the ANSI/ISO SQL standard and works across different database systems. It's the standardized way to escape/quote identifiers in SQL when needed. Manually tested with SQLite and Postgres.


xoxys:
Yeah.... we need to implement it driver-specific. What a scrap.

anbraten:
Could we you add some test.

xoxys:
There are tests and thats why the pipine fails for mysql or what do you mean?

xoxys:
I've added an explicit test case now to also verify the pipeline information from a feed item. This fails as intended with the current code base from `main`:

```
=== RUN   TestGetPipelineQueue
    ~/Devel/golang/woodpecker/woodpecker/server/store/datastore/feed_test.go:75:
        	Error Trace:	~/Devel/golang/woodpecker/woodpecker/server/store/datastore/feed_test.go:75
        	Error:      	Not equal:
        	            	expected: "abc123"
        	            	actual  : "pipelines.commit"

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-abc123
        	            	+pipelines.commit
        	Test:       	TestGetPipelineQueue
--- FAIL: TestGetPipelineQueue (0.00s)
FAIL
FAIL	go.woodpecker-ci.org/woodpecker/v3/server/store/datastore	0.009s
```

anbraten (approved):

PR #4762, opened 2025-01-23
Unsanitize user and org names in DB

### LINKED ISSUE #3614: CamelCase usernames get added as lower-case into DB and result in access issues
### Component

server

### Describe the bug

When a username is in camel-case, e.g. `MyUser`, the user get's added as `myuser` into the DB (e.g. the `orgs` table).

This causes access issues during use. Users only see the generic "an unknown error occured" when accessing their own user org and other access-restricted parts within WP.

We have hundreds of these cases in the Codeberg instance (https://codeberg.org/Codeberg-CI/feedback/issues/149). Additionally, I also verified this behavior with a new user on my own private instance.

Not sure if it is forge-related, but it at least affects Gitea/Forgejo.

### System Info

```shell
WP 2.4.1 but at least 2.3.x and likely also previous versions are affected
```

### Additional context

_No response_

### Validations

pat-s:
fix #3614

As discussed in chat, preferred to use non-sanitized values everywhere.

- Store user and org names using the casing provided by the forge
- Allow to search for users / orgs using case insensitive names
- Removed all sanitation in the code
- AFAICS only the org names were stored sanitized in the DB. User names and repo names are not affected


server/store/datastore/migration/024_unsanitize_org_and_user_names.go
```diff
@@ -0,0 +1,61 @@
+// Excerpt from:
+// Copyright 2024 Woodpecker Authors
```
xoxys, line 2:
```suggestion
// Copyright 2025 Woodpecker Authors
```

server/store/datastore/migration/024_unsanitize_org_and_user_names.go
```diff
@@ -0,0 +1,61 @@
...
+		type org struct {
+			Name string `xorm:"TEXT 'name'"`
+		}
+
+		if err := sess.Sync(new(user), new(org)); err != nil {
+			return fmt.Errorf("sync new models failed: %w", err)
+		}
+
+		// get all users
+		var us []*user
```
xoxys, line 41:
```suggestion
		var users []*user
```

server/store/datastore/migration/024_unsanitize_org_and_user_names.go
```diff
@@ -0,0 +1,61 @@
...
+		if err := sess.Find(&us); err != nil {
+			return fmt.Errorf("find all repos failed: %w", err)
+		}
+
+		for _, user := range us {
```
xoxys, line 46:
```suggestion
		for _, user := range users {
```

server/store/datastore/migration/024_unsanitize_org_and_user_names.go
```diff
@@ -0,0 +1,61 @@
...
+			Name string `xorm:"TEXT 'name'"`
+		}
+
+		if err := sess.Sync(new(user), new(org)); err != nil {
+			return fmt.Errorf("sync new models failed: %w", err)
+		}
+
+		// get all users
+		var us []*user
+		if err := sess.Find(&us); err != nil {
```
xoxys, line 42:
```suggestion
		if err := sess.Find(&users); err != nil {
```

server/store/datastore/org.go
```diff
@@ -28,13 +28,24 @@ func (s storage) OrgCreate(org *model.Org) error {
...
+	orgs, err := s.OrgList(nil)
+	if err != nil {
+		return err
+	}
+
+	for _, existingOrg := range orgs {
+		if strings.EqualFold(existingOrg.Name, org.Name) {
+			return fmt.Errorf("organization name already exists")
+		}
+	}
```
xoxys, line 41: We store orgs now case-sensitive in the DB, why do we check here for case-insensitive duplicates?

server/store/datastore/org.go
```diff
@@ -84,7 +80,6 @@ func (s storage) OrgFindByName(name string) (*model.Org, error) {
 
 func (s storage) orgFindByName(sess *xorm.Session, name string) (*model.Org, error) {
 	// sanitize
-	name = strings.ToLower(name)
 	org := new(model.Org)
 	return org, wrapGet(sess.Where("name = ?", name).Get(org))
```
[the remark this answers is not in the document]
xoxys: This can in theory return multiple results now, how is this handled then?

xoxys:
After thinking of it again Im not sure if thats the best approach. Cant really say why but I see a lot of potential for issues and corner cases. Looks like this issue only occures with forgejo/gitea and the api returns mixed capitalization while the forge only supports case-insensitive values. What do you think?

pat-s:
If all users are sanitized, the user list in the UI would also reflect a non-optimal state.

The simplest fix which also wouldn't require a migration would probably be to just sanitize the comparison call of `user.Login` and `org.Name`?

anbraten:
Would store the forges casing in the database and allow api / url routes (the db find functions like `orgFindByName`) to return values case insensitive (similar to how the forges handle it).

xoxys:
OK with it, but in that case Im still wondering how we want to handle it if multiple case-insensitive items are returned from the DB?

xoxys:
If they would or not doesn't matter IMO. In worst case we end up with "Foo" and "foo" in the DB (for whatever reason) as we store it case-sensitive but expect it to be case-insensitive. If we can add constraint to the column to prevent this, I'm fine with it, but otherwise this could lead to issues later.

pat-s:
Can somebody please summarize the resolution here now, so I can make the required adjustments?

xoxys:
> Would store the forges casing in the database and allow api / url routes (the db find functions like `orgFindByName`) to return values case insensitive (similar to how the forges handle it).

Thats the way to go. I would like to ensure we can not store the same forge/org combination twice with differene casing, @anbraten wanted to check the code, see the last comment.

pat-s:
So this PR is waiting for a review from @anbraten now, is this correct? AFAICS the PR is currently

- storing the raw casing for org/usernames
- sanitizing API calls

Please let me know if this needs more time overall, then I'll build a standalone image and apply this as a hotfix, so the issue on CB is solved (as there are already a few pending users waiting for a fix before they can be onboarded).

xoxys:
For someone to verify that we have proper unique constraints in place.

anbraten:
Checked the code and can't see a place that it would add it twice. There is some issue with multiple forges mixing orgs I've detected, but that's something else.

xoxys:
Ok I dont have time to look into this on my own. Feel free to go on if you think the current state is fine.

xoxys:
In best case we add a test case. Try to write "org1/foo" and "org1/Foo" to the db. If it fails, we are good to go if not we should still add db level restrictions.

pat-s:
Added two tests which explicitly check that attempting to add a case-sensitive duplicate for orgs and users results in an error.

xoxys:
Thanks!

xoxys (dismissed):

server/store/datastore/migration/common.go
```diff
@@ -170,7 +170,7 @@ func alterColumnNull(sess *xorm.Session, table, column string, null bool) error
 	dialect := sess.Engine().Dialect().URI().DBType
 	switch dialect {
 	case schemas.MYSQL:
-		sql := fmt.Sprintf("SHOW COLUMNS FROM `%s` WHERE lower(field) = '%s'", table, strings.ToLower(column))
+		sql := fmt.Sprintf("SHOW COLUMNS FROM `%s` WHERE lower(field) = '%s'", table, column)
```
anbraten, line 173: Would skip this in this PR to avoid accidentally side effects and put it into a separate one.

server/store/datastore/migration/024_unsanitize_org_and_user_names.go
```diff
@@ -0,0 +1,60 @@
...
+		type org struct {
+			Name string `xorm:"TEXT 'name'"`
+		}
+
+		if err := sess.Sync(new(user), new(org)); err != nil {
+			return fmt.Errorf("sync new models failed: %w", err)
+		}
+
+		// get all users
+		var users []*user
```
anbraten, line 40: Should use pagination to allow smooth migrations on larger instances

server/store/datastore/migration/024_unsanitize_org_and_user_names.go
```diff
@@ -0,0 +1,60 @@
...
+			return fmt.Errorf("find all repos failed: %w", err)
+		}
+
+		for _, user := range users {
+			userOrg := &org{}
+			_, err := sess.Where("name = ?", user.Login).Get(userOrg)
+			if err != nil {
+				return fmt.Errorf("getting org failed: %w", err)
+			}
+			if user.Login != userOrg.Name {
```
anbraten, line 51: Here I would add a check logging an error if the forge ids don't match and skip in that case.

pat-s: Why skip? This is the actual important part of this migration, adjusting the orgName if it doesn't match `user.Login`. Why log an error and skip?

server/store/datastore/migration/024_unsanitize_org_and_user_names.go
pat-s: Even large instances should have only xxxx users (e.g. Codeberg). Getting this amount of users without pagination in one call should be acceptable? I would worry more about pagination for getting all pipeline logs or similar but I think we don't need pagination here. 

I only know the shared_utils.Paginate() but I don't wanna make the code more complex than it really needs to be.

pat-s: That said, feel free to add pagination if you think it is really needed and/or you know directly what the best way would be.

server/store/datastore/migration/024_unsanitize_org_and_user_names.go
anbraten: You can habe multiple forges and the same org Name in both therefore I would only update them if the user and org both belong to the same forge.

server/api/login.go
```diff
@@ -183,29 +183,34 @@ func HandleAuth(c *gin.Context) {
 	// create or set the user's organization if it isn't linked yet
 	if user.OrgID == 0 {
 		// check if an org with the same name exists already and assign it to the user if it does
-		if org, err := _store.OrgFindByName(user.Login); err == nil && org != nil {
+		// TODO: find the org by name and forgeID directly
+		org, err := _store.OrgFindByName(user.Login)
```
anbraten, line 187: I've noticed this bug as the error was not handled by the `if` below as it was inline in the previous if.

anbraten:
As the current migration will only fix user orgs, I am thinking of a global repair api endpoint that could do some additional tasks like:
- updating user names from forge
- org names from forge
- checking each user has an user org
- updating user permissions
- doing repo repairs

server/api/login.go
xoxys: Good catch

xoxys:
Code LGTM, have you tested the PR locally?

xoxys:
@anbraten Something is wrong. I have applied this PR to an existing WP instance. While I was able to log-in with the user before this PR, this now fails:

> woodpecker-server   | {"level":"error","time":"2025-02-04T07:54:23Z","caller":"/build/server/api/login.go:228","message":"user org is not from the same forge  woodpecker"}

server/api/login.go
```diff
@@ -217,10 +222,18 @@ func HandleAuth(c *gin.Context) {
 			c.Redirect(http.StatusSeeOther, server.Config.Server.RootPath+"/login?error=internal_error")
 			return
 		}
+
+		// this should never happen, but if it does we should make admins aware of it
+		if org != nil && org.ForgeID != forgeID {
+			log.Error().Err(err).Msgf("user org is not from the same forge  %s", user.Login)
```
xoxys, line 228:
```suggestion
			log.Error().Err(err).Msgf("user org is not from the same forge: user:  %s", user.Login)
```

pat-s:
> but the table is called migration. This looks like migrations at all are no longer working?

I've seen this error before a few times.
Not sure how "blocking" it is as all past migrations were successfully applied in all my instances.
But it looks like this is a typo, yes. The table is called "migration".

> While I was able to log-in with the user before this PR, this now fails:

This is worrisome. Can you check which forge ID got assigned to the user org? Actually, tests should catch this.

server/store/datastore/org.go
```diff
@@ -84,9 +81,8 @@ func (s storage) OrgFindByName(name string) (*model.Org, error) {
 
 func (s storage) orgFindByName(sess *xorm.Session, name string) (*model.Org, error) {
 	// sanitize
-	name = strings.ToLower(name)
 	org := new(model.Org)
-	return org, wrapGet(sess.Where("name = ?", name).Get(org))
+	return org, wrapGet(sess.Where("LOWER(name) = ?", strings.ToLower(name)).Get(org))
```
pat-s, line 85: This might be a mistake? I thought we don't want to sanitize calls/entries in the DB but only to the forge API.

Also removing sanitation in l87 and then re-adding it seems strange.

xoxys:
> This is worrisome. Can you check which forge ID got assigned to the user org? Actually, tests should catch this.

```
woodpecker_server=# select * from orgs;
 id | forge_id |    name    | is_user | private 
----+----------+------------+---------+---------
  1 |        0 | woodpecker | t       | f
```

`forge_id` is 0 while only one forge exists with id `1`:

```
woodpecker_server=# select * from forges;
 id | type  |             url              |                client                |                      client_secret                       | skip_verify | oauth_host | additional_options 
----+-------+------------------------------+--------------------------------------+----------------------------------------------------------+-------------+------------+--------------------
  1 | gitea | http://woodpecker-gitea:3000 | **** | **** | f           |            | {}
(1 row)
```

anbraten:
Strange, I started to add a method to the storage interface to directly get an org by name and forge-id, that might prevent such cases. Forge-id 0 sounds like not set at all.

xoxys:
Yeah maybe we mixed in too much different fixes into this PR? Can we revert it to focus on the sanitizing issue and move other improvements to another PR?

I have a backup of the DB before the migration from this PR so I can also re-test this PR.

server/store/datastore/org.go
anbraten: The idea is to have searching case insensitive and storing case sensitive (the same way how forges do it, you can use both ways in urls etc and it shows the casing in the UI etc the user used).

> Also removing sanitation in l87 and then re-adding it seems strange.

It was previously missing the lower part on both sides. The change makes it unique with other calls.

pat-s:
I don't know where this PR stands now and what there is do to - appreciated if somebody could clarify or apply the necessary actions.

xoxys:
Im more changes need to be done for https://github.com/woodpecker-ci/woodpecker/pull/4817 can we revert those changes here so we can merge it?

server/api/repo.go
```diff
@@ -120,6 +120,7 @@ func PostRepo(c *gin.Context) {
 
 	// find org of repo
 	var org *model.Org
+	// TODO: find org by name and forge id
```
xoxys, line 123:
```suggestion
```

server/store/datastore/org.go
```diff
@@ -28,11 +27,10 @@ func (s storage) OrgCreate(org *model.Org) error {
 }
 
 func (s storage) orgCreate(org *model.Org, sess *xorm.Session) error {
-	// sanitize
-	org.Name = strings.ToLower(org.Name)
 	if org.Name == "" {
 		return fmt.Errorf("org name is empty")
 	}
+
```
xoxys, line 33:
```suggestion
```

xoxys (approved):

pat-s:
@xoxys Will you test this again after merge? Is a patch release planned after merge?

xoxys:
Tested the PR already. Can login again but had no time to test fresh install and migration again.

Zubairrahman11:
- Thanks! Then
   -
   - 0 sounds good.
   - Makes sense. Thanks for clarifying!
   - Agree, a patch release doesn't seem necessary.

Zubairrahman11:
- This email thread is a discussion about a pull request to sanitize
   user and org names in the database.
   - The pull request was approved and merged.
   - A patch release is not necessary.

PR #4763, opened 2025-01-23
fix: remove lower casing from orgs

### LINKED ISSUE #3614: CamelCase usernames get added as lower-case into DB and result in access issues

anbraten:
closes #3614 

After doing some research, those are some things we found out:
- (most) forges are not case-sensitive in urls (or redirect to the configured case) => we should allow searching case insensitive
- there seems to be NO concrete reason we converted org names to lower in #1873

# Changes
- [x] remove sanitizing from org names
- [ ] add migration: fetch org names from forge(s) and update in db once

### DIFF
--- a/server/store/datastore/org.go
+++ b/server/store/datastore/org.go
@@ -28,8 +28,6 @@ func (s storage) OrgCreate(org *model.Org) error {
 }
 
 func (s storage) orgCreate(org *model.Org, sess *xorm.Session) error {
-	// sanitize
-	org.Name = strings.ToLower(org.Name)
 	if org.Name == "" {
 		return fmt.Errorf("org name is empty")
 	}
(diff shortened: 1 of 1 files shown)


pat-s:
Hm, just saw that now. #4762 has also a migration included.

PR #4768, opened 2025-01-23
Add allow list for approvals

### LINKED ISSUE #336: Extend Gated Feature with an Allow/Block List

qwerty287:
closes #336 

Add a list of users who's pipelines never require an approval.

grafik


server/pipeline/gated.go
```diff
@@ -31,6 +35,11 @@ func needsApproval(repo *model.Repo, pipeline *model.Pipeline) bool {
 		return false
 	}
 
+	// skip if user is allowed
+	if slices.Contains(repo.ApprovalAllowedUsers, pipeline.Author) {
```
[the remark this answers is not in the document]
anbraten: We could add a comment like:

```suggestion
        //  It's enough to check the username as the repo matches the forge of the pipeline already (no username clashes from different forges possible)
	if slices.Contains(repo.ApprovalAllowedUsers, pipeline.Author) {
```

anbraten:
Could you add a screenshot. Might also be clever if we start adding some short description of new features & breaking changes to the description, so we could use that for release blog posts and for users checking the PR.

web/src/assets/locales/en.json
```diff
@@ -509,7 +509,11 @@
     "none_desc": "Every event triggers pipelines, including pull requests. This setting can be dangerous and is only recommended for private instances.",
     "forks": "Pull request from forked repository",
     "pull_requests": "All pull requests",
-    "all_events": "All events from forge"
+    "all_events": "All events from forge",
+    "allowed_users": {
+      "allowed_users": "Allowed users",
+      "desc": "Pipelines authored by users listed here never require an approval."
```
xoxys, line 515:
```suggestion
      "desc": "Pipelines created by the listed users never require approval."
```

web/src/views/repo/settings/General.vue
```diff
@@ -18,22 +18,26 @@
...
+            </div>
+            <div class="flex gap-2">
+              <TextField :id="id" v-model="newImage" @keydown.enter.prevent="addNewImage" />
+              <Button type="button" color="gray" start-icon="plus" @click="addNewImage" />
+            </div>
           </div>
-        </div>
+        </template>
+        <template #description>
+          <p class="text-sm">
```
xoxys, line 37: I have refactored this part a bit more in https://github.com/woodpecker-ci/woodpecker/pull/4773 maybe we can merge this one first?

xoxys (commented):
Code LGTM to me otherwise

lafriks:
Could we also have option to specify, that users that has write access to repo should not require approval?

xoxys:
I still believe we should just implement a policy engine instead of implementing every use case on our own.

server/pipeline/gated.go
```diff
@@ -31,6 +35,12 @@ func needsApproval(repo *model.Repo, pipeline *model.Pipeline) bool {
 		return false
 	}
 
+	// skip if user is allowed
+	//  It's enough to check the username as the repo matches the forge of the pipeline already (no username clashes from different forges possible)
```
xoxys, line 39:
```suggestion
	// It's enough to check the username as the repo matches the forge of the pipeline already (no username clashes from different forges possible)
```

xoxys (approved):
Just a small nit.

xoxys (approved):

PR #4771, opened 2025-01-24
Tag pipeline with source information

jenrik:
Indirectly adds labels to Kubernetes pods for use with NetworkPolicies and other integrations.

### DIFF
--- a/pipeline/backend/kubernetes/pod.go
+++ b/pipeline/backend/kubernetes/pod.go
@@ -31,7 +31,8 @@ import (
 )
 
 const (
-	StepLabel            = "step"
+	StepLabelOld         = "step"
+	StepLabel            = "woodpecker-ci.org/step"
 	podPrefix            = "wp-"
 	defaultFSGroup int64 = 1000
 )
@@ -115,6 +116,7 @@ func podLabels(step *types.Step, config *config, options BackendOptions) (map[st
 	if step.Type == types.StepTypeService {
 		labels[ServiceLabel], _ = serviceName(step)
 	}
+	labels[StepLabelOld], err = stepLabel(step)
 	labels[StepLabel], err = stepLabel(step)
 	if err != nil {
 		return labels, err
--- a/server/model/repo.go
+++ b/server/model/repo.go
@@ -140,6 +140,10 @@ type RepoPatch struct {
 
 type ForgeRemoteID string
 
+func (r ForgeRemoteID) String() string {
+	return string(r)
+}
+
 func (r ForgeRemoteID) IsValid() bool {
 	return r != "" && r != "0"
 }
--- a/server/pipeline/stepbuilder/stepBuilder.go
+++ b/server/pipeline/stepbuilder/stepBuilder.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"maps"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/oklog/ulid/v2"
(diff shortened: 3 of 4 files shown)


server/model/repo.go
```diff
@@ -140,6 +140,10 @@ type RepoPatch struct {
 
 type ForgeRemoteID string
 
+func (r ForgeRemoteID) String() string {
```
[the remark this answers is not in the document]
jenrik: I didn't want to assume this type will continue to a alias for type string in the future, but I can change it no problem

server/pipeline/stepbuilder/stepBuilder.go
```diff
@@ -194,6 +195,19 @@ func (b *StepBuilder) genItemForWorkflow(workflow *model.Workflow, axis matrix.A
 		maps.Copy(item.Labels, b.DefaultLabels)
 	}
 
+	// "woodpecker-ci.org" namespace is reserved for internal use
+	for key, _ := range item.Labels {
+		if strings.HasPrefix(key, "woodpecker-ci.org") {
+			delete(item.Labels, key)
+		}
+	}
```
xoxys, line 203: Should we at least add a warning to the logs?

xoxys:
As discussed heavily in https://github.com/woodpecker-ci/woodpecker/pull/4670, metadata labels are good to have in general for multiple scenarios. I would like to keep container backends as close as possible and as docker supports labels as well it would be good to add it to both backends. However, this doesn't necessarily have to be part of this PR.

jenrik:
@xoxys the labels are added to pipeline before it hits any backend, so this should just work with docker without any further modification

xoxys:
> so this should just work with docker without any further modification

But we would have to add the step labels as container labels during container creation in the same way as they are added as pod lables for kubernetes or am I wrong?

xoxys:
Are the pipeline labels used for the pipeline <-> agent scheduling? Can it somehow affect that?

jenrik:
I believe the Kubernetes backend relies on the Pod name, but I will have to check and make sure that it doesn't also rely on the `step` label

jenrik:
@xoxys I have checked the code for the Kubernetes backend and it doesn't use the `step` label for associating pipeline steps to pods, that is all done through the name of the pod

pipeline/backend/kubernetes/pod.go
```diff
@@ -31,7 +31,8 @@ import (
 )
 
 const (
-	StepLabel            = "step"
+	StepLabelOld         = "step"
```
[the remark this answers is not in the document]
jenrik: I have added a deprecation notice in the code, but I can't find it anywhere in the documentation so I have nowhere to warn people except for maybe a changelog somewhere

server/model/repo.go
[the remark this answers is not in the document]
jenrik: Replaced with plain cast to string

server/pipeline/stepbuilder/stepBuilder.go
jenrik: Added log line for drop pipeline labels

server/pipeline/stepbuilder/stepBuilder.go
```diff
@@ -194,6 +195,20 @@ func (b *StepBuilder) genItemForWorkflow(workflow *model.Workflow, axis matrix.A
 		maps.Copy(item.Labels, b.DefaultLabels)
 	}
 
+	// "woodpecker-ci.org" namespace is reserved for internal use
```
pat-s, line 198: Does it need to have the `.org` suffix? I'd rather not have a valid URL in there but only the app name.

pat-s (commented):
Assigning the labels to the pod instead of the container will work for this one but makes it more complicated to eventually implement the "one pod per workflow" approach which uses one pod with many containers for a workflow.

server/pipeline/stepbuilder/stepBuilder.go
xoxys: Why do you think so? Its pretty common (e.g. `app.kubernetes.io` labels) and I would prefer it this way.

pat-s: app.kubernetes.io doesn't resolve to anything. Anyhow, no strong opinion.

The other comment is more important, though.

xoxys:
> Assigning the labels to the pod instead of the container will work for this one but makes it more complicated to eventually implement the "one pod per workflow" approach which uses one pod with many containers for a workflow.

Was there a discussion about this feature already?

pat-s:
Didn't say anything related to blocking. 
It was more a comment related to "why not use container labels instead of pod ones because of XY".

xoxys:
Ok. I'm also not against labels adding labels to containers in general, just not sure if it is really helpful for the problems from the PR description.

jenrik:
Do note that Kubernetes only support labeling on pods, and in many regards treats all the containers in a given pod as the same unit

pat-s:
Ok didn't know that. That is actually an issue then for the "one pod per workflow" idea.

xoxys:
@jenrik Could you look into the linter and test failures?

jenrik:
@xoxys they should both be fixed now

xoxys (approved):

PR #4814, opened 2025-02-05
Optimize repository list loading to return also latest pipeline info

### LINKED ISSUE #4427: New repos list triggers rate limiting
### Component

server, web-ui

### Describe the bug

Testing new repo list, sometimes I get errors `: max connections reached: 10`.
Having 7 repos:
1. for each repo there are requests to `pipeline` endpoint
2. somehow they are doubled, so I have 14 requests  to `pipeline` endpoint

### Steps to reproduce

1. Woodpecker + Gitea
2. Create some repos in Gitea, 10 for example
3. Enable them in Woodpecker
4. Open repositories list
6. Optionally, see `: max connections reached: 10` errors
7. Open browser console, network
8. See, that UI tries to load pipeline info for each repo twice
Screenshot 2024-11-21 133824

### Expected behavior

1. One request for last pipelines info for all repos like `pipelines/last?repo_ids=[1,2,3,4,5]`
2. Or send last pipeline info with repo info in response to repos list
3. Else at least fix those doubled requests
4. In addition to 3, reduce the pagination page size

### System Info

```shell
WP `next-350082cd19`, Gitea `1.22.3`, Postgres `16`, Kubernetes `v1.30.6+k3s1`
```

### Additional context

https://github.com/woodpecker-ci/woodpecker/pull/4386#issuecomment-2490781907

### Validations

lafriks:
Implement for user repos list to return also last pipeline information for enabled repositories

Fixes #4427


lafriks:
Tested in production, now repo list opens instantly 🎉

server/api/user.go
```diff
@@ -140,6 +140,26 @@ func GetRepos(c *gin.Context) {
 		return
 	}
 
+	repoIDs := make([]int64, len(activeRepos))
```
anbraten, line 143: Would it make sense to use an extended model just for this api endpoint to keep the normal pipeline model lean. Sth like:

```go
type PipelineWithLastPipeline struct {
	model.Pipeline
	LastPipeline *model.Pipeline `json:"last_pipeline"`
}
```

server/api/user.go
```diff
@@ -140,6 +140,26 @@ func GetRepos(c *gin.Context) {
...
+	for i, repo := range activeRepos {
+		repoIDs[i] = repo.ID
+	}
+
+	pipelines, err := _store.GetRepoLatestPipelines(repoIDs)
```
anbraten, line 148: What do you think about a query parameter to include the latest pipeline and leaving it out by default?

server/api/user.go
xoxys: What's the (real) benefit?

server/api/user.go
xoxys: Do you worry about the performance or something else?

server/api/user.go
anbraten: Openapi would imply on all other endpoints that a repo could include a pipeline which is never the case. In addition the model is used for grpc etc and would imply extra properties which are never used / assigned.

server/api/user.go
lafriks: I did not find any case where it would be called without needing this, so I don't see much point in this currently

server/api/user.go
lafriks: @anbraten changed

anbraten (approved):

PR #4849, opened 2025-02-14
docs: remove latest from docker compose example

### LINKED ISSUE #4835: Docker compose deployment documentation is outdated
### Component

other

### Describe the bug

The documentation at https://woodpecker-ci.org/docs/administration/deployment-methods/docker-compose the documentation is outdated and the example uses the latest tag instead of v3, despite stating further down that latest is no longer recommended or updated.

### Steps to reproduce

1. Go to https://woodpecker-ci.org/docs/administration/deployment-methods/docker-compose
2. Make sure the documentation is for v3
3. Try the example and end up with v2.8

### Expected behavior

The documentation should state the tag as v3 and not latest

### System Info

```shell
System Agnostic.
```

### Additional context

Image

### Validations

mark-pitblado:
As `latest` is no longer supported for v3, it should not be in the v3 documentation for the example docker compose config file.

Closes #4835

### DIFF
--- a/docs/docs/30-administration/05-deployment-methods/10-docker-compose.md
+++ b/docs/docs/30-administration/05-deployment-methods/10-docker-compose.md
@@ -7,7 +7,7 @@ It relies on a number of environment variables that you must set before running 
 ```yaml title="docker-compose.yaml"
 services:
   woodpecker-server:
-    image: woodpeckerci/woodpecker-server:latest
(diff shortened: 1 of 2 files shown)


xoxys:
I really don't like that we dropped `latest`... Now we have to remember to bump this in the docs on every new major release.

xoxys (approved):

Issue #3326, opened 2024-02-05
Templates for manual workflows


CodeDoctorDE:
### Clear and concise description of the problem

When running many manual workflows, it is not very handy to input the keys on every run.

### Suggested solution

Create a system for github action like form where you add default values, checkboxes and more.
![image](https://github.com/woodpecker-ci/woodpecker/assets/20452814/feeb85ea-37fd-4cd2-86e8-783b5859347d)
The syntax should be like in github actions:
```yaml
    inputs:
      next_version:
        default: "0.0.0"
        description: "The next version to be released"
        required: false
      stable:
        default: false
        description: "Is this version stable?"
        type: boolean
        required: false
```

Maybe adding an extra type of event for tis.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

zc-devs:
I propose to just display [ENV](https://woodpecker-ci.org/docs/usage/environment) vars from workflow and set default value defined in yaml.
```yaml
 steps:
   - name: build
     image: golang
     environment:
       - CGO=0
       - GOOS=linux
       - GOARCH=amd64
       - DISABLE_CACHE
     commands:
       - go build
       - go test
```

![Screenshot 2024-02-05](https://github.com/woodpecker-ci/woodpecker/assets/127358482/f753caa8-b768-4f88-a9bf-3f1290faa1b0)

zc-devs:
>make a opt-in/out for this for internal env vars?

Do you mean [built-in-environment-variables](https://woodpecker-ci.org/docs/usage/environment#built-in-environment-variables). I'm afraid, it is not gonna work. Just look, how much vars Server sends to Agent :) (if you on Kubernetes, just describe pod).

smainz:
Just one thing to consider: 

We are using the manual pipeline execution to perform builds too, but there are much more environment variables in the steps than need to be set / overriden while performing a build / deployment.

Would a config like 
```yml
   environement-input: 
     - BUILD_NUMER: <proposed value>
       MATCH_REX: <some regex to validate the input>
```
   
be an option?

At the moment we have a seperate step to validate the ENV vars to overcome typos.

[...]

PR #4861, opened 2025-02-16
Parameters for manual pipelines

arthurpro:
The Parameters feature introduces a flexible and dynamic way to manage environment variables for manual pipelines.

Key Features:

- Users can create, update, and delete parameters.
- Supports various parameter types, including boolean, string, text, password, single and multiple-choice options.
- Parameter values are configured during manual pipeline creation and passed as environment variables to the pipeline.
- **Environment variables** collapsible panel added to pipeline **Config** tab (see screenshots)


arthurpro:
@qwerty287 yes, I looked at #3326 and I really wanted something similar to GitHub workflow inputs. but for that to work we would need to have woodpecker pipelines fully loaded and parsed ahead of running a manual pipeline. that made me rethink the approach. And I found the least invasive and more elegant way to simply plug into existing Additional pipeline variables in the manual pipeline window. 
Also I borrowed the idea of Parameters from Jenkins pipelines because I'm using Woodpecker while to go away from Jenkins.
Thanks for your feedback! I'll remove the migration.

server/router/api.go
```diff
@@ -126,6 +126,13 @@ func apiRoutes(e *gin.RouterGroup) {
 					// requires push permissions
 					repo.DELETE("/logs/:number", session.MustPush, api.DeletePipelineLogs)
 
+					// requires push permissions
+					repo.GET("/parameters", api.GetParameterList)
```
lafriks, line 130: Shouldn't the parameters list also require push privs?

xoxys:
@arthurpro Sorry for the delay. Could you resolve the merge conflicts and look at the open feedback?

arthurpro:
@xoxys sure, I will take a look at that during this week

qwerty287:
closing as of https://github.com/woodpecker-ci/woodpecker/pull/5632

PR #4863, opened 2025-02-16
Docs improvements

sznowicki:
As signalled over Matrix I prepared some docs changes that I believe would make me less confused during my initial setups where I wanted to have two agents: docker and local and pick the proper one per pipeline jobs.

On top added some references here and there in places where I felt confused as well, while not being that experienced with linux installation process and systemd setups.

Thanks again for help yesterday!

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,6 +8,15 @@ This website is built using [Docusaurus 3](https://docusaurus.io/), a modern sta
 pnpm install
 ```
 
+## Before your first build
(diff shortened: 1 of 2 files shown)


xoxys:
Thanks! After looking into this, I have the feeling we should refactor the installation guide at all. The added installation instructions are not specific for the local backend and should be part of the installation instructions for package-based installations. However, this docs part is pretty messy, there is only a section for [third party packages](https://woodpecker-ci.org/docs/administration/deployment-methods/third-party) while DEB and RPM packages are officially supported...

PR #4870, opened 2025-02-18
Update plugins.json

maurerle:
add hugo static site generator to plugins.
Plugin was forked from https://github.com/drone-plugins/drone-hugo


maurerle:
Thanks @qwerty287 - I fixed this in https://github.com/maurerle/woodpecker-hugo/commit/907f03a1350def69914c60b6f7aee81c6cd01ba5

Should be fine now?
Otherwise, please state so :)

maurerle:
Thanks! Fixed! :)

PR #4871, opened 2025-02-18
plugins: Add SSH/SCP plugin

maurerle:
This is a popular drone plugin which also works well with woodpecker. I am using it to deploy my hugo page to a webserver through scp.

Seems like there once was a deployment plugin for woodpecker which did something similar: https://github.com/woodpecker-ci/woodpecker/commit/87c762f222279dbf0339318dfb42a158fa2bcaff


maurerle:
@qwerty287 can you check this again please?
Now that the pr to the docs is merged

Issue #4887, opened 2025-02-22
Webhook not working in 3.1.0 with Forgejo


srohr-luther:
### Component

server

### Describe the bug

Incoming webhooks are not working in 3.1.0 for Forgejo.

### Steps to reproduce

1. fresh install of Forgejo (tried 9.0.3 and 10.0.1)
2. fresh install of Woodpecker Server following https://woodpecker-ci.org/docs/administration/forges/forgejo
3. enable some repo
4. pushing to repository in Forgejo just gives "failure to parse token from hook" (see additional context)

### Expected behavior

I would expect that triggering builds via webhook from the forge works.

### System Info

```shell
{
    "source": "https://github.com/woodpecker-ci/woodpecker",
    "version": "3.1.0"
}
```

### Additional context

woodpecker-server[7155]: {"level":"error","error":"token is unverifiable: error while executing keyfunc: sql: no rows in result set","time":"2025-02-22T14:14:23+01:00","message":"failure to parse token from hook"}

3.0.1 works, 3.1.0 does not.

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

srohr-luther:
Login, adding/enabling repos, triggering builds all works; it's just the webhook that is borked somehow.

I'm using Forgejo und Woodpecker since a while ago (multiple versions of both) and never had an issue like this.

shared/token/token.go and server/api/hook.go don't look like they had any changes between 3.0.1 and 3.1.0 so I'm at a loss here.

Will try to find out more over the next days.

srohr-luther:
Yes, it does.

oivoo-dev:
I might be mistaken, but I think this might be the changes related to webhooks not working: https://github.com/woodpecker-ci/woodpecker/compare/v3.0.1...v3.1.0#diff-3cd73db84155aa0447a17dadc8f50271d646fcf6132606cf2ab703cc7b80eb38L151

I am using Bitbucket  DC deplyed in Kubernetes and it seems that the repo-id in the access token is always set to 0.

Example"

```{
  "repo-id": "0",
  "type": "hook"
}```

eiger3970:
I'm also affected on v3.13.0, not just v3.1.0

PR #4827, opened 2025-02-08
Store/delete repos after forge communication

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/4730

@mzampetakis Can you test this?

PR #4897, opened 2025-02-24
Use forge IDs for hook tokens

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/4887

Bug introduced by #4827

The repo id is not available when the hook is created, instead use the forge ids.

### DIFF
--- a/server/api/hook.go
+++ b/server/api/hook.go
@@ -249,7 +249,7 @@ func PostHook(c *gin.Context) {
 
 func getRepoFromToken(store store.Store, t *token.Token) (*model.Repo, error) {
 	if t.Get("repo-forge-remote-id") != "" {
-		// check if we have a forge remote ID and forge ID
+		// TODO: use both the forge ID and repo forge remote ID
 		/*forgeID, err := strconv.ParseInt(t.Get("forge-id"), 10, 64)
 		if err != nil {
 			return nil, err


server/api/hook.go
```diff
@@ -248,10 +248,21 @@ func PostHook(c *gin.Context) {
 }
 
 func getRepoFromToken(store store.Store, t *token.Token) (*model.Repo, error) {
-	// try to get the repo by the repo-id
-	repoID, err := strconv.ParseInt(t.Get("repo-id"), 10, 64)
+	if t.Get("forge-remote-id") == "" {
+		// get the repo by the repo-id
+		// TODO remove in next major
```
xoxys, line 253:
```suggestion
		// TODO: remove in next major
```

xoxys (approved):

xoxys (approved):

server/api/hook.go
```diff
@@ -248,10 +248,21 @@ func PostHook(c *gin.Context) {
...
+
+	// check if we have a forge remote ID and forge ID
+	/*forgeID, err := strconv.ParseInt(t.Get("forge-id"), 10, 64)
 	if err != nil {
 		return nil, err
-	}
-	return store.GetRepo(repoID)
+	}*/
+
+	return store.GetRepoForgeID(model.ForgeRemoteID(t.Get("forge-remote-id")))
```
anbraten, line 267: Would change the code to check for existence of `forge-remote-id` first and use that if available. If that's not the case check for `repo-id` and use that and so on.

server/api/repo.go
```diff
@@ -150,7 +151,8 @@ func PostRepo(c *gin.Context) {
 
 	// creates the jwt token used to verify the repository
 	t := token.New(token.HookToken)
-	t.Set("repo-id", strconv.FormatInt(repo.ID, 10))
+	t.Set("forge-remote-id", string(forgeRemoteID))
```
anbraten, line 154: `forge-remote-id` was a bit confusing for me in the first moment as it does not tell its the repos `forge-remote-id`, maybe sth like `repo-forge-remote-id` would be better.

server/api/hook.go
anbraten: Using `!=` would be a bit more direct and cleaner IMO.

anbraten (approved):

anbraten (approved):

Issue #4341, opened 2024-11-08
Conversion extension or pipeline modification "on the fly"?


nmapx:
### Clear and concise description of the problem

I'm currently moving from Drone to Woodpecker and while discovering differences I've noticed there is no such thing as conversion extension here. I guess it was introduced into Drone when Woodpecker was already there but I'm not sure. Maybe you could clarify more, or if it was removed why so? In my opinion it's a very useful and specific feature that allows more than just static pipelines. I'm talking here about "paved roads" / "golden paths" approach which allows creating complex yet very easy to maintain pipelines.

### Suggested solution

Well, maybe it doesn't need to be conversion extension thing just like it is in Drone but something that would allow modifying pipeline syntax on the fly. This way we could easily introduce support for other languages (Starlark & Jsonnet) - it would eventually be converted into Yaml using this approach.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

zc-devs:
https://woodpecker-ci.org/docs/administration/advanced/external-configuration-api
https://github.com/woodpecker-ci/woodpecker/pull/1396
https://github.com/woodpecker-ci/woodpecker/discussions/3277
https://github.com/woodpecker-ci/woodpecker/discussions/2254

zc-devs:
> Before the run or restart of any pipeline Woodpecker will make a POST request to an external HTTP API sending the current repository, build information and all current config files retrieved from the repository

> still sending yaml syntax to the external API. There is no way to initialize the repository with pipeline in different syntax than yaml

I'm not a ~real welder~ user of External API :) If I remember correctly, WP indeed looks up only for `yaml` and `yml` files. But I don't know whether it reads/parses them or not. You can try to save `jsonnet` content in a file with `yaml` extension 😆

zc-devs:
> WP indeed looks up only for yaml and yml config files

server/services/config/forge.go:116

> trying to parse the file(s) before sending them further

server/services/setup.go:83
server/services/config/combined.go:36
server/pipeline/create.go:81
server/pipeline/create.go:94

> will make a POST request to an external HTTP API sending ... all current config files retrieved from the repository

[...]

PR #4908, opened 2025-02-28
Include forge type in netrc

qwerty287:
to fix https://github.com/woodpecker-ci/woodpecker/issues/4341 by allowing external services to know which forge type they would have to query


xoxys (approved):

xoxys:
As we have existing tests for the forges, should we just add it and check that it is set correctly? E.g. https://github.com/woodpecker-ci/woodpecker/blob/26ae32ac5fe7de4441c81f2fd356340947c60758/server/forge/github/github_test.go#L68

xoxys (approved):

Issue #4918, opened 2025-03-01
Cannot enable new repos after upgrading from v3.1.0 to v3.2.0. Duplicate entry for key 'UQE_orgs_name'


DHandspikerWade:
### Component

server

### Describe the bug

After upgrading to 3.2.0, I am no longer able to enable new repos in the server dashboard. Attempting results in a error notification and the message "Error 1062 (23000): Duplicate entry 'devin' for key 'UQE_orgs_name'" being written to the server logs with "devin" being admin and only user.  Also happens in the current `next` version.

Rolling back to 3.1.0, repos can be enabled without issue. 

![Image](https://github.com/user-attachments/assets/e5880058-8783-4820-8653-433dcfd7d39f)

### Steps to reproduce

1. Navigate to the `/repos` page.
2. Click the "Add repository" button.
3. Click "Enable" on any listed repository not currently enabled. 
4. Receive error message

### Expected behavior

I should be able to enable a repo without error. 

### System Info

```shell
{
  "source": "https://github.com/woodpecker-ci/woodpecker",
  "version": "next-1232ceb794"
}
```

### Additional context

- Reproducible with `woodpeckerci/woodpecker-server:v3.2.0` and `woodpeckerci/woodpecker-server:next-1232ceb794` container images.
- Running in a K3S cluster
- Gitea forge  version 1.23.3
- MariaDB database

Server debug log: [server.debug.log](https://github.com/user-attachments/files/19040797/server.debug.log)

Kubernetes Manifest: 

```
---

apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: woodpecker-server
  namespace: woodpecker
  labels:
    app.kubernetes.io/name: woodpecker
    app.kubernetes.io/component: server
spec:
  serviceName: woodpecker-server-headless
  revisionHistoryLimit: 2
  replicas: 1
  minReadySeconds: 20
  selector:
    matchLabels:
      app.kubernetes.io/name: woodpecker
      app.kubernetes.io/component: server
  template:
    metadata:
      labels:
        app.kubernetes.io/name: woodpecker
        app.kubernetes.io/component: server
    spec:
      priorityClassName: please-keep-running
      restartPolicy: Always
      terminationGracePeriodSeconds: 100
      containers:
      - name: server
        image: woodpeckerci/woodpecker-server:v3.2.0
        imagePullPolicy: IfNotPresent
        env:
          - name: WOODPECKER_DATABASE_DRIVER
            value: mysql
[...]

PR #4817, opened 2025-02-07
Fix org assign on login

anbraten:
# Changes

- ...

PR #4923, opened 2025-03-03
Fix unique constraint for orgs

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/4918

In https://github.com/woodpecker-ci/woodpecker/pull/4817 we changed the org lookup from `name="name"` to `name="name AND forge_id="forgeid"` (https://github.com/woodpecker-ci/woodpecker/pull/4817/files#diff-eb104e63f04fbe20f88199ac5e15aa7164690f990773a43cedb458c083b16260R84) while the unique constraint was still `"UQE_orgs_name" UNIQUE, btree (name)`.

With this change, the constrained was changes to `"UQE_orgs_s" UNIQUE, btree (forge_id, name)`

### DIFF
--- a/server/model/org.go
+++ b/server/model/org.go
@@ -17,8 +17,8 @@ package model
 // Org represents an organization.
 type Org struct {
 	ID      int64  `json:"id,omitempty"       xorm:"pk autoincr 'id'"`
-	ForgeID int64  `json:"forge_id,omitempty" xorm:"forge_id"`
-	Name    string `json:"name"               xorm:"UNIQUE 'name'"`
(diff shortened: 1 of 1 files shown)


xoxys:
Fair. I checked my local test instance again. I had an entry in the orgs table with forge_id=0. Don't ask me how it gets there, but this has triggered the error message from the issue with v3.2 while it has not with v3.1. 

```
woodpecker_server=# select * from orgs;
 id | forge_id |    name    | is_user | private 
----+----------+------------+---------+---------
  1 |        0 | woodpecker | t       | f
  2 |        1 | testorg    | f       | f
  3 |        1 | testorg2   | f       | f
```

I might got trapped by this weird state, however, the constraint should still be fixed and match the changed lookup.

PR #4924, opened 2025-03-03
Add migration to fix zero forge_id in orgs table

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/4918

- drive-by fix that unifies the migration file names
- add migration for `forge_id=0` in `orgs` table

I have not worked much with XORM and even less with the migrations, so please let me know if this is wrong.


lafriks (approved):
Code looks good but what is the source of such records being created?

PR #4967, opened 2025-03-16
fix(backend/docker): add a permission fix function before each step

### LINKED ISSUE #4952: docker agent not setting right permissions when skip_clone is true
### Component

agent

### Summary
When using the following combination:
- docker backend
- rootless Image
- `skip_clone: true`
The step fails with the message
```
mkdir: can't create directory '/woodpecker/src/': Permission denied
```

### Reason
The agent creates a docker volume and attaches it when creating the container for the pipeline step. The volume gets mounted to /woodpecker by default. When `skip_clone` is `true` the permissions for this folder are not correctly set and therefore the pipeline step fails during the entrypoint script where the /woodpecker/src/user/repo folder gets created.

### Steps to reproduce

1. use docker agent with default config
2. start pipeline step with `skip_clone: true` and a rootless image like `curlimages/curl`
3. see the failed step

### Expected behavior

Set the folder permission to 777 like it gets done when cloning a repo:
https://github.com/woodpecker-ci/plugin-git/blob/main/plugin.go#L33

### System Info

```shell
woodpecker-agent-1:
    container_name: woodpecker-agent-1
    image: woodpeckerci/woodpecker-agent:v3.3.0
    user: root
    command: agent
    restart: unless-stopped
    privileged: true
    volumes:
    environment:
      - WOODPECKER_SERVER=woodpecker-grpc.apps.yolokube.de:443
      - WOODPECKER_AGENT_SECRET=${WOODPECKER_AGENT_SECRET}
      - WOODPECKER_MAX_WORKFLOWS=4
      - WOODPECKER_FILTER_LABELS="backend=docker"
      - WOODPECKER_BACKEND_DOCKER_ENABLE_IPV6=true
      - WOODPECKER_GRPC_SECURE=true
      - WOODPECKER_GRPC_VERIFY=true
```

### Additional context

_No response_

### Validations

Tom3201:
Closes #4952

This pull request includes a workaround for the "write-denied" permission errors for rootless image steps in the docker backend. The following features were added:
- Function to spawn a busybox container, create the workdir and set the permission to 777
- Integration of the function in each step as a subroutine so that the actual container + image can be prepared in parallel


xoxys:
I dont see another way around it. The clone plugin can "fix" it because its running as root and can chown 777 the workspace dir. However, I commented in the issue.

zc-devs:
> Just running an additional step/container without asking seems

as a solution to me, if it's one 'setup' container per workflow (like `clone`).

xoxys:
Closing as discussed in the issue. @Tom3201 Thanks again for your effort!

PR #4968, opened 2025-03-16
Remove woodpecker prefix from env var title in docs

xoxys:
Remove the `WOODPECKER_` prefix from the title of each env var in the docs. This makes the sidebar more readable. Also, slightly restructured the layout of the env var docs:

image


anbraten (commented):
lgtm in general. There might be a small risk that someone tries the headline as env name now.

anbraten (approved):

anbraten:
Sounds good to just give it a try and adjust if necessary.

Issue #525, opened 2021-11-21
secrets aren't preserving newlines encoded as strings within json fields


eleith:
### Component

server

### Describe the bug

reproducible steps

1. add a secret with the value `{"hi":"there\nshould be no new\nline control character"}`
2. in a pipeline, inject the secret as `VALUE` and run the command `echo $${VALUE}`
3. note that the newlines are now interpreted as control characters and aren't preserved in the string, if you were to save the value into a file, the file would look like the follow

 ```
{"hi":"there
should be no new
line control character"}
```

expected

1. the string would exactly match the format as entered `{"hi":"there\nshould be no new\nline"}`

### System Info

```shell
next-d02dfe99
```

### Additional context

```shell
the problem i'm trying to solve is to save a json file as a secret and in the pipeline, echoing that string back into a file.

i have tried manually adding the secret in the web-ui and using the cli (with the '@' curl syntax) and they all produced the same result.

i have verified in the database directly, the value is as expected (`{"hi":"there\nshould be no new\nline"}`)
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy).

anbraten:
I don't think we can easily support / solve this. Some users expect secrets to have proper new-lines (e.g. for ssh-keys). Replacing new-line with its escape version `\n` wouldn't work for them in that case I think.

eleith:
to document my workaround, i currently take the secret, pipe it through sed and then store that in a file

```
echo $${SECRET} | sed -z 's/\n/\\\n/g' | sed -z 's/\\\n$//g' > /tmp/secret.json
```

now, any commands in my pipeline that need the secret can be pointed to a file, accessible to my pipeline

eleith:
if JSON secrets can not be supported, it might be worth stating that in the [docs](https://woodpecker-ci.org/docs/usage/secrets#adding-secrets). 

it was unexpected for me to find that json was not being preserved as i saw it in the secrets UI, especially when i tried the curl syntax as suggested in the documentation:


[...]

PR #4974, opened 2025-03-17
Rework secrets docs and document multiline secrets

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/525


xoxys:
I dropped the multi-line part for now. I can't reproduce the issue (see my comment in https://github.com/woodpecker-ci/woodpecker/issues/525). It just works as expected and `\n` control chars are preserved and not converted to newlines.

PR #4980, opened 2025-03-18
Apply refactoring to v3 docs

xoxys:
Copies next docs to v3 versioned docs.


xoxys:
What does it mean? 😅

xoxys:
Understood and yes we can do this. But what's the benefit? There is no difference in the docs for these 3.x minor versions. IDK if it is really worth it...

xoxys:
Looks like a few maintainers agree here so can we merge this PR and stick to 3.x docs for now? Otherwise Im going to close the PR and create the docs for every missing v3 minor version to get forward.

xoxys:
Nevermind, I dont think that makes sense as we would have to change more to get 3.x docs instead of 3.4.x. Ill continue with the current version scheme.

PR #4992, opened 2025-03-22
Add missing docs for 3.x minor versions

xoxys:
- Copy missing docs for `3.4`, `3.3` and `3.2` from the corresponding tag `docs/docs` folder to versioned docs
- Drop versioned docs for `2.7` and only keep latest 2.x docs for compatibility reasons


anbraten (approved):
Haven't checked every changed files. What I checked lgtm

xoxys:
Thanks for reviewing

PR #5001, opened 2025-03-24
Add pagination component to UI

xoxys:
Supersedes: https://github.com/woodpecker-ci/woodpecker/pull/2977

### DIFF
--- a/web/components.d.ts
+++ b/web/components.d.ts
@@ -88,6 +88,7 @@ declare module 'vue' {
     NumberField: typeof import('./src/components/form/NumberField.vue')['default']
     OrgRegistriesTab: typeof import('./src/components/org/settings/OrgRegistriesTab.vue')['default']
     OrgSecretsTab: typeof import('./src/components/org/settings/OrgSecretsTab.vue')['default']
+    Pagination: typeof import('./src/components/layout/Pagination.vue')['default']
     Panel: typeof import('./src/components/layout/Panel.vue')['default']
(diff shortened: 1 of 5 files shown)


anbraten:
@xoxys Do you have any plans to use this component directly?

xoxys:
Yes, Im planning to make use of it in the pipelines list view. Why do you ask?

anbraten:
> Yes, I'm planning to make use of it in the pipelines list view. Why do you ask? Something to consider?

Would suggest to add it in this PR directly then, so we can test it with some actual UI.

xoxys:
With infinite scrolling, you mean load next page if we reached the end of the current view? We can I dont care which method. But this is also not implemented somewhere? Or do we use infinite scrolling already for some views?

xoxys:
Well no. This is not what I understand under "infinite scrolling". If the branches page gets opened, we just load all branches from the forge at once. We do it via pagination but this still means multiple requests were sent immediately to get all branches.

This is IMO a bad practice and will have massive performance issues if you have to load hundreds of pages, e.g. for pipelines. What should be done here is loading only the first page and show a "load more" button to load the next page on request. Or, if the user scrolled down to the current viewport, automatically request the next page and dynamically extend the viewport to show also items from the next page.

xoxys:
Oh it **is** using `useInfiniteScroll` composition already :facepalm: alright then I was wrong sorry :sweat_smile: in that case we should just use it for pipelines as well. Ill give it a try. Thanks for the hint!

qwerty287:
@xoxys we still need this?

xoxys:
I dont think so.

PR #4982, opened 2025-03-18
Add back cursor-pointer to pipeline step list buttons

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/discussions/2162#discussioncomment-12538209

PR #5002, opened 2025-03-24
Use pointer cursor for icon buttons

anbraten:
similar to #4982


xoxys:
Why the change to the icon buttons? Those are native `a` elements and have a pointer anyway.

xoxys:
Switching in some places to `const btnClass` approach looks inconsistent to me. If we dont want to use the inline style apply approach anymore fine but I would prefer to apply it everywhere then. The new approach makes it impossible to find out whats defined in the class by hovering over it:

image

xoxys:
I reverted the change to use `btnClass` for the given reason. This can be done for and discussed in a separate PR, hope you are ok with that.

PR #5010, opened 2025-03-24
Use badge value instead of label for single values

xoxys:
Single value badges look a bit "massive"  make them a little lighter.

Before:
image

After:
image


lafriks:
How will it look in dark mode?

xoxys:
image

lafriks (approved):

lafriks:
Conflicts but otherwise lgtm

lafriks (approved):

Issue #5003, opened 2025-03-24
Tasks won't start


rubenelshof:
### Component

server

### Describe the bug

None of the tasks in any repo will start currently.
This started happening about 2 days ago I think.
Unfortunately I can't pin point where it exactly is going wrong.
The message with `queue: pending right now` and `queue: trying to assign task` are in a continous loop.

### Steps to reproduce

-

### System Info

```shell
next-296bd5b437
```

### Additional context

```
{"level":"debug","time":"2025-03-24T14:07:12Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/router/middleware/session/repo.go:157","message":"dionysussg granted &{UserID:1 RepoID:11 Repo:<nil> Pull:true Push:true Admin:true Synced:1742824989 Created:1684141068 Updated:1742825228} permission to dionysussg/ci-test"}
{"level":"debug","time":"2025-03-24T14:07:12Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/queue/fifo.go:419","message":"queue: trying to remove 3622"}
{"level":"error","error":"sql: no rows in result set","time":"2025-03-24T14:07:12Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/pipeline/cancel.go:60","message":"queue: evict_at_once: [3622]"}
{"level":"debug","ip":"","latency":16.324537,"method":"POST","path":"/api/repos/11/pipelines/58/cancel","status":204,"time":"2025-03-24T14:07:12Z","user-agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36","time":"2025-03-24T14:07:12Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/router/middleware/logger.go:59"}
{"level":"debug","ip":"","latency":3.61538,"method":"GET","path":"/api/badges/11/status.svg","status":200,"time":"2025-03-24T14:07:14Z","user-agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36","time":"2025-03-24T14:07:14Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/router/middleware/logger.go:59"}
[...]

PR #5011, opened 2025-03-24
Fix pipeline source information

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/5003

We need to distinguish between labels used for agent filtering and those that should be used by the backend, e.g. for network policies on Kubernetes. Especially when used for security related things like netpols, it should not be possible to override such labels by end users in the pipeline config. This PR does:

- only use the prefix `woodpecker-ci.org/` for WP-managed labels, those labels are now excluded from the agent filters
- revert all labels used for agent filtering to original state, those labels can still be overwritten by end users in the pipeline config

That also makes all label migrations obsolete. I have also tested that agent filtering is working again with the example filtering from the linked issue.


xoxys:
@jenrik FYI, if you also want to take a look.

docs/docs/20-usage/20-workflow-syntax.md
```diff
@@ -594,34 +594,24 @@ For more details check the [matrix build docs](./30-matrix-workflows.md).
...
-
-| Label                              | Description                                                        | Woodpecker managed |
-|------------------------------------|--------------------------------------------------------------------|--------------------|
-| `woodpecker-ci.org/forge-id`       | Internal Forge identifier                                          | 🤖 yes             |
-| `woodpecker-ci.org/repo-forge-id`  | Repository identifier from the Forge                               | 🤖 yes             |
-| `woodpecker-ci.org/repo-id`        | Internal repository identifier                                     | 🤖 yes             |
-| `woodpecker-ci.org/repo-name`      | Repository display name (excluding project/organization)           | 🤖 yes             |
-| `woodpecker-ci.org/repo-full-name` | Repository display name (including project/organization)           | 🤖 yes             |
-| `woodpecker-ci.org/branch`         | Git branch name                                                    | 🤖 yes             |
-| `woodpecker-ci.org/org-id`         | Internal organization/project identifier                           | 🤖 yes             |
```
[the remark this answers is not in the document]
xoxys: Yes but not in this place. Ill take care.

cmd/agent/core/agent.go
```diff
@@ -247,12 +247,10 @@ func run(ctx context.Context, c *cli.Command, backends []types.Backend) error {
 
 	// set default labels ...
 	labels := make(map[string]string)
-	for label := range pipeline.ManagedLabels {
-		labels[label] = "*"
-	}
```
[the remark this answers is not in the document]
xoxys: This works now as I have dropped completely from the agent filter func, see https://github.com/woodpecker-ci/woodpecker/pull/5011/files#diff-fa3f3d2bbef57c87a86886007de51bf6efd76e6820b8cf279dcb398017a3293fR28

xoxys:
Thanks. Would you like to give it a test on your own? We can create a PR container image automatically if that would help.

jenrik:
Please do that, and I will try to get it deployed to our staging environment today

mariusheine:
@xoxys Where can I find this PR container image? I would also like to test it because I think we have the same issue atm.

xoxys:
I had to trigger the CI again, after https://ci.woodpecker-ci.org/repos/3780/pipeline/26131 is done the image is published on dockerhub. I can post the link later.

mariusheine:
> I had to trigger the CI again, after https://ci.woodpecker-ci.org/repos/3780/pipeline/26131 is done the image is published on dockerhub. I can post the link later.

Nevermind. I found the place where it is pushed. I was just confused because nothing was pushed in the pipeline

xoxys:
Image ready: `docker pull woodpeckerci/woodpecker-server:pull_5011-alpine`

mariusheine:
@xoxys For me it is still not working. We are using your autoscaler as well. Maybe there is a problem as well with the labels. The autoscaler keeps creating and draining agents infinitely but is never able to assign tasks to it...

So let's see what @jenrik says.

xoxys:
Thanks for the feedback. Can you share your filter config? Then I'll test it with the autoscaler as well. Just did a test with a single static agent yesterday and a simple hostname filter, at least this was working.

mariusheine:
> Thanks for the feedback. Can you share your filter config? Then I'll test it with the autoscaler as well. Just did a test with a single static agent yesterday and a simple hostname filter, at least this was working.

Sorry for the dumb question but where I do find the filter config? I can share the docker compose of our deployment with you. Our workflows do not set special labels or filters .

xoxys:
Have you set anything for `labels` in your pipeline config? (see https://github.com/woodpecker-ci/woodpecker/issues/5003#issuecomment-2749218779) If not, I'm a bit confused as without a filter config everything was working even without the PR. We use the autoscaler for ci.woodpecker-ci.org and don't see any issue with scheduling pipelines.

As your issue seems to be different, I would suggest opening a dedicated issue to keep this PR more readable.

mariusheine:
> Have you set anything for `labels` in your pipeline config? (see [#5003 (comment)](https://github.com/woodpecker-ci/woodpecker/issues/5003#issuecomment-2749218779)) If not, I'm a bit confused as without a filter config everything was working even before this PR. We use the autoscaler for ci.woodpecker-ci.org and don't see any issue with scheduling pipelines.
> 
> As your issue seems to be different, I would suggest opening a dedicated issue to keep this PR more readable.

I will open a new issue. Thx for your help!

jenrik:
We have a problem with the `woodpecker-ci.org/repo-full-name` label, apparently Kubernetes doesn't like labels that contain `/` characters.

```
Pod "wp-01jq67hehrv8ypjvcmp0nw28j9" is invalid: metadata.labels: Invalid value: "PATCHMGMT/patch-management-web": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

xoxys:
But this was already the case before this PR and affects all metadata labels not only this one?

jenrik:
Annotations can not be used for filtering in Kubernetes, so they are useless for if wanna configure NetworkPolicies. Also annotations have the same limitations on what they can contain

xoxys:
Ah yeah, my bad. But then we can't use the labels on docker backend without transformation... I would say we switch to `org.woodpecker` as prefix. So labels would be `org.woodpecker.repo-full-name`.

jenrik:
Kubernetes label keys are allowed to contain `/` characters it is just the key that can not.
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

xoxys:
Now I got it :D so it does not affect all labels but only the full name one as this one contains `/` in the value. In that case just drop it and use repo name and repo owner as two separate labels?

jenrik:
@xoxys I have created #5013 to add a more general fix for the Kubernetes label issue

xoxys:
@rubenelshof could you give `woodpeckerci/woodpecker-server:pull_5011-alpine` a test?

docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
```diff
@@ -6,6 +6,20 @@ toc_max_heading_level: 2
...
+## Metadata labels
+
+Woodpecker adds some labels to the pods to provide additional context to the workflow. These labels can be used for various purposes, e.g. for simple debugging or as selectors for network policies.
+
+The following metadata labels are supported:
+
+- `woodpecker-ci.org/forge-id`
+- `woodpecker-ci.org/repo-forge-id`
+- `woodpecker-ci.org/repo-id`
+- `woodpecker-ci.org/repo-name`
```
anbraten, line 18: Just to confirm, are there actually use-cases for `repo-forge-id`, `repo-id` and `repo-name` if you could use `repo-full-name`? I could image users want to filter for a specific repo, an org and maybe a forge. Having one unique selector for each like id or name should be sufficient, shouldn't it?

xoxys: Make it human readable. There was a huge discussion around how to debug containers as its pretty hard to find out which container/pod handles a step or workflow.

rubenelshof:
> @rubenelshof could you give `woodpeckerci/woodpecker-server:pull_5011-alpine` a test?

I have tested it and confirm it works for me.

jenrik:
I will try to get it deployed to our staging environment later today (European Central Time) and report back

jenrik:
The Kubernetes backend is working again, I'm able run a pipeline without errors

docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
anbraten: Will check on the discussion and open a PR before the next release if necessary.

anbraten (approved):

PR #5013, opened 2025-03-25
Replace illegal characters in Kubernetes labels

jenrik:
This is a workaround for Kubernetes labels being more restrictive that what we allow in Woodpecker.

### DIFF
--- a/pipeline/backend/kubernetes/pod.go
+++ b/pipeline/backend/kubernetes/pod.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"regexp"
 	"strings"
 
 	"github.com/rs/zerolog/log"
@@ -100,6 +101,11 @@ func podMeta(step *types.Step, config *config, options BackendOptions, podName s
 	return meta, nil
 }
 
+var (
+	k8sLabelValueSanitizer = regexp.MustCompile("[^a-zA-Z0-9-]+")
+	k8sLabelKeySanitizer   = regexp.MustCompile(`[^a-zA-Z0-9-_./]+`)
+)
+
 func podLabels(step *types.Step, config *config, options BackendOptions) (map[string]string, error) {
 	var err error
 	labels := make(map[string]string)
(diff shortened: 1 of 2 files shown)


xoxys:
Just reuse the logic from https://github.com/woodpecker-ci/woodpecker/blob/a9ea85742b2f4c146f7eaa0033c205f8eeb82616/pipeline/backend/kubernetes/pod.go#L132C33-L134?

jenrik:
That is a good idea. I have switched it over to use `toDNSName` for labels, and removed the sanitization of label keys because the `toDNSName` function removes `/` characters.

pipeline/backend/kubernetes/pod.go
```diff
@@ -108,7 +108,11 @@ func podLabels(step *types.Step, config *config, options BackendOptions) (map[st
 		// Only copy user labels if allowed by agent config.
 		// Internal labels are filtered on the server-side.
 		if config.PodLabelsAllowFromStep || strings.HasPrefix(k, pipeline.InternalLabelPrefix) {
-			labels[k] = v
+			sanitizedValue, err := toDNSName(v)
+			if err != nil {
+				return nil, err
+			}
+			labels[k] = sanitizedValue
```
xoxys, line 115:
```suggestion
			labels[k], err = toDNSName(v)
			if err != nil {
				return labels, err
			}
```

Can we not skip the extra spep here as well? I would also return `labels` on errors instead of `nil` to follow the existing implementation.

jenrik: fixed in 41dd58c573284b7c7e2c43b71e45403e0892035a

xoxys: Only partially :) The suggestion also drops the extra assignment `sanitizedValue, err := toDNSName(v)`

pipeline/backend/kubernetes/secrets.go
```diff
@@ -260,7 +260,11 @@ func registrySecretLabels(step *types.Step, config *config) (map[string]string,
 		// Only copy user labels if allowed by agent config.
 		// Internal labels are filtered on the server-side.
 		if config.PodLabelsAllowFromStep || strings.HasPrefix(k, pipeline.InternalLabelPrefix) {
-			labels[k] = v
+			sanitizedValue, err := toDNSName(v)
```
xoxys, line 263: Same here.

xoxys:
Can we add a test case for this? We should prevent adding labels with illegal chars in the future.

jenrik:
We could probably pull the full regex from Kubernetes' source code and use that as part of a unit test

xoxys:
I'm totally fine to hardcode the test regex from the Kubernetes docs.

PR #5023, opened 2025-03-27
Use existing WorkingDir for clone steps

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/2510, https://github.com/woodpecker-ci/woodpecker/issues/4952

This PR ensures an existing path is used as `WorkingDir` for clone steps to ensure container environments that don't create missing working directories automatically can still execute the clone step to ensure the full workspace path is created.

As now the mount path is used as working directory, this should always exist.

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -56,8 +56,6 @@ MD013:
   tables: false
   # Include headings
   headings: true
-  # Include headings
-  headers: true
   # Strict length checking
   strict: false
   # Stern length checking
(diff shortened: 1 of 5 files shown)


xoxys:
I dont know why. At least the official clone plugin does not rely on it. We can add an. Extra step and check if the official clone plugin is used only change it in that case and log a deprecation warning in case a custom clone plugin is used. In that case the fix will not work for custom clone plugins of course which would affect at least me 🙈

lafriks:
Why are making breaking changes for one platform on expense on breaking other? 😞

xoxys:
No worries :) I just missed this side effect. It was definitely not intended to force users to change hundreds of pipeline configs again.

PR #5034, opened 2025-03-31
Bump golangci-lint to v2

xoxys:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ check-xgo: ## Check if xgo is installed
 
 install-tools: ## Install development tools
 	@hash golangci-lint > /dev/null 2>&1; if [ $$? -ne 0 ]; then \
-		go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest ; \
+		go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest ; \
 	fi ; \
 	hash gofumpt > /dev/null 2>&1; if [ $$? -ne 0 ]; then \
 		go install mvdan.cc/gofumpt@latest; \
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -129,7 +129,7 @@ check-xgo: ## Check if xgo is installed
 
 install-tools: ## Install development tools
 	@hash golangci-lint > /dev/null 2>&1; if [ $$? -ne 0 ]; then \
-		go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest ; \
+		go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest ; \
```
anbraten, line 132: As using the `go install` is not recommended, should we change to curl with sth like here?

```suggestion
		 test -s $(LOCALBIN)/golangci-lint && $(LOCALBIN)/golangci-lint version | grep -q $(GOLANG_CI_LINT_VERSION) || \
                curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s $(GOLANG_CI_LINT_VERSION) \
```

xoxys: If we want to version it now, yes. But we dont need curl, we can just use `$(GO) run $(GOLANGCI_LINT_PACKAGE) run` directly IMO. However I would do it in a dedicated PR then across all repos...

xoxys: Can you give a reference for `As using the go install is not recommended,`? Cant find why.

anbraten: Don't think `go run` should be used as well: https://github.com/golangci/golangci-lint/issues/5641 https://golangci-lint.run/welcome/install/#install-from-sources

xoxys: Ok it's not recommended for golangci not in general. However, I don't see a real reason for us, as it works without issues. However, if someone wants to refactor it I'm not against it.

PR #5056, opened 2025-04-02
Document rootless container requirements for skip-clone

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/4952


docs/docs/20-usage/20-workflow-syntax.md
```diff
@@ -710,6 +710,10 @@ steps:
 
 ## `skip_clone`
 
+:::warning
+The default clone step is executed as `root` to ensure that the workspace directory can be accessed by any user (`0777`). This is necessary to allow rootless step containers to write to the workspace directory. If a rootless step container is used with `skip_clone`, the user must ensure a suitable workspace directory that can be accessed by the unprivileged container use, e.g. `/tmp`.
```
zc-devs, line 714: S from SOLID. And [could has been run rootless](https://github.com/woodpecker-ci/woodpecker/issues/2510#issuecomment-2773221104).

xoxys: Can you please communicate in full sentence? We are mixing up two cases here:

- if rootless step containers are used with `skip_clone`, we cant fix (or dont want as explained in the discussion already) permissions
- if rootless steps are used with the regular clone step at least the official clone plugin takes care of setting the correct permission

This is what this PR documents. The linked issue has noting to do with this, as its about container engines that dont create missing workdirs automatically regardless of any permission issue.

lafriks:
Maybe we need sample on how to do that?

xoxys:
I can link to the existing workspace config or add a dedicated example here. What do you think?

lafriks:
I think a dedicated example in the docs would be best option

6543:
we could also add an option to the clone plugin that it does chown to a set user/grup id if set

xoxys:
> we could also add an option to the clone plugin that it does chown to a set user/grup id if set

This is about `skip-clone` so no you cant.

lafriks:
Maybe we could have a workspace preparation docker image that would be run as root and do necessary tasks to prepare everything so that the skip clone would work and also clone steps could be run as non-root images?

lafriks:
Maybe even a woodpecker server could act as docker registry to serve container image for this so that it would work in instances that have no internet connection

zc-devs:
https://github.com/woodpecker-ci/woodpecker/pull/5056#discussion_r2025605751
https://github.com/woodpecker-ci/woodpecker/pull/5056#issuecomment-2784618118

Finally! Well done, @lafriks.

xoxys:
This is drifting away too far from the PR purpose. For now, we agreed in the linked issue to not implement any extra step or similar approaches because it adds a lot of complexity. Please open a new issue/discussion for such ideas.

lafriks:
Sorry for commenting that here and I'm not against for now just updating docs but in long-term we could need this anyway to also implement more advanced features like for moving assets between multiple pipelines etc

xoxys:
Sure, I'm also not against it as well as long as it is a reliable and maintainable implementation. But this requires an agreement on how to design such a feature first, and obviously someone to work on the implementation (in the last few months, there haven't been that many people). In the meantime, we should document the possible workaround.

lafriks (approved):

PR #5066, opened 2025-04-03
fix(server/router/api.go): org member permissions to org secrets

Levy-Tal:
## Issue
Organization members(not org admin) with repository write access cannot view global secrets and organization secrets in the repository settings page. They receive an error when trying to access organization secrets, despite having the necessary permissions to use them.

## Steps to Reproduce
1. Create a global secret and an organization secret
2. Log in as an organization member (not admin)
3. Go to one of the organization repo's settings and then secrets

### Expected Behavior
As a repository writer and organization member, you should see:
1. Repository secrets
2. Organization secrets
3. Global secrets

### Actual Behavior
Only repository secrets are visible, and the following error appears:

Error when accessing secrets

## Root Cause
Upon investigation, the error comes from `/api/orgs/{number}/secrets`:

API error trace

In the `api.go` file, the route `/secrets` can only be accessed by organization admins due to the middleware:
```go
org.Use(session.MustOrgMember(true))
```

However, organization members should be able to view (but not modify) organization secrets they can use in their repositories.

## Solution
Move the GET endpoint for `/secrets` before the admin middleware and limit it to organization members (not requiring admin status).

## Code Changes
```diff
org := orgBase.Group("")
{
+	org.GET("/secrets", session.MustOrgMember(false), api.GetOrgSecretList)
	org.Use(session.MustOrgMember(true))
	org.DELETE("", session.MustAdmin(), api.DeleteOrg)
-	org.GET("/secrets", api.GetOrgSecretList)
	org.POST("/secrets", api.PostOrgSecret)
	org.GET("/secrets/:secret", api.GetOrgSecret)
	org.PATCH("/secrets/:secret", api.PatchOrgSecret)
```

## Result
After applying this change, organization members can now see all secrets they have access to:

Fixed view showing all accessible secrets

### DIFF
--- a/server/router/api.go
+++ b/server/router/api.go
@@ -60,10 +60,10 @@ func apiRoutes(e *gin.RouterGroup) {
 
 				org := orgBase.Group("")
 				{
+					org.GET("/secrets", session.MustOrgMember(false), api.GetOrgSecretList)
 					org.Use(session.MustOrgMember(true))
 					org.DELETE("", session.MustAdmin(), api.DeleteOrg)
 
-					org.GET("/secrets", api.GetOrgSecretList)
 					org.POST("/secrets", api.PostOrgSecret)
 					org.GET("/secrets/:secret", api.GetOrgSecret)
 					org.PATCH("/secrets/:secret", api.PatchOrgSecret)


Levy-Tal:
Thanks for the feedback!

Regarding the concern that everyone can now see the secrets — the change introduces two specific behaviors:

Organization members can list global secrets:
If someone is an organization member, they are by definition a Woodpecker user, so it makes sense for them to be able to list global secrets. These are intended to be available to all users.
Organization members can list organization secrets:
If you're part of the organization, you should be able to list its organization-level secrets — but not secrets from other repositories in the org.
I've verified this: users do not see secrets from other repos unless they are members of those repos.
Also, if a user has write access to a repo, they can already run pipelines and inspect environment variables — including the secret names — so the exposure level doesn't change with this PR.

Importantly, a regular organization member should not be blocked with an error just for visiting the Secrets tab in a repo they have access to. That creates confusion and a poor UX, especially when they are legitimately allowed to use these secrets in their pipelines.

And if an organization admin wants to limit access further, they don’t have to use org-level secrets. They can rely solely on per-repo secrets for finer control. But having the option to define shared secrets at the org level is a useful feature, especially when multiple repos need access to the same credentials — and it should be expected that org members (who can use these secrets) should also be able to see which ones are available.

As for the suggestion to allow access only to users with admin rights on any of the org’s repos — I think that might create confusion. In many real-world setups, all developers have write access to certain repos. Requiring admin permissions on a completely different repo just to see shared secrets doesn’t seem intuitive or necessary, especially when write access already gives them the ability to use those secrets in pipelines.

Let me know if you'd like to explore a middle-ground solution, but I believe this change aligns well with how secrets are used and scoped today.

Levy-Tal:
Thanks — I see where the confusion is coming from.

Just to clarify: in session.MustOrgMember(false), the false doesn't mean "non-member allowed" — it controls whether the user needs to be an admin or not. So the function still requires the user to be an org member, but not necessarily an org admin.
Reference: [session.MustOrgMember source](https://github.com/woodpecker-ci/woodpecker/blob/main/server/router/middleware/session/user.go#L123)

As for the case of external contributors with admin access to a specific repo — that’s definitely an interesting edge case and probably worth a broader discussion.

Could you help clarify why you think admin access to a repo should be required to view org secrets, rather than just write access? If someone can trigger pipelines with write access (and therefore use org secrets), they effectively already have visibility into what secrets are configured (at least by name). So from a practical security standpoint, requiring admin access for visibility seems a bit inconsistent with actual usage.

Happy to collaborate on finding a balance that keeps things secure without overcomplicating access.

Levy-Tal:
Got it I understand what you mean now.

Just to add a related point: currently, in the UI, if I’m not an admin on any repo, I can’t see the global secrets. But I can fetch them via the API, which makes sense — I can’t edit them, but I can list them, because they are global and I’m a regular Woodpecker user.

I confirmed this with a simple curl command with a user with no admin permissions:

`curl -i ${WOODPECKER_SERVER}/api/secrets -H "Authorization: Bearer ${WOODPECKER_TOKEN}"`

Relevant code reference:
https://github.com/woodpecker-ci/woodpecker/blob/50687387fbd47715a10dd5b40f4fd88947f4a2d8/server/router/api.go#L188

This same logic also applies to organization secrets — if you're a member of the organization, you can already fetch them through the API, even if the UI hides them. So the UI change here is just making that capability visible and consistent, not introducing new exposure.

Levy-Tal:
Today, only organization admins can list the organization secrets, so they are unusable for regular organization users (repo admins or not). They just don't know that they exist.

Levy-Tal:
Do you want me to implement something like this?

`org.GET("/secrets", session.MustOrgMember(false), session.MustRepoAdminInOrg(), api.GetOrgSecretList)`

However, changing the organization api path (/orgs) to be repo-based may complicate the current permission model. 

Let me know how you'd like to proceed!

server/router/middleware/session/user.go
```diff
@@ -166,3 +166,48 @@ func MustOrgMember(admin bool) gin.HandlerFunc {
...
+		repos, err := store.FromContext(c).OrgRepoList(org, nil)
+		if err != nil {
+			log.Error().Err(err).Msg("Failed to fetch repositories for organization")
+			c.String(http.StatusInternalServerError, "Failed to fetch repositories")
+			c.Abort()
+			return
+		}
+
+		// Check if the user has admin permissions on any repository in the organization
+		for _, repo := range repos {
```
lafriks, line 197: This could result in hundreds of database queries on large instances

Levy-Tal: I agree, repo-based queries in the /org path are unnecessary. It brings more complications, especially since users with only write access (and not admin) can already trigger pipelines and indirectly use the secrets. 

The previous approach ensures that all organization members—who are expected to use these secrets—can list them via the API, even if they're not repo admins.

@qwerty287 is that what you ment when you said it should be repo-based?

Levy-Tal:
Let me know what you want to do, regarding this issue. Any suggestions will be appreciated. Thanks.

lafriks:
Imho there should be a single request that would return deduplicated list with type what kind of secret that is (so repo secret endpoint would also return org/global secrets in single requests and rights would be checked as they are now)

Levy-Tal:
Sounds like a good idea. I'm not very familiar with Vue, so perhaps we could close this pull request and open an issue instead. It would be great if someone with more frontend experience could pick it up.

PR #5098, opened 2025-04-17
Slugify plugin names used for urls

xoxys:
URLs are pretty ugly now and look even worse if they are used for links somewhere, as it contains a lot of URL encoding.

grafik

This PR introduces slugify to create proper URLs and also handles special chars and Unicode symbols properly. This change will break existing links to the plugin docs but I would accept this trade-off and fix them afterwards.


anbraten (approved):

xoxys:
You are too fast and have not verified it in the preview :D the card links are broken. Ill fix it in a follow up PR

PR #5102, opened 2025-04-17
Fix example value for `WOODPECKER_GRPC_ADDR` in autoscaler docs

gsaslis:
In case `https://` is included, I get an error: 

`could not create new auth interceptor: rpc error: code = Unavailable desc = dns: A record lookup error: lookup https on 127.0.0.11:53: server misbehaving`

it is trying to lookup `https` as the host.


docs/versioned_docs/version-3.5/30-administration/10-configuration/40-autoscaler.md
```diff
@@ -27,7 +27,7 @@ services:
       - WOODPECKER_MIN_AGENTS=0
       - WOODPECKER_MAX_AGENTS=3
       - WOODPECKER_WORKFLOWS_PER_AGENT=2 # the number of workflows each agent can run at the same time
-      - WOODPECKER_GRPC_ADDR=https://grpc.your-woodpecker-server.tld # the grpc address of your woodpecker server, publicly accessible from the agents
+      - WOODPECKER_GRPC_ADDR=grpc.your-woodpecker-server.tld # the grpc address of your woodpecker server, publicly accessible from the agents. See https://woodpecker-ci.org/docs/administration/configuration/server#caddy for an example of how to expose. Do not include "https://" in the value.
```
anbraten, line 30:
```suggestion
      - WOODPECKER_GRPC_ADDR=grpc.your-woodpecker-server.tld # the grpc address of your woodpecker server, publicly accessible from the agents. See https://woodpecker-ci.org/docs/administration/configuration/server#caddy for an example of how to expose. Do not include `https://` in the value.
```

anbraten (approved):

xoxys (changes requested):
Could you please fix this in `next` docs as well? https://github.com/woodpecker-ci/woodpecker/blob/main/docs/docs/30-administration/10-configuration/40-autoscaler.md?plain=1

docs/versioned_docs/version-3.5/30-administration/10-configuration/40-autoscaler.md
xoxys: This is a comment in a yaml codeblock. Using backticks here doesn't make sense IMO

xoxys (approved):

anbraten (approved):

PR #5104, opened 2025-04-18
Set dynamic page titles

### LINKED ISSUE #5051: Update the page title when users navigate through the web UI

At the moment the page title is hardcoded to "Woodpecker" and it seems it is never updated when users navigate through the app. As a result the browser's history ui looks something like this:

Image

Which isn't really that useful.

### Suggested solution

Ideally each page navigation updates the page title, so the history entries (and bookmarks) are more descriptive.

We usually use a structure like this in our apps, but feel free to use whatever makes sense:

| Page |  Page title  |
|---|---|
|  Repository overview  | `Repositories \| Woodpecker` |
| Repo activity tab | `[repo name] \| Woodpecker` |
| Repo general settings | `General \| Settings \| [repo name] \| Woodpecker` |

### Alternative

_No response_

### Additional context

Woodpecker 3.4.0

### Validations

lukashass:
Closes #5051 

### TODO

- [x] set titles of all pages

### Alternative

I considered setting the page title in vue-router, but decided against it since we would have to move the loading of repository and other relevant data into the router to have the names, etc. available there.


lukashass:
Any suggestions before I continue with all the other pages?

web/src/compositions/useWPTitle.ts
```diff
@@ -0,0 +1,7 @@
+import { useTitle } from "@vueuse/core"
+import type { Ref } from "vue";
+import { computed } from "vue"
+
+export function useWPTitle(elements: Ref<string[]>) {
```
anbraten, line 5: Might be more readable to use sth like: 


```suggestion
export function useWPTitle(...elements: Ref<string>) {
```

web/src/views/Repos.vue
```diff
@@ -64,6 +66,8 @@ const reposLastActivity = computed(() => sortReposByLastActivity(searchedRepos.v
 onMounted(async () => {
   await repoStore.loadRepos();
 });
+const { t } = useI18n()
+useWPTitle(computed(() => [t('repositories.title'),]));
```
anbraten, line 70: Can't think of a better way to do this, so lgtm.

anbraten (approved):

web/src/views/admin/AdminSettingsWrapper.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <Scaffold enable-tabs>
     <template #title>
-      {{ $t('settings') }}
+      {{ $t('admin.settings.settings') }}
```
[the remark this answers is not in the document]
lukashass: There is, but this new key now translates to "Admin Settings" which I think is useful to distinguish the different settings pages.

PR #5127, opened 2025-04-26
Fix args docs for admin secrets

qwerty287:


xoxys:
Why do we need to remove `ArgsUsage` from the subcommands?

xoxys:
Because its for global secrets.

xoxys (approved):

PR #5129, opened 2025-04-26
Explain tasks

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/discussions/5064


xoxys (approved):
Same here, maybe worth it to add to current version docs as well

xoxys (approved):

PR #5156, opened 2025-05-07
Add Bluesky post plugin

qwerty287:


xoxys:
Im ok to add it but the mastodon plugin is broken since months now and this one is unmaintained since the initial commit... I would really like to see someone taking care, especially for "special" plugins written in typescript. If now I would reject adding more plugins as official plugins to the CB org in the future.

xoxys (approved):

anbraten:
Added extra logs to the new plugin versions and updated our tokens.

PR #5162, opened 2025-05-09
fix(bitbucketdc): build status not reported on PR builds

### LINKED ISSUE #5158: PR build status does not get updated on BitBucket Datacenter
### Component

server

### Describe the bug

When build finishes(either successfully or with failure) the status doesn't get updated, even though the commit status of the last commit on the PR is updated just fine.

I was able to trace back this bug to this PR: https://github.com/woodpecker-ci/woodpecker/pull/4724

Now I see the need for the `Ref` to be set when updating the build status but it turns out that `pipeline.Ref` doesn't have a correct value. In my case it would set only the ref as follows: `refs/pull-requests/<pullRequestID>/from`. I was able to find that `Ref` gets set [here](https://github.com/woodpecker-ci/woodpecker/blob/main/server/forge/bitbucketdatacenter/convert.go#L122) and I believe this is wrong, it's not a "valid" git ref and if ref is set to an invalid value BitBucket will not set the build status for that PR(https://community.atlassian.com/forums/Bitbucket-questions/Re-Builds-not-appearing-in-pull-requests/qaq-p/1805991/comment-id/65864#M65864).

I suggest that we change the code from above to:
```go
Ref:       fmt.Sprintf("refs/heads/%s", ev.PullRequest.Source.DisplayID),
```
I believe this would fix the issue, and I would like to work on it if possible.

Adding: @Levy-Tal as he worked on PR mentioned above and @langecode because he is the main contributor behind the BitBucket Datacenter forge integration.

### Steps to reproduce

1. Open a PR with woodpecker pipelines on BitBucket Datacenter
2. Trigger a build on the last commit of that PR
3. Wait for the build to finish
4. Observe that there is no status on the PR itself

### Expected behavior

There should be a build status on the PR.

### System Info

```shell
Woodpecker version: v3.5.2
Helm chart version: 3.1.0
```

### Additional context

_No response_

### Validations

rruzicic:
Fixes #5158

### DIFF
--- a/.github/release_template.md
+++ b/.github/release_template.md
@@ -29,5 +29,6 @@
 - [ ] Release the Helm Chart. If renovate has not created the upgrade PR already, manually trigger it from the Dependency Dashboard.
 - [ ] Announce release in relevant chats and on social media platforms
   - [ ] Mastodon (check if already posted from the release pipeline)
+  - [ ] Bluesky (check if already posted from the release pipeline)
   - [ ] Matrix
   - [ ] Twitter
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -10,7 +10,7 @@ tasks:
       WOODPECKER_BACKEND_DOCKER_NETWORK: ci_default
     init: |
       # renovate: datasource=golang-version depName=golang
-      GO_VERSION=1.24.2
+      GO_VERSION=1.24.3
       rm -rf ~/go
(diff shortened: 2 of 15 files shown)


rruzicic:
Sorry, at first I wasn't really able to run the tests but I figured it now and the last commit should fix the tests.

PR #5178, opened 2025-05-14
fix(bitbucketdc): build status gets incorrectly reported on multi workflow builds

### LINKED ISSUE #5177: Statuses get incorrectly reported for multi workflow pipelines on BitBucket
### Component

server

### Describe the bug

When a pipeline with multiple workflows starts a correct status is sent at first BitBucket correctly displays that build is in progress. When the build finishes however, only one build is displayed as finished but the other one is still in progress marking the whole commit as in progress(this is internal BitBucket logic). 

Here is a pipeline that has two workflows:

Image

And here is the outcome on BitBucket:

Image

I asked @langecode if he can take a look and we realized that there is a problem in BitBucket forge implementation of `Status` function. It looks like that for each call of `Status` function we send a status of the whole pipeline when we should've sent a status of individual workflows. I.e. change status of https://github.com/woodpecker-ci/woodpecker/blob/main/server/forge/bitbucketdatacenter/bitbucketdatacenter.go#L314

To:
```go
State:       convertStatus(workflow.State)
```

### Steps to reproduce

1. Run woodpecker with BitBucket Datacenter forge
2. Create a pipeline that has multiple workflows
3. Let the pipeline finish on woodpecker
4. Observe the build statuses on the commit in BitBucket

### Expected behavior

_No response_

### System Info

```shell
Woodpecker version: v3.5.2
BitBucket Datacenter version: v8.18.1
```

### Additional context

_No response_

### Validations

rruzicic:
Fixes #5177


6543 (approved):

langecode:
Good finding - nice to see it fixed 👍

Issue #5195, opened 2025-05-23
Kubernetes backend leaks secrets to Kubernetes apiserver audit logs


hhamalai:
### Component

agent

### Describe the bug

Currently the defined repository / organization secrets are injected on frontend compile phase as environment variables, used by the backend. In case of Kubernetes backend, these environment variables are logged on Kubernetes apiserver audit logs, often monitored by cluster administrators or SOC experts who should not learn these secrets unnecessarily. 

### Steps to reproduce

1. Install woodpecker server & agent, with kubernetes backend
2. Define a workflow with a repository secret containing a specific secret value, with a step using the secret
3. Execute the workflow, ensure the step using the secret is executed
4. Check cluster environment logs, e.g. apiserver audit logs for the specific secret value

### Expected behavior

Secret values should not be leaked to infrastructure logs; kubernetes secrets could be used to provide steps with specific secrets. 

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"dev"}
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

PR #5196, opened 2025-05-23
kubernetes: prevent secrets from leaking to api-server logs

hhamalai:
In order to prevent secrets from leaking to Kubernetes api server logs, create step secrets as Kubernetes secret resource when needed and refer to those secrets from environment variable with valueFrom & secretKeyRef.

As the woodpecker secrets might be used in different steps with different environment variable names, and as currently the secret injection happens on the frontend compile step, another object "secretMapping" is passed in the step structure, used to filter environment variables of the step for secrets.

The end result is that the environment variable containing the secret, previously looked like:
```
    - name: VERY_SECRET
      value: v3ry_s3cr3t
```   
becomes:
```
    - name: VERY_SECRET
      valueFrom:
        secretKeyRef:
          key: VERY_SECRET
          name: wp-01jvxvr94ww1c1nrxah01hkt9g-step-secret
```   
and there's a secret during the step execution:
```
apiVersion: v1
data:
  VERY_SECRET: v3ry_s3cr3t #as a base64 encoded value
kind: Secret
metadata:
  name: wp-01jvxvr94ww1c1nrxah01hkt9g-step-secret
  namespace: woodpecker
  ...
type: Opaque
```

Will not affect functionality of other backends (local, docker), though the secret mapping metadata is also passed there.

Fixes: #5195


pat-s:
Please avoid force pushes in the PR.

pat-s (approved):
LGTM

pat-s:
@woodpecker-ci/maintainers @hhamalai 

I had to revert this PR in Crow as it caused the following issues which I could pin down to this commit via bisecting:

- List strutcs as the `logins` struct in the buildx plugin cannot be parsed anymore and fail with `ERR execution failed error="could not unmarshal logins: invalid character 'D' looking for beginning of value"`
- Also buildx plugin: The `build_args` settings value is not parsed anymore when there is a map containing `from_secret` in there and instead ignored silently. 

I'd strongly advise to check these cases and adjust as necessary (I didn't had time for that yet and only reverted the PR) before including this in a new release.

xoxys:
As you merged this PR can you create a revert PR for WP as well?

PR #5199, opened 2025-05-24
fix: docker compose docs

shelldandy:
add proper port to the `WOODPECKER_SERVER` env variable in line with the rest of the docker config from the doc


shelldandy:
nvm i now understand you were talking about the other endpoint haha

PR #5207, opened 2025-05-27
Add direnv

6543:
https://direnv.net/

just some smal devx impruvement ...


6543:
if you have envdir enabled it will setup the dev enfironment (in this case via nix dev shell) on entering the folder (shell) automatically ... if you dont it just gets ignored ... so its an conviniense thing not a must have

6543:
it is shell / distro independent ...

https://direnv.net/#prerequisites

qwerty287:
I know, but you wrote `use flake` in the `.envrc`. And this is definitely _not_ shell independent.

pat-s (changes requested):
I've been using `direnv` for a long time already and I am quite confused about this PR.

1. `.envrc` should always be in `.gitignore` as it will contain sensitive secrets which are loaded from it
2. Yes, there should (only) be env vars in `.envrc` AFAIK the tool

Maybe `direnv` can do even more but I am sure non-nix users will encounter issues if such statements are in there. Note that adding this will add a message prompt for any direnv user to `direnv allow` this directory after it has found an `.envrc` file. This will then likely fail or init something not all users want.

In general: There should be a general policy on how to proceed with all of the nix-specific PRs in the future. Adding general dev tool config files isn't a problem per se, but they should either be static or update themselves through renovate (if needed).

6543:
->  #5235

PR #5227, opened 2025-05-31
Add pipeline author and avatar env vars

qwerty287:
https://github.com/woodpecker-ci/woodpecker/pull/4626#discussion_r2116850787


pat-s (commented):
What use case does the avatar have?

While it's surely not harming, it extends the list and makes it bigger without a real use case (?).

anbraten (approved):

qwerty287:
Can we get a decision here? I don't care tbh. @6543 what do you say? You originally suggested this.

xoxys (approved):

xoxys:
Also not sure about the use case but its just an env var and it doesnt increase the code complexity so why not.

pipeline/frontend/metadata/types.go
```diff
@@ -55,6 +55,8 @@ type (
 		Commit     Commit `json:"commit,omitempty"`
 		Parent     int64  `json:"parent,omitempty"`
 		Cron       string `json:"cron,omitempty"`
+		Author     string `json:"author,omitempty"`
```
anbraten, line 58: This change needs a grpc version bump.

anbraten (commented):
I saw it being used by notification plugins in drone to show the author and their avatar

pipeline/frontend/metadata/types.go
qwerty287: Really? Do we pass this data to the agent? I think just the final env vars are passed

pat-s (approved):

pipeline/frontend/metadata/types.go
qwerty287: Checked again, seems to be not used by RPC.

xoxys (approved):

nmapx:
Have you noticed it overrides the commit author? I don't know if that's how it supposed to work, looks weird.
If you trigger the pipeline manually on a commit that you did not pushed it's gonna say that "prev commit author" is the person that actually pushed the commit woodpecker is actually processing.

PR #5251, opened 2025-06-12
Fix doc links in agent settings

mmatous:
Hi, this small PR will fix broken links in https://woodpecker.example.com/admin/agents submenus.

### DIFF
--- a/web/src/components/agent/AgentForm.vue
+++ b/web/src/components/agent/AgentForm.vue
@@ -41,7 +41,7 @@
       <InputField
         v-slot="{ id }"
         :label="$t('admin.settings.agents.capacity.capacity')"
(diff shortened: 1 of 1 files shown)


qwerty287:
Thanks! The anchor is still wrong. The link is https://woodpecker-ci.org/docs/administration/configuration/agent#max_workflows (with just `max_workflows` as anchor)

anbraten:
Are we going to use the latest links instead of next ones in general?

mmatous:
Anchor fixed. The links led to /next bc the originals were that way, but @anbraten is correct. Every link in the UI leads to latest, so I modified these accordingly.

anbraten (approved):

Issue #3070, opened 2023-12-29
Kubernetes: smart(er) agent affinity


pat-s:
## Scenario

- Mixed arm/amd k8s cluster
- One agent for arm, one agent for amd

When starting a build with 4 concurrent workflows, two go to the arm agent and two to the amd agent (assuming an empty queue).

The issue now is that within the cluster there are way more free resources on nodes of a specific architecture. This result in

- two builds getting scheduled on nodes which start instantly
- The other two builds cause the autoscaler to do work and spin up new nodes of the architecture, which is low on resources

In fact, though, all resource requests would have been satisfiable with existing resources of *one* node architecture in the cluster. The issue is that by having multi-arch agents in WP, 50% of the scheduling logic is taken away from the cluster and hence causing unnecessary resources to be provisioned.

## Options

I wonder how this can be done smarter. I see the following options:

- Give the scheduling entirely to the cluster when using the k8s backend and don't do a prior agent assignment (not sure if this would even be possible)
- Somehow infer free resources on the cluster nodes of the respective agent and do a standalone scheduling (e.g. in the case above this would mean 3/1 or 4/0 in favor of one architecture).
- Honor predefined matrix variables/labels which indicate an architecture preference but don't limit the job to this one (e.g. sometimes `buildx` is used to build an amd64 only image -> it doesn't make much sense to let this be build by an arm node)

NB: I am aware that I could use a `nodeSelector` for manual arch assignment, however I would actually prefer to not do this as the resource availability changes dynamically. And I would like to be able to sent jobs to whatever architecture (if feasible and space exists). 

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

PR #5266, opened 2025-06-20
Add Agent-level Tolerations setting

scottshotgg:
This PR adds the ability to specify tolerations at the agent level similar to annotations, labels, node selector, etc. I know this ability is available from within a pipeline but as an administrator I should be able to set this at a global level as well as prevent any pipeline from overriding this.
In similar vein to how annotations are done, I have introduced two new env variables and updated the documentation to match:
- `WOODPECKER_BACKEND_K8S_POD_TOLERATIONS`
- `WOODPECKER_BACKEND_K8S_POD_TOLERATIONS_ALLOW_FROM_STEP`

I initially started this feature because I have a "runner" node in my k8s cluster specifically for jobs like this (crons, ci/cd, etc) so that the core compute on the worker nodes isn't effected by random variadic scaling patterns. I felt that the ability to define these tolerations and node selectors at the individual step level was really nice and could come in handy, but the inability to define them at a global level was _necessary_; I needed a simple way to say that every pipeline should be scheduled to a particular node, which `WOODPECKER_BACKEND_K8S_POD_NODE_SELECTOR` was good for - however it is a bit underwhelming because you are then left with a pod that has been told to schedule onto a node for which it has no toleration for the nodes taint unless you include this in _every_ pipeline, _every_ step, etc.

I think this could be a useful feature that others could benefit from. Overall I found it simple to implement - thanks for the easy code to work in, and thanks for Woodpecker!


qwerty287:
Hello and sorry that it took a long for us to answer.
This looks good with my very limited k8s knowledge, but you added the docs also for the old versions. Only the files in `docs/docs` should get changed.

scottshotgg:
Ahh okay, so don't change the `versioned` docs - that makes sense. I'll fix that up

pat-s:
With proper rules applied, this might be a solution to https://github.com/woodpecker-ci/woodpecker/issues/3070.

scottshotgg:
@pat-s 
I didn't thoroughly read through your issue, but if your goal is to have a builder for arm based images hosted on an arm node, and a corollary one on amd - it could definitely be a solution for that.

If you need/want an example or two I can try to help you out.

scottshotgg:
@qwerty287 
I've restored the `docs/versioned_docs` folder back to main. I would really like to merge this in soon if that's possible. Right now I am running this branch but would like to run the official image, etc

qwerty287:
This looks good to me, but could you add the backend options parameter to the schema? https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/yaml/linter/schema/schema.json

And I'd like to have a review from someone else, I don't have much knowledge of k8s. @woodpecker-ci/maintainers

scottshotgg:
@qwerty287 
Done!

pat-s:
> I didn't thoroughly read through your issue, but if your goal is to have a builder for arm based images hosted on an arm node, and a corollary one on amd - it could definitely be a solution for that.

No, that's not the scope of the linked issue at all.

docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
```diff
@@ -371,6 +371,24 @@ Determines if Pod annotations can be defined from a step's backend options.
...
+### BACKEND_K8S_POD_TOLERATIONS
+
+- Name: `WOODPECKER_BACKEND_K8S_POD_TOLERATIONS`
+- Default: none
+
+Additional tolerations to apply to worker Pods. Must be a YAML object, e.g. `[{"effect":"NoSchedule","key":"jobs","operator":"Exists"}]`.
+
+---
+
+### BACKEND_K8S_POD_TOLERATIONS_ALLOW_FROM_STEP
```
pat-s, line 392: I'd name this `BACKEND_K8S_POD_TOLERATIONS_ALLOW_OVERRIDE_FROM_STEP`

docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
```diff
@@ -371,6 +371,24 @@ Determines if Pod annotations can be defined from a step's backend options.
...
+
+- Name: `WOODPECKER_BACKEND_K8S_POD_TOLERATIONS_ALLOW_FROM_STEP`
+- Default: `false`
```
pat-s: I think this should always be `true`. When `false`, the this would be a breaking change as `tolerations` set in the `backend_options` would not have any effect anymore.
While I am OK with adding a hard admin-override one for this (if you need it for some use case), I really don't think the default should be that user-level don't take precedence.

docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
qwerty287: It's the same naming for other options: https://woodpecker-ci.org/docs/administration/configuration/backends/kubernetes#backend_k8s_pod_labels_allow_from_step

docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
scottshotgg: Ok I can do that. If you have tests for the annotations that you could point me to and I can replicate for tolerations that would be great.

scottshotgg: Done

scottshotgg:
@pat-s 
Do you have any tests that you could point me to? I'd preferably like to replicate the tests for annotations since this feature is 99% the same as that one with the exception of the default value

qwerty287:
I guess this test is the one you're looking for:

https://github.com/woodpecker-ci/woodpecker/blob/5de2d26b8de3b61ee3933e7c64f3582d65afb572/pipeline/backend/kubernetes/pod_test.go#L198

xoxys (approved):
LGTM after tests were added.

qwerty287:
@scottshotgg do you plan to add the tests?

pipeline/backend/kubernetes/flags.go
```diff
@@ -73,12 +73,24 @@ var Flags = []cli.Flag{
...
 		Sources: cli.EnvVars("WOODPECKER_BACKEND_K8S_POD_ANNOTATIONS_ALLOW_FROM_STEP"),
 		Name:    "backend-k8s-pod-annotations-allow-from-step",
 		Usage:   "whether to allow using annotations from step's backend options",
 		Value:   false,
 	},
+	&cli.BoolFlag{
+		Sources: cli.EnvVars("WOODPECKER_BACKEND_K8S_POD_TOLERATIONS_ALLOW_FROM_STEP"),
+		Name:    "backend-k8s-pod-tolerations-allow-from-step",
+		Usage:   "whether to allow using tolerations from step's backend options",
+		Value:   true,
```
6543, line 92: it this has security implcations it should default to false ...

qwerty287: No: https://github.com/woodpecker-ci/woodpecker/pull/5266#discussion_r2183041989

scottshotgg:
I will try to work on the tests at some point

marcusramberg:
I'm interested in getting this merged.  Here's a patch to add tests for tolerations (as well as fixing a failing existing test) 

https://gist.github.com/marcusramberg/c18a22a191efba8810151c1d2d7fbb34

scottshotgg:
Thank you @marcusramberg for implementing those tests - I've been pretty busy as of late.

@xoxys @qwerty287
Are we g2g now?

xoxys:
@scottshotgg @marcusramberg Tests failed, can you take a look?

PR #5275, opened 2025-06-27
Use pipeline number in title

qwerty287:


xoxys:
Whats the motivation?

qwerty287:
The ID is the database id. It is not the number that is counted per repo, it is counted globally. We never show the id anywhere else.

Example here:

xoxys:
Thanks 👍

xoxys (approved):

6543:
-> updated pull description with clarification :) - thx for the fix

Issue #5031, opened 2025-03-30
Segmentation fault on woodpecker-agent startup


Nothing4You:
### Component

agent

### Describe the bug

Since I'm on 3.4.0, the agent runs into a segmentation fault on startup.

I just updated from woodpecker 2.7.0 to 3.4.0.
I'm using [picus](https://github.com/windsource/picus) 0.7.2 to deploy VMs for agents as needed.
It does not appear to be a picus issue: https://github.com/windsource/picus/issues/29

### Steps to reproduce

1. Create woodpecker agent with the compose config attached below
2. Watch it crash

My woodpecker server has pending jobs for the agent in case that's relevant.

### Expected behavior

no crash

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.4.0"}



services:

  woodpecker-agent:
    image: woodpeckerci/woodpecker-agent:v3.4.0-alpine
    command: agent
    restart: always
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      - WOODPECKER_GRPC_SECURE=true
      - WOODPECKER_BACKEND=docker
      - WOODPECKER_SERVER=[...]:443
      - WOODPECKER_FILTER_LABELS=platform=linux/amd64,backend=docker,repo=*,org-id=*
      - WOODPECKER_AGENT_SECRET=[...]


Woodpecker runs behind nginx.
```

### Additional context

```
2025-03-30T16:28:30.809785618Z {"level":"info","time":"2025-03-30T16:28:30Z","message":"log level: info"}
2025-03-30T16:28:31.036192576Z {"level":"info","time":"2025-03-30T16:28:31Z","message":"starting Woodpecker agent with version '3.4.0' and backend 'docker' using platform 'linux/amd64' running up to 1 pipelines in parallel"}
2025-03-30T16:28:31.676659905Z panic: runtime error: invalid memory address or nil pointer dereference
2025-03-30T16:28:31.676698591Z 	panic: runtime error: invalid memory address or nil pointer dereference
2025-03-30T16:28:31.676703160Z [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa7745f]
2025-03-30T16:28:31.676706736Z
2025-03-30T16:28:31.676710003Z goroutine 66 [running]:
[...]

PR #5299, opened 2025-07-05
Simplify backend types

qwerty287:
prevent crashes like https://github.com/woodpecker-ci/woodpecker/issues/5031

fix https://github.com/woodpecker-ci/woodpecker/issues/5031


6543 (changes requested):
nice cleanup!

as backend types changed the grpc version needs an bump

6543:
else lgtm

qwerty287:
Done. I guess however we could also just drop that version if possible. It is not supported to use different agent/server versions anyways.

6543:
it was introduced for people who run the dev branch to have an clear error to debug as that version field do not changes its size and position at the binary level ... so you get no strange stack traces but clear error messages

6543 (approved):
thx ❤️

6543:
@qwerty287 i personally would not get rid of that version field just for the sace of less issues from users. but if you want to automate it, you could change that field to string and pass through the exact woodpecker commit ...

but that would make the version check way more strict ...

PR #5305, opened 2025-07-05
Prevent secrets from leaking to Kubernetes API Server logs

hhamalai:
The previous attempt was lacking proper handling for secrets used in complex types, which lead to issues as mentioned by @pat-s with e.g. [woodpeckerci/plugin-docker-buildx](https://hub.docker.com/r/woodpeckerci/plugin-docker-buildx)  plugin but also applied to any use case having anything but the secret value alone in environment variable's value. 

With the [fix](f15cef5b7ed18a892bb776dcce741ab2013659aa), instead of just the referenced secret, the whole environment variable value is stored in the SecretMapping struct carried to agent, which (in case of kubernetes backend) can setup the pod with secretFrom references to step secret.

original commit: 41a456c7614e287565996523b6b73b68d39bf078
original pr: https://github.com/woodpecker-ci/woodpecker/pull/5196#issuecomment-3031175946

I've also tested that the buildx plugin is now running with `logins` and `build_args` settings, with a usage example from: https://woodpecker-ci.org/plugins/docker-buildx:
```
+ /usr/local/bin/dockerd --data-root /var/lib/docker --host=unix:///var/run/docker.sock --dns 10.244.0.102
Logging in with username 'harrmi' to registry 'https://index.docker.io/v1/'Detected registry credentials
+ /usr/local/bin/docker version
Client:
...
 OS/Arch:           linux/arm64
...
+ /usr/local/bin/docker buildx create --name builder --use --config /tmp/buildkit.toml
builder
...
+ /usr/local/bin/docker buildx build --rm=true -f Dockerfile . --pull=true --build-arg ARCH=Linux_ARM64 --build-arg VER=1000 --build-arg DOCKER_IMAGE_CREATED=2025-07-05T11:40:34Z --build-arg SOURCE_DATE_EPOCH=1751714283 --output type=oci,dest=hhamalai-hello.tar -t codeberg.org/hhamalai/hello:latest --label org.opencontainers.image.created=2025-07-05T11:40:33Z 
...
```

### DIFF
--- a/docs/pnpm-lock.yaml
+++ b/docs/pnpm-lock.yaml
@@ -10,19 +10,19 @@ importers:
     dependencies:
       '@docusaurus/core':
         specifier: ^3.8.1
-        version: 3.8.1(@docusaurus/faster@3.8.1(@docusaurus/types@3.8.1(@swc/core@1.12.9)(acorn@8.15.0)(react-dom@19.1.0(react@19.1.0))(react@19.1.0)))(@mdx-js/react@3.1.0(@types/react@19.1.8)(react@19.1.0))(@rspack/core@1.4.4)(@swc/core@1.12.9)(acorn@8.15.0)(debug@4.4.1)(react-dom@19.1.0(react@19.1.0))(react@19.1.0)(typescript@5.8.3)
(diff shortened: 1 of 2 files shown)


xoxys (approved):
Thanks for your work and the provided fix. LGTM

PR #5309, opened 2025-07-09
feat(k8s): k8s namespace per organization

henkka:
# Problem
Our organization requires the ability to isolate different teams (organizations in Woodpecker CI) into separate Kubernetes namespaces within our CI cluster. This enables better hardening for additional features that we're supplementing Woodpecker CI with for our teams, providing stronger security boundaries between team workloads.

# Solution
This PR introduces optional Kubernetes namespace isolation functionality behind a feature flag. Once enabled, each teams CI workloads will be deployed to their dedicated namespace, providing:

- **Enhanced Security**: Stronger isolation between team workloads
- **Better Hardening**: Improved security posture for supplemental features we're adding to Woodpecker CI
- **Resource Management**: Better control over resource allocation per team
- **Compliance**: Meets organizational security requirements for multi-tenant environments

# Implementation Details
- Feature is disabled by default to maintain backward compatibility
- Existing behavior remains unchanged when feature flag is not enabled

# Status
This is currently a **proposal/RFC** to gather feedback from maintainers on the approach and feasibility of this feature. The feature has been successfully tested once `woodpecker-agent` ServiceAccount is given the following additional permissions

<details>
  <summary>Additional Kubernetes permissions</summary>
  
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: woodpecker-agent
rules:
- apiGroups:
  - ""
  resources:
  - persistentvolumeclaims
  - services
  - secrets
  verbs:
  - create
  - delete
- apiGroups:
  - ""
  resources:
  - pods
  verbs:
  - watch
  - create
  - delete
  - get
  - list
- apiGroups:
  - ""
  resources:
  - pods/log
  verbs:
  - get
- apiGroups:
  - ""
  resources:
  - namespaces
  verbs:
  - get
  - list
  - create
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: woodpecker-agent
subjects:
- kind: ServiceAccount
  name: woodpecker-agent
  namespace: woodpecker
roleRef:
  kind: ClusterRole
  name: woodpecker-agent
  apiGroup: rbac.authorization.k8s.io
```
  
</details>

# Next Steps
If maintainers approve this direction, we will:
1. Create a corresponding PR for the Woodpecker CI Helm Chart repo to support this feature flag
2. Add documentation and tests

We're happy to iterate on this based on your feedback.

### DIFF
--- a/pipeline/backend/kubernetes/kubernetes.go
+++ b/pipeline/backend/kubernetes/kubernetes.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"runtime"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/rs/zerolog/log"
@@ -56,6 +57,7 @@ type kube struct {
 
 type config struct {
 	Namespace                   string
+	EnableNamespacePerOrg       bool
 	StorageClass                string
 	VolumeSize                  string
 	StorageRwx                  bool
@@ -68,6 +70,14 @@ type config struct {
 	SecurityContext             SecurityContextConfig
 	NativeSecretsAllowFromStep  bool
 }
+
+func (c *config) K8sNamespace(owner string) string {
+	if c.EnableNamespacePerOrg {
+		return strings.ToLower(fmt.Sprintf("%s-%s", c.Namespace, owner))
+	}
+	return c.Namespace
+}
+
 type SecurityContextConfig struct {
 	RunAsNonRoot bool
 	FSGroup      *int64
@@ -88,6 +98,7 @@ func configFromCliContext(ctx context.Context) (*config, error) {
 		if c, ok := ctx.Value(types.CliCommand).(*cli.Command); ok {
 			config := config{
 				Namespace:                   c.String("backend-k8s-namespace"),
+				EnableNamespacePerOrg:       c.Bool("backend-k8s-namespace-per-org"),
 				StorageClass:                c.String("backend-k8s-storage-class"),
 				VolumeSize:                  c.String("backend-k8s-volume-size"),
 				StorageRwx:                  c.Bool("backend-k8s-storage-rwx"),
@@ -191,7 +202,16 @@ func (e *kube) getConfig() *config {
 func (e *kube) SetupWorkflow(ctx context.Context, conf *types.Config, taskUUID string) error {
 	log.Trace().Str("taskUUID", taskUUID).Msgf("Setting up Kubernetes primitives")
 
-	_, err := startVolume(ctx, e, conf.Volume)
+	namespace := e.config.K8sNamespace(conf.Stages[0].Steps[0].Owner)
+
+	if e.config.EnableNamespacePerOrg {
+		err := ensureNamespaceExists(ctx, e, namespace)
+		if err != nil {
+			return err
+		}
+	}
+
+	_, err := startVolume(ctx, e, conf.Volume, namespace)
 	if err != nil {
 		return err
 	}
@@ -269,7 +289,7 @@ func (e *kube) WaitStep(ctx context.Context, step *types.Step, taskUUID string) 
 		}
 	}
 
-	si := informers.NewSharedInformerFactoryWithOptions(e.client, defaultResyncDuration, informers.WithNamespace(e.config.Namespace))
+	si := informers.NewSharedInformerFactoryWithOptions(e.client, defaultResyncDuration, informers.WithNamespace(e.config.K8sNamespace(step.Owner)))
 	if _, err := si.Core().V1().Pods().Informer().AddEventHandler(
 		cache.ResourceEventHandlerFuncs{
 			UpdateFunc: podUpdated,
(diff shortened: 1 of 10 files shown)


qwerty287:
Hello and thanks for your contribution.

To me this looks quite nice, but I'm not experienced with k8s so I'd wait for other maintainers to check this out.

Just one point already: this shouldn't use the owner name, but it's ID, because in the future we want to add multi-forge support, so it will be possible to have the same org name from different forges. If the name should be included for readability reasons, it could be prefixed with the ID (e.g. `prefix-5-some-org`).

henkka:
> Hello and thanks for your contribution.
> 
> To me this looks quite nice, but I'm not experienced with k8s so I'd wait for other maintainers to check this out.
> 
> Just one point already: this shouldn't use the owner name, but it's ID, because in the future we want to add multi-forge support, so it will be possible to have the same org name from different forges. If the name should be included for readability reasons, it could be prefixed with the ID (e.g. `prefix-5-some-org`).

Thank you for the quick comments! I modified the pull request to accommodate the future plans in the project with commit id 0eb14221bdbb0421e04d7e1aebdc53ca9b121c13

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -68,6 +71,14 @@ type config struct {
 	SecurityContext             SecurityContextConfig
 	NativeSecretsAllowFromStep  bool
 }
+
+func (c *config) K8sNamespace(orgID int64) string {
```
xoxys: Can we rename this method to `GetNamespace`?

pipeline/backend/kubernetes/namespace.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+
+	"github.com/rs/zerolog/log"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ensureNamespaceExists(ctx context.Context, engine *kube, namespace string) error {
```
xoxys: Can we follow the naming scheme and name this method `mkNamespace`? Would also be nice to add basic tests similar to e.g. https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/backend/kubernetes/volume_test.go

xoxys (commented):
What happens if orgs got removed or the option is getting disabled again? Old org-related namespaces will never get cleaned up and exist forever? Do we want to handle this case?

henkka:
> Old org-related namespaces will never get cleaned up and exist forever? 

Correct! In the current implementation we're just creating the namespace if it doesn't exist.

> Do we want to handle this case?

Personally I wouldn't include logic to cleanup/delete the namespaces automatically, as users might have some other resources in the namespaces that the Woodpecker is not aware of, but I'm open to include this behavior if required

xoxys:
> Personally I wouldn't include logic to cleanup/delete the namespaces automatically, as users might have some other resources in the namespaces that the Woodpecker is not aware of, but I'm open to include this behavior if required

Makes sense to me, but we should at least document this behavior.

pipeline/backend/kubernetes/kubernetes.go
henkka: 👍 https://github.com/woodpecker-ci/woodpecker/pull/5309/commits/905304755478a6869d46ad515796ccd9140b290d

pipeline/backend/kubernetes/namespace.go
henkka: 👍 https://github.com/woodpecker-ci/woodpecker/pull/5309/commits/0f6cb312b6b391e3999b8ccd4843c51dd1ba6443

henkka:
> > Personally I wouldn't include logic to cleanup/delete the namespaces automatically, as users might have some other resources in the namespaces that the Woodpecker is not aware of, but I'm open to include this behavior if required
> 
> Makes sense to me, but we should at least document this behavior.

Good point! I updated the documentation with https://github.com/woodpecker-ci/woodpecker/pull/5309/commits/4bfc131e88f72b52c851d2ae2783f186c234c592

qwerty287:
@henkka could you checkout the linters https://ci.woodpecker-ci.org/repos/3780/pipeline/27524/34

henkka:
> @henkka could you checkout the linters https://ci.woodpecker-ci.org/repos/3780/pipeline/27524/34

 @qwerty287 oops, fixed with https://github.com/woodpecker-ci/woodpecker/pull/5309/commits/adf5850cae67031b37b1db15d9e45dcbc90e85ee

```
➜  woodpecker git:(k8s-namespace-separation) make lint
Running golangci-lint
golangci-lint run
0 issues.
```

PR #5328, opened 2025-07-14
Add and edit additional forges in UI

### LINKED ISSUE #425: Set gitea-client and gitea-secret in web UI

If you start Gitea together with Woodpecker using a docker compose file, there is no way to retrieve the OAuth data from Gitea. And now since #365 has been merged, which is great btw, you would get an obvious error when trying to access woodpecker.

This means that after you installed Gitea completely, you need to get the OAuth client and secret, and change the compose setup to add these two values.

### Suggested solution

It would be nice to set Gitea OAuth2 environment variables on the first start of Woodpecker in the webui, if the enviroment variables are emtpy.

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [X] Read the [docs](https://woodpecker-ci.github.io/docs/intro).

anbraten:
closes #425

image

image

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -537,7 +537,7 @@
   "bitbucket_dc": "Bitbucket Data Center",
   "gitea": "Gitea",
   "forgejo": "Forgejo",
-  "addon": "Addon-Forge",
+  "addon": "Addon",
   "forge_type": "Forge type",
   "oauth_client_id": "OAuth Client ID",
   "oauth_client_secret": "OAuth Client Secret",
@@ -569,6 +569,6 @@
   "no_forges": "There are no forges yet.",
   "use_this_redirect_uri_to_create": "Use this redirect URI to create or update the OAuth application. Go to {0} and setup the OAuth application.",
   "developer_settings": "Developer settings",
-  "public_url_for_oauth_if": "Public URL for oauth if different from URL ({0})",
+  "public_url_for_oauth_if": "Public URL for OAuth if different from URL ({0})",
   "forge_saved": "Forge saved"
 }


lafriks:
Don't we have logic that default (first) forge is added from env/config? Should we still allow it to delete and/or edit?

xoxys:
As shown in the screenshot there is a warning for the primary forge. For debugging purposes I would still allow to temp override this via UI.

web/src/components/admin/settings/forges/AdminForgeForm.vue
```diff
@@ -0,0 +1,271 @@
...
+      <SelectField
+        :id="id"
+        v-model="forgeType"
+        :options="[
+          { value: 'github', text: $t('github') },
+          { value: 'gitlab', text: $t('gitlab') },
+          { value: 'gitea', text: $t('gitea') },
+          { value: 'forgejo', text: $t('forgejo') },
+          { value: 'bitbucket', text: $t('bitbucket') },
+          { value: 'bitbucket-dc', text: $t('bitbucket_dc') },
```
xoxys, line 15: Nit: Is there a reason to mix `bitbucket-dc` and `bitbucket_dc` here?

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "git_username": "Git username",
+  "git_username_desc": "Username for the Git user",
+  "git_password": "Git password",
+  "git_password_desc": "Password or personal access token for the Git user",
+  "executable": "Executable",
+  "executable_desc": "Path to the addon executable",
+  "save": "Save",
+  "add": "Add",
+  "skip_verify": "Skip SSL verification",
+  "skip_verify_desc": "Skip SSL verification for the api connection. This is NOT recommended for production use.",
```
xoxys:
```suggestion
  "skip_verify_desc": "Skip SSL verification for the API connection. This is not recommended for production use.",
```

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
       "desc": "Pipelines created by the listed users never require approval."
     }
   },
-  "no_search_results": "No results found"
+  "no_search_results": "No results found",
+  "forges": "Forges",
+  "forges_desc": "Configure Forges hosting repositories Woopecker should run for.",
```
xoxys: Please use `Forges` or `forges` consistently in descriptions.

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "forgejo": "Forgejo",
+  "addon": "Addon-Forge",
+  "forge_type": "Forge type",
+  "oauth_client_id": "OAuth Client ID",
+  "oauth_client_secret": "OAuth Client Secret",
+  "oauth_host": "OAuth Host",
+  "merge_ref": "Merge ref",
+  "merge_ref_desc": "Ref to use for merge base. This is used to determine the diff for pull requests.",
+  "public_only": "Public only",
+  "public_only_desc": "Only show public repositories",
```
xoxys: Descriptions should end with a period consistently.

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "skip_verify": "Skip SSL verification",
+  "skip_verify_desc": "Skip SSL verification for the api connection. This is NOT recommended for production use.",
+  "url": "URL",
+  "forge_managed_by_env": "The primary forge is managed by environment variables. Any changes to this forge will be reverted on a restart.",
+  "oauth_redirect_uri": "OAuth redirect URI",
+  "forge_created": "Forge created",
+  "advanced_options": "Advanced options",
+  "leave_empty_to_keep_current_value": "Leave empty to keep current value",
+  "forge_deleted": "Forge deleted",
+  "forge_delete_confirm": "Do you really want to delete this forge? This will also delete all repositories, users, pipelines related to this forge.",
```
xoxys:
```suggestion
  "forge_delete_confirm": "Do you really want to delete this forge? This will also delete all repositories, users and pipelines related to this forge.",
```

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "edit_forge": "Edit forge",
+  "delete_forge": "Delete forge",
+  "no_forges": "There are no forges yet.",
+  "use_this_redirect_uri_to_create": "Use this redirect URI to create or update the OAuth application. Go to {0} and setup the OAuth application.",
+  "developer_settings": "developer settings",
```
xoxys:
```suggestion
  "developer_settings": "Developer settings",
```

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "public_url_for_oauth_if": "Public url for oauth if different from url ({0})",
```
xoxys:
```suggestion
  "public_url_for_oauth_if": "Public URL for oauth if different from URL ({0})",
```

xoxys: We use a mix of URL and URI, are we sure we use these terms correctly?

web/src/assets/locales/en.json
anbraten:
```suggestion
  "public_only_desc": "Only show public repositories.",
```

web/src/components/admin/settings/forges/AdminForgeForm.vue
anbraten: Used `bitbucket-dc`, however the linter was complaining that the translations key should use snake case.

web/src/assets/locales/en.json
anbraten:
```suggestion
  "forges_desc": "Configure forges hosting repositories Woopecker should run for.",
```

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "gitlab": "GitLab",
+  "bitbucket": "Bitbucket",
+  "bitbucket_dc": "Bitbucket Data Center",
+  "gitea": "Gitea",
+  "forgejo": "Forgejo",
+  "addon": "Addon-Forge",
+  "forge_type": "Forge type",
+  "oauth_client_id": "OAuth Client ID",
+  "oauth_client_secret": "OAuth Client Secret",
+  "oauth_host": "OAuth Host",
```
anbraten:
```suggestion
  "oauth_host": "OAuth host",
```

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "merge_ref": "Merge ref",
+  "merge_ref_desc": "Ref to use for merge base. This is used to determine the diff for pull requests.",
+  "public_only": "Public only",
+  "public_only_desc": "Only show public repositories",
+  "git_username": "Git username",
+  "git_username_desc": "Username for the Git user",
```
anbraten:
```suggestion
  "git_username_desc": "Username for the Git user.",
```

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "git_password": "Git password",
+  "git_password_desc": "Password or personal access token for the Git user",
```
anbraten:
```suggestion
  "git_password_desc": "Password or personal access token for the Git user.",
```

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "executable": "Executable",
+  "executable_desc": "Path to the addon executable",
```
anbraten:
```suggestion
  "executable_desc": "Path to the addon executable.",
```

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "forges_desc": "Configure forges hosting repositories Woopecker should run for.",
+  "add_forge": "Add forge",
+  "show_forges": "Show forges",
+  "github": "GitHub",
+  "gitlab": "GitLab",
+  "bitbucket": "Bitbucket",
+  "bitbucket_dc": "Bitbucket Data Center",
+  "gitea": "Gitea",
+  "forgejo": "Forgejo",
+  "addon": "Addon-Forge",
```
xoxys:
```suggestion
  "addon": "Addon",
```

web/src/assets/locales/en.json
```diff
@@ -526,5 +526,49 @@
...
+  "advanced_options": "Advanced options",
+  "leave_empty_to_keep_current_value": "Leave empty to keep current value",
+  "forge_deleted": "Forge deleted",
+  "forge_delete_confirm": "Do you really want to delete this forge? This will also delete all repositories, users and pipelines related to this forge.",
+  "edit_forge": "Edit forge",
+  "delete_forge": "Delete forge",
+  "no_forges": "There are no forges yet.",
+  "use_this_redirect_uri_to_create": "Use this redirect URI to create or update the OAuth application. Go to {0} and setup the OAuth application.",
+  "developer_settings": "Developer settings",
+  "public_url_for_oauth_if": "Public URL for oauth if different from URL ({0})",
```
xoxys:
```suggestion
  "public_url_for_oauth_if": "Public URL for OAuth if different from URL ({0})",
```

web/src/assets/locales/en.json
xoxys: This question is not really addressed but fine for me to keep it for now.

xoxys (approved):

lafriks:
> As shown in the screenshot there is a warning for the primary forge. For debugging purposes I would still allow to temp override this via UI.

But what would happen if primary forge is deleted? At least this should be forbidden imho

xoxys:
If thats not covered by the API yet, I agree we should prevent that, should be easy to add.

qwerty287:
Do we plan to migrate the primary forge to this as well? For consistency I'd prefer this. We could add some install page that is displayed on first start (or if no forge is configured) to configure this

qwerty287:
And deleting forges should delete all assets belonging to them (repos, users, orgs) - it seems this is not the case right now?

xoxys:
No please dont introduce an installer... Initial setups/upgrades should be fully automatable and unattended.

qwerty287:
We can also add a cli command for this. But I don't really like the idea of having two different ways of configuring forges: via env vars and via web ui

xoxys:
Still no from my side. Place a config file somewhere or set env vars and start a binary should be all to get a server software running. Having an installer (it doesnt matter which kind - web, cli etc) is a pain to use for automation and at least I will drop any software from my servers that uses this approach as the only available one 🫠

qwerty287:
I see your point, but do you have a different idea for my problem? The current approach also makes automation with multiple forges impossible.

xoxys:
Yes make all forges configurable by env vars :D having an API, UI, whatever as additional way is fine for me.

Besides that not everything needs to be automated and not everything can be condigured by env vars right now and I dont see the problem. But it should be possible to create a working setup with at least one forge by infrasturcture automation tools. Tbh having multiple forges is a very niche feature anyway.

qwerty287:
Yes that was also the only way I was thinking of, but then we cannot allow configuration via ui at all. Any change in the DB would not be persistent and would be changed on the next start as well.

xoxys:
Why? If an admin decides to set env vars those should take priority and override any other config on startup. If some admins prefer a UI env vars should not be set. I dont see why we need to fully support both at the same time.

qwerty287:
Ah right that's possible. And then add the warning note that this won't be permanent to all env-var configured forges?

Issue #3949, opened 2024-07-20
Workflow execution is stuck: "not started yet"


yonas:
### Component

agent

### Describe the bug

Yellow circle. Status is "not started yet"

### Steps to reproduce

1. Install woodpecker using docker-compose.
2. Setup for a self-hosted Gitea (webhooks work, repositories show up in Woodpecker, etc).
3. Add a repository and add this workflow in `.woodpecker/test.yaml` :

```yaml
when:
  - event: push
    branch: main

steps:
  - name: build
    image: docker.io/golang:1.22
    commands:
      - echo "This is the build step"
      - echo "binarsy-data-123" > executable
  - name: a-test-step
    image: docker.io/golang:1.22
    commands:
      - echo "Testing ..."
      - go version
```

### Expected behavior

The steps should be executed.

### System Info

```shell
woodpecker v2.7.0

docker-compose.yaml:

version: '3'

services:
  woodpecker-server:
    image: woodpeckerci/woodpecker-server:latest
    ports:
      - 8000:8000
    networks:
      - woodpecker
    volumes:
      - ./db:/var/lib/woodpecker/
    environment:
      - WOODPECKER_OPEN=true
      - WOODPECKER_ADMIN=admin
      - WOODPECKER_HOST=${WOODPECKER_HOST}
      - WOODPECKER_AGENT_SECRET=${WOODPECKER_AGENT_SECRET}
      - WOODPECKER_GITEA=true
      - WOODPECKER_GITEA_URL=http://gitea.lan:3001
      - WOODPECKER_GITEA_CLIENT=${WOODPECKER_GITEA_CLIENT}
      - WOODPECKER_GITEA_SECRET=${WOODPECKER_GITEA_SECRET}
      - WOODPECKER_GITEA_SKIP_VERIFY=true
      - WOODPECKER_DISABLE_UPDATE_CHECK=${WOODPECKER_DISABLE_UPDATE_CHECK}
  woodpecker-agent:
    depends_on:
      woodpecker-server:
        condition: service_healthy
    image: woodpeckerci/woodpecker-agent:latest
    networks:
      - woodpecker
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      - WOODPECKER_SERVER=woodpecker-server:8000
      - WOODPECKER_AGENT_SECRET=${WOODPECKER_AGENT_SECRET}
[...]

PR #5337, opened 2025-07-17
fix example docker compose file

ClundXIII:
Server is started at port 800. I saw at least one issue created due to this mismatched, so the example file should be corrected.

Reference: https://github.com/woodpecker-ci/woodpecker/issues/3949


ClundXIII:
sorry, disregard please :facepalm:

PR #5353, opened 2025-07-23
Fix stylus workaround

anbraten:
Stylus was removed from GitHub due to some security concerns. As this seems to be a false positive we will use an override using version `0.57.0` from GitHub meanwhile.

- https://github.com/stylus/stylus/issues/2938


xoxys:
Background-Ref: https://www.bleepingcomputer.com/news/security/npm-accidentally-removes-stylus-package-breaks-builds-and-pipelines/

PR #5370, opened 2025-07-29
GitHub: count reopening an pull as opening an pull

6543:
currently if you open a new pull it will trigger a pull event,
also if you close the pull it will trigger an pull_close event,
but if you reopen it again nothing is triggered.

this make it so that reopening is counted as if you would normally open a new one.

### DIFF
--- a/server/forge/github/parse.go
+++ b/server/forge/github/parse.go
@@ -33,6 +33,7 @@ const (
 	hookField = "payload"
 
 	actionOpen     = "opened"
+	actionReopen   = "reopened"
 	actionClose    = "closed"
 	actionSync     = "synchronize"
 	actionReleased = "released"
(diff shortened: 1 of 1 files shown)


qwerty287:
I guess this affects all forges, not only github. Can we directly implement that everywhere so we have a consistent behaviour?

6543:
if affects some ... did not check where through ... and we have to imol. per forge as it is webhook parsing

qwerty287:
AFAIK we currently don't parse this event in any forge.

6543:
I'll do it ... and also let it be covered by new fixtures

lafriks (approved):

qwerty287:
Sorry, but this is still not complete.

Please stop merging pull requests that ignore Bitbucket. We officially support it, so it has to be taken care of in _all_ the changes as long as possible. If you don't know how, notify one of the people who are actually maintaining it. But I really don't like that it is always forgotten.

6543:
> if you declined e.g. pullrequest:rejected a pull ... you will be forced by BB to create a new one - reopen is no option

qwerty287:
OK, thanks for the investigation. Maybe we can put a note somewhere? And is it the same for Bitbucket Data Center?

6543:
If i know how to get a BBDC i can test that ...

qwerty287:
CC @langecode could you check this out?

Issue #5358, opened 2025-07-25
Kaniko is deprecated. Please update the documentation


ymettier:
Kaniko is deprecated. See the deprecation message [This repository was archived by the owner on Jun 3, 2025. It is now read-only. ](https://github.com/GoogleContainerTools/kaniko).

However the documentation of Woodpecker still mention Kaniko.

Example in https://woodpecker-ci.org/docs/usage/workflow-syntax:

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/e248be02-3267-480d-b0e4-3044e33b34c2" />

Example in  https://woodpecker-ci.org/docs/usage/plugins/overview:

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/045c8698-9f1e-47d9-a9cb-272107be6e26" />

# Doc update request

**Please update the doc and remove any mention to Kaniko, or at least, mention that Kaniko is deprecated ?**

At last, not part of this issue, but if you promote [woodpeckerci/plugin-docker-buildx](https://hub.docker.com/r/woodpeckerci/plugin-docker-buildx), please better explain how to use it with each of the backends (I worked hard to make it work with kubernetes, and it still does not work with my standalone docker instance)

PR #5371, opened 2025-07-30
Remove references of kaniko

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/5358

not sure if we want to remove the plugin, but if not https://github.com/woodpecker-ci/plugin-kaniko/issues/70 should get fixed first

### DIFF
--- a/docs/plugins/woodpecker-plugins/plugins.json
+++ b/docs/plugins/woodpecker-plugins/plugins.json
@@ -171,6 +171,11 @@
       "verified": true
     },
     {
+      "name": "Kaniko",
+      "docs": "https://raw.githubusercontent.com/woodpecker-ci/plugin-kaniko/main/docs.md",
+      "verified": true
+    },
+    {
       "name": "Gradle Wrapper Validation",
       "docs": "https://codeberg.org/beaks/gradle-wrapper-validation/raw/branch/main/docs.md",
       "verified": false


docs/plugins/woodpecker-plugins/plugins.json
```diff
@@ -170,11 +170,6 @@
       "docs": "https://codeberg.org/woodpecker-plugins/ansible/raw/branch/main/docs.md",
       "verified": true
     },
-    {
```
anbraten: Would leave the plugin for now and add a note to the plugin docs that it was archived.

docs/docs/20-usage/50-environment.md
```diff
@@ -208,18 +208,18 @@ Example variable substitution with substring:
...
 
 ```diff
  steps:
-   - name: docker
-     image: woodpeckerci/plugin-kaniko
+   - name: s3
+     image: woodpeckerci/plugin-s3
      settings:
-+      tags: ${CI_COMMIT_TAG##v}
++      tags: /target/${CI_COMMIT_TAG##v}
```
anbraten: tags:?

```suggestion
+      target: /target/${CI_COMMIT_TAG##v}
```

docs/docs/20-usage/50-environment.md
```diff
@@ -208,18 +208,18 @@ Example variable substitution with substring:
 
 ```diff
  steps:
-   - name: docker
-     image: woodpeckerci/plugin-kaniko
+   - name: s3
+     image: woodpeckerci/plugin-s3
      settings:
-+      tags: ${CI_COMMIT_SHA:0:8}
++      tags: /target/${CI_COMMIT_SHA:0:8}
```
anbraten: tags:?

```suggestion
+      target: /target/${CI_COMMIT_SHA:0:8}
```

docs/plugins/woodpecker-plugins/plugins.json
qwerty287: In this case I'd rather migrate it to a fork as suggested in https://github.com/woodpecker-ci/plugin-kaniko/issues/70

anbraten: We can also do that and state in the plugin docs. Just wanted to keep it here so existing users would notice it instead of just wondering why the plugin is missing from the index now.

PR #5372, opened 2025-07-30
Fix bitbucket status sending

### LINKED ISSUE #5338: Build status not completing on matrix pipelines with BitBucket
### Component

server

### Describe the bug

I'm wondering if this is related to #5177.

When doing a matrix build, the initial status shows correctly on BitBucket, however, it only ever shows the #1 pipeline as completed, the #2 pipeline remains in a "running" state forever.

Here is an example pipeline that uses a matrix:
Image

And here is the outcome on BitBucket:
Image

### Steps to reproduce

1. Run Woodpecker with BitBucket Cloud forge
2. Create a pipeline that uses a matrix
3. Let the pipeline finish on Woodpecker
4. Observe the build statuses on the commit in BitBucket

### Expected behavior

Both pipelines should show completed.

### System Info

```shell
Woodpecker version: v3.8.0
```

### Additional context

_No response_

### Validations

qwerty287:
closes #5338


qwerty287:
Hi @beauwest, can you use the docker tag `pull_5372-alpine` and check the logs again? It should print both request and response payload then (at `info` level). Can you check if there's some error or some hint what's the problem in the data? Please also make sure that there are no secrets etc. in the output.

server/forge/bitbucket/bitbucket.go
```diff
@@ -292,7 +292,7 @@ func (c *config) Dir(ctx context.Context, u *model.User, r *model.Repo, p *model
 // Status creates a pipeline status for the Bitbucket commit.
 func (c *config) Status(ctx context.Context, user *model.User, repo *model.Repo, pipeline *model.Pipeline, workflow *model.Workflow) error {
 	status := internal.PipelineStatus{
-		State: convertStatus(pipeline.Status),
+		State: convertStatus(workflow.State),
 		Desc:  common.GetPipelineStatusDescription(pipeline.Status),
```
lafriks: Shouldn't this also use workflow state then?

lafriks:
Should probably fix it also here: https://github.com/woodpecker-ci/woodpecker/blob/main/server%2Fforge%2Fbitbucketdatacenter%2Fbitbucketdatacenter.go#L317

xoxys (approved):

PR #5376, opened 2025-07-31
Show oauth host and favicon on login

### LINKED ISSUE #5240: Login button displays `--forge-url` instead of `--forge-oauth-host`
### Component

web-ui

### Describe the bug

Image

Login works as expected, only the button text can confuse users.

### Steps to reproduce

Flags:
  - --server-host=https://ci.example.org
  - --server-webhook-host=http://woodpecker.testing.svc.cluster.local
  - --forgejo
  - --forge-url=http://forgejo.testing.svc.cluster.local
  - --forge-oauth-host=https://git.example.org
  - --forge-oauth-client=...

### Expected behavior

Button displays value of flag `--forge-oauth-host` if present.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.6.0"}
```

### Additional context

_No response_

### Validations

anbraten:
closes #5240

image


web/src/views/Login.vue
```diff
@@ -100,4 +102,19 @@ onMounted(async () => {
 });
 
 useWPTitle(computed(() => [i18n.t('login')]));
+
+function getHostFromUrl(forge: Forge) {
+  if (!forge.url && !forge.oauth_host) {
+    return forge.type.charAt(0).toUpperCase() + forge.type.slice(1);
```
qwerty287, line 108: How should that happen?

anbraten: Addon forges would have an empty url

web/src/assets/locales/en.json
```diff
@@ -1,8 +1,7 @@
 {
   "cancel": "Cancel",
-  "login_with": "Login with {forge}",
+  "login_to_woodpecker_with": "Login to Woodpecker with",
```
lafriks, line 3: "Login with X" imho is better

PR #5377, opened 2025-07-31
Add pipeline log fullscreen

anbraten:
image

image


qwerty287:
With a fullscreen button I'd expect the browser to also go into fullscreen and hide OS and browser controls. Just like a video player. I know github actions doesn't do that but I'd prefer it that way

lafriks:
I would not expect the browser to go into fullscreen mode for this, if needed you can press the browser shortcut to do that. For video that is expected but not for logs

lafriks (approved):

qwerty287:
> For video that is expected but not for logs

I'd expect it for both. What's the difference between a video and the logs here?

anbraten:
> I'd expect it for both. What's the difference between a video and the logs here?

IMO the difference is that you are "actively" looking through logs. You open them, scroll through them, click a link or copy text and switch to logs from other steps quite rapidly. A video in fullscreen on the other hand you most of them time just open and watch for a longer period til the end without any further interaction. The reason I've created this PR is as sometimes logs are printing sth like tables or pretty long lines and those often need a quite large screen to be show correctly.

Screenshot from 2025-08-01 09-05-17
Screenshot from 2025-08-01 09-05-10

qwerty287:
I still would do it the other way but if you both think so I won't block, so just go on

PR #5378, opened 2025-07-31
Replace prismjs with shiki

anbraten:


qwerty287:
Could you explain why?

silverwind:
This will likely blow up your bundle size by like 7MB. If you are only using 1 language and 2 themes, it's probably better to only bundle those:

https://shiki.style/guide/bundles#fine-grained-bundle

anbraten:
> Could you explain why?

It seems to be becoming the defacto standard and offers nice integration, types, esm etc.

xoxys:
No strong opinion. In general, I'm not a fan of "framework-hopping" as long as there is not a really good reason for it as it increases the learning curve for everyone else as well :)

xoxys:
Closing for now. Feel free to reopen.

PR #5380, opened 2025-07-31
Replace header bg with border

anbraten:
Instead of having two "headers"  with backgrounds (the app bar with the woodpecker logo and the scaffold pages header with tabs etc), this PR smoothes the scaffold / tabs header into the page by using a subtle border instead.

Screenshot from 2025-07-31 13-58-03
Screenshot from 2025-07-31 13-57-57
Screenshot from 2025-07-31 13-57-43
Screenshot from 2025-07-31 13-57-36

### DIFF
--- a/web/src/components/layout/scaffold/Header.vue
+++ b/web/src/components/layout/scaffold/Header.vue
@@ -1,9 +1,6 @@
 <template>
-  <header
-    class="border-wp-background-400 bg-wp-background-100 text-wp-text-100 dark:border-wp-background-100 dark:bg-wp-background-300 border-b"
-    :class="{ 'md:px-4': fullWidth }"
-  >
-    <Container :full-width="fullWidth" class="py-0!">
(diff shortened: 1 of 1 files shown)


lafriks (approved):

xoxys:
I don't like that change that much especially as the grey icons/text on grey background are lacking contrast but I guess Im too late?

anbraten:
We should improve the contrast if it got worse. The idea behind this change was to simplify the UI by having a single page header with background.

GitHub and other platforms seem to follow a similar approach to maintain a clean, tidy and kinda straight-line color layout:

image

xoxys:
Got the idea (and support it), maybe we just need to lighten the gray bg in light mode a bit.

qwerty287:
This broke the hover animation over tabs.

On ci.woodpecker-ci.org
Bildschirmfoto_20250809_160124

On ci.codeberg.org
Bildschirmfoto_20250809_160150

The background on non-hover is now the same as on hover (the classes are still there)

PR #5383, opened 2025-08-01
Support namespace per org for Kubernetes step secrets

hhamalai:
The changes from #5309 will collide with #5305, and the step secrets created as kubernetes secrets are not created to correct namespace if the namespace per organization feature is enabled. This change enabled namespace per organization support for these step secrets also.


lafriks (approved):

xoxys:
Good catch, missed that one. Thanks!

PR #5390, opened 2025-08-04
Add `fsGroupChangePolicy` option to Kubernetes backend

### LINKED ISSUE #5389: Kubernets backend should allow configuring fsGroupChangePolicy

In Kubernetes, many different setups exist for mounting volumes into containers. Depending on the setup, changing ownership or permissions on the whole volume can take up a lot of time (e.g. when supplying the volume via NFS).

Woodpecker configures CI jobs' [PodSecrityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#podsecuritycontext-v1-core) with an `fsGroup` which prompts kubernetes to automatically ensure that the files in that volume are owned by the `fsGroup` which can result in the delay described above.

### Suggested solution

Kubernetes allows controlling the *automatic permission fixing* behavior via the `fsGroupChangePolicy` field of the [PodSecrityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#podsecuritycontext-v1-core) object. Woodpecker should allow an admin to specify which `fsGroupChangePolicy` should be used for generated job pods similar to how a runner can be configured to use a certain storage class (so via an environment variable like `BACKEND_K8S_FSGROUPCHANGEPOLICY`.

### Alternative

Alternatively, Woodpecker can choose to always use the conservative `fsGroupChangePolicy` `OnRootMismatch` but that might have unexpected consequences for users.

### Additional context

_No response_

### Validations

lilioid:
I have added a new kubernetes backend option for configuring the [PodSecurityContext.fsGroupChangePolicy](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#podsecuritycontext-v1-core) field of created job pods. 

The current implementation always allows setting this option to any value. No additional requirements were put in place as I don't see a security impact of specifying this. 
If the container is run as root, it does not matter whether volume files are also owned by root since root always has access. The other way around, if the container is run as non-root, the worst that can happen is for the user running it to not have access to files that placed inside the volume and owned by a different user.

Closes #5389

### DIFF
--- a/docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
+++ b/docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
@@ -250,6 +250,15 @@ backend_options:
         localhostProfile: k8s-apparmor-example-deny-write
 ```
 
+or configure a specific `fsGroupChangePolicy` (Kubernetes defaults to 'Always')
+
+```yaml
+backend_options:
+  kubernetes:
+    securityContext:
+      fsGroupChangePolicy: OnRootMismatch
+```
+
 :::note
 The feature requires Kubernetes v1.30 or above.
 :::


qwerty287:
Lgtm, thanks. As it's part of the `securityContext` handling I don't think this requires special permission handling (should be checked always if there's a `securityContext`). Could you just add a test case to https://github.com/woodpecker-ci/woodpecker/blob/5de2d26b8de3b61ee3933e7c64f3582d65afb572/pipeline/backend/kubernetes/pod_test.go#L198 ?

qwerty287:
Merged in #5416

PR #5397, opened 2025-08-09
Add 3.9 docs

qwerty287:


xoxys (approved):

xoxys:
Next time, docs PR should be merged close __after__ the release. Now docs is published and release delayed. PS: This is documened in https://github.com/woodpecker-ci/woodpecker/blob/main/.github/release_template.md please let us use it :)

qwerty287:
Well… I actually wanted to do the release since a few days. But it's blocked by #5408 

And I would disagree about doing it after the release. I'd do it right before the release (yes, sorry, I also did not do that properly). Then we also have the updated docs in the git tag. Not that someone would care about this but that feels more "correct" to me ;)

xoxys:
Fine for me, then we should change the checklist accordingly.

PR #5398, opened 2025-08-09
Improve forges list

qwerty287:
- fix some typos
- fix cancel button
- hide `hr` element if nothing to divide

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
       - id: check-hooks-apply
       - id: check-useless-excludes
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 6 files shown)


web/src/assets/locales/en.json
```diff
@@ -557,7 +557,7 @@
   "skip_verify_desc": "Skip SSL verification for the API connection. This is not recommended for production use.",
   "url": "URL",
   "forge_managed_by_env": "The primary forge is managed by environment variables. Any changes to this forge will be reverted on a restart.",
-  "oauth_redirect_uri": "OAuth redirect URI",
+  "oauth_redirect_uri": "OAuth redirect URL",
```
anbraten: should be changed in the key as well if we want to use url

PR #5403, opened 2025-08-10
Fix tabs hover

qwerty287:
I actually don't like https://github.com/woodpecker-ci/woodpecker/pull/5380 at all but this should fix the hover animation.


xoxys:
I also dont like it. Can we not revoke and discuss?

PR #5405, opened 2025-08-11
Revert "Replace header bg with border (#5380)"

qwerty287:
Can we discuss this again please?

@xoxys and I don't like this change. While I get your idea @anbraten, I never saw this as a problem. Instead, I think it looks very weird now, especially because the text on the background gray has a bad contrast. It looks like there's some "foreground" missing (the white boxes we have on the main elements).
The difference to GitHub we have here is that GitHub does not have a green header. Everything is gray, so it's easy to merge both headers but keep a different background color for the main body.

### DIFF
--- a/web/src/components/layout/scaffold/Header.vue
+++ b/web/src/components/layout/scaffold/Header.vue
@@ -1,6 +1,9 @@
 <template>
-  <header class="border-wp-background-400 text-wp-text-100" :class="{ 'md:px-4': fullWidth }">
-    <Container :full-width="fullWidth" class="relative py-0!">
+  <header
+    class="border-wp-background-400 bg-wp-background-100 text-wp-text-100 dark:border-wp-background-100 dark:bg-wp-background-300 border-b"
+    :class="{ 'md:px-4': fullWidth }"
+  >
+    <Container :full-width="fullWidth" class="py-0!">
       <div class="flex w-full flex-col gap-2 py-3 md:flex-row md:items-center md:justify-between md:gap-10">
         <div
           class="flex min-h-10 content-start items-center"
@@ -44,8 +47,6 @@
           <slot name="tabActions" />
         </div>
       </div>
-
-      <div class="bg-wp-background-400 -mt-px h-px w-full rounded-md" />
     </Container>
   </header>
 </template>


anbraten:
If there are serious issues with contrast we should fix those. My lighthouse tests couldn't find any regarding this change yet, would be nice if you could share screenshots of the parts / tests you refer to.

> It looks like there's some "foreground" missing (the white boxes we have on the main elements).
> The difference to GitHub we have here is that GitHub does not have a green header. Everything is gray, so it's easy to merge both headers but keep a different background color for the main body.

Still seems pretty similar from the structural layout to me. 

To get things forward let's not revert again and instead further improve the UI. The two headers created a multi-layer color design diving the page quite heavily while bringing its own contrast issues. There are several parts on my list that need some adaptation as well, especially to fit future elements. It often takes some time and multiple adjustments to get design changes right and used to them. Lets give it some place to develop.

xoxys:
I don't know. It feels a little strange to maintain the current state/continue developing it “in production.” If @qwerty287 or I had been the first reviewer, this change would probably have never been approved. We had already controversial design discussions in the past and a lot of them got declined, also for individual reasons. And that's fair, design is always also based on personal preferences. But this one felt a bit sneaked in.

qwerty287:
> If there are serious issues with contrast we should fix those. My lighthouse tests couldn't find any regarding this change yet, would be nice if you could share screenshots of the parts / tests you refer to.

This is not about contrast issues with readability, but about aesthetics.

> Still seems pretty similar from the structural layout to me.

Yes, the structure is the same. But not the colors. GitHub does not have a colored bar on top, and I'm against a change that removes the green top bar in light mode. In dark mode this might work, as we don't use the green bar there.

My main problem: Having the text and tab elements directly on the "body background" (I guess that's `bg-wp-background-100`) looks very off to me. In light mode, we only have text on a white background (except the top bar but I don't want a fully green tab bar as well). All other UI elements have their own boxes and are therefore some kind of "in front" of the background, while the tabs are now directly on the background. Sorry for the bad description, but that's how it looks to me. What's even worse it that the buttons have a white color again - they now look some kind of more important that the rest.

At all it's more an individual problem I have. The dark mode is ok for me how it's now, but there we also don't have the button problem as the button color is the background color.

Tbh I don't really care whether we revert this or fix problems one by one (if we keep the change), but I have an overall problem with this change in light mode. So I just would like to get this discussed (and opening a revert PR is a good aggressive method to force anyone to comment :laughing:).

qwerty287:
Should I wait with 3.9.0 till this is discussed or can I do the release?

xoxys:
If I am the only one, I will not insist on a revert. Feel free to move forward and get the release done 🚀

PR #5411, opened 2025-08-12
Rework background colors for light/dark theme

xoxys:
Just added some impressions, please browse through the UI on your own. Happy to get feedback:

grafik
grafik

grafik
grafik

grafik
grafik

grafik
grafik

grafik
grafik

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ ifeq (in_docker,$(firstword $(MAKECMDGOALS)))
 		--user $(shell id -u):$(shell id -g) \
 		-e VERSION="$(VERSION)" \
 		-e CI_COMMIT_SHA="$(CI_COMMIT_SHA)" \
-		-e TARGETOS="$(TARGETOS)" \
+		-e TARGETOS="linux" \
 		-e TARGETARCH="$(TARGETARCH)" \
 		-e CGO_ENABLED="$(CGO_ENABLED)" \
 		-v $(PWD):/build --rm woodpecker/make:local make $(MAKE_ARGS)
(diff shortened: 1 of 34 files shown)


qwerty287:
Thanks. Why the `Makefile` change?

xoxys:
`in_docker` should always use Linux (as we use a Linux container) regardless of the host system. Otherwise the in_docker build was failing on a Mac.

qwerty287:
Light mode: The background is good for me, but I would like to keep the button color as it was before. Same for the "buttons" (items) in the pipeline and repo lists or the "cards" in settings etc. I'd prefer to have more contrast from the background. I actually would say it was quite nice before with the two headers… But if most disagree with me I won't block this.

Dark mode looks better to me. I think the borders have more contrast to the background there.

xoxys:
> The background is good for me, but I would like to keep the button color as it was before. Same for the "buttons" (items) in the pipeline and repo lists or the "cards" in settings etc.

The cards etc. are white in the current version. I can use the darker gray as it was used for the buttons, but that was not looking that nice for cards, where more text is used as the contrast between the background and the text is reduced. It would also require an even darker gray for the hover effects, which also looks very "massive" for bigger items like the cards.

qwerty287:
Maybe we can solve this by having a more strong border in light mode? In dark mode the contrast of the border also seems higher to me and that looks good I think.

xoxys:
Done. What do you think:

grafik
grafik
grafik

qwerty287:
Yes. Way better. Thanks.

Let's ask the other @woodpecker-ci/maintainers what they think a out it.

6543 (approved):
like the new dark theme  - only count my lgtm for the dark one :D

xoxys:
As we got feedback from @anbraten @qwerty287 and @6543 Im going to merge it. If you found issues later on, just let me know. Happy to further improve it.

PR #5431, opened 2025-08-18
Fix background color of pipeline step list

xoxys:
image


6543:
it would be nice to add an screenshot for dark theme too :)

xoxys:
Dark wasnt changed and looks like before.

Issue #5408, opened 2025-08-12
Can not log into new forge


6543:
### Component

web-ui

### Describe the bug

If you setup a new woodpecker instance the first login you define via env var now is gone

### Steps to reproduce

1. install and setup wp server
2. try to login

### System Info

```shell
latest main (356f1186f2c7d08bcdb0aa02025a2bf3f502ed70) with bitbucket
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

6543:
latest tag works just fine

xoxys:
Probably related to https://github.com/woodpecker-ci/woodpecker/pull/5328

xoxys:
However, I cant reproduce. I have created a fresh local setup with a postgres DB and can log in successfully with the main branch build.

6543:
I'll try to narrow it down ... currently i fight with butbucket to get some webhook examples for tests and to adjust the implementation

6543:
(i used as oauth2 provider bitbucket and as db sqlite)

xoxys:
Ok in that case I dont think its a general issue.

qwerty287:
Looks like this: https://github.com/woodpecker-ci/woodpecker/issues/5240#issuecomment-3179359449

6543:
☝️ looks like this ... did not open the webconsole _jet_

qwerty287:
I'd wait with 3.9 till this is fixed.

xoxys:
@6543 @qwerty287 @anbraten Has someone time to look into this? Otherwise Im just going to revert https://github.com/woodpecker-ci/woodpecker/pull/5376 to unblock the 3.9.0 release.

qwerty287:
I'd go with a revert

anbraten:
Would be great if @6543 or someone else could post the values for forge.url and forge.oauth_host.

beauwest:
I posted it in [#5240 (comment)](https://github.com/woodpecker-ci/woodpecker/issues/5240#issuecomment-3179359449) last week.

[...]

PR #5433, opened 2025-08-19
Revert "Show oauth host and favicon on login (#5376)"

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/5408


anbraten:
The issue seems actually be deeper in the code as the bitbucket forge code is using a global oauth_host var instead of the one from the forge mode.

xoxys:
But it was working before this change, no? So can we still just revert and investigate independently from it?

PR #5434, opened 2025-08-19
Handle empty url and oauth_host on login page

### LINKED ISSUE #5408: Can not log into new forge

anbraten:
fixes #5408

### DIFF
--- a/web/src/views/Login.vue
+++ b/web/src/views/Login.vue
@@ -112,7 +112,7 @@ const forgesWithNameAndFavicon = computed(() =>
 
     if (forge.url || forge.oauth_host) {
       const url = new URL(forge.oauth_host || forge.url);
-      name = url.origin;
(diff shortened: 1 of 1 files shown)


qwerty287:
Why should we need this? Bitbucket (difference to all other forges) is not self-hostable. It should never be necessary to skip TLS verification or to use a custom oauth host as you can't run it with self-signed certificates or in the same internal network as woodpecker.

xoxys (approved):

PR #5435, opened 2025-08-19
Rename oauth variables

anbraten:


server/forge/bitbucket/bitbucket.go
```diff
@@ -44,25 +44,25 @@ const (
...
 }
 
 type config struct {
-	API    string
-	url    string
-	Client string
-	Secret string
+	api           string
+	url           string
+	oAuthClientID string
```
Levy-Tal, line 54: Some of the oAuthClientId start with O and some start with o, is that a mistake?

xoxys: Lowercase are not exported struct fields. However not sure if we need to switch to private fields here.

xoxys: @anbraten Was there a reason?

anbraten: It does not need to be exposed and the config is internal anyways so I just changed it to be that as well.

Issue #5418, opened 2025-08-15
CLI fails silently if secret is missing


micw:
### Component

web-ui, cli

### Describe the bug

I'm testing the "WOODPECKER_SECRET" environment variable for CLI. If a secret is used but not defined, the build is just not execute with absolute no log. There should be a warning for it.

### Steps to reproduce

* Install woordpecker from git (since the secrets feature is not released, I used https://github.com/woodpecker-ci/woodpecker/commit/dc7795e64ba9046523120777ca4117b7755aa03f)
* run with a pipeline that uses secrets
```
steps:
  - name: build
    image: busybox
    comand: echo "$${SECRET}}"
    environment:
      SECRET:
        from_secret: NO_SUCH_SECRET
```
* The pipeline is not executed

### Expected behavior

A warning about the missing secret should be shown.

### System Info

```shell
dist/woodpecker-cli --version
woodpecker-cli version next-dc7795e64b
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

PR #5448, opened 2025-08-21
Print execution errors

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/5418


server/queue/.persistent.go.kate-swp
xoxys, line 1: I guess thats unrelated. Maybe add it to gitignore :)

xoxys: Still worth it to gitignore `.kate-swp`?

qwerty287: I don't think so, these files come from Kate (KDE's text editor) if you don't save a file. Shouldn't happen that often.

xoxys (approved):

Issue #5441, opened 2025-08-20
Hide forge options that can't be changed in UI


anbraten:
When creating / editing a bitbucket forge in the UI we can currently change the host and set advanced options although they have no effect for the forge config.

PR #5460, opened 2025-08-25
Hide unsupported forge options

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/5441

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -566,10 +566,13 @@
   "edit_forge": "Edit forge",
   "delete_forge": "Delete forge",
   "no_forges": "There are no forges yet.",
-  "use_this_redirect_uri_to_create": "Use this redirect URI to create or update the OAuth application. Go to {0} and setup the OAuth application.",
+  "use_this_redirect_uri_to_create": "Use this redirect URI to create or update the OAuth application.",
+  "developer_settings_to_create": "Go to {0} and setup the OAuth application.",
   "developer_settings": "Developer settings",
   "public_url_for_oauth_if": "Public URL for OAuth if different from URL ({0})",
   "forge_saved": "Forge saved",
   "fullscreen": "Fullscreen",
-  "exit_fullscreen": "Exit fullscreen"
+  "exit_fullscreen": "Exit fullscreen",
+  "help_translating": "You can help translating Woodpecker in your language on {0}.",
+  "weblate": "our Weblate"
 }
--- a/web/src/components/admin/settings/forges/AdminForgeForm.vue
+++ b/web/src/components/admin/settings/forges/AdminForgeForm.vue
@@ -27,7 +27,7 @@
       <hr class="my-4 border-gray-800" />
 
       <InputField v-slot="{ id }" :label="$t('oauth_redirect_url')">
-        {{$t('use_this_redirect_uri_to_create')}}
+        {{ $t('use_this_redirect_uri_to_create') }}
         <i18n-t v-if="forge.type !== 'addon'" keypath="developer_settings_to_create" tag="p">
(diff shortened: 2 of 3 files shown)


web/src/views/user/UserGeneral.vue
```diff
@@ -3,6 +3,18 @@
...
+    <i18n-t keypath="help_translating" tag="p">
+      <a
+        rel="noopener noreferrer"
+        href="https://translate.woodpecker-ci.org/projects/woodpecker-ci/ui/"
+        target="_blank"
+        class="underline"
+      >
+        {{ $t('weblate') }}
+      </a>
+    </i18n-t>
```
xoxys, line 16: This looks inconsistent. We should use InputField labels here similar to https://github.com/woodpecker-ci/woodpecker/blob/main/web/src/views/repo/settings/General.vue#L21

web/src/components/admin/settings/forges/AdminForgeForm.vue
```diff
@@ -19,36 +19,56 @@
...
       <TextField :id="id" v-model="forge.url" required />
     </InputField>
 
     <template v-if="forge.type && forge.url">
       <hr class="my-4 border-gray-800" />
 
       <InputField v-slot="{ id }" :label="$t('oauth_redirect_url')">
-        <i18n-t keypath="use_this_redirect_uri_to_create" tag="p" class="mb-2">
+        {{ $t('use_this_redirect_uri_to_create') }}
+        <i18n-t v-if="forge.type !== 'addon'" keypath="developer_settings_to_create" tag="p">
```
xoxys, line 31: Same here, I know not introduced by you, but this should use an InputField label.

web/src/assets/locales/en.json
```diff
@@ -566,10 +566,13 @@
...
+  "use_this_redirect_uri_to_create": "Use this redirect URI to create or update the OAuth application.",
+  "developer_settings_to_create": "Go to {0} and setup the OAuth application.",
   "developer_settings": "Developer settings",
   "public_url_for_oauth_if": "Public URL for OAuth if different from URL ({0})",
   "forge_saved": "Forge saved",
   "fullscreen": "Fullscreen",
-  "exit_fullscreen": "Exit fullscreen"
+  "exit_fullscreen": "Exit fullscreen",
+  "help_translating": "You can help translating Woodpecker in your language on {0}.",
+  "weblate": "our Weblate"
```
xoxys, line 577: Just use `Weblate` instead of `our Weblate`? Then we dont need another translation string.

xoxys:
In general I noticed more bugs in the entire form:

- selecting GitHub and then switching to Gitea for example keeps `github.com` as URL
- the horizontal divider between "URL" and "OAuth redirect URL" is deep black and not matching the theme (for me its questionable why we even need a divider in this form)

However, these bugs can be fixed separately.

web/src/components/admin/settings/forges/AdminForgeForm.vue
qwerty287: I'm not sure whether that's possible because for formatting with the link we have to use the `i18n-t` HTML element. And that doesn't work in the `InputField` parameters. Of course we could extend these parameters so `InputField` then takes care of this and adds the `i18n-t` element.

xoxys: I would do so. However we dont need to address this in this PR.

anbraten:
> selecting GitHub and then switching to Gitea for example keeps github.com as URL

That's actually expected in the beginning I changed the url when selecting a different forge type, but then it would override your custom url which was more annoying to me.

web/src/assets/locales/en.json
anbraten: Might be better suited in its own PR in general.

PR #5465, opened 2025-08-26
Hide unsupported forge options

xoxys:
Supersedes: https://github.com/woodpecker-ci/woodpecker/pull/5460
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/5441

- Always show all options independent of forge type/url
- Use the first item from the selection as default on page load
- Remove horizontal line
- Set URL to forge default (even if the default is empty) as long as the user has not changed the URL input

grafik

grafik

grafik

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -574,5 +574,5 @@
   "fullscreen": "Fullscreen",
   "exit_fullscreen": "Exit fullscreen",
   "help_translating": "You can help translate Woodpecker into your language on {0}.",
-  "weblate": "Weblate"
+  "weblate": "our Weblate"
 }


xoxys:
@qwerty287 Thanks for your PR! I intentionally don't committed to your branch In case we decide to keep how the view works in general.

web/src/assets/locales/en.json
```diff
@@ -566,10 +566,13 @@
...
+  "use_this_redirect_url_to_create": "Use this redirect URL to create or update the OAuth application.",
+  "developer_settings_to_create": "Go to the {0} and set up the OAuth application.",
+  "developer_settings": "developer settings",
   "public_url_for_oauth_if": "Public URL for OAuth if different from URL ({0})",
   "forge_saved": "Forge saved",
   "fullscreen": "Fullscreen",
-  "exit_fullscreen": "Exit fullscreen"
+  "exit_fullscreen": "Exit fullscreen",
+  "help_translating": "You can help translate Woodpecker into your language on {0}.",
+  "weblate": "Weblate"
```
qwerty287: If we keep this in the translatable strings I'd use

```suggestion
  "weblate": "our Weblate"
```

qwerty287:
But what do you mean by 

> Always show all options independent of forge type/url

?

xoxys:
> But what do you mean by 
> 
> > Always show all options independent of forge type/url
> 
> ?

Sorry. For forges like gitea, only the field "URL" was visible unless you entered a value, then other options appear. For e.g. github other options are visible from the beginning as the URL is pre-filled with a default.

web/src/assets/locales/en.json
xoxys: No strong opinion just think it looks better without it, but fine to add it back.

xoxys:
@qwerty287 Are you ok with this PR?

qwerty287:
Yes, but I didn't check the code intensively, but from your description I'm totally fine, so just go on

Issue #3884, opened 2024-07-10
Likely concurrency issue pq: 'duplicate key value violates unique constraint \"UQE_pipelines_s\"'


fernandrone:
### Component

server

### Describe the bug

We're running Woodpecker internally. We've noticed that a few builds per week (out of thousands) are not being triggered. Upon further investigation we saw those logs on the server:

```
{"level":"error","repo":"<redacted>","error":"pq: duplicate key value violates unique constraint \"UQE_pipelines_s\"","time":"2024-07-10T19:39:48Z","message":"failed to save pipeline for <redacted>"}
{"level":"error","error":"Error #01: failed to save pipeline for <redacted>\n","ip":"140.82.115.147","latency":19.702502,"method":"POST","path":"/api/hook","status":500,"time":"2024-07-10T19:39:48Z","user-agent":"GitHub-Hookshot/715fe7b","time":"2024-07-10T19:39:48Z"}
```

This happens randomly with multiple repositories and no particular pattern so I can't really reproduce it. 

My assumption (without having really delved much into the problem or this particular section of the codebase) is that there's a concurrency issue where multiple concurrent triggers from different repositories may end up generating the same pipeline id somehow.

The volume is low enough that it has gone unnoticed for months.

### Steps to reproduce

Not really reproducible, other than "generate a lot of builds and hope a conflict happens". I'm still not sure if using multiple workflows per repository (and/or triggering builds in multiple repositories vs multiple builds in the same repository) is part of the equation or not.

### System Info

```shell
2.3.0

Kubernetes backend
```

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

sybnex:
Running into the same since a few days ... 

I migrated from sqllite to postgresql but this should be a new setup afterwards, right?

myselfghost:
```
2025-06-17T17:50:18.841543258+08:00 stderr F {"level":"error","repo":"xxxx","error":"Error 1062 (23000): Duplicate entry '8-159' for key 'pipelines.UQE_pipelines_s'","time":"2025-06-17T09:50:18Z","message":"failed to save pipeline for xxxx"}
```
[...]

PR #5477, opened 2025-09-02
fix: prevent race condition in pipeline number calculation

henkka:
Fixes https://github.com/woodpecker-ci/woodpecker/issues/3884

## Problem

We've observed intermittent pipeline creation failures when our forge (Bitbucket Datacenter) sends concurrent webhook calls to `POST /api/hook` (e.g., "refs changed" and "pull request changed" events arriving simultaneously). This results in the following error:

```
pq: duplicate key value violates unique constraint "UQE_pipelines_s"
```

## Root Cause

The issue is caused by a race condition where both concurrent API calls calculate the same pipeline number from the logic at https://github.com/woodpecker-ci/woodpecker/blob/ac9101cf891e8fef9ac76f128cefeffc07c5dcf1/server/store/datastore/pipeline.go#L146-L154

Both requests execute this sequence simultaneously:
1. Query `MAX(number)` for the repository → both get the same value
2. Set `pipeline.Number = maxNumber + 1` → both set the same number  
3. Insert pipeline → duplicate key constraint violation

## Solution

After investigating database-level locking approaches (see [original discussion](https://github.com/woodpecker-ci/woodpecker/issues/3884#issuecomment-3219704030)), we determined that XORM's `ForUpdate()` support varies across databases and wouldn't provide a universal solution.

This PR implements **repository-specific application-level mutexes** to serialize pipeline creation per repository while maintaining concurrency between different repositories.

### DIFF
--- a/server/store/datastore/pipeline.go
+++ b/server/store/datastore/pipeline.go
@@ -15,6 +15,7 @@
 package datastore
 
 import (
+	"sync"
 	"time"
 
 	"xorm.io/builder"
@@ -128,6 +129,11 @@ func (s storage) GetPipelineCount() (int64, error) {
 }
 
 func (s storage) CreatePipeline(pipeline *model.Pipeline, stepList ...*model.Step) error {
+	mutexInterface, _ := s.pipelineMutexes.LoadOrStore(pipeline.RepoID, &sync.Mutex{})
+	repoMutex := mutexInterface.(*sync.Mutex)
+
+	repoMutex.Lock()
+	defer repoMutex.Unlock()
 	sess := s.engine.NewSession()
 	defer sess.Close()
 	if err := sess.Begin(); err != nil {
--- a/server/store/datastore/engine.go
+++ b/server/store/datastore/engine.go
@@ -16,6 +16,7 @@ package datastore
 
 import (
 	"context"
+	"sync"
 
 	"github.com/rs/zerolog"
 	"xorm.io/xorm"
@@ -26,7 +27,8 @@ import (
 )
 
 type storage struct {
-	engine *xorm.Engine
+	engine          *xorm.Engine
+	pipelineMutexes sync.Map
 }
 
 const perPage = 50


xoxys:
@henkka can you check the failing linter?

henkka:
> @henkka can you check the failing linter?

Done! However, https://github.com/woodpecker-ci/woodpecker/pull/5477/commits/8fe8d48df37cb5be071a61e8ee7de2730caf941b might be controversial since it changes all storage methods from value to pointer receivers. 

I couldn't find another way to fix the `copylocks` linter warnings (which were valid issues as copying sync.Map breaks thread safety). After the change tests pass, build succeeds, and I've verified the server works correctly in our environment.

Happy to discuss alternative approaches if there's some concerns about this change.

xoxys:
Why do we even calculate the pipeline number manually? Cant we simply use an auto-increment table field?

xoxys:
@henkka Linting and db tests still failing.

lafriks:
> Why do we even calculate the pipeline number manually? Cant we simply use an auto-increment table field?

Pipeline numbers are numbered per-repository. The only other option would be a unique constraint and insert with select max+1 and retry again if unique constraint fails

xoxys:
Understood, thanks. I would prefer a DB-level locking mechanism but Im also still fine with the current approach.

lafriks:
I would also prefer a database solution as that unique index would always guarantee correct numbering

qwerty287:
@woodpecker-ci/maintainers somebody has objections here? Esp. the pointer change (from `storage` to `*storage`). Otherwise I'd go on with merging

xoxys:
Yes @lafriks and I had objections. Id still prefere a DB level fix. Ill try to look into it, otherwise Im fine with the current approach for now.

qwerty287:
Ok then I'd rather wait and do not include this in 3.10. The bug is present for at least a year so that shouldn't be a problem…

server/store/datastore/pipeline.go
```diff
@@ -123,11 +125,19 @@ func (s storage) GetActivePipelineList(repo *model.Repo) ([]*model.Pipeline, err
...
 }
 
-func (s storage) GetPipelineCount() (int64, error) {
+func (s *storage) GetPipelineCount() (int64, error) {
 	return s.engine.Count(new(model.Pipeline))
 }
 
-func (s storage) CreatePipeline(pipeline *model.Pipeline, stepList ...*model.Step) error {
+func (s *storage) CreatePipeline(pipeline *model.Pipeline, stepList ...*model.Step) error {
+	mutexInterface, _ := s.pipelineMutexes.LoadOrStore(pipeline.RepoID, &sync.Mutex{})
```
6543, line 133: where do you unload it ... else we have a memory leak

6543 (changes requested):

6543:
a database solution would be to set a write lock to that table at the beginning.

the only downside ... we have to write DBMS specific sql:
- mysql / mariadb:
  - https://dev.mysql.com/doc/refman/8.0/en/lock-tables.html
  - https://mariadb.com/kb/en/lock-tables/
  - https://mariadb.com/kb/en/unlock-tables/
- postgress:
  - https://www.postgresql.org/docs/current/sql-lock.html
- sqlite:
  - https://www.sqlite.org/lockingv3.html

6543:
| Lock Type | You Can | Others Can | Others Cannot |
|-----------|---------|------------|---------------|
| **WRITE** | Read & Write | Nothing | Read or Write |
| **READ** | Read only | Read | Write |

6543:
ok looking into documentation...

sqlite just needs `BEGIN EXCLUSIVE TRANSACTION;` at session start.

all others just need to append `… FOR UPDATE` to the select.

xoxys:
Yeah I know. Will you create a PR then?

6543:
tested `FOR UPDATE` ... it creates read lock so the other transaction waits and the fails :/

I go with table lock ...

lafriks:
Why not use unique index and retry on unique index exception?

6543:
sql is kind of an mess but the rest works just fine :) -> #5534

PR #5478, opened 2025-09-02
fix: changed schema definition for "backend_options.kubernetes.tolerations" to accept an array of objects

### LINKED ISSUE #5462: Linting `steps.XXX.backend_options.kubernetes.tolerations` fails
### Component

other

### Describe the bug

Lint a pipeline using `backend_options.kubernetes.tolerations` fails with `Invalid type. Expected: object, given: array`

### Steps to reproduce

```
# ...

variables:
  &kubernetes
      tolerations:
        - key: 'node-type'
          operator: 'Equal'
          value: 'pipeline'
          effect: 'NoSchedule'

steps:
  foo:
    backend_options:
      kubernetes:
        <<: *kubernetes
  bar:
    backend_options:
      kubernetes:
        <<: *kubernetes
# ...
```

### Expected behavior

Lint works as expected without warnings.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.9.0"}
```

### Additional context

_No response_

### Validations

klausi85:
Fixes #5462

Also added a regression test.


pipeline/frontend/yaml/linter/schema/schema.json
```diff
@@ -647,9 +647,12 @@
...
         "tolerations": {
-          "type": "object",
-          "additionalProperties": {
-            "type": ["boolean", "string", "number"]
+          "type": "array",
+          "items": {
+            "type": "object",
+            "additionalProperties": {
+              "type": ["boolean", "string", "number"]
+            }
```
siad007: @klausi85 Would you mind to also include a reference to the items used by this array? Something like:
```
        "tolerations": {
          "description": "The tolerations section defines a list of references to the native Kubernetes tolerations",
          "type": "array",
          "items": {
            "$ref": "#/definitions/step_kubernetes_toleration"
          },
          "minLength": 1
        },
```
And additional:
```
    "step_kubernetes_toleration": {
      "description": "XXX",
      "type": "object",
      "properties": {
        "key": {
          "type": "string"
        },
        // ...
      }
    }
```
Toleration struct: https://github.com/woodpecker-ci/woodpecker/blob/bd7314e6ba0e091ba1e54e9dc76cc121c4d600f8/pipeline/backend/kubernetes/backend_options.go#L30-L36

klausi85: @siad007 I've added the suggested schema changes and updated the test.

Issue #3386, opened 2024-02-13
Improve cli


anbraten:
The woodpecker cli could get some polishing:

- [x] updater #3382 
  - [x] background update checker
  - [x] update command with `--force` option
- [ ] interactive setup #3384 
  - [x] setup command 
  - [ ] command to enter server-url on first start 
  - [x] improved auth (opens browser with UI, send back token to localhost via POST) 
- [ ] follow step logs when step is still running
- [ ] improve error handling for missing args (missing repo id / name results in an parseInt error atm)
- [ ] improve outputs
  - [ ]  use some table layouts
- [ ] watch options
  - [ ] agents list
  - [ ] agent tasks
  - [ ] pipeline status
- [ ] improve exec command
  - [ ] support multiple workflow files
  - [ ] exec with metadata from UI #1155 
- [ ] detect repo from `.git/config` in pwd #910

xoxys:
@woodpecker-ci/maintainers I would like to work on `improve error handling for missing args (missing repo id / name results in an parseInt error atm)` How do we want to handle it? Do we really want to continue using args? IMO, it would be better to drop args completely and just use flags. Any reason to keep using args?

anbraten:
I think its quite common practice to use args (see kubectl, docker, ...) therefore I would suggest to stick to it and add a empty-string check with a more descriptive error in front of the `parseInt`.

xoxys:
Obviously, but I could also name a lot of tools without args. We could have a way more clean code without arg validation, as flag validation and required flags is a built-in feature of urfave/cli. However, if args is preferred, that works for me as well.

lafriks:
I don't know if urfave/cli still checks that required arguments are not empty if required (you can provide argument but provide it with empty value) so check for empty value could still be needed

xoxys:
There is no required args feature, just a required flags option.

jinnatar:
While the new setup flow seems nice, it utterly fails on a headless machine. As a quick fix I'd recommend echoing the URL that it's trying to apparently open in a browser or otherwise provide a flow where I can open the URL on another device and feed back a token on the cli.

6543:
yes we should check if the env has an wayland or xorg session, and if not we print out this sugestion and fail as we run "headless" ...

6543:
I have a dedicated `cli exec tui` summary issue: https://github.com/woodpecker-ci/woodpecker/issues/6488

where you can track what the current state & plan is :)

PR #5481, opened 2025-09-04
feat: expose listing available organizations via woodpecker-go / CLI

hhamalai:
Currently neither the woodpecker cli nor woodpecker-go don't expose functionality to expose listing Organizations while it's available on the HTTP API.

Mentioned in #3386

### DIFF
--- a/cli/org/org_list.go
+++ b/cli/org/org_list.go
@@ -1,0 +1,65 @@
+package org
+
+import (
+	"context"
+	"os"
+	"strings"
+	"text/template"
+
+	"github.com/urfave/cli/v3"
+	"go.woodpecker-ci.org/woodpecker/v3/cli/common"
+	"go.woodpecker-ci.org/woodpecker/v3/cli/internal"
+	"go.woodpecker-ci.org/woodpecker/v3/woodpecker-go/woodpecker"
+)
+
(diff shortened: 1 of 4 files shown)


cli/org/org_list.go
```diff
@@ -0,0 +1,65 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// Template for org list items.
+var tmplOrgList = "\x1b[33m{{ .Name }} \x1b[0m" + `
+Organization ID: {{ .ID }}
+{{- if .IsUser }}
```
qwerty287: This should never be necessary. The endpoint should only return those with `is_user=false`

qwerty287:
Thanks! Thinking about this again, this should actually go to the `admin` right? There are also user list etc. and in the web ui the org list is also under the admin settings.

hhamalai:
> Thinking about this again, this should actually go to the `admin` right? There are also user list etc. and in the web ui the org list is also under the admin settings.

You're right, the org package requires forge admin permissions, while listing organizations requires Woodpecker admin role. Moved this under admin command.

qwerty287:
Thanks. You have to regenerate the mock client though: https://ci.woodpecker-ci.org/repos/3780/pipeline/28156/34#L239

PR #5494, opened 2025-09-09
fix: allow spaces in `WOODPECKER_PLUGINS_PRIVILEGED`

Gusted:
In v3.8.0 the following was parsed correctly:
```
WOODPECKER_PLUGINS_PRIVILEGED="docker.io/woodpeckerci/plugin-docker-buildx:latest, docker.io/woodpeckerci/plugin-docker-buildx:6.0.2"
```

But in v3.9.0 this would not be parsed correctly and results in the following linter error:
```
The formerly privileged plugin `woodpeckerci/plugin-docker-buildx:6.0.2` is no longer privileged by default, if required, add it to `WOODPECKER_PLUGINS_PRIVILEGED`
```

Regression of 59a855c7b9529a64d66461be100617b70d9d0d3e

### DIFF
--- a/cli/admin/secret/secret_add.go
+++ b/cli/admin/secret/secret_add.go
@@ -41,10 +41,16 @@ var secretCreateCmd = &cli.Command{
 		&cli.StringSliceFlag{
 			Name:  "event",
 			Usage: "secret limited to these events",
+			Config: cli.StringConfig{
+				TrimSpace: true,
+			},
 		},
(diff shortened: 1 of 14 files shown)


qwerty287:
Thanks!

I guess we have more like this (the `StringSliceFlag`, I'm not sure if we have more but I'd be surprised if not), could you update this as well? If not I can take care of in the next days.

Issue #5168, opened 2025-05-10
Skipping branches without a Woodpecker config broken


xoxys:
Branches without a woodpecker config should be ignored as implemented in https://github.com/woodpecker-ci/woodpecker/pull/2949

Current state (at least for gitea):

![Image](https://github.com/user-attachments/assets/2f773a0d-d046-4ff8-8309-31e92041a5e0)

Version: `next-11abc7c44b`

xoxys:
This bug is pretty annoying as it also occures on every PR close event now...

6543:
well the config fetcher might use custom errors to signal that the config fetch is to be ignored but ... we did not pass the correct error upwards the stack :/

 #5518

6543:
#5518 might actually solve this or at least we know why it fails now 😆

PR #5501, opened 2025-09-11
Show readable error

qwerty287:
to help debugging issues like https://github.com/woodpecker-ci/woodpecker/issues/5168.

### DIFF
--- a/server/pipeline/create.go
+++ b/server/pipeline/create.go
@@ -88,7 +88,7 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline 
 		return nil, ErrFiltered
 	} else if configFetchErr != nil {
 		log.Error().Str("repo", repo.FullName).Err(configFetchErr).Msgf("error while fetching config '%s' in '%s' with user: '%s'", repo.Config, pipeline.Ref, repoUser.Login)
-		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge: %w", err))
+		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge: %s", err.Error()))
 	}
 
 	pipelineItems, parseErr := parsePipeline(_forge, _store, pipeline, repoUser, repo, forgeYamlConfigs, nil)


server/pipeline/create.go
```diff
@@ -88,7 +88,7 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline
 		return nil, ErrFiltered
 	} else if configFetchErr != nil {
 		log.Error().Str("repo", repo.FullName).Err(configFetchErr).Msgf("error while fetching config '%s' in '%s' with user: '%s'", repo.Config, pipeline.Ref, repoUser.Login)
-		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge: %w", err))
+		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge: %s", err.Error()))
```
6543, line 91:
```diff
-		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge: %s", err.Error()))
+		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge (%w): %s", err, err.Error()))
```

as now it is redable but we can not pass up errors to be asserted/catched specialy ahndled etc!!!

6543: normally %w should be equal %s but allow to get the error back if needed or assert if it is the underlaying.
it it does not format nicely this is the better way to do to not lose that bit of information

anbraten: Could you create a PR for you suggestion. Might be best to add a test for it as well.

6543: https://github.com/woodpecker-ci/woodpecker/pull/5518

6543: a unit test would not catch those issues.

full integration test suite will, but i paused my work on that :/

xoxys: Don't understand your comments. Have you read the linked issue? As you can see, the error message is not properly formatted at all and therefor error information got lost. Yes, wrapping errors is better to be able to assert some error types, but what's your suggestion to keep error wrapping **and** fix the linked issue where the error message was not shown at all?

6543: PS: yes you probably could write an unit test but that's just for this codeblock and it can just be done wront at all other places at another time -> thats why reviews are here :D

i dont think it is an hard issue and i did not test if we use some errors.Is or errors.As up in the caller stack, but its more a contract that somebody might at some point do so - e.g. we should expect golangs error handling system to work

6543: @xoxys mv to -> https://github.com/woodpecker-ci/woodpecker/pull/5518#issuecomment-3297356299

PR #5507, opened 2025-09-14
Add log service addon

qwerty287:
provide a more flexible approach to #5482 
Works just like the forge addons.


xoxys:
> but is it also possible to load more than one file at a time?

IMO that's a must for an addon system.

Levy-Tal:
Cool, this still doesn’t solve the problem that the addon can’t know if the step is finished from the log chunk or from the *model.Step when the last logAppend is called.  
Would it make sense to add a log lifecycle, like `logStart` or `logFinished`?  
Or do you think it would be better to create a separate pull request for this?

Levy-Tal:
Also, I really like the addon approach!  
Just throwing out an idea — this could also be extended to a “secret store”  
For example, someone could securely save secrets in Vault, AWS Secrets Manager, etc.

qwerty287:
> IMO that's a must for an addon system.

The system supports having multiple plugins in one binary. But that would require everyone who uses multiple plugins to compile their own correct "mix" of plugins.

> this could also be extended to a “secret store”

For secrets, I'd rather use an http-based approach because then, this can also be added on a repo-level and not only globally.

qwerty287:
> Cool, this still doesn’t solve the problem that the addon can’t know if the step is finished from the log chunk or from the *model.Step when the last logAppend is called.  
> Would it make sense to add a log lifecycle, like `logStart` or `logFinished`?  
> Or do you think it would be better to create a separate pull request for this?

Should be done in a separate PR.

My idea would be to add a new function to the interface like `StepFinished`.

xoxys:
> The system supports having multiple plugins in one binary. But that would require everyone who uses multiple plugins to compile their own correct "mix" of plugins.

No way around it? That's a pretty bad UX.

qwerty287:
No, I don't know. Didn't try it. Maybe (I hope so) it also works with more than one plugin binary.

anbraten:
> No, I don't know. Didn't try it. Maybe (I hope so) it also works with more than one plugin binary.

It should be possible to load multiple binaries, otherwise that would be to difficult for 99% of our users to run such plugins and no-one would really share a helpful plugin, resulting in the feature not being used a lot.

In general I am still not 100% happy with the addon / plugin mechanisms we have atm:
- HTTP extensions (config service): 
  - (+) pretty easy to implement
  - (+) can be easily shared (container image, public hosted api, some binary you ship, ... endless ways to share such a service)
  - (-) hosting them requires additional containers etc or external hosting (needs to be allowed by the admin)
- Binary addons (as used by the addon forge):
  - (+) can be quite powerful and are probably a bit more performant than http plugins
  - (-) difficult to share as they need to be compiled with same version of the server / architecture  
  - (-) need to be installed by admins, can not be installed by users as necessary for #3349
 
Some notes of my thoughts on the plugin topic (__no concrete ideas / directions yet__):

We had a look at https://github.com/knqyf263/go-plugin some time ago, there seems to be sth similar in istio: https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin
There was also the idea to use agents to host containers that would run them like pipeline services that never stop (but that would be quite some journey as well).

qwerty287:
> (-) difficult to share as they need to be compiled with same version of the server / architecture

No. This affefcts Go's internal plugins system. This one uses RPC and as long as this does not change there should be no incompatibiltiy.

qwerty287:
Tested it, it is working with both log and forge addon configured. So no problem there. Also added some more docs.

What we need to think about is where we put the user docs for addons as that's quite an advanced feature. Right now it was below the forges but that doesn't fit anymore after this.

xoxys:
> What we need to think about is where we put the user docs for addons as that's quite an advanced feature. Right now it was below the forges but that doesn't fit anymore after this.

I would move it to Configuration > Addons

docs/docs/30-administration/10-configuration/10-server.md
```diff
@@ -1121,7 +1121,11 @@ Disable version check in admin web UI.
 - Name: `WOODPECKER_LOG_STORE`
 - Default: `database`
 
-Where to store logs. Possible values: `database` or `file`.
+Where to store logs. Possible values:
+
+- `database`: stores the logs in the database
+- `file`: stores logs in JSON files on the files system
+- `addon`: uses a separate addon to store logs. See [Addons](./100-addons.md#log)
```
xoxys:
```suggestion
- `addon`: uses an [addon](./100-addons.md#log) to store logs
```

docs/docs/92-development/100-addons.md
```diff
@@ -38,7 +20,7 @@ This example will use the Go language.
 
 Directly import Woodpecker's Go packages (`go.woodpecker-ci.org/woodpecker/v3`) and use the interfaces and types defined there.
 
-In the `main` function, just call `"go.woodpecker-ci.org/woodpecker/v3/server/forge/addon".Serve` with a `"go.woodpecker-ci.org/woodpecker/v3/server/forge".Forge` as argument.
+In the `main` function, just call the `Serve` in the corresponding addon package with the service as argument (see [below](#addon-types)).
```
xoxys:
```suggestion
In the `main` function, just call the `Serve` method in the corresponding [addon package](#addon-types) with the service as argument.
```

server/services/log/addon/client.go
```diff
@@ -0,0 +1,116 @@
...
+		HandshakeConfig: HandshakeConfig,
+		Plugins: map[string]plugin.Plugin{
+			pluginKey: &Plugin{},
+		},
+		Cmd: exec.Command(file),
+		Logger: &logger.AddonClientLogger{
+			Logger: log.With().Str("addon", file).Logger(),
+		},
+	})
+	// TODO: defer client.Kill()
```
xoxys, line 44: Just a copy/paste from addon forges, but why is this not implemented but a todo?

qwerty287: Because it has to be executed when the server exits. And I think there was no logic for this when I built the forge add-on. But I think there's something now for this so I can actually check this out. Would it be fine in a separate PR?

xoxys: Yes sure 👍

Issue #2948, opened 2023-12-14
Link queue entries to pipeline


anbraten:
Its sometimes pretty difficult to debug a queue entry and understand what it is about. Therefore it might help to link the pipeline the queue entry is related to.

6543:
TODOs:
- [ ] model.task: add workflow_id to get a link in qeueu view
- [ ] change webui acordingly

Issue #5469, opened 2025-08-29
In the queue list, use the agent name rather than the agent id for the badge.


dessalines:
### Clear and concise description of the problem

The tasks queue currently shows a badge, which shows the agent running the task:

<img width="314" height="147" alt="Image" src="https://github.com/user-attachments/assets/7626d7ca-ba88-4fa7-b6f5-6baf65ce0efa" />

This id number isn't helpful for viewing, as people don't have the IDs of all their CI runners memorized. It'd be much better to use the agent name (or alias / whatever you tagged it as) instead, and possibly default to the ID if no name is present.

### Suggested solution

Use the agent name / alias instead of an id number for the queue badge.

### Validations

- [x] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/versions]
- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

PR #5516, opened 2025-09-15
Show human readable information in queue info

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/2948, https://github.com/woodpecker-ci/woodpecker/issues/5469

Bildschirmfoto 2025-09-15 um 15 18 46


server/store/datastore/migration/026_add_task_fields.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"src.techknowlogick.com/xormigrate"
+	"xorm.io/xorm"
+)
+
+var addTaskFields = xormigrate.Migration{
+	ID: "add-task-fields",
+	MigrateSession: func(sess *xorm.Session) (err error) {
+		// Add the name column to the tasks table
```
anbraten: The bean syncing should handle this already.

server/pipeline/queue.go
```diff
@@ -33,8 +33,11 @@ func queuePipeline(ctx context.Context, repo *model.Repo, pipelineItems []*stepb
 			continue
 		}
 		task := &model.Task{
-			ID:     fmt.Sprint(item.Workflow.ID),
-			Labels: make(map[string]string),
+			ID:         fmt.Sprint(item.Workflow.ID),
```
anbraten, line 36: tasks are passed to agents and probably need a grpc version bump therefore.

server/queue/fifo.go
```diff
@@ -198,23 +199,50 @@ func (q *fifo) Extend(_ context.Context, agentID int64, taskID string) error {
...
+func (q *fifo) Info(c context.Context) InfoT {
 	q.Lock()
 	stats := InfoT{}
 	stats.Stats.Workers = len(q.workers)
 	stats.Stats.Pending = q.pending.Len()
 	stats.Stats.WaitingOnDeps = q.waitingOnDeps.Len()
 	stats.Stats.Running = len(q.running)
 
+	// Create a map of agent IDs to names
+	agentNames := make(map[int64]string)
```
anbraten: Would suggest to query the agent endpoint for the names using the api instead for normalized data structures.

server/store/datastore/migration/026_add_task_fields.go
xoxys: Good point, in that case we can drop the entire migration?

server/queue/fifo.go
6543: the only concern is, that this mixes queue logic ans store logic as the queue was not aware of the store and you could move out the implementation to a completly different backend ... If we have only one caller of Info() that needs this info we should move the query logic of that info to that place ...

xoxys: Whats the benefit of an extra http call here instead of talking to the store directly?

anbraten: Normalized data in the api helps with data handling in the UI as for example updating an agent would result in automatically showing the updated agent name and thereby creates cleaner api models. In the end its a tradeoff between fewer api calls and mixed up data models.

web/src/views/admin/AdminQueue.vue
```diff
@@ -57,18 +57,33 @@
...
               />
-              <span>{{ task.id }}</span>
+              <span>{{ task.name }}</span>
             </div>
-            <div class="flex items-center">
-              <span class="ml-auto flex gap-2">
-                <Badge v-if="task.agent_id !== 0" :label="$t('admin.settings.queue.agent')" :value="task.agent_id" />
+            <div class="ml-auto flex items-center gap-2">
+              <span class="flex gap-2">
+                <Badge
```
anbraten: We could add a link to the agent for this badge.

xoxys: Not really. The agent itself has no linkable URL, we can just link to the agents list view. Dont know how helpful this is.

anbraten: I see, we would need to add separate routes for the agent settings first. Will open a PR for that.

xoxys: But then you will end up in the edit view of this agent? Still dont know if thats a good approach.

xoxys: Anyway can be done in a dedicated PR.

server/pipeline/queue.go
xoxys: Done.

xoxys:
@anbraten @6543 Can I get another review? 👼

server/model/task.go
```diff
@@ -25,12 +25,16 @@ import (
...
 type Task struct {
 	ID           string                 `json:"id"           xorm:"PK UNIQUE 'id'"`
+	Name         string                 `json:"name"         xorm:"'name'"`
 	Data         []byte                 `json:"-"            xorm:"LONGBLOB 'data'"`
 	Labels       map[string]string      `json:"labels"       xorm:"json 'labels'"`
 	Dependencies []string               `json:"dependencies" xorm:"json 'dependencies'"`
 	RunOn        []string               `json:"run_on"       xorm:"json 'run_on'"`
 	DepStatus    map[string]StatusValue `json:"dep_status"   xorm:"json 'dependencies_status'"`
-	AgentID      int64                  `json:"agent_id"     xorm:"'agent_id'"`
+	AgentID      int64                  `json:"agent_id"     xorm:"-"`
```
anbraten: I assume now the agent-id would not survive a restart and for example the autoscaler would get a hickup then.

xoxys: The task is only stored in the DB as long as it is pending and at this state it does not have an agent ID assigned anyway. As soon as a queue task is getting started it is removed from the table. So I assumed this is useless, but I can change it back.

xoxys: This is just the tasks table, not the agents table.

qwerty287:
I don't have time for a proper review right now, but I was wondering: What happens if you change the name of an agent? Is this reflected on the queue page?

xoxys:
No with the current way this does not update the name for __already running__ tasks.

xoxys:
To get the agent name updated immediately, we would need to look up the agent name by ID on every queue info call. There are multiple methods to do it:

1) Using the store as done with https://github.com/woodpecker-ci/woodpecker/pull/5516/commits/9d406f986a8ad80088481fcc77d4176e47a84b4b#diff-2d7687e7cf4d8db83805f8950671e37eb0af7d3cf0f470b5ae0b78ad93985401 
2) Using an http call to GET `/agents/{agent_id}` If we want to use the http call instead of the store call, which base URL should be used? Making an external call to the public URL is IMO.

server/model/task.go
xoxys: Reverted

xoxys:
OK I think I found a better way. The agent name lookup is now done on the API level within the `GetQueueInfo` method. It still uses a store call instead of http call, however I guess on this level its acceptable. This also reflects agent name changes immediatly.

qwerty287:
Thanks, could we do that on sql level with join?

xoxys:
No, at least I dont see a way how this could work (easily).

qwerty287:
Ah you're right, the queue info is not loaded from db, sorry.
But at least some kind of cache for the agents would be good I think? So we don't query the DB multiple times for the same agent. So with a `map[int]*model.Agent`

xoxys:
I had this already but this end up with a lot more code. But I can push it again.

anbraten:
> I had this already but this end up with a lot more code. But I can push it again.

As this introduces such new issues and to keep the api clean, I would still suggest querying the `/agents` list from the UI for now. We can still add an api property later on, but removing one would be a breaking change again.

xoxys:
Ok here we go :D added some helper functions to declutter the enrichment while using the cache map approach.

xoxys:
> As this introduces such new issues and to keep the api clean, I would still suggest querying the /agents list from the UI for now. We can still add an api property later on, but removing one would be a breaking change again.

Why? IMO the latest change is a good one and I dont see any benefit to add it to the UI instead.

qwerty287:
I agree to @xoxys. Querying the agents list should not working as well because of pagination (of course you could work around this by just loading all pages).

6543 (approved):
backend code ack - thanks :)

anbraten:
> Why? IMO the latest change is a good one and I dont see any benefit to add it to the UI instead.

Adding a property as agent_name to the model while having agent_id already is a mix of models IMO as this property adds no new information that was not queryable before and is rather adjusting the model to match our UI views than providing REST models. Feels more like the graphql approach to me. We also need to keep in mind that those properties are send to the agent via grpc (should not be to crazy for a string yet) and api property removals are considered breaking if we change plans. Won't block if the rest is fine with it.

6543:
well the problem is that either we have models reused or we have type definition for each interface that marks a border to another component of or software to make things more modulare and have cleare seperation of concerns ...

... as there was already the complaint why we have so many "similar" type definitions and the server component is kind of tangled anyway right now i dont care much ... about that concern

about that I'd love to have modules more clear destinct but at the end of the day it still is a single pice of software that we get by cobine our components ...

this here dont save agent_name to database so only if we wana change the json consumers are affected ...

_most probably offtopic but the reason why i'm fine with that mix (through if someone works on better cristalizing out components ... i would welcome that)_

xoxys:
Alright, if the `Task` model is that fragile (due to the usage in rpc calls), we should not publish it via API directly. I have added a custom type `QueueTask` now used in the API and enrich this one with agent names.

xoxys:
@anbraten @6543 what do you think about the current approach?

web/src/assets/locales/en.json
```diff
@@ -341,7 +342,10 @@
           "badge": "org"
         },
         "version": "Version",
-        "last_contact": "Last contact",
+        "last_contact": {
+          "last_contact": "Last contact",
+          "badge": "last_contact"
```
qwerty287: I think this just should be lowercase
```suggestion
          "badge": "last contact"
```

server/api/hook.go
```diff
@@ -37,16 +37,91 @@ import (
...
+	name, exists := agentNameMap[agentID]
+	if exists {
+		return name, true
+	}
+
+	// 2. If not in cache, query the store.
+	agent, err := store.AgentFind(agentID)
+	if err != nil || agent == nil {
+		// Agent not found or an error occurred.
+		return "", false
```
qwerty287, line 106: Should we report that error? At least it should get logged. I think it would also be fine to let the http request fail then. The agent should exist.

qwerty287:
I like how it's now.

web/src/assets/locales/en.json
xoxys: Badges are kinda technical key value pairs, I would keep the underline.

qwerty287: I'd disagree... What about others? @woodpecker-ci/maintainers

xoxys: No strong opinion on that

server/api/hook.go
xoxys: Done

xoxys:
@woodpecker-ci/maintainers Would like to get a final review

6543 (approved):
ACK!!!

exactly how it should be (backendwise)

web/src/lib/api/types/queue.ts
```diff
@@ -1,10 +1,14 @@
 export interface Task {
   id: number;
+  name: string;
```
qwerty287, line 4: Can we rename this to `workflow_name`?

web/src/views/admin/AdminQueue.vue
```diff
@@ -57,18 +57,29 @@
...
               </span>
             </div>
+            <div class="ml-2 flex items-center gap-2">
+              <IconButton
+                icon="chevron-right"
+                :title="$t('repo.pipeline.view')"
+                class="h-8 w-8"
+                :to="{
+                  name: 'repo-pipeline',
+                  params: { repoId: task.repo_id, pipelineId: task.pipeline_id },
```
qwerty287: This only links to the pipeline, not the workflow - I'd directly link to the workflow…

xoxys: Thats not possible. There is not workflow route in the UI

qwerty287: It is. E.g. the links that are sent to the forge: https://ci.woodpecker-ci.org/repos/3780/pipeline/28362/1

It looks exactly like the step route but the pid `1` actually is the workflow pid. In the UI, always the first step is opened (so usually the clone step)

web/src/lib/api/types/queue.ts
xoxys: I would keep it. The entire task object is meant to be workflow-related, e.g. the ID is already the workflow ID. Just renaming name now to workflow_name makes the model semantically more confusing.

web/src/views/admin/AdminQueue.vue
xoxys: I still dont know how to construct this URL on the UI.... I have the workflow id and the pipeline id

xoxys: Got it... Had to extend the Info API even further.

anbraten:
Tasks that have not been picked by an agent point to a not existing pipeline: https://ci.woodpecker-ci.org/repos/3780/pipeline/0/4 (works as soon as an agent picks the task)

The border could be using some darker color in dark-mode:
image

FYI @xoxys

xoxys:
Thx for the feedback, will take care 👍

PR #5518, opened 2025-09-16
Use correct error when config fetching fails

6543:
https://github.com/woodpecker-ci/woodpecker/pull/5501/files#r2351742200

normally %w should be equal %s but allow to get the error back if needed or assert if it is the underlaying.
it it does not format nicely this is the better way to do to not lose that bit of information

### DIFF
--- a/server/pipeline/create.go
+++ b/server/pipeline/create.go
@@ -88,7 +88,7 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline 
 		return nil, ErrFiltered
 	} else if configFetchErr != nil {
 		log.Error().Str("repo", repo.FullName).Err(configFetchErr).Msgf("error while fetching config '%s' in '%s' with user: '%s'", repo.Config, pipeline.Ref, repoUser.Login)
-		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge: %s", err.Error()))
+		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge (%w): %s", err, err.Error()))
 	}
 
 	pipelineItems, parseErr := parsePipeline(_forge, _store, pipeline, repoUser, repo, forgeYamlConfigs, nil)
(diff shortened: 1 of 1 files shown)


server/pipeline/create.go
```diff
@@ -88,7 +88,7 @@ func Create(ctx context.Context, _store store.Store, repo *model.Repo, pipeline
 		return nil, ErrFiltered
 	} else if configFetchErr != nil {
 		log.Error().Str("repo", repo.FullName).Err(configFetchErr).Msgf("error while fetching config '%s' in '%s' with user: '%s'", repo.Config, pipeline.Ref, repoUser.Login)
-		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge: %s", err.Error()))
+		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge (%w): %s", err, err.Error()))
```
anbraten: Please add a test for it

6543: a unit test would not catch those issues.

full integration test suite will, but i paused my work on that :/

PS: yes you probably could write an unit test but that's just for this codeblock and it can just be done wront at all other places at another time -> thats why reviews are here :D

i dont think it is an hard issue and i did not test if we use some errors.Is or errors.As up in the caller stack, but its more a contract that somebody might at some point do so - e.g. we should expect golangs error handling system to work

6543:
%w should be like %s but something here got wrong:

image

this might be due to json.Marshal not use err.Error() but some generic stringing atempt.
anyway if we serialize errors to json we should better manually convert them to string ... I have not looked into where it gets serialized

qwerty287:
Could you explain why there is this weird `%!w()` then? If we always have that it's useless at all

6543:
My guess is: it has to do wit json serialisation ...

6543:
normally this works just fine

# example: https://play.golang.com/p/8OUzOaka0J0

6543:
# found it!!!!

https://play.golang.com/p/OEJXRlQM47Y

```
package main

import (
	"fmt"
)

func main() {
	var baseErr error = nil

	testErr := fmt.Errorf("some other but wrap that: %w", baseErr)
	fmt.Println("Wrapped error:", testErr)
}
```
```
Wrapped error: some other but wrap that: %!w(<nil>)
```

server/pipeline/create.go
6543:
```suggestion
		return nil, updatePipelineWithErr(ctx, _forge, _store, pipeline, repo, repoUser, fmt.Errorf("could not load config from forge: %w", configFetchErr))
```

anbraten (approved):

6543:
if you have a better wording ... please rename the title :)

6543:
as for #5168 we might actually fix it with this or at least know why it fails now :laughing: :rofl:

6543:
we need integration test suite :laughing: -> #3262

PR #5519, opened 2025-09-16
Fix dead page and cleanup router

anbraten:
The `org-settings` was just showing an empty page as there is no "general" page, I therefore linked the first page (secrets).

Added types for the route meta and changed the default to require authentication.

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="app bg-wp-background-100 dark:bg-wp-background-300 m-auto flex h-full w-full flex-col">
-    <router-view v-if="blank" />
+    <router-view v-if="layout === 'blank'" />
     <template v-else>
       <Navbar />
       <main class="relative flex h-full min-h-0">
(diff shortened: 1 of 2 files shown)


xoxys (changes requested):
This change breaks the login screen. It just doesnt load anymore and hangs forever.

qwerty287:
`org-settings` was added as "meta-screen" and to have it consistent with other settings.

If we have general settings for orgs they should go there…

xoxys:
Yeah but then we should add an empty dummy page with at least a header. Just showing a blank screen without even the tab navigation for org settings looks broken.

anbraten:
We can change it to a general page if we have one, therefore I left the route name currently being a redirect to secrets.

Currently it looks like this and made me initially thinking the page is broken somehow: https://ci.woodpecker-ci.org/orgs/2/settings

anbraten:
Did some adjustments and fixed the issue.
What I've tested:
- login to Woodpecker
- open a public page `/repos/8990/branches` and click on login in navbar => redirects to original page
- open a public page and go to home => redirects to login
- open /login while being authenticated => redirects to home
- open a private page  `/repos/8990/settings`  without being logged in => redirects to login and returns to page after login

xoxys:
Ill give it a test. Thanks.

PR #5529, opened 2025-09-20
Fix prettier configs

qwerty287:


qwerty287:
Somebody has an idea why this is failing? https://ci.woodpecker-ci.org/repos/3780/pipeline/28352/9

Plugin and prettier are installed in the same (global) directory

xoxys (approved):

PR #5530, opened 2025-09-20
Add `StepFinished` to log service

qwerty287:
Allow to notify the log service that a step finished.

See #5507 #5482


Makefile
```diff
@@ -116,7 +116,8 @@ generate: install-tools generate-openapi ## Run all code generations
 	CGO_ENABLED=0 go generate ./...
 
 generate-openapi: install-tools ## Run openapi code generation and format it
-	go run github.com/swaggo/swag/cmd/swag fmt
```
xoxys: I think we can revert this change, I addressed the issue with an exclude in the mockery PR.

xoxys (approved):

PR #5533, opened 2025-09-21
Fix dir not found handling

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/5168


anbraten:
Would this work for Gitlab, Github and the others as well?

xoxys:
Already done on other forges https://github.com/woodpecker-ci/woodpecker/blob/main/server%2Fforge%2Fgithub%2Fgithub.go#L258

PR #5534, opened 2025-09-21
Aquire write lock for pipeline creation into the database

### LINKED ISSUE #3884: Likely concurrency issue pq: 'duplicate key value violates unique constraint \"UQE_pipelines_s\"'

6543:
closes #5477
fixes #3884

### DIFF
--- a/server/store/datastore/pipeline.go
+++ b/server/store/datastore/pipeline.go
@@ -15,10 +15,13 @@
 package datastore
 
 import (
+	"fmt"
+	"sync"
 	"time"
 
 	"xorm.io/builder"
 	"xorm.io/xorm"
+	"xorm.io/xorm/schemas"
 
 	"go.woodpecker-ci.org/woodpecker/v3/server/model"
 )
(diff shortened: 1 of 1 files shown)


server/store/datastore/pipeline.go
```diff
@@ -127,7 +130,16 @@ func (s storage) GetPipelineCount() (int64, error) {
 	return s.engine.Count(new(model.Pipeline))
 }
 
+// createPipelineSQLiteLock is needed for CreatePipeline if sqlite is used
+// as sqlite don't support to upgrade the transaction lock to be exclusive
```
6543:
```suggestion
// as sqlite don't support to upgrade the transaction lock to be exclusive and xorm don't support EXCLUSIVE transactions jet
```

xoxys:
@6543 I like this approach. Can you fix lint and tests?

6543:
the DB error is an xorm issue ... let me look into it - I think it might lead to an upstream fix ;)

server/store/datastore/pipeline.go
```diff
@@ -143,15 +156,41 @@ func (s storage) CreatePipeline(pipeline *model.Pipeline, stepList ...*model.Ste
...
 	}
 
+	// get write lock
+	// TODO: upstream that to xorm
+	switch s.engine.Dialect().URI().DBType {
+	case schemas.SQLITE:
+		// we have an exclusive lock via createPipelineSQLiteLock already
+
+	case schemas.MYSQL:
+		if _, err := sess.Exec("LOCK TABLE `pipelines` WRITE"); err != nil {
```
6543, line 167: we cant do that or we dont have support for mariadb cluster

6543: https://mariadb.com/docs/server/reference/sql-statements/transactions/lock-tables#limitations

also tested -> ack :/

xoxys:
What about the soft-lock approach https://github.com/woodpecker-ci/woodpecker/pull/5477#issuecomment-3315966457 from @lafriks

6543:
thats how gitea does it and my next aproach ...

xoxys:
Thanks for working on this task!

hhamalai:
Is this fix being actively worked with @6543 optimistic locking approach? Our builds often fails to start from webhooks due to this issue. Either this or #5477 should be worked to finish to resolve the issue, but which one core maintainers currently prefer?

6543:
merged self contained hotfix: https://github.com/woodpecker-ci/woodpecker/pull/6067#issuecomment-3854045200 till i get time to do this here propperly

6543:
NOTE: a propper locking should use sql (beside sqlite3) as woodpecker-server at some point should be culser capapble

PR #5540, opened 2025-09-22
Fix formatting

xoxys:
We should still version golangci-lint used in makefile.


xoxys:
> We should still version golangci-lint used in makefile.

Done now.

6543:
thanks!

PR #5543, opened 2025-09-23
Add packageManager to packages.json

6543:
see e.g. https://ci.woodpecker-ci.org/repos/3780/pipeline/28429/9


xoxys:
Its just a warning, if we add it everybody has to use the same pnpm version and I dont see any benefit. If we still want to have it, we should revert https://github.com/woodpecker-ci/woodpecker/pull/4581 as well.

6543:
ok

Issue #5402, opened 2025-07-30
Support affinity for job pods


dunn:
Is it possible to specify affinity rules for job pods?  I currently only have RWO volumes, so in order for docker-in-docker to work, I need both the docker server and worker pods to be running on the same node, since they need to both mount the certificate volume.

xoxys:
IIRC not possible yet. This would need to be implemented in the Kubernetes backend of the server. Ill move the issue.

qwerty287:
Is that related or even a duplicate of https://github.com/woodpecker-ci/woodpecker/issues/3070 ?

henkka:
Based on my analysis of the codebase, I believe `taskUUID` serves as the workflow identifier that groups related steps together, which  when attached to the pod labels, we can use as a affinity/anti-affinity rules. If this understanding is correct, I have a potential solution implemented in this branch: https://github.com/woodpecker-ci/woodpecker/compare/main...henkka:woodpecker:k8s-job-affinity

if anyone can confirm that I'm on the right track it would be nice. I'm happy to clean up the implementation and submit a proper pull request.

6543:
taskUUID == workflow Identifyer yes :)

qwerty287:
I guess this can be closed after https://github.com/woodpecker-ci/woodpecker/pull/5854

PR #4796, opened 2025-01-31
Tag pipeline with source information

jenrik:
_This is the seconds attempt trying to get this to work._

Indirectly adds labels to Kubernetes pods for use with NetworkPolicies and other integrations.

TODO:
- [x] Add new labels to agent filters https://github.com/woodpecker-ci/woodpecker/blob/00aa968afa9c66b23d5c91bd7cc32098ead142b5/server/grpc/filter.go#L32
- [x] Include pipeline labels in Workflow RPC, so they can be added to containers and pods

PR #5544, opened 2025-09-23
Add task UUID label to Kubernetes pods

henkka:
Slightly relates to https://github.com/woodpecker-ci/woodpecker/issues/5402 and https://github.com/woodpecker-ci/woodpecker/issues/3344

# Problem

When using the Kubernetes backend with ReadWriteOnce PersistentVolumes, volumes can only be attached to one Kubernetes worker node at a time. This creates scheduling conflicts when workflow steps are scheduled to run on different nodes (due to worker node size constraints for example).

# Background

The somewhat ideal solution would be to implement Kubernetes PodAffinity and PodAntiAffinity rules directly in Woodpecker to ensure:
- All pods in the same workflow run on the same node (to share PVs)
- Pods from different workflows avoid the same node (to prevent conflicts)

However, a previous attempt to add this functionality ([PR #3345](https://github.com/woodpecker-ci/woodpecker/pull/3345)) was declined as there's different views how the root problem should be resolved. In our opinion, https://github.com/woodpecker-ci/woodpecker/discussions/3366 sounds like a good approach, but at this moment we cannot afford to wait for implementing such a major change.

# Proposed Solution

Instead of modifying Woodpecker's core scheduling logic, this PR enables external solutions by adding a `woodpecker-ci.org/task-uuid` label to each pod. This label:

- Contains the workflow/task UUID (identical for all pods in the same workflow)
- Allows external tools like Kubernetes [MutatingAdmissionWebhooks](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) to inject appropriate affinity rules to the pods scheduled by Woodpecker Agent
- Provides a clean separation of concerns between Woodpecker and cluster-specific scheduling policies

TL;DR: We figured this approach is the path of least resistance that allows Woodpecker CI users to overcome the background issue without big changes to Woodpecker CI project itself.

### DIFF
--- a/docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
+++ b/docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
@@ -19,6 +19,8 @@ The following metadata labels are supported:
 - `woodpecker-ci.org/repo-full-name`
 - `woodpecker-ci.org/branch`
 - `woodpecker-ci.org/org-id`
+- `woodpecker-ci.org/task-uuid`
+- `woodpecker-ci.org/step`
 
 ## Private registries
 


docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
```diff
@@ -19,6 +19,8 @@ The following metadata labels are supported:
 - `woodpecker-ci.org/repo-full-name`
 - `woodpecker-ci.org/branch`
 - `woodpecker-ci.org/org-id`
+- `woodpecker-ci.org/task-uuid`
+- `woodpecker-ci.org/step`
```
henkka, line 23: https://github.com/woodpecker-ci/woodpecker/pull/4796 forgot to update this, so I added it to the docs as well

henkka:
Thank you for the really quick turnaround on this pull request! We're amazed by how well this project is run and how easily we can contribute to it.

PR #5545, opened 2025-09-23
Display workflow link in queue only on assigned tasks

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/pull/5516#issuecomment-3324192429

grafik

### DIFF
--- a/web/src/views/admin/AdminQueue.vue
+++ b/web/src/views/admin/AdminQueue.vue
@@ -32,7 +32,9 @@
           :key="task.id"
           class="bg-wp-background-200! dark:bg-wp-background-200! mb-2 flex-col items-center gap-4"
         >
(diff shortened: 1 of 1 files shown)


anbraten:
Not sure if this fix is ideal. The pipeline is known as soon as we create it from the webhook call. I think the issue is coming from https://github.com/woodpecker-ci/woodpecker/blob/32a70847f3f11465aedfd9863c3022329739f457/server/api/hook.go#L127-L129 skipping the pipeline number assignment in case there was no agent https://github.com/woodpecker-ci/woodpecker/blob/32a70847f3f11465aedfd9863c3022329739f457/server/api/hook.go#L144 
Instead it should only skip the pipeline number assignment in case of `task.PipelineID` being 0, but that should actually never be the case.

@xoxys

xoxys:
Fair point, Ill take another look 👍

PR #5547, opened 2025-09-23
Migrate to mockery v3

xoxys:
The inline go configuration for mockery was deprecated and was removed in v3. Migrate mocks to mockery packages config. I decided to adapt the mockery config we already use in `woodpecker/autosclaer` even if that leads to some more changes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -37,7 +37,7 @@
               glibc.static
               gofumpt
               golangci-lint
-              go-mockery_2
+              go-mockery
               protobuf
               sqlite
               go-swag # for generate-openapi


6543:
haha thanks for tackling this ...
... could you also change tje mockerly version pinned in the flake? (or should i justvpush onto your branch?)

xoxys:
@6543 The PR is ready to merge from my side. Feel free do commit nix stuff to this branch, but you can also do it in a separate one.

qwerty287:
Thanks, this warning is bothering me for a while already but I never was motivated enough to fix it 😉

PR #5550, opened 2025-09-25
Add back-off retry for pod log streaming to kubernetes backend

henkka:
# Problem

Woodpecker CI fails with 

```
rpc error: code = Unknown desc = workflow finished with error Get "https://10.0.9.147:10250/containerLogs/...": remote error: tls: internal error
```

when trying to read container logs from newly provisioned Kubernetes worker nodes. This occurs because:

- New worker nodes require CSR (Certificate Signing Request) approval before kubelet can serve TLS requests
- Woodpecker attempts to read logs immediately
- Race condition: log collection happens faster than certificate provisioning

This can be reproduced by ensuring that the Woodpecker step causes creation of new Kubernetes worker node (we are using Karpenter that handles it), where from time to time it fails with the error.

https://repost.aws/questions/QUK8WLbLYlSs2lKw__7h-uOQ/eks-remote-error-tls-internal-error-when-running-kubectl-logs-command led to us to understand the issue better, and the fact that when provisioning a new worker node, the `kubectl get csr --watch` shows 15s period of time when the CSR is pending

```
csr-d5fmg   0s      kubernetes.io/kubelet-serving   system:node:<node>  <none>              Pending
csr-d5fmg   15s     kubernetes.io/kubelet-serving   system:node:<node>   <none>              Approved
csr-d5fmg   15s     kubernetes.io/kubelet-serving   system:node:<node>   <none>              Approved,Issued
```

```mermaid
sequenceDiagram
    participant WP as Woodpecker CI
    participant K8s as Kubernetes API
    participant Node as New Worker Node
    participant Kubelet as Kubelet
    
    WP->>K8s: Create build pod
    K8s->>Node: Schedule pod
    Node->>Kubelet: Start container
    Note over Node: CSR not yet approved
    WP->>Kubelet: GET /containerLogs (too early!)
    Kubelet-->>WP: TLS: internal error
    Note over Node: CSR gets approved later

```

# Solution

Added retry logic with exponential backoff to the `TailStep` function in the Kubernetes backend to improve resiliency when fetching log streams. This prevents Woodpecker from marking steps as failed due to temporary issues (such as our specific issue with newly provisioned nodes where kubelet certificates are not yet valid when Woodpecker is already trying to tail the logs).

### DIFF
--- a/pipeline/backend/kubernetes/kubernetes.go
+++ b/pipeline/backend/kubernetes/kubernetes.go
@@ -46,6 +46,10 @@ const (
 	EngineName = "kubernetes"
 	// TODO: 5 seconds is against best practice, k3s didn't work otherwise
 	defaultResyncDuration = 5 * time.Second
+	initialRetryDelay     = 1 * time.Second
+	maxRetryDelay         = 5 * time.Second
+	maxRetryDuration      = 1 * time.Minute
+	backoffMultiplier     = 2
 )
 
 var defaultDeleteOptions = newDefaultDeleteOptions()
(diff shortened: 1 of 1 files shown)


qwerty287:
Thanks. I cannot tell anything about how this fixes the issue as I don't know much a out k8s, but just one note: for backoffs we use  github.com/cenkalti/backoff/v5 at other points already. No need to reimplement it.

henkka:
> Thanks. I cannot tell anything about how this fixes the issue as I don't know much a out k8s, but just one note: for backoffs we use github.com/cenkalti/backoff/v5 at other points already. No need to reimplement it.

🙏 modified to use the mentioned library for the backoff logic -> https://github.com/woodpecker-ci/woodpecker/pull/5550/commits/86b7bba525b48c0140766c20833fce280762bc84

Basically this adds resiliency to Woodpecker Kubernetes backend so it doesn't fail steps so easily when there are temporary issues with fetching the log stream (e.g. in our case the mentioned issue with Woodpecker tailing the logs before the certificates for the kubelet are valid, and that error causing Woodpecker to consider the step as failed)

PR #5552, opened 2025-09-26
Enable completion support for cli

marcusramberg:
This PR enables the built in urfave autocomplete generator.

ref <https://cli.urfave.org/v3/examples/completions/customizations/>
I'm using the customization to ensure it's included in the help text.
Also skipping the setup check to ensure you can generate completion at install
time.


qwerty287:
Thanks! Just for understanding, why don't we use the defaults and hide it?

PS: Added Norwegian to our Weblate as you requested.

marcusramberg:
> Thanks! Just for understanding, why don't we use the defaults and hide it?
> 
> 
> 
> PS: Added Norwegian to our Weblate as you requested.

Thank you. My thinking was that it's better if it appears in the help menu if someone downloads the binary and wants to add completion for their shell. The completion help also includes specific instructions for each shell. If you prefer that it's hidden I can update the PR. I'm planning to do a update to the nix package as well to use the completion once it's in a release.

xoxys (approved):

PR #5559, opened 2025-09-27
Trace errors during SetupWorkflow, make service step setup errors visible to user

hhamalai:
Trace errors during SetupWorkflow.  Currently all the steps are traced when executed, but due to how pipelines on kubernetes backend orchestrated its workflow resources, the service steps are created during SetupWorkflow, and the creation is prone to errors.

A specific example is when service steps are missing explicit port definitions:
- In Docker backend: This works fine even without explicit ports
- In Kubernetes backend: Service creation fails the whole pipeline, no error description on UI

The lack of descriptive error has raised questions, usually requiring Woodpecker Admin to provide the user with reason and solution to fix the issue.

Currently:
image

With this PR:
image


qwerty287:
Thanks! Instead of using a callback, I'd rather introduce a new error type that includes the step and the error data. Then we can check this on return. For example similar to https://github.com/woodpecker-ci/woodpecker/blob/main/server/forge/types/errors.go#L43

hhamalai:
> I'd rather introduce a new error type that includes the step and the error data. Then we can check this on return. For example similar to https://github.com/woodpecker-ci/woodpecker/blob/main/server/forge/types/errors.go#L43

Great idea to keep the interfaces clean, cheers!

I made the changes you suggested and tested that the outcome is same as with previous approach.

qwerty287:
Thanks, can you reformat your code?

hhamalai:
> Thanks, can you reformat your code?

formated and checked that make lint passes without errors

PR #5569, opened 2025-09-29
local backend: fix windows cmd.exe command escaping

6543:
## Problem

The local backend's cmd.exe implementation has a bug where commands containing special characters (`&`, `|`) or newlines are **partially executed during the echo phase** instead of being safely displayed.

### Current Behavior

1. Before executing a command, the backend echoes it to show what will run
2. However, the echo statement itself interprets `&` and `|` as command separators
3. This causes parts of the command to execute prematurely during echo, before the actual intended execution
4. Commands with newlines are similarly broken - the newline causes cmd.exe to treat it as multiple commands during the echo phase

### Example

Given this YAML command:
```yaml
commands:
  - echo "test" & echo "unintended"
```

**Expected:**
1. Echo displays: `+ echo "test" & echo "unintended"` (display only)
2. Then execute: `echo "test" & echo "unintended"` (actual execution)

**What actually happens:**
1. During echo, the `&` causes `echo "unintended"` to execute immediately
2. Then the full command executes again

This bug makes multi-line commands impossible to use with cmd.exe, as newlines in the echo statement cause premature command execution.

## Solution

Replace special characters in the **echo output only** with symbolic representations:
- `&` → `\AND`
- `|` → `\OR`
- `\n` → `\\n`

The actual command execution remains unchanged and uses the original characters.

### Trade-off

The echoed output is not 100% identical to what executes, but:
- Users can always refer to their YAML for the exact command
- The symbolic representation is understandable enough
- This fixes the premature execution bug
- Multi-line commands now work correctly

### Why `shellescape.Quote()` Isn't Sufficient

The existing `shellescape.Quote()` function doesn't fully protect against cmd.exe's special character interpretation in echo statements. Even with quoting, `&` and `|` can break out of quoted strings in certain contexts, and newlines cannot be safely echoed without this additional escaping layer.

## Changes

- Added proper error handling for empty shell and empty command list
- Fixed cmd.exe echo to escape special characters with symbolic representations
- Added comprehensive test suite covering all shell types and edge cases
- Multi-line commands now work correctly with cmd.exe

## Screenshots

image
image


pipeline/backend/local/command.go
```diff
@@ -17,25 +17,47 @@
...
 	"os"
 	"strings"
 
 	"al.essio.dev/pkg/shellescape"
 )
 
+var (
+	ErrNoShellSet   = errors.New("no shell was set")
+	ErrNoCmdSet     = errors.New("no commands where set")
+	ErrNoPosixShell = errors.New("assumed posix shell but test failed")
```
6543:
```suggestion
```

xoxys:
Tbh I don't understand the root cause. Can you please improve the PR description and describe the issue more precisely?

6543:
@xoxys done with the magic of ai ... i think that did a better job as i have ever could ... just had to clarify stuff for the 🦜

kmk142789 (approved):

6543:
yep :/

i you have please tell me :D

I had following ideas:
print out `+ ` without newline and dont attach `@` to the commands to be executed so cmd normally prints out the commands inside the batch before exec but:
printing out string without newline or carrierreturn seems to be a whole nother level of hack for cmd
and if we dont add the `+ ` prefix we have no way to differentiate between what is an command to be executed or the command output :/

so in the end CMD is old and just a mess ... but still used :sweat:

PR #5570, opened 2025-09-29
local backend: test shells if unknown

6543:
currently if we don't know the shell we just assume posix.
this adds a small test, to ensure it is and fail gracefully before doing weird stuff.
## Test Conf
```yaml
skip_clone: true
steps:
  build:
    image: "true"
    commands:
      - echo "building..."
```
## Before
image

## After
image

### DIFF
--- a/pipeline/backend/local/command_test.go
+++ b/pipeline/backend/local/command_test.go
@@ -153,8 +153,8 @@ func TestProbeShellIsPosix(t *testing.T) {
 		err := probeShellIsPosix("nonexistentshell12345")
 		if assert.ErrorIs(t, err, &ErrNoPosixShell{}) {
 			assert.Equal(t,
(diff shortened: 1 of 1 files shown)


6543:
Base work for  #5582

pipeline/backend/local/errors.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// ErrNoPosixShell indicates that a shell was assumed to be POSIX-compatible but failed the test.
+type ErrNoPosixShell struct {
+	Shell string
+	Err   error
+}
+
+func (e *ErrNoPosixShell) Error() string {
+	return fmt.Sprintf("Shell %q was assumed as posix shell but test failed: %v\n(if you want support for it, open an issue at woodpecker project)", e.Shell, e.Err)
```
xoxys:
```suggestion
	return fmt.Sprintf("Shell %q was assumed to be a Posix shell, but test failed: %v\n(if you want support for it, please open an issue)", e.Shell, e.Err)
```

PR #5572, opened 2025-09-29
Fix language selector by only using the language part to get the display name

marcusramberg:
Adding nb_NO here broke the selector as Intl.DisplayNames does not understand it

### DIFF
--- a/web/src/views/user/UserGeneral.vue
+++ b/web/src/views/user/UserGeneral.vue
@@ -49,10 +49,13 @@ const { locale, t } = useI18n();
 const { storeTheme } = useTheme();
 
 const localeOptions = computed(() =>
-  SUPPORTED_LOCALES.map((supportedLocale) => ({
-    value: supportedLocale,
-    text: new Intl.DisplayNames(supportedLocale, { type: 'language' }).of(supportedLocale) || supportedLocale,
-  })),
+  SUPPORTED_LOCALES.map((supportedLocale) => {
+    const language = supportedLocale.split('_')[0];
+    return {
+      value: supportedLocale,
+      text: new Intl.DisplayNames(language, { type: 'language' }).of(language) || supportedLocale,
+    };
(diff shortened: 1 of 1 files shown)


6543:
what happens if we have two xx_XX locales at some point?

as i understand this is a hotfix...

qwerty287:
As @6543 says, this introduces new issues.

We already have `zh-Hant` and `zh-Hans` - maybe the problem is using an underscore instead of a hyphen? Could you try renaming the file to `nb-NO`? We then would have to check how this is reflected into weblate.

marcusramberg:
> what happens if we have two xx_XX locales at some point?

Here's the full list of supported languages in weblate https://github.com/WeblateOrg/language-data/blob/main/languages.csv
If you were to add multiple variants of Arabic for instance  they would all show up as Arabic in the list with this fix (but would probably just break account settings without it). The reason the locale for my language is nb_NO is we have another official language in Norway (nn_NO)


> as i understand this is a hotfix...

Yepp. Account settings is currently broken in main. Would have been really nice to have some basic integration tests that at least ensure all the pages can load without js errors.

marcusramberg:
> As @6543 says, this introduces new issues.
> 
> We already have `zh-Hant` and `zh-Hans` - maybe the problem is using an underscore instead of a hyphen? Could you try renaming the file to `nb-NO`? We then would have to check how this is reflected into weblate.

Hmm, you're right, that is the better fix. 

image

marcusramberg:
Should I open a new PR for that, or does it need to be fixed from inside weblate to maintain the mapping to the translation?

marcusramberg:
> what happens if we have two xx_XX locales at some point?
> 
> as i understand this is a hotfix...

It's not broken in the latest release as Norwegian was added after it was tagged btw.

marcusramberg:
Closing in favor of #5573

PR #5573, opened 2025-09-30
Rename Norwegian to nb-NO to fix language selector

marcusramberg:
Not sure how this will affect the mapping in Weblate, but I see it was already done this way for the chinese translations.

Replaces #5572


anbraten (approved):

6543:
thanks

6543:
note: weplate does notice it via git

image

PR #5578, opened 2025-09-30
Support matrix environ badges only with no key-values

xoxys:
Fixes:

grafik

After that change:

image

### DIFF
--- a/web/src/components/atomic/Badge.vue
+++ b/web/src/components/atomic/Badge.vue
@@ -1,15 +1,19 @@
 <template>
   <span class="inline-flex items-center text-xs font-medium">
     <span
-      v-if="label !== undefined"
-      class="border-wp-state-neutral-100 bg-wp-state-neutral-100 rounded-l-full border-1 py-0.5 pr-1 pl-2 whitespace-nowrap text-white"
+      v-if="label"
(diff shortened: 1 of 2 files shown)


anbraten:
Are empty values somehow important / do they have an effect on sth?

xoxys:
Kinda, in this case the plugin config is set dynamically from the matrix env. That means one pipeline is using `tag_suffix: "insecure"` and another one `tag_suffix: ""`. However, if we want to display empty values, how should they look like?

6543:
add an cursive _empty_ ?

xoxys:
Yes we can do that if we prefer it that way.

6543:
yes please - as droping it would hide information that the user could need to know for fully understanding what's happening

xoxys:
Do we want to have the string translated? Or should we use something like *null* without translations? Translating this string could make it very long.

anbraten:
We could also avoid the "white" part and make the gray round on the right in case its empty. Would not use an english string in there. If you want a string we could use "-" as well, however that could always lead to someone thinking that's the actual value.

6543:
the idea to make a simple lable without the second part is a better idea no need for i18 problems

PR #5580, opened 2025-09-30
Announce only stable releases

xoxys:


.woodpecker/social.yaml
```diff
@@ -4,6 +4,7 @@ depends_on:
 
 when:
   - event: tag
+  - evaluate: 'CI_COMMIT_TAG matches "^v?[0-9]+\.[0-9]+\.[0-9]+$"'
```
anbraten:
```suggestion
  - event: tag
    evaluate: 'CI_COMMIT_TAG matches "^v?[0-9]+\.[0-9]+\.[0-9]+$"'
```

.woodpecker/social.yaml
```diff
@@ -4,6 +4,7 @@ depends_on:
 
 when:
   - event: tag
+    evaluate: CI_COMMIT_TAG matches "^v?[0-9]+\\.[0-9]+\\.[0-9]+$"
```
6543: doesn't this also match v1.2.3 ?

and your intention is to only run it on vx.x.0 right?

xoxys: No the goal is to just exclude pre releases.

PR #5583, opened 2025-10-01
Document pipeline backend engine interface precisely

6543:
Changes:
 - Document pipeline backend engine interface precisely
 - Add debug log to docker backend on ContainerWait return values

### DIFF
--- a/pipeline/backend/types/backend.go
+++ b/pipeline/backend/types/backend.go
@@ -130,7 +130,7 @@ type Backend interface {
 	WaitStep(ctx context.Context, step *Step, taskUUID string) (*State, error)
 
 	// DestroyStep cleans up resources associated with a step.
-	// This is called after WaitStep completes, or if the workflow is cancelled.
+	// This is called after WaitStep completes, or if the workflow is canceled.
 	//
 	// Implementations should:
 	// - Stop the step if still running
--- a/pipeline/backend/types/backend.go
+++ b/pipeline/backend/types/backend.go
@@ -25,17 +25,18 @@ import (
 
 // Backend defines the mechanism for orchestrating workflows and their steps.
 //
-// A Backend instance may be created multiple times per agent, depending on the
-// configured parallel workflow capacity. Each instance handles one workflow at
-// a time, but multiple steps within that workflow may execute concurrently.
+// A Backend instance is created once per agent and must handle multiple
+// workflows concurrently, depending on the configured parallel workflow
+// capacity. Each workflow may have multiple steps executing concurrently.
 //
 // Thread Safety and Isolation:
 //
 //   - Each workflow must have a unique taskUUID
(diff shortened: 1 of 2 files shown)


6543:
use:
```sh
go doc go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/types.Backend
```

to generate the golang doc string

pipeline/backend/types/backend.go
```diff
@@ -21,38 +23,134 @@ import (
...
+	//
+	// Must be safe to call even if StartStep failed or the step was never started.
 	DestroyStep(ctx context.Context, step *Step, taskUUID string) error
 
-	// DestroyWorkflow destroys the workflow environment.
+	// DestroyWorkflow cleans up all workflow-level resources.
+	// This is called after all steps have been destroyed (via DestroyStep).
+	//
+	// Implementations should:
+	// - Remove workflow-specific workspaces, networks, or namespaces
```
6543:
```suggestion
	//
	// Implementations should:
	// - Destroy steps still running in the background (detached steps and services) 
	// - Remove workflow-specific workspaces, networks, or namespaces
```

pipeline/backend/types/backend.go
```diff
@@ -21,38 +23,134 @@ import (
 	"github.com/urfave/cli/v3"
 )
 
-// Backend defines a container orchestration backend and is used
-// to create and manage container resources.
+// Backend defines the mechanism for orchestrating workflows and their steps.
+//
+// A Backend instance may be created multiple times per agent, depending on the
+// configured parallel workflow capacity. Each instance handles one workflow at
```
qwerty287: I don't think works like this. It just re-uses the backend instance for all the workflows.

See https://github.com/woodpecker-ci/woodpecker/blob/main/cmd/agent/core/agent.go#L184

That's the only point we init the backend and this is only done once.

pipeline/backend/types/backend.go
```diff
@@ -21,38 +23,134 @@ import (
...
+	// The taskUUID uniquely identifies this workflow and must be used to
+	// isolate this workflow's resources from other concurrent workflows on
+	// the same backend instance or host.
+	//
+	// Implementations should:
+	// - Create isolated workspaces, networks, or namespaces
+	// - Initialize shared volumes or storage
+	// - Ensure the setup doesn't interfere with other running workflows
+	//
+	// Note: Only one workflow is executed at a time per Backend instance,
```
qwerty287: Here as well. This doesn't work like this.

pipeline/backend/types/backend.go
6543: thanks somehow remembered it wrong :sweat_smile:

6543:
as we now have a good description of expections .... we should check our backends.

the local backend does not match that so fix is here: #5582

PR #5588, opened 2025-10-01
Fix repo extension router and icon

xoxys:
Fixes the router (https://github.com/woodpecker-ci/woodpecker/pull/3349#issuecomment-3357641195) and add unique icon

image

### DIFF
--- a/web/src/views/repo/settings/Extensions.vue
+++ b/web/src/views/repo/settings/Extensions.vue
@@ -1,5 +1,5 @@
 <template>
-  <Settings :title="$t('extensions')" :description="$t('extensions_description')" docs-url="docs/usage/registries">
+  <Settings :title="$t('extensions')" :description="$t('extensions_description')" docs-url="docs/usage/extensions">
     <form @submit.prevent="saveExtensions">
       <InputField :label="$t('extensions_signatures_public_key')">
         <pre class="code-box">{{ signaturePublicKey }}</pre>


xoxys:
As I look at it, the UI has some more glitches, fixed now:

image

web/src/views/repo/settings/Extensions.vue
```diff
@@ -1,17 +1,13 @@
...
-    </div>
-
-    <div class="flex flex-col">
-      <span class="text-color font-bold">{{ $t('extensions_signatures_public_key') }}</span>
-      <span class="text-color">{{ $t('extensions_signatures_public_key_description') }}</span>
-      <pre class="code-box">{{ signaturePublicKey }}</pre>
-    </div>
-
-    <div class="mt-4 flex flex-col border-t-1 dark:border-gray-600">
+    <div class="mt-4 flex flex-col">
```
anbraten: The div could even be removed as there is only a single form inside (at least what I did recently 🙈).

xoxys: Good catch, thanks!

web/src/views/repo/settings/Extensions.vue
```diff
@@ -1,27 +1,18 @@
 <template>
   <Settings :title="$t('extensions')" :description="$t('extensions_description')" docs-url="docs/usage/registries">
```
anbraten:
```suggestion
  <Settings :title="$t('extensions')" :description="$t('extensions_description')" docs-url="docs/usage/extensions">
```

Issue #5585, opened 2025-10-01
[UX] Mark Disabled Agents in Agent-List


6543:
<img width="1838" height="238" alt="Image" src="https://github.com/user-attachments/assets/019979b2-3cd5-4958-9222-d4918e8e0b22" />

if you have a lot of agents it's hard to tell - you have to click into each agent config menue :/

PR #5593, opened 2025-10-02
Add disabled badge to agents

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/5585

grafik

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -319,6 +319,7 @@
         },
         "no_schedule": {
           "name": "Disable agent",
+          "disabled": "Disabled",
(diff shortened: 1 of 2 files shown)


web/src/assets/locales/en.json
```diff
@@ -319,6 +319,7 @@
         },
         "no_schedule": {
           "name": "Disable agent",
+          "disabled": "Disabled",
```
anbraten: New translations should be on top-level.

xoxys: Why that?

anbraten: We started it some time ago to avoid duplicates.

PR #5595, opened 2025-10-03
feat(ui): repoStore::loadRepos to return all repos if user is system admin

henkka:
# Problem
Currently, administrators face a UX limitation when managing repositories across the Woodpecker instance:
- Admins can access any repository directly via URL
- However, they cannot see repositories in listings unless they're members of the respective organization
- This creates an inconsistent experience and makes repository discovery difficult for administrative tasks

# Solution
This PR enhances the repository listing functionality for administrators by:
- Displaying all repositories available in the instance, regardless of organization membership
- Maintaining existing access controls for non-admin users
- Providing administrators with a complete view of the instance for better management capabilities

Quick demo of the current UX (before the page refresh its Vue's internal state showing the entry, and after hard refresh its empty list)

https://github.com/user-attachments/assets/12aaa1f2-4654-43bd-a9f5-fb4ba8f93459


qwerty287:
Hi @henkka, thanks for the PR. There is already a overview over all repos in the admin settings - isn't that enough?

henkka:
> Hi @henkka, thanks for the PR. There is already a overview over all repos in the admin settings - isn't that enough?

The admin settings overview is useful, but it's not really integrated into the normal workflow.

The issue is that I can already access any repo directly if I know the repo id, but I can't see them in the regular repo listings. So I end up having to either bookmark repos or remember repo ids (with teams currently migrating to Woodpecker and having build issues, I'm constantly checking their build status to anticipate where our help is needed the most), which is pretty clunky.

Basically, this just makes the UI match what the permissions already allow, without changing any actual access controls.

xoxys (approved):

PR #5611, opened 2025-10-07
Enhance Bitbucket Datacenter build status reporting

MartinSchmidt:
Improve build status reporting by adding duration, date added, and parent context to the status response.
The parent field enables `Required builds` to be set in bitbucket datacenter, to block merges to main before tests pass.


server/forge/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -318,6 +322,9 @@ func (c *client) Status(ctx context.Context, u *model.User, repo *model.Repo, pi
 		URL:         common.GetPipelineStatusURL(repo, pipeline, workflow),
 		Key:         common.GetPipelineStatusContext(repo, pipeline, workflow),
 		Description: common.GetPipelineStatusDescription(workflow.State),
+		Duration:    uint64((pipeline.Finished - pipeline.Started) * millisecondsInSecond),
```
6543, line 325: there should be a const in the time package

6543: never mind https://cs.opensource.google/go/go/+/refs/tags/go1.25.2:src/time/time.go;l=933 is not usefull for us :/

MartinSchmidt:
@6543 i fixed the linting errors.

PR #5612, opened 2025-10-07
fix: find users with forge-id

anbraten:

### DIFF
--- a/server/api/users.go
+++ b/server/api/users.go
@@ -57,7 +57,7 @@ func GetUsers(c *gin.Context) {
 //	@Param			Authorization	header	string	true	"Insert your personal access token"	default(Bearer <personal access token>)
 //	@Param			login			path	string	true	"the user's login name"
 func GetUser(c *gin.Context) {
-	user, err := store.FromContext(c).GetUserLogin(c.Param("login"))
+	user, err := store.FromContext(c).LookupUserByLogin(c.Param("login"))
 	if err != nil {
 		handleDBError(c, err)
 		return
@@ -87,7 +87,7 @@ func PatchUser(c *gin.Context) {
 		return
 	}
 
-	user, err := _store.GetUserLogin(c.Param("login"))
+	user, err := _store.LookupUserByLogin(c.Param("login"))
 	if err != nil {
 		handleDBError(c, err)
 		return
@@ -159,7 +159,7 @@ func PostUser(c *gin.Context) {
 func DeleteUser(c *gin.Context) {
 	_store := store.FromContext(c)
 
-	user, err := _store.GetUserLogin(c.Param("login"))
+	user, err := _store.LookupUserByLogin(c.Param("login"))
 	if err != nil {
 		handleDBError(c, err)
 		return
--- a/server/store/datastore/user.go
+++ b/server/store/datastore/user.go
@@ -18,8 +18,6 @@ import (
 	"errors"
 	"fmt"
 
-	"xorm.io/xorm"
-
 	"go.woodpecker-ci.org/woodpecker/v3/server/model"
 	"go.woodpecker-ci.org/woodpecker/v3/server/store/types"
 )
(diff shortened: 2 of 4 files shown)


server/store/datastore/user.go
```diff
@@ -29,21 +27,32 @@ func (s storage) GetUser(id int64) (*model.User, error) {
...
-func (s storage) GetUserLogin(login string) (*model.User, error) {
-	return s.getUserLogin(s.engine.NewSession(), login)
+func (s storage) GetUserByLogin(forgeID int64, login string) (*model.User, error) {
+	sess := s.engine.NewSession()
+	user := new(model.User)
+	return user, wrapGet(sess.Where("forge_id = ? AND login=?", forgeID, login).Get(user))
 }
 
-func (s storage) getUserLogin(sess *xorm.Session, login string) (*model.User, error) {
+// TODO: replace with more explicit GetUserByLogin / GetUserByRemoteID calls
```
6543: @anbraten could you describe what this todo is for

as i interprete it as of now: its a todo mark for the callers of that function ... right?

it should be clear if we want to get rid of this func long term or if we want to discourage the callers to migrate to something else ...

6543:
PS: we have a simmilar problem if a organisation is renamed ...

6543:
PS: PS: I'll address the test errors tomorow ... but the constrain should be here now ... as we use it as key for users now

server/store/datastore/user.go
anbraten: @6543 The intention was a note like a deprecation that we try to switch to the more explicit functions.

anbraten:
code lgtm

6543:
@anbraten I fixed more multi-user forge stuff you might review again :)

server/api/users.go
```diff
@@ -56,8 +62,23 @@ func GetUsers(c *gin.Context) {
...
+//	@Param			forge_id		query	string	true	"specify forge (else default will be used)"
+//	@Param			forge_remote_id	query	string	false	"specify user id at forge (else fallback to login)"
 func GetUser(c *gin.Context) {
-	user, err := store.FromContext(c).GetUserLogin(c.Param("login"))
+	forgeID, err := strconv.ParseInt(c.DefaultQuery("forge_id", fmt.Sprint(defaultForgeID)), 10, 64)
+	if err != nil {
+		c.AbortWithStatus(http.StatusBadRequest)
+		return
+	}
+	forgeRemoteID := model.ForgeRemoteID(c.Query("forge_remote_id"))
```
6543, line 73: need to look how conversion is done in other places ... as `model.ForgeRemoteID(val)` != `model.ForgeRemoteID(fmt.Sprint(val))`

6543: it is https://play.golang.com/p/NXh3mO2zt9W

6543:
@anbraten could you write lgtm or deny again :)

I know review dont work on own pulls :sweat:

xoxys:
I dont like to introduce new cli flags using id's. Thats weird to use not human friendly and should be avoided.

6543:
then I'll remove that cli change part for now

server/model/repo.go
```diff
@@ -45,9 +45,9 @@ func (mode ApprovalMode) Valid() bool {
 type Repo struct {
 	ID      int64 `json:"id,omitempty"                    xorm:"pk autoincr 'id'"`
 	UserID  int64 `json:"-"                               xorm:"INDEX 'user_id'"`
-	ForgeID int64 `json:"forge_id,omitempty"              xorm:"forge_id"`
+	ForgeID int64 `json:"forge_id,omitempty"              xorm:"UNIQUE(forge) forge_id"`
```
anbraten, line 48: unrelated change that should be in its own PR

6543:
@xoxys did remove the cli patch

xoxys:
See my comments in the chat.

server/model/repo.go
xoxys, line 48: Does it require a migration?

6543: xorm.Sync takes kare of

server/api/users.go
```diff
@@ -87,14 +108,26 @@ func PatchUser(c *gin.Context) {
 		return
 	}
 
-	user, err := _store.GetUserLogin(c.Param("login"))
-	if err != nil {
+	if in.ForgeID < defaultForgeID {
```
anbraten, line 111: This could be a kinda breaking change, I would stick with the lookup method I had before for now. We can add this improvement later on

6543:
ugh ... I can strip it out :sweat:

6543:
well thanks :sweat_smile: I still was going to split it out but that makes it way easyer for me :)

xoxys:
As we have done the release without this PR, need to overcomplicate things IMO. Just keep in mind if we have to revert partially this needs to be done manually. If I misunderstood out discussion in the chat and overruled you @anbraten, sorry for that, this was not on purpose.

6543:
no problem :) just ping me if there is an issue with this exact patch

6543:
created snapshot and deployed on our own ci ...

PR #5613, opened 2025-10-08
Add deployment target suggestions from pipeline configuration

mofr93:
### Add deployment target suggestions from pipeline configurations
This PR enhances the deploy pipeline popup by extracting deployment targets from pipeline 
configurations and presenting them as suggestions to users.

### Changes

- Auto-discovery: Parses pipeline YAML files to find CI_PIPELINE_DEPLOY_TARGET values used in conditional expressions

- Smart input: TextField now supports both typing custom values and selecting from discovered targets via HTML datalist

- Better UX: Users can see available deployment targets without having to remember or check pipeline configs manually

### Technical Details

- Extracts targets from patterns like CI_PIPELINE_DEPLOY_TARGET == "production"

- Base64 decodes pipeline configs to parse YAML content

Skærmbillede 2025-10-08 kl  19 12 40
Skærmbillede 2025-10-08 kl  19 21 47

### DIFF
--- a/web/src/components/layout/popups/DeployPipelinePopup.vue
+++ b/web/src/components/layout/popups/DeployPipelinePopup.vue
@@ -40,9 +40,9 @@
 </template>
 
 <script lang="ts" setup>
+import { decode } from 'js-base64';
 import { computed, onMounted, ref, toRef } from 'vue';
 import { useRouter } from 'vue-router';
-import { decode } from 'js-base64';
 
 import Button from '~/components/atomic/Button.vue';
 import InputField from '~/components/form/InputField.vue';
@@ -68,24 +68,25 @@ const router = useRouter();
 
 const pipelineConfigs = requiredInject('pipeline-configs');
 
-const decodedConfigs = computed(() =>
-  pipelineConfigs.value?.map((config) => ({
-    ...config,
-    data: decode(config.data), // Decode base64 to readable YAML
-  })) ?? []
+const decodedConfigs = computed(
+  () =>
+    pipelineConfigs.value?.map((config) => ({
+      ...config,
+      data: decode(config.data), // Decode base64 to readable YAML
+    })) ?? [],
 );
 
 // Extract CI_PIPELINE_DEPLOY_TARGET values
 const deployTargetsFromConfigs = computed(() => {
   const targets = new Set<string>();
 
-  decodedConfigs.value.forEach(config => {
+  decodedConfigs.value.forEach((config) => {
     const yamlContent = config.data;
     // Look for evaluate: CI_PIPELINE_DEPLOY_TARGET == "somevalue" patterns
     const targetMatches = yamlContent.match(/CI_PIPELINE_DEPLOY_TARGET\s*==\s*["']([^"']+)["']/g);
 
     if (targetMatches) {
-      targetMatches.forEach(match => {
+      targetMatches.forEach((match) => {
         // Extract the value between quotes
         const valueMatch = match.match(/["']([^"']+)["']/);
         if (valueMatch && valueMatch[1]) {
@@ -99,9 +100,9 @@ const deployTargetsFromConfigs = computed(() => {
 });
 
 const deployTargetOptions = computed(() => {
-  return deployTargetsFromConfigs.value.map(target => ({
+  return deployTargetsFromConfigs.value.map((target) => ({
     value: target,
-    text: target
+    text: target,
   }));
 });
 


xoxys:
I'm not sure if parsing CI_PIPELINE_DEPLOY_TARGET from pipeline config is a common/generic pattern. Not everyone is using `evaluate` to check the deployment target. Implementing such mechanics that are not generic and somehow specific to individual setups is IMO a bad idea.

xoxys:
After thinking about it a bit more, Im going to close this issue for the mentioned reason.

wolffberg:
@xoxys

We have had this feature on our wishlist for quite some time as it's very frustrating for developers working in many different repos all with different pipelines having to browse the Woodpecker YAML files every day to check what deployment targets are available before they can deploy.

This can be especially tedious on days where Dependabot pushes to tens or hundreds of repos and everything needs to get to nonprod and prod fast.

Can you see another way this could be implemented if `CI_PIPELINE_DEPLOY_TARGET` parsing is not viable?

xoxys:
Yes, having https://github.com/woodpecker-ci/woodpecker/pull/5632 would help I guess. Right now its only designed for manual pipeline runs but should be possible to extend it to deployments.

wolffberg:
Without much knowledge of the Woodpecker codebase I don't know for sure if it would help even if extended.

Would https://github.com/woodpecker-ci/woodpecker/pull/5632 allow our developers to select from a pre-populated list of deploy targets e.g. prod, test, staging, etc. Defined as code in the repo?

xoxys:
There is an extended description in the origin PR https://github.com/woodpecker-ci/woodpecker/pull/4861

> Would https://github.com/woodpecker-ci/woodpecker/pull/5632 allow our developers to select from a pre-populated list of deploy targets e.g. prod, test, staging, etc. Defined as code in the repo?

Yes it would allow selections from pre-defined parameters, but not managed by code. Parameters can be managed by UI (or API, CLI, ...)

wolffberg:
I don't feel it's maintainable to ask developers to make manual changes in the UI or script API calls when the targets (or parameters) could already be inferred from versioned YAML files. This also makes it much harder to debug if something is changed and deploys stop working.

Would you consider making parameters and targets available for autocompletion through YAML files instead of UI or API calls?

mofr93:
@xoxys 

Our developers are using deployments with targets and they are requesting this feature. The parameter solution is useful for some things. But I don't think maintaining a list of build parameters (per pipeline) is better when we already have the available targets from eg. CI_PIPELINE_DEPLOY_TARGET in the yaml config regarding the deployments dialog.

xoxys:
My main issue is: Using CI_PIPELINE_DEPLOY_TARGET in the pipeline config is highly optional and not a generic solution. What if someone is using plugin steps for the deployments and the plugin is doing the entire CI_PIPELINE_DEPLOY_TARGET validation logic? In that case, not a single occurrence of CI_PIPELINE_DEPLOY_TARGET in the pipeline config is required and the logic of this PR will not work. 

I understand your use case, however I still believe we should not implement features that don't work generic and highly depend on the used pipeline config.

mofr93:
@xoxys I see your point with the CI_PIPELINE_DEPLOY_TARGET as it can be used in different way. And I agree it is not the most generic way to do it.

But just to clarify I'm talking strictly CI_PIPELINE_DEPLOY_TARGET in evaluation conditions. 


Let me show you an example where we would get the list of
- all
- prod
- production

```yaml
steps:
  - name: deploy prod
    image: my-argo-cd-plugin-image
    pull: true
    environment:
      APP_CONTAINER_NAME: app
when:
  - evaluate: CI_PIPELINE_DEPLOY_TARGET == "all" || CI_PIPELINE_DEPLOY_TARGET == "prod" || CI_PIPELINE_DEPLOY_TARGET == "production"
    event:
      - deployment
labels:
  environment_type: prod
  platform: linux/amd64
```

While this would not do anything
```yaml
steps:
  - name: deploy prod
    image: my-argo-cd-plugin-image
    pull: true
    environment:
      APP_CONTAINER_NAME: app
when:
  - evaluate: CI_PIPELINE_DEPLOY_TARGET contains "prod"
    event:
      - deployment
labels:
  environment_type: prod
  platform: linux/amd64
```

Alternatively is it possible we can get a field in the yaml spec to define this list which we then can use in the select box of the deployment dialog?

wolffberg:
If I understand this PR correctly, this would not break pipelines using a plugin step for deployments. It would only enhance the built-in way of deploying using Woodpecker.

If you really think this is too generic, I still think a way to define deployment targets using YAML is crucial for maintainability. Do you have another way this could be done, preferrably using the deployment targets on the deploy pop-up (might as well use it when it's there).

Maybe a root field in the YAML files with a list or something similar?

xoxys:
No, I don't have any idea yet. I would recommend opening a https://github.com/woodpecker-ci/woodpecker/discussions/categories/request-a-feature and discuss possible proposals there.

PR #5616, opened 2025-10-09
Add branch filter to cli pipeline purge

eikemeier:
This patch adds a `--branch` option to `woodpecker-cli pipeline purge` to enable purging only the pipeline for one branch, while keeping the other pipelines intact.

Use case would be a the pipelines of a merged and possibly deleted development branch, while keeping the pipelines of the release branch intact.

~Also, make the `older-than` parameter optional, since having to specify `--older-than 1s` is not useful.~


xoxys:
Thanks.

> Also, make the older-than parameter optional, since having to specify --older-than 1s is not useful.

I would disagree here. Deleting everything by default without a user input is pone to errors. Actively setting it to 1s is IMO the better way in cases users really want to delete everything.

eikemeier:
> > Also, make the older-than parameter optional, since having to specify --older-than 1s is not useful.
> 
> I would disagree here. Deleting everything by default without a user input is pone to errors.

The default for `keep-min` is still at `10`, so it's not _“everything by default”_.

> Actively setting it to 1s is IMO the better way in cases users really want to delete everything.

I would vote for making `dry-run` the default then, if that's an issue. I get the cautiousness, but I'm unsure why especially `older-than` should help here.

xoxys:
In any case, this will make this PR breaking, and we would have to wait for v4 before we can merge it. Not sure if it is really worth it, or should be done in a dedicated PR.

eikemeier:
> In any case, this will make this PR breaking, and we would have to wait for v4 before we can merge it. Not sure if it is really worth it, or should be done in a dedicated PR.

Is there a list for _“planned breaking changes in v4”_? I'm too lazy to make a separate PR for this small improvement. The `1s` workaround works, although it's stupid.

6543:
I think it's fine as is now ... but @ owners have the last word :eyes:

xoxys:
> Is there a list for “planned breaking changes in v4”? I'm too lazy to make a separate PR for this small improvement. The 1s workaround works, although it's stupid.

I'll take care. Thanks!

cli/pipeline/purge.go
```diff
@@ -72,25 +76,31 @@ func pipelinePurge(c *cli.Command, client woodpecker.Client) (err error) {
...
-	if err != nil {
-		return err
+	var before time.Time
+	if olderThan != "" {
+		duration, err := time.ParseDuration(olderThan)
+		if err != nil {
+			return err
+		}
+
+		before = time.Now().Add(-duration)
```
xoxys: This part should be reverted as well.

eikemeier: This fixes a bug in the original code:

The duration is calculated including the time it takes for the `fetchPipelinesToKeep` query, which might be surprising: Assuming in a script I request to delete everything older than 10s, sparing the current run, but the `fetchPipelinesToKeep` query is slow, taking ten seconds itself - then everything is deleted.

Maybe the current time should be taken before `NewClient` is called. I can include that. Does the code have other bugs or changes in behavior, or what would be the reason to revert to the previous calculation? I do not assume someone could really depend on the old way to calculate the time.

What would be a change is that `--older-than ""` is now possible - is that breaking?

xoxys: Was not aware of this bug. Good catch. In that case, let's keep it that way.

> What would be a change is that --older-than "" is now possible - is that breaking?

Is it? For required string flags, the cli lib should still prevent that.

eikemeier: > > What would be a change is that --older-than "" is now possible - is that breaking?
> 
> Is it? For required string flags, the cli lib should still prevent that.

It wouldn't, since it is a valid string.

With the current changes, only valid durations are accepted. Also, the error message is better (starts with “can't parse older-than:”).

Introduced a `time.Time` parameter to `pipelinePurge` for the start time, which is taken as early as possible - assuming `urfave/cli` is fast. This parameter can be zero (used in tests) in which case the “older-than” parameter has no effect.

Could be used in future tests to have a deterministic test call. As far as this commit is concerned it is fine, and better than using a non-deterministic `time.Now()` in tests.

Issue #5221, opened 2025-05-30
Inconsistency between `woodpecker-cli repo add` and `woodpecker-cli repo update`


krlmlr:
Thank you for providing and maintaining this tool.

From the help:

```sh
woodpecker-cli repo add [command [command options]] <forge-remote-id>
woodpecker-cli repo update [command [command options]] <repo-id|repo-full-name>
```

What would it take to accept a `repo-full-name` for `woodpecker-cli repo add` ? It's a bit awkward to get the `forge-remote-id` for a repo.

I've searched the issue tracker for "repo add" and didn't find an existing discussion.

PR #5631, opened 2025-10-11
Add repo filter options to GetRepos api

xoxys:
Add repo filter options to `GetRepos` API. Right now, only filtering by name is implemented. The filter is useful to enable repos by name via CLI (https://github.com/woodpecker-ci/woodpecker/issues/5221). In that case getting the full repo list from the server is unnecessary.

### DIFF
--- a/server/store/datastore/repo.go
+++ b/server/store/datastore/repo.go
@@ -155,8 +155,7 @@ func (s storage) RepoList(user *model.User, owned, active bool, f *model.RepoFil
 		sess = sess.And(builder.Eq{"repos.active": true})
 	}
 	if f != nil && f.Name != "" {
-			sess = sess.And(builder.Eq{"repos.name": f.Name})
-		}
+		sess = sess.And(builder.Eq{"repos.name": f.Name})
 	}
 	return repos, sess.
 		Asc("full_name").
--- a/server/store/datastore/repo_test.go
+++ b/server/store/datastore/repo_test.go
@@ -152,11 +152,45 @@ func TestRepoList(t *testing.T) {
 		assert.NoError(t, store.PermUpsert(perm))
 	}
 
-	repos, err := store.RepoList(user, false, false, nil)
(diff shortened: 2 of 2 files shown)


server/store/datastore/repo.go
```diff
@@ -154,6 +154,11 @@ func (s storage) RepoList(user *model.User, owned, active bool) ([]*model.Repo,
 	if active {
 		sess = sess.And(builder.Eq{"repos.active": true})
 	}
+	if f != nil {
+		if f.Name != "" {
```
anbraten:
```suggestion
	if f != nil && f.Name != "" {
```

xoxys: It was intended that way as it allows to easily add more filter options.

server/store/datastore/repo.go
```diff
@@ -143,7 +143,7 @@ func (s storage) deleteRepo(sess *xorm.Session, repo *model.Repo) error {
 
 // RepoList list all repos where permissions for specific user are stored
 // TODO: paginate
-func (s storage) RepoList(user *model.User, owned, active bool) ([]*model.Repo, error) {
+func (s storage) RepoList(user *model.User, owned, active bool, f *model.RepoFilter) ([]*model.Repo, error) {
```
anbraten, line 146: Just a nit.

```suggestion
func (s storage) RepoList(user *model.User, owned, active bool, filter *model.RepoFilter) ([]*model.Repo, error) {
```


Should owned / active be part of the filter as well?

server/store/datastore/repo.go
anbraten: ahh I see, would extend it when we add the next one

server/store/datastore/repo.go
xoxys: We named it f in multiple other places... Thats why I was choosing this name. I would like to keep it. Feel free to harmonize in a dedicated pr.

PR #5635, opened 2025-10-13
flake.lock: Update

6543:
Flake lock file updates:

• Updated input 'nixpkgs':
    'github:nixos/nixpkgs/2bb57ed7e7fc2b7ce3b39c6934e4fb069dc7922d' (2025-07-23)
  → 'github:nixos/nixpkgs/78c9baf72eea1e60a52faf255741ffa257799757' (2025-10-13)


xoxys:
Im a bit annoyed by these protobuf version string changes everywhere...

6543:
the flake had protobuff pined for at least 3 months ... and if you use that version it wont change

as for using your system/own dev environment ... well before we had no official protobuv version pinned so i would say if you regenerate stuff use `git add -p` ...

... anyway i thought adding them here is fine as we do bump them from time to time so why not pin it to the defined flake dev version ...

PR #5636, opened 2025-10-13
Document Forge interface precisely

6543:
just document interface precisely

### DIFF
--- a/server/forge/forge.go
+++ b/server/forge/forge.go
@@ -51,7 +51,7 @@ import (
 // Error Handling:
 // - types.ErrIgnoreEvent: Skippable webhook events
 // - types.RecordNotExist: Resource not found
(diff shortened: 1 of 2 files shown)


server/forge/forge.go
```diff
@@ -23,71 +28,134 @@ import (
...
+// Thread Safety:
+// Implementations must be safe for concurrent use. Methods receive context.Context
+// for cancellation/timeout. Do not maintain user-specific state; user context is
+// passed via *model.User parameter.
+//
+// Authentication:
+// OAuth2-based authentication is assumed. Tokens are refreshed 30 minutes before
+// expiry via the optional Refresher interface.
+//
+// Webhook Processing Flow:
```
qwerty287: This should go to the `Hook` func

PR #5637, opened 2025-10-13
Modernize a couple of loops, fix incorrect function docs

marcusramberg:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,7 +55,7 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/yaronf/httpsign v0.3.2
 	github.com/zalando/go-keyring v0.2.6
-	gitlab.com/gitlab-org/api/client-go v0.156.0
+	gitlab.com/gitlab-org/api/client-go v0.157.0
 	go.uber.org/multierr v1.11.0
 	golang.org/x/crypto v0.43.0
 	golang.org/x/net v0.46.0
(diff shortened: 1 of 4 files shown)


cmd/agent/core/agent.go
```diff
@@ -283,8 +283,7 @@ func run(ctx context.Context, c *cli.Command, backends []types.Backend) error {
 		}
 	})
 
-	for i := 0; i < maxWorkflows; i++ {
-		i := i
+	for i := range maxWorkflows {
```
6543, line 287: can you add a code comment that links that in modern golang version on a loop the value is done by copybyval not as we had in the past copybyref (witch explains the `i := i` statement, to get a local scoped var witch stays)

this is needed here as we pass that var into a go routine

server/grpc/rpc.go
```diff
@@ -73,9 +74,7 @@ func (s *RPC) Next(c context.Context, agentFilter rpc.Filter) (*rpc.Workflow, er
 	}
 
 	// enforce labels from server by overwriting agent labels
-	for k, v := range agentServerLabels {
-		agentFilter.Labels[k] = v
-	}
+	maps.Copy(agentFilter.Labels, agentServerLabels)
```
6543, line 77: does copy let keys from agentFilter.Labels that dont exist at agentServerLabels stay or are they removed?
-> did not look into the sourcecode or documentation just a question to make sure this works propperly

6543: https://cs.opensource.google/go/go/+/refs/tags/go1.25.2:src/maps/maps.go;l=58-66

PR #5641, opened 2025-10-14
Improve log.CopyByLine to be more robust

marcusramberg:
More robust error handling in the function
update the test to use assert.Eventually to make it more resistant to timing flakiness


pipeline/log/utils.go
```diff
@@ -45,17 +45,17 @@ func CopyLineByLine(dst io.Writer, src io.Reader, maxSize int) error {
 	r := bufio.NewReader(src)
 
 	for {
-		// TODO: read til newline or maxSize directly
```
6543, line 48: todo was not solved jet ...

... if you remove you should link via commit pinned link to that codeblock and create a issue to track it

6543: @marcusramberg

anbraten: Would just keep the todo if its not resolved yet.

marcusramberg: I'd rather just resolve the TODO directly, but I have to confess I'm not sure what it suggests.  Are you looking for something like this? 

```go
func CopyLineByLine(dst io.Writer, src io.Reader, maxSize int) error {
	r := bufio.NewReader(src)

	for {
		line, err := r.ReadBytes('\n')
		start := 0
		for start < len(line) {
			end := min(start+maxSize, len(line))
			_, werr := dst.Write(line[start:end])
			if werr != nil {
				return werr
			}
			start = end
		}
		if err != nil {
			if err == io.EOF {
				return nil
			}
			return err
		}
	}
}
```
- That gets rid of the writeChunks method completely and just writes to maxSize - Or is the idea that there's a different way to do reads that respects maxSize? I guess I could also just restore the TODO.

6543: I only remember rougth but the core idea is to have as mouch as fesable copyed from logs while not waiting.
and the break between a single copy should if possible be at a newline ...

... for a more specific answer i would have to look into the pull that added the todo and the related code ...

marcusramberg: Thanks, I've restored the TODO comment now.

6543 (approved):

pipeline/log/utils.go
6543: -> https://github.com/woodpecker-ci/woodpecker/issues/5646

so we can propperly document and dicusst about this :)

PR #5673, opened 2025-10-23
Also create image preview on label change only

6543:
currently you have to push to a pull after you added the label to get a preview image.
this changes it by also create a preview on adding the preview label alone

example: https://ci.woodpecker-ci.org/repos/3780/pipeline/29226


qwerty287 (approved):
It will also run for any other label change once the `build_pr_images` is added but that's not possible in a different way I think

PR #5618, opened 2025-10-10
Switch to GitHub REST API to load changed files

xoxys:
Supersedes: https://github.com/woodpecker-ci/woodpecker/pull/5597

- added some small wording/var name adjustment
- added test cases using GitHub mock client

PR #5681, opened 2025-10-23
Fix nil pointer dereference during GitHub Hook parsing

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/5680
regression of  #5618


xoxys:
It can only be a bug or a regression. As the PR that introduced this issue was released already its a bug - regression label removed.

PR #5693, opened 2025-10-26
Switch from BoolTrue to optional.Option[bool]

6543:
as per https://github.com/woodpecker-ci/woodpecker/pull/1835#issuecomment-3446222018

the BoolTrue is cursed as it exloites the bool type and use it inferted ... this mostly works but if you de-/serialize it things get messed up.

instead now we store if it was set or not and what it value was and make the defaulting to true if not set an explicit thing in our if logics ... this also alows us to reserialize it as much as we want

### DIFF
--- a/shared/optional/serialization.go
+++ b/shared/optional/serialization.go
@@ -37,7 +37,7 @@ func (o *Option[T]) UnmarshalYAML(value *yaml.Node) error {
 	return nil
 }
 
-func (o Option[T]) MarshalYAML() (interface{}, error) {
+func (o Option[T]) MarshalYAML() (any, error) {
 	if !o.Has() {
 		return nil, nil
 	}


pipeline/frontend/yaml/parse_test.go
```diff
@@ -275,6 +275,7 @@ func TestReSerialize(t *testing.T) {
             - test.zig
         - path:
             exclude: a
+            on_empty: true
```
6543: yes this refactor is also kind of an bugfix as it preserfes optional bool `true` if explizite as explizite :)

in its core func it does not matter as when removed it will default to true in the old and new case :)

6543:
the difference is before we default to true by serializing now we default to true by in code bool operations

pipeline/frontend/yaml/parse_test.go
6543: well i think we should let serialization clean up config so did 1a7d7c4faf8c9f934b7f5129d872f73db21d0fb7 to restore old behaviour

qwerty287:
I'm not that sure about this change. Yes, the `BoolTrue` logic is weird, but using it is quite easy as you just do `b.Bool()`. Now you have to do `!b.Has() || b.Value()`. This is quite error-prone I think, it's likely you forget something at some point, esp. the `Has()`. What would be better would be something like `b.ValueWithDefault(true)` so it returns `true` if it's not set.

6543:
if it's about the lib being external we can just copy it into wp

qwerty287:
No, that's not my problem. I just don't like the syntax. If you would have a func `ValueWithDefault(true)` that returns `true` if the value isn't set that would be fine for me.

6543:
> No, that's not my problem. I just don't like the syntax. If you would have a func `ValueWithDefault(true)` that returns `true` if the value isn't set that would be fine for me.

hah nice we come up with the same idea ... was just going to suggest that to you :)

pipeline/frontend/yaml/constraint/path.go
```diff
@@ -97,7 +104,8 @@ func (c *Path) Match(v []string, message string) bool {
 
 	// return value based on 'on_empty', if there are no commit files (empty commit)
 	if len(v) == 0 {
-		return c.OnEmpty.Bool()
+		// on_empty is true if not set or value is true
+		return !c.OnEmpty.Has() || c.OnEmpty.Value()
```
6543:
```suggestion
		return c.OnEmpty.ValueOrDefault(true)
```

pipeline/frontend/yaml/constraint/constraint.go
```diff
@@ -132,6 +133,13 @@ func (when *When) UnmarshalYAML(value *yaml.Node) error {
 
 // MarshalYAML implements custom Yaml marshaling.
 func (when When) MarshalYAML() (any, error) {
+	// clean up local if true make it none as it will default to true
```
6543:
```suggestion
	// clean up local if true make it none as we will default to true
```

pipeline/frontend/yaml/constraint/path.go
```diff
@@ -67,18 +68,24 @@ func (c Path) MarshalYAML() (any, error) {
...
 		len(c.IgnoreMessage) == 0 &&
-		c.OnEmpty.Bool() {
+		c.OnEmpty.ValueOrDefault(true) {
 		if len(c.Include) == 0 {
 			return nil, nil
 		}
 		return yamlBaseTypes.StringOrSlice(c.Include), nil
 	}
+
+	// clean up on_empty if true make it none as it will default to true
```
6543:
```suggestion
	// clean up on_empty if true make it none as we will default to true
```

6543:
done

pipeline/frontend/yaml/constraint/constraint.go
```diff
@@ -20,6 +20,7 @@ import (
 	"path"
 	"slices"
 
+	"github.com/6543/go-optional/v2"
```
anbraten: As the implementation sounds not too crazy, could we include it in WP directly? That way we avoid bottlenecks with external libs from solo-devs that might need adjustments at some point.

6543:
imported

qwerty287:
Linter fails but otherwise lgtm

PR #5700, opened 2025-10-31
docs: add warning about 27-axis matrix limit

1001Josias:
This PR adds a documentation warning about the current matrix axis limit in Woodpecker.

Although the [source code defines a limit of axes per workflow](https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/yaml/matrix/matrix.go#L105-L109), this limitation is not [mentioned in the docs](https://woodpecker-ci.org/docs/usage/matrix-workflows).

This update clarifies why some entries in large matrix configurations may not trigger executions once the limit is reached, helping users avoid spending unnecessary time debugging their pipelines.


6543:
the limit is a drone relict ... we should increase it to a bigger number :)

-> 1000 ? (that should not limit most usecases anymore but prevent infinite loop stufff)

6543:
and i dont like that it's silently droped we should have an error then

6543:
I created an issue so it does not get lost

PR #5717, opened 2025-11-04
On set/get of repo make sure forge_id is set and on fetch respected

6543:
Currently, we identify repos by the repos remote id and just ignore the forge id, but nothing prevents different forges to have same ids for an repo. We now ensure the forgeID is set and respected when a repo is tried to be fetched etc.


server/api/login.go
```diff
@@ -315,7 +315,10 @@ func updateRepoPermissions(c *gin.Context, user *model.User, _store store.Store,
 	}
 
 	for _, forgeRepo := range repos {
-		dbRepo, err := _store.GetRepoForgeID(forgeRepo.ForgeRemoteID)
+		// make sure forgeID is set
+		forgeRepo.ForgeID = forgeID
+
```
anbraten, line 320: Why would a repo forge-id change here? If they are wrong we should migrate them.

server/api/user.go
```diff
@@ -127,6 +127,9 @@ func GetRepos(c *gin.Context) {
 
 		var repos []*model.Repo
 		for _, r := range _repos {
+			// make sure forgeID is set
+			r.ForgeID = user.ForgeID
```
anbraten, line 131: Why do we need to adjust the id here?

server/forge/bitbucket/bitbucket.go
```diff
@@ -50,6 +50,7 @@ type Opts struct {
 }
 
 type config struct {
+	id            int64
```
anbraten, line 53: To avoid confusion what this id is used for we should name it `forgeID` or add a comment (same for the other forges)

```suggestion
	forgeID            int64
```

server/forge/bitbucket/bitbucket_test.go
```diff
@@ -34,7 +34,7 @@ import (
 )
 
 func TestNew(t *testing.T) {
-	forge, _ := New(&Opts{OAuthClientID: "4vyW6b49Z", OAuthClientSecret: "a5012f6c6"})
+	forge, _ := New(0, &Opts{OAuthClientID: "4vyW6b49Z", OAuthClientSecret: "a5012f6c6"})
```
anbraten, line 37: ids start with 1

```suggestion
	forge, _ := New(1, &Opts{OAuthClientID: "4vyW6b49Z", OAuthClientSecret: "a5012f6c6"})
```

server/forge/bitbucketdatacenter/bitbucketdatacenter_test.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 func TestNew(t *testing.T) {
-	forge, err := New(Opts{
+	forge, err := New(0, Opts{
```
anbraten, line 30: ids start with 1

server/forge/forgejo/forgejo_test.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 func TestNew(t *testing.T) {
-	forge, _ := New(Opts{
+	forge, _ := New(0, Opts{
```
anbraten, line 34: ids start with 1

server/forge/gitea/gitea_test.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 func TestNew(t *testing.T) {
-	forge, _ := New(Opts{
+	forge, _ := New(0, Opts{
```
anbraten, line 35: ids start with 1

server/api/repo.go
```diff
@@ -78,6 +78,7 @@ func PostRepo(c *gin.Context) {
 		c.String(http.StatusInternalServerError, "Could not fetch repository from forge.")
 		return
 	}
+	from.ForgeID = user.ForgeID
```
anbraten, line 81: We already set ForgeID for new repos already in line 99

server/forge/bitbucket/bitbucket_test.go
6543: well yes but this are tests ... and we dont use the id ...

server/api/repo.go
6543: well not exactly see https://github.com/woodpecker-ci/woodpecker/pull/5806/commits/e02ddca0402c725bfaa481201576c10bc21f24ab

server/api/user.go
6543: same:

no that has nothing to do with db ... i just dont trust all forge adapters and forge addons to behave correctly and this ensures the ID is set correctly on our side

server/api/login.go
6543: no that has nothing to do with db ... i just dont trust all forge adapters and forge addons to behave correctly and this ensures the ID is set correctly on our side

server/api/user.go
anbraten: >  i just dont trust all forge adapters and forge addons to behave correctly and this ensures the ID is set correctly on our side

We should fix those records and the related code if that's really the case. In-fly migrations especially those that would not be saved to the db will never fix the underlying problem and are likely to be missed in other places and we would still wonder why it works in some other cases.

6543: well should i add an hard fail instead of an fix here then?

anbraten: @6543 Yes I would replace in-flight updates with a check like forgeID==0 that fails and returns an error and only update for repo update

anbraten: @6543 Yes I would replace in-flight updates with a check like forgeID==0 that fails and returns an error and only update for repo update

PR #5721, opened 2025-11-05
Fix workflow path filter for GitHub

chamburr:
The workflow path filter for conditional execution stopped working for GitHub since v3.11.0 (most probably #5618). As a result, all push workflows are triggered for every commit, even when the paths specified are not changed. This (untested) pull request attempts to fix the bug, which I believe is due to incorrect order of arguments.


kapilcdave:
Good job chamburr

Issue #5685, opened 2025-10-24
Agent reports incorrect version to server


DHandspikerWade:
### Component

agent

### Describe the bug

While running the agent via the `woodpeckerci/woodpecker-agent:v3.11.0` container image, the agent reports it's version as "next-7058dbcc96". 

### Steps to reproduce

1. Setup both the server and agent using the container image tagged with `v3.11.0`
2. Wait for the initialization
3. Check the agent's version either by reading its logs or navigating to `/admin/agents`

### Expected behavior

I would expect the agent to report with the stable version number. 

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.11.0"}

Agent manifest:

apiVersion: apps/v1
kind: StatefulSet
metadata:
  labels:
    app.kubernetes.io/component: agent
    app.kubernetes.io/name: woodpecker
    app.kubernetes.io/version: 3.11.0
  name: woodpecker-agent
  namespace: woodpecker
spec:
  persistentVolumeClaimRetentionPolicy:
    whenDeleted: Retain
    whenScaled: Retain
  podManagementPolicy: OrderedReady
  replicas: 1
  revisionHistoryLimit: 2
  selector:
    matchLabels:
      app.kubernetes.io/component: agent
      app.kubernetes.io/name: woodpecker
  serviceName: woodpecker-agent-headless
  template:
    metadata:
      labels:
        app.kubernetes.io/component: agent
        app.kubernetes.io/name: woodpecker
        app.kubernetes.io/version: 3.11.0
      name: woodpecker-agent
      namespace: woodpecker
    spec:
      containers:
      - env: [ REDACTED ]
        image: woodpeckerci/woodpecker-agent:v3.11.0
        imagePullPolicy: IfNotPresent
        name: agent
        ports:
        - containerPort: 3000
          name: http
          protocol: TCP
        resources:
          requests:
            cpu: 100m
            memory: 128Mi
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /etc/woodpecker
          name: agent-config
      dnsPolicy: ClusterFirst
      priorityClassName: realtime
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      serviceAccount: ci-agent
      serviceAccountName: ci-agent
      terminationGracePeriodSeconds: 30
      volumes:
      - name: agent-config
        persistentVolumeClaim:
[...]

PR #5722, opened 2025-11-05
agent: log injected version string at startup

Utkarsh9571:
This patch adds a startup log line to the Woodpecker agent that prints the injected version string.

- Uses `-ldflags` to inject `Version` at build time
- Logs the version using `zerolog` in `main.go`
- Helps validate build provenance and patch testing
- Safe to merge; does not affect agent registration or backend handshake

Example output:
{"level":"info","message":"Woodpecker Agent Version: localtest"}

This supports contributor validation for Issue #5685.


Utkarsh9571:
Hi team! Just checking in on this patch — it adds a startup log line to the agent that prints the injected version string, helping validate build provenance and patch testing.

- Scope: agent only
- No impact on registration or backend handshake
- Fully contributor-safe and reproducible

Would appreciate a quick review or workflow approval when possible. Thanks for your time!

qwerty287:
Thanks, but this doesn't look right. The version is stored at `version/version.go`.

Utkarsh9571:
@qwerty287 Thanks for the feedback! I've updated the patch to use `version.String()` from `version/version.go` for logging the agent version, as suggested. The custom `Version` variable and `-ldflags` logic have been removed for consistency.

Let me know if you'd like any further changes!

qwerty287:
Thanks, please revert any change that is not related.
- Restore the docker example
- remove the version.go file
- use tabs instead of spaces.

Checkout the diff to see how your change currently looks like, then you'll see what's wrong.

Also, I'd put the log statement into cmd/agent/core/agent.go.

Utkarsh9571:
@qwerty287 Thanks for the detailed feedback! I think i've made the required changes:

- Moved the version log to `cmd/agent/core/agent.go` as requested
- Removed `version/version.go`
- Restored `docker-compose.example.yaml`
- Reverted unrelated changes and ensured tab-based formatting with `gofmt`

Let me know if anything else needs adjustment!

6543:
that was the wrong `version.go` i would say

now server and cli have no way of knowing what version they are 😅

Utkarsh9571:
@6543 yeah, thanks for pointing that out man. I misunderstood the version.go context — happy to restore version reporting properly. Would you prefer I reintroduce the version string via `ldflags` and wire it into the CLI output for both agent and server? I can patch that cleanly.

6543:
it's already done so via our build chain just use it, doesnt have to be as complex ;)

6543:
https://github.com/woodpecker-ci/woodpecker/blob/14320372e3e2ca621486d647c99c513a818aa038/Makefile#L40

Utkarsh9571:
yeah, do you need me to do something else in this? i'm kinda new to open source, so don't know much how all that works.

qwerty287:
Just reverting all the changes except `cmd/agent/core/agent.go` should do it.

Utkarsh9571:
@qwerty287 i made a commit, is it good now?

qwerty287:
Sorry, but no.

**All** the files in the diff (you can see that in the files tab) except `cmd/agent/core/agent.go` have to have the same content as in `main`. So this file must be the only one that's shown there.

Utkarsh9571:
@qwerty287 bro, this branch is messed up, i created another branch and opened another pr #5724 

so check that out

PR #5728, opened 2025-11-06
add events query parameter to badge url

### LINKED ISSUE #4505: Wrong state icon
### Component

web-ui

### Describe the bug

We have a good state build, but the WEB UI shows a bad build result icon:
image

We have newer seen this problem before!

### Steps to reproduce

...don't know...

### Expected behavior

_No response_

### System Info

```shell
{
"version": "2.7.3"
}
```

### Additional context

_No response_

### Validations

tuxmainy:
Badges are currently only available for push events. To make them also available for other events, a URL query parameter `events` is added which can be set to one or multiple (separated by comma) events. This was also added into the repo badge settings page for easier generation of badges:

image

This change was discussed in https://github.com/woodpecker-ci/woodpecker/issues/4505

closes #4505 

PS: This is my very first time writing anything in go or vue. Please review carefully!


qwerty287:
Nice!

This looks quite good already. I'll do a full review later when I have more time, but just from a quick look on it two notes:
1. Your changes are not formatted correctly. Run `make format` in the root and `pnpm run format` in `web`.
2. Don't add translations in the PRs. You can translate the strings on our weblate once it's merged.

web/src/views/repo/settings/Badge.vue
```diff
@@ -98,7 +114,7 @@ const badgeContent = computed(() => {
   }
 
   if (badgeType.value === 'html') {
-    return `<a href="${baseUrl}${repoUrl.value}" target="_blank">\n  <img src="${baseUrl}${badgeUrl.value}" alt="status-badge" />\n</a>`;
+    return `<a href="${baseUrl}${repoUrl.value}" target="_blank">\n  <img src="${baseUrl}${badgeUrl.value.replace('&', '&amp;')}" alt="status-badge" />\n</a>`;
```
qwerty287, line 126: Why this change?

web/src/views/repo/settings/Badge.vue
```diff
@@ -113,5 +129,16 @@ watch(repo, () => {
 });
 
 const { t } = useI18n();
+
+const badgeEventsOptions: CheckboxOption[] = [
```
qwerty287, line 142: The push event should be selected by default.

And I'd say it must be necessary that at least one event must be selected as well.

web/src/assets/locales/de.json
qwerty287, line 1: As I said, revert this change and translate in weblate afterwards.

server/store/datastore/pipeline.go
```diff
@@ -35,11 +35,13 @@ func (s storage) GetPipelineNumber(repo *model.Repo, num int64) (*model.Pipeline
 	).Get(pipeline))
 }
 
-func (s storage) GetPipelineBadge(repo *model.Repo, branch string) (*model.Pipeline, error) {
+func (s storage) GetPipelineBadge(repo *model.Repo, branch string, events []string) (*model.Pipeline, error) {
 	pipeline := new(model.Pipeline)
+
```
qwerty287: No need for a newline here.

```suggestion
```

server/api/badge.go
```diff
@@ -72,7 +73,14 @@ func GetBadge(c *gin.Context) {
...
 
-	pipeline, err := _store.GetPipelineBadge(repo, branch)
+	// Events to lookup in the pipe, set multiple separated by comma
+	// If none given, fallback to former default "push"
+	events := c.Query("events")
+	if len(events) == 0 {
+		events = string(model.EventPush)
+	}
+
+	pipeline, err := _store.GetPipelineBadge(repo, branch, strings.Split(events, ","))
```
qwerty287: This should be validated and if there is at least one event that does not exist a 400 error should be returned.

Should be possible to do like this:
```suggestion
	// Events to lookup, multiple separated by comma
	var events []*model.WebhookEvent
	eventsQuery := c.Query("events")

	// If none given, fallback to default "push"
	if len(events) == 0 {
		events = []*model.WebhookEvent{model.EventPush)}
	} else {
		for _, event := range strings.Split(events, ",") {
			e := model.WebhookEvent(event)
			if err := e.Validate(); err != nil {
				c.AbortWithError(http.StatusBadRequest, err)
				return
			}
		}
	}

	pipeline, err := _store.GetPipelineBadge(repo, branch, events)
```

And then also change the `GetPipelineBadge` signature to use `[]*model.WebhookEvent`.

docs/docs/20-usage/80-badges.md
```diff
@@ -15,4 +15,11 @@ The status badge displays the status for the latest build to your default branch
...
 
-Please note status badges do not include pull request results, since the status of a pull request does not provide an accurate representation of your repository state.
+The badge represents the state of the last push event. If you'd like to respect other or further events, you can add the `events` query parameter:
+
+```diff
+-<scheme>://<hostname>/api/badges/<repo-id>/status.svg
++<scheme>://<hostname>/api/badges/<repo-id>/status.svg?events=manual,cron
+```
+
+Please note status badges should not include pull request results, since the status of a pull request does not provide an accurate representation of your repository state.
```
qwerty287: Can we move this line over l. 18 and change to something like this
```suggestion
By default status badges do not include pull request results, since the status of a pull request does not provide an accurate representation of your repository state.
```

web/src/views/repo/settings/Badge.vue
```diff
@@ -113,5 +129,16 @@ watch(repo, () => {
...
+  { value: WebhookEvents.Push, text: t('repo.pipeline.event.push') },
+  { value: WebhookEvents.Tag, text: t('repo.pipeline.event.tag') },
+  { value: WebhookEvents.Release, text: t('repo.pipeline.event.release') },
+  { value: WebhookEvents.PullRequest, text: t('repo.pipeline.event.pr') },
```
qwerty287, line 146: There's also `PullRequestClosed` and `PullRequestMetadata` (for secrets these are not shown as they are merged into a single option but here it's good to split them I think).

web/src/views/repo/settings/Badge.vue
tuxmainy: ampersands need to be escaped to produce valid html/xml

server/api/badge.go
tuxmainy: If I understood correctly, it wasn't possible to use a reference ([]*model.WebhookEvent) as the WebhookEvents are consts and cannot be used by addresses. So I changed it to []model.WebhookEvent.

web/src/views/repo/settings/Badge.vue
tuxmainy: > The push event should be selected by default.

done

> And I'd say it must be necessary that at least one event must be selected as well.

The implementation falls back to Push event if no parameter is given. This is necessary to not break already used badge URLs. So requiring an event would be "a lie" as it's not working like this. Also requiring an event would make the UI more complex in style and error handling for something that is not needed.

tuxmainy:
>     1. Your changes are not formatted correctly. Run `make format` in the root and `pnpm run format` in `web`.

done

web/src/views/repo/settings/Badge.vue
qwerty287: I know, but in UI, if you unselect all options, it will fallback to this default and use `push`. So in this case `push` should be auto-selected again. Just that it's clear that "nothing selected" means "only push".

lafriks: But then you can't copy url without params

qwerty287: You can, in the beginning without changing anything.

I don't think most will care, but then I'd just remove the URL param if only push is selected.

lafriks: Yes, removing param if only push is selected would be best option

tuxmainy: So, the UI should behave as follows, did I understand correctly?

1. when starting, Push checkbox is preselected, badge URL has no events parameter
2. when selecting additional or other events, badge URL has events parameter according to selection
3. when only selecting Push checkbox again, badge URL has no events parameter
4. when deselecting all events checkboxes, the Push event should be selected automatically

I can live with that, but it feels crazy complicated. Would a hint "if you don't select anything, push is default" be sufficient?

qwerty287: Yes, that's what we mean. I prefer this as the explicite way, I wouldn't like just a hint.
This basically is also how the branch selector works (besides the point that you can't unselect it).

tuxmainy: But I need a hint :)

I have following done so far which handles 1, 2, 3 and half of 4. Point 4 only works, if, and only if, push is not checked and you uncheck another event (e.g. tag). If push is checked and you uncheck push then nothing happens in the UI but the value is added in events. At his point, things become bad...

```diff
diff --git a/web/src/views/repo/settings/Badge.vue b/web/src/views/repo/settings/Badge.vue
index ab3517055..af4fa47ac 100644
--- a/web/src/views/repo/settings/Badge.vue
+++ b/web/src/views/repo/settings/Badge.vue
@@ -31,7 +31,7 @@
       <SelectField :id="id" v-model="branch" :options="branches" required />
     </InputField>
     <InputField v-slot="{ id }" :label="$t('repo.settings.badge.events')">
-      <CheckboxesField :id="id" v-model="events" :options="badgeEventsOptions" />
+      <CheckboxesField :id="id" v-model="events" :options="badgeEventsOptions" @update:model-value="eventsChanged" />
     </InputField>
 
     <div v-if="badgeContent" class="flex flex-col space-y-4">
@@ -94,7 +94,10 @@ const badgeUrl = computed(() => {
   }
 
   if (events.value.length > 0) {
-    params.push(`events=${encodeURIComponent(events.value.join(','))}`);
+    // dont set events parameters, if only WebhookEvents.Push is selected, as this is the default behaviour
+    if (events.value.length !== 1 || events.value.at(0) !== WebhookEvents.Push) {
+      params.push(`events=${encodeURIComponent(events.value.join(','))}`);
+    }
   }
 
   return `${rootPath}/api/badges/${repo.value.id}/status.svg${params.length > 0 ? `?${params.join('&')}` : ''}`;
@@ -143,4 +146,11 @@ const badgeEventsOptions: CheckboxOption[] = [
 ];
 
 useWPTitle(computed(() => [t('repo.settings.badge.badge'), repo.value.full_name]));
+
+function eventsChanged() {
+  if (events.value.length === 0) {
+    // This does not set checkbox when only Push was selected. Why?
+    events.value.push(WebhookEvents.Push);
+  }
+}
```

Do I need to inform the UI element that the value has changed again? As said, first time vue.

PS: `@update:model-value` only gets raised when a value is unchecked due to the nature of [adding the value this way](https://github.com/woodpecker-ci/woodpecker/blob/a7f275c0032c5a3a27cec96e0a6504877ba46438/web/src/components/form/CheckboxesField.vue#L39), value is not really changed. But anyway, getting called when a value is removed is totally fine for this.

PPS: I am off until tomorrow evening (UTC)

tuxmainy: Made it work by adding `await nextTick()`. Please test before resolving, thanks.

qwerty287:
Thanks, works quite nice so far - just another point regarding the checkboxes: Would it be possible to disable the push checkbox if there's nothing else selected? Because right now it feels weird clicking the box and nothing happens as it's just re-selected.
Then I also think the `nextTick` should no longer be necessary.

tuxmainy:
> Thanks, works quite nice so far - just another point regarding the checkboxes: Would it be possible to disable the push checkbox if there's nothing else selected? Because right now it feels weird clicking the box and nothing happens as it's just re-selected. Then I also think the `nextTick` should no longer be necessary.

Push checkbox is now disabled if only push is selected. And yes, `nextTick` is removed. Also I added the description `default`, so the user might better understand why this checkbox is behaving special.

Is using `!important` in CSS appropriate?

qwerty287:
You should be able to use `disabled:bg-wp-control-neutral-300 disabled:border-wp-control-neutral-200` as CSS classes from tailwind. You don't have to define your own.

tuxmainy:
> You should be able to use `disabled:bg-wp-control-neutral-300 disabled:border-wp-control-neutral-200` as CSS classes from tailwind. You don't have to define your own.

Used said classes and removed the ones I defined from style.css. But I still override the color of the checkmark as the default white has low contrast and is hard to see (and also fails if you use auto dark mode plugin in browser).

web/src/components/form/Checkbox.vue
```diff
@@ -65,6 +67,10 @@ const id = (Math.random() + 1).toString(36).substring(7);
   opacity: 0;
 }
 
+.checkbox:disabled::before {
+  border-color: var(--color-gray-600);
```
xoxys: Does this work in dark mode properly as well?

tuxmainy: Did not, using a color which is changing in respect of dark/light mode

lafriks (approved):

PR #5666, opened 2025-10-21
Add support for sidecar containers and volumes in Kubernetes backend

MartinSchmidt:
Introduce functionality to support sidecar containers in the Kubernetes backend, allowing additional containers to run alongside the main step container.
Implement volume support for sidecars to facilitate shared resources between containers. Provide examples and configuration options in the documentation.

Solves #3977

PR #5729, opened 2025-11-07
Add support for sidecar containers and volumes in Kubernetes backend

MartinSchmidt:
Continuation of #5666, reverted in #5726

@qwerty287 Here is the PR again. 

I hear we have two aspects you want me to look at:

- Volume mount
- Security (privileged) perspective

## Volumes

The current way i implemented the volumes for sidecars **only allow for EmptyDir mounts**, which is basically a folder that is shared between the sidecar and the main container.
Sidecars is more of a _pattern_, than an actual sidecar. All containers in a Pod in k8s are actually equal. They share the same network namespace, as in, see each other as on the same host, but still run in isolated user namespaces. They are scheduled as one "unit", and one can create EmptyDir volumes on the pod, which the containers can mount, within which they can share files locally, so there is no way to mount external volumes in this new implementation.

## Privileged

I read through some of the code, and i think I found out that you use the [linter.go](https://github.com/woodpecker-ci/woodpecker/blob/1109f87d7d0dbde6a520f4bd88ede12c37a28c67/pipeline/frontend/yaml/linter/linter.go#L250) to verify if the container is allowed to be scheduled, correct? 

I am now trying to think about a solution, since, I wouldn't like to embed specific backend stuff into the linter.go, since that would be wrong. But i can clearly see the issue here, that i introduced a whole new place where reference to images can be located, which is only relevant for the specific backend.

Any suggestions?

### DIFF
--- a/cmd/agent/core/agent.go
+++ b/cmd/agent/core/agent.go
@@ -65,6 +65,8 @@ var (
 )
 
 func run(ctx context.Context, c *cli.Command, backends []types.Backend) error {
+	log.Info().Str("version", version.String()).Msg("Starting Woodpecker agent")
+
 	agentCtx, ctxCancel := context.WithCancelCause(ctx)
 	stopAgentFunc = func(err error) {
 		msg := "shutdown of whole agent"
--- a/web/src/components/atomic/Badge.vue
+++ b/web/src/components/atomic/Badge.vue
@@ -2,7 +2,7 @@
   <span class="inline-flex items-center text-xs font-medium">
     <span
       v-if="label"
-      class="border-wp-state-neutral-100 bg-wp-state-neutral-100 rounded-l-full border-1 py-0.5 pl-2 whitespace-nowrap text-white"
+      class="border-wp-state-neutral-100 bg-wp-state-neutral-100 rounded-l-full border-1 py-0.5 pr-1 pl-2 whitespace-nowrap text-white"
       :class="{
         'rounded-r-full pr-2': !value,
       }"


MartinSchmidt:
I just spent more time looking into it, and when I also see how to check for if the sidecar is a plugin, to enable run it as privileged by default, this would also need changes..

I am beginning to see a lot of **cracks** form with the original intent, and that it might just be best to actual look towards using VMs for the workers that require this instead, or look at more custom image (our current solution), this currently allows all of the developers to run privileged, this is not pretty, and undermines some of the ideas, but works.

```yaml
when:
  - event: pull_request
steps:
  - name: test
    image: docker:24.0.9
    privileged: true
    commands:
      - dockerd-entrypoint.sh  1> /dev/null 2>  /dev/null &
      - apk add openjdk21
      - ./mvnw test
```

I can see that this idea of creating a sidecar, is not in line with the architecture as a whole, where every container scheduled, is a step (also services and clone) and that it would either require restructuring, based on the requirement for a specific backend, since we would basically be introducing a new place for containers.

Alternative solution would be to isolated changes to only be in the kubernetes backend, and rewrite something about how these are done. We could "reuse" services and allow them to be "taken" as a sidecar instead, something like. Then not schedule it as an individual service, but inject it as sidecars. But now we would be running them not separately, and I have a trouble following how information about the expected services are sent back to the UI

```yaml
steps:
  - name: build-with-docker
    image: docker:cli
    commands:
      - docker build -t my-app .
      - docker run --rm my-app
    backend_options:
      kubernetes:
        sidecars:
        - docker-in-docker

services: # would not run separately since the backend_option above, but linting and so on would work as expected with current model
  - name: docker-in-docker
    image: docker:dind
    privileged: true
```

qwerty287:
Thanks.

I also just noticed the trusted configuration doesn't work well currently with the backends (I have #4675 open that would actually allow this to be way more flexible).

Right now it might be enough (for the security issues) to allow to disable the sidecar feature completely, so users can enable it if they have a safe environment that cannot be abused. Similar to e.g. https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/backend/kubernetes/flags.go#L108.

But do I read that correctly now that it's not possible to mount arbitrary dirs from the host system using the sidecars? Because that's how it looked to me (but I have no idea about how k8s works).

MartinSchmidt:
@qwerty287 I have spent some time going through different solutions, but all of them comes back to the point that trying to do this would go against the architecture you have in woodpecker. 

I went back into thinking mode, and looking at different ways to solve it instead of forcefully trying to implement sidecar containers (specifically for using test-containers in a docker-in-docker setup) and I think I found a simple solution.

I made a PR #5764 for headless services  and that should solve the issues, now dind can run in its own service, and we don't have to predefine the ports.

PR #5205, opened 2025-05-27
feat(bitbucketdatacenter): support for pull request file changes

henkka:
Closes #1138.

This pull request brings the missing "Get pull request file changes" functionality for the Bitbucket Datacenter forge.

The `replace` directive from the `go.mod` shall be removed once https://github.com/neticdk/go-bitbucket/pull/17 is merged and new version released in the `neticdk/go-bitbucket` library, but I figured I might open this pull request already for review.

PR #5730, opened 2025-11-07
Support for file changes in Bitbucket Cloud

MeurillonGuillaume:
This pull-request implements the `when.path` filter for Bitbucket Cloud by requesting the changed files for a commit when a `/api/hook` is received.

Similar to #5205, but for the Bitbucket Cloud forge.


qwerty287:
Thanks! Looks very good already. Could you check the formatting? https://ci.woodpecker-ci.org/repos/3780/pipeline/29586/16

And a question: How does that work with pull requests?

lafriks (approved):

MeurillonGuillaume:
@qwerty287 thanks for the heads-up, I fixed the formatting of the docs.

For pull-requests the source commit is set as Pipeline Commit:
https://github.com/woodpecker-ci/woodpecker/blob/300e2a083bb9fb72cda4b76b6c2d56f531c4917f/server/forge/bitbucket/convert.go#L167-L169

and so it will work the same as the Commit Hook as the diffstat is loaded after parsing the Hook content. I just tested it on my instance and everything worked as expected.

qwerty287:
OK, but for PRs it should load the changes for the full PR. So all commits together. That's how the other forges work.

MeurillonGuillaume:
@qwerty287 Good catch. I've updated the PR so that it lists all changes between the source & destination branch using the [diffstat API](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-commits/#api-repositories-workspace-repo-slug-diffstat-spec-get).

server/forge/bitbucket/bitbucket.go
```diff
@@ -410,6 +410,28 @@ func (c *config) Hook(ctx context.Context, req *http.Request) (*model.Repo, *mod
...
+		if pr == nil {
+			return nil, nil, fmt.Errorf("can't run hook against empty PR information")
+		}
+
+		// List all changes between source & destination branch
+		pl.ChangedFiles, err = client.ListChangedFiles(repo.Owner, repo.Name, fmt.Sprintf("%s..%s", pr.PullRequest.Source.Branch.Name, pr.PullRequest.Dest.Branch.Name))
+		if err != nil {
+			return nil, nil, err
+		}
+	default:
```
qwerty287: Is this empty default actually necessary? (I'm not sure as well, would have to try)

qwerty287 (commented):
Thanks. Seems good so far, just a simple question

server/forge/bitbucket/bitbucket.go
MeurillonGuillaume: The `default` case is optional, I'll leave it out. I probably added it as a force of habit 😅

PR #5748, opened 2025-11-13
feat(bitbucketserver): get changes from all commits in a single push event

hhamalai:
Currently, on a push event, only the changed files from the last commit are used  e.g. to check path conditions:
```yaml
# ... pipeline
when:
  - path:
      include: ['*.md']
``` 

In general the push event can contain multiple commits, which currently leads to missed pipeline executions. An example with the snippet above would a user commiting changes to their *.md files, after which making other commits targeting other files. Pushing commits would not trigger the pipeline above if the last commit would not have any changes to .md files.

The proposed change uses the range of commits (From&To) in the webhook payload with Bitbucket data center compare commits API to list the changed files across a set of commits.

In case the branch is created/deleted, bitbucket reports commit as 0000000000000000000000000000000000000000". The current implementation then checks changes against the default branch. This has its own problems and might cause too wide executions, but the bitbucket webhook implementation also truncates the list of commits to include in the payload to  max 5 commits, so there are no information in the payload that could be used as an anchor point as "the earliest" commit/reference in case of new branch.

This change requires changes to neticdk/go-bitbucket (pr open https://github.com/neticdk/go-bitbucket/pull/37)


server/forge/bitbucketdatacenter/parse.go
```diff
@@ -17,10 +17,10 @@ type HookResult struct {
 	Payload  []byte
 }
 
-func parseHook(r *http.Request, baseURL string) (*HookResult, error) {
+func parseHook(r *http.Request, baseURL string) (_ *HookResult, currCommit, prevCommit string, _ error) {
```
qwerty287: This shouldn't be required.

```suggestion
func parseHook(r *http.Request, baseURL string) (*HookResult, string, string, error) {
```

server/forge/bitbucketdatacenter/parse_test.go
```diff
@@ -60,10 +64,12 @@ func Test_parseHook(t *testing.T) {
 		req.Header.Set("Content-Type", "application/json")
 		req.Header.Set("X-Event-Key", "repo:refs_changed")
 
-		result, err := parseHook(req, "https://bitbucket.example.com")
+		result, curCommit, prevCommit, err := parseHook(req, "https://bitbucket.example.com")
 
 		assert.NoError(t, err)
 		assert.NotNil(t, result)
+		assert.NotEmpty(t, curCommit)
+		assert.NotEmpty(t, prevCommit)
```
qwerty287: Can you also check for the correct value?

hhamalai:
> And then let's wait for the lib to merge the PR.

new release is out from go-bitbucket library, and this PR is updated to use it

Issue #3977, opened 2024-07-25
K8S Backend - Ability to add a side container to a step (pod)


eliasscosta:
### Clear and concise description of the problem

Sometimes we have a test step where they need to run inside a docker image (dind), because a few tests are written with the TestContainer library. 

To achieve it, we need to configure one step like this: 

```yaml
  allTests:
    image: docker:24.0.9
    commands:
      - apk add openjdk17
      - dockerd-entrypoint.sh --registry-mirror=http://myprivate-mirror.ior:5000 1> /dev/null 2> /dev/null &
      - ./gradlew test myTEST
    when:
      event: push
```

This example is in the easy mode, but sometimes we need to install more than one jdk or the dind image doesn't have the library in the OS packages.

### Suggested solution

The solution was to add into the `backend_options` a way to create a side container with the main step container.

```yaml
  allTests:
    image: openjdk17
    commands:
      - ./gradlew test myTEST
    when:
      event: push
    backend_options:
      kubernetes:
        container: 
          - image: docker:dind
            env: []
```

Then both containers can start and we can access the docker in the local host and execute the tests with the TestContainers.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

MartinSchmidt:
I am having the same issue when trying to use test containers and dind, since it creates a lot of containers, which needs port to be forwarded, which is not really a good solution as is. Other would be to run a single thick container, running both the test and dind, but it just seems wrong.

I created a PR #5666 with a suggestion, it is still a draft, but would love some feedback. I still need to test it on my cluster.

xoxys:
Why does it need special handling on k8s? Is there a reason not to use services/detached steps?

MartinSchmidt:
@xoxys  Services i k8s require one to forward specific ports, and these ports are with testcontainers not always know ahead of time, and can be quite many.

Sidecars containers all run in the same pod, and network wise is localhost within the pod, so this makes it easy to handle these usecases.

It also would enable one to use empty volume for mounting the docker socket between the containers.

xoxys:
Fixed by https://github.com/woodpecker-ci/woodpecker/pull/5764

PR #5764, opened 2025-11-18
Add support for headless Kubernetes services

MartinSchmidt:
Introduce an [headless services](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) for the entire workflow run. This can resolve the name for all steps similar to how docker does within the workflow. 

These are used for all steps (except services and detached, that has ports specified, since those use the ClusterIP services as until now)

Update tests to validate this new functionality, and added a snipped in the doc.

This enables a use case when one is using test-containers in a dind (docker-in-docker) since these require many ports, which one would prefer not to have to pre-define.

This should solve the underlying issue behind #3977 albiet in another way that sidecars.

### DIFF
--- a/server/api/repo.go
+++ b/server/api/repo.go
@@ -78,7 +78,6 @@ func PostRepo(c *gin.Context) {
 		c.String(http.StatusInternalServerError, "Could not fetch repository from forge.")
 		return
 	}
-	from.ForgeID = user.ForgeID
 	if !from.Perm.Admin {
 		c.String(http.StatusForbidden, "User has to be a admin of this repository")
 		return
@@ -88,6 +87,7 @@ func PostRepo(c *gin.Context) {
 		return
 	}
 
+	from.ForgeID = user.ForgeID
 	if enabledOnce {
 		repo.Update(from)
 	} else {
@@ -96,7 +96,6 @@ func PostRepo(c *gin.Context) {
 		repo.AllowPull = server.Config.Pipeline.DefaultAllowPullRequests
 		repo.AllowDeploy = false
 		repo.CancelPreviousPipelineEvents = server.Config.Pipeline.DefaultCancelPreviousPipelineEvents
-		repo.ForgeID = user.ForgeID // TODO: allow to use other connected forges of the user
 	}
 	repo.IsActive = true
 	repo.UserID = user.ID
--- a/server/forge/bitbucket/bitbucket.go
+++ b/server/forge/bitbucket/bitbucket.go
@@ -50,7 +50,7 @@ type Opts struct {
 }
 
 type config struct {
-	id            int64
+	forgeID       int64
 	api           string
 	url           string
 	oAuthClientID string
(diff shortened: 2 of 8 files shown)


MartinSchmidt:
Just remembered that the service cluster ip might be returned another place and used as hosts, so this might not work as intended.. I will look into it.

qwerty287:
@MartinSchmidt is this ready for review from your side?

MartinSchmidt:
@xoxys i made a different solution for the type of usecase from the #3977

I have tested it on a k8s cluster with the following config:

```yaml
when:
  - event: push
    branch: main

services:
  - name: docker
    image: docker:28.5.1-dind
    privileged: true
    environment:
      DOCKER_TLS_CERTDIR: /woodpecker/dind-certs

steps:
  - name: test-docker
    image: docker:cli
    environment:
      DOCKER_HOST: 'tcp://docker:2376'
      DOCKER_CERT_PATH: '/woodpecker/dind-certs/client'
      DOCKER_TLS_VERIFY: '1'
    commands:
      - sleep 5 # wait for docker daemon to be ready
      - docker run hello-world
```

result, everything works as expected

image

MartinSchmidt:
> qwerty287

sure is :)

xoxys:
> @xoxys i made a different solution for the type of usecase from the #3977
> 
> I have tested it on a k8s cluster with the following config:

That's exactly what I was talking about :D nice that it works for you

MartinSchmidt:
@xoxys quick question, I was sitting and adding some additional tests, and actually though Woodpecker supported port mapping in the services, but I can see that is not the case, since that is not the case, I would say the entire functionality of the original services is no longer needed, since they then add no functionality, but add code and complexity. 
I would like some feedback on what you think would be best, strip it out?

Look [here](https://github.com/MartinSchmidt/woodpecker/pull/1) for my thoughts on it

qwerty287:
This looks good to me in general from the code (without the additional PR you linked), but I can't tell anything about how it works with k8s. Maybe @woodpecker-ci/maintainers someone else could check it out.

xoxys:
@MartinSchmidt  Sorry for the delay, can you explain why a headless service is needed at all? Why do we need to resolve to the Pod IP instead of the cluster IP of a regular service? Resolving the IP by DNS should work for both service types, so is there any benefit to use a headless service instead of a regular one?

xoxys:
> This enables a use case when one is using test-containers in a dind (docker-in-docker) since these require many ports, which one would prefer not to have to pre-define.

Nevermind, it's because of this requirement.

>  I was sitting and adding some additional tests, and actually though Woodpecker supported port mapping in the services, but I can see that is not the case

True, as `targetPort` == `port` in the regular service implementation there is no real advantage over headless services. In that case I agree, we can drop it and replaces it by headless services overall 👍

xoxys:
Should we do the replacement in this PR? Feels a bit weird to introduce a new feature and refactor it again in a direct follow-up PR but no strong opinion.

MartinSchmidt:
@xoxys is the security thing in the frontend okay? since it is not something I have been changing.
Other than that everything should be ready :)

xoxys:
Yes thats fine, unrelated to your change.

MartinSchmidt:
So is there more i need to do before it can be merged? :)

PR #5767, opened 2025-11-19
Improve repair endpoints

anbraten:
- repair endpoint was not using the updated user in case it changed
- repair all endpoint was failing when having issues with a repair of a single repo

### DIFF
--- a/server/api/repo.go
+++ b/server/api/repo.go
@@ -649,7 +649,7 @@ func RepairAllRepos(c *gin.Context) {
 	}
 
 	if len(failedRepos) > 0 {
-		c.JSON(http.StatusInternalServerError, map[string]interface{}{
+		c.JSON(http.StatusInternalServerError, map[string]any{
 			"error":        "failed to repair some repos",
 			"failed_repos": failedRepos,
 		})


server/api/repo.go
```diff
@@ -635,98 +636,113 @@ func RepairAllRepos(c *gin.Context) {
...
-			_ = c.AbortWithError(http.StatusInternalServerError, err)
-		}
-		return
+		log.Error().Err(err).Msgf("cannot assign user to repo '%s'", repo.FullName)
+		return err
 	}
 
-	// creates the jwt token used to verify the repository
+	// refresh user token if necessary
+	forge.Refresh(c, _forge, _store, repoUser)
```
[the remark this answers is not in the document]
anbraten: forge.Refresh updates repoUser by ref, no need to fetch it again

server/api/repo.go
```diff
@@ -635,98 +636,112 @@ func RepairAllRepos(c *gin.Context) {
...
 	for _, r := range repos {
-		repairRepo(c, r, false, true)
-		if c.Writer.Written() {
-			return
+		// updatePermissions is false as RepoListAll does not load permissions
+		updatePermissions := false
+		err := repairRepo(c, r, updatePermissions)
+		if err != nil {
+			hadRepairErr = true
+			_ = c.Error(err)
```
qwerty287, line 646: Do I read that correctly that it will also the user which repos failed? So in the HTTP response.

qwerty287 (approved):

server/api/repo.go
anbraten: I've changed it to return an array of the failed repo ids.

PR #5769, opened 2025-11-19
fix(bitbucketdatacenter): fix CI_COMMIT_PULL_REQUEST

### LINKED ISSUE #5768: Pull-request ID not available as environment variable in Bitbucket Datacenter
### Component

server

### Describe the bug

When trying to use the environment variable `CI_COMMIT_PULL_REQUEST` in woodpecker 3.11, with Bitbucket Datacenter, it is an empty string.

The reason for this is that the variable is populated from the _number_ present in the `ref`, which is extracted in [line 110 of environment.go](https://github.com/woodpecker-ci/woodpecker/blob/9ec0b5cdcb592378bd99a0478b6a71ae3896127b/pipeline/frontend/metadata/environment.go#L110).

This works for bitbucket and other forges, presumably because they have been coded to include the name, e.g. github, which [constructs the ref](https://github.com/woodpecker-ci/woodpecker/blob/9ec0b5cdcb592378bd99a0478b6a71ae3896127b/server/forge/github/parse.go#L200) as ["refs/pull/%d/head"](https://github.com/woodpecker-ci/woodpecker/blob/9ec0b5cdcb592378bd99a0478b6a71ae3896127b/server/forge/github/convert.go#L43).

In Bitbucket server, this value is just taken from the pull-request event payload.

As far as I know, the Bitbucket datacenter equivalent is `refs/pull-requests/%d/from`, which is [what it is in Bitbucket cloud](https://github.com/woodpecker-ci/woodpecker/blob/9ec0b5cdcb592378bd99a0478b6a71ae3896127b/server/forge/bitbucket/convert.go#L170).

I would like to change the bitbucket server component to use this same approach to construct the "from" ref, so that the pull-request ID may be forwarded to pipelines.  I would be happy to contribute and test this in a live system.

### Steps to reproduce

1. Have Woodpecker installed configured to use Bitbucket Server.
2. Open a pull-request to trigger a workflow, it might get an ID of `1`.
3. Print the CI_COMMIT_PULL_REQUEST environment variable.

What happens: `CI_COMMIT_PULL_REQUEST=` (the value is unset or an empty string)

### Expected behavior

Expected value: `CI_COMMIT_PULL_REQUEST=1`

### System Info

```shell
Woodpecker 3.11.0
Data Center 9.4.13
```

### Additional context

* The bitbucket server event payload (like github's event) does not include this refs/pull-requests/123/from, so it is not possible to discover.  It must be constructed.
* The bitbucket server format "refs/pull-requests/123/from" is (as far as I know) not documented anywhere, and seeing as Datacenter is going away, I doubt this (or the fact that _the refs are actually there_) will change.

### Validations

mogsie:
fixes #5768

The CI_COMMIT_PULL_REQUEST environment variable was not filled out for Bitbucket Datacenter edition. This was because this ID comes from the `ref`, which for this forge was set to the branch name of the pull-request, and not the reference that contains the pull-request ID.

The pullRequest.fromRef.ID is "refs/heads/branch-name".

Bitbucket datacenter also publishes this as "refs/pull-requests/123/from"

This change switches from using the ref from the event payload to the one that contains the pull-request ID in the path, allowing the CI_COMMIT_PULL_REQUEST variable to be set correctly.


Levy-Tal:
There is a reason why they changed it, look at #5162

mogsie:
Oh, I see.  So I'd have to find a different approach to solving this.  It's almost as if adding an explicit _pull request ID_ to some struct would be better than actually parsing the ref.  I see that #4626 would provide `Index` that would make this easy to fix.  In the meantime, perhaps there is a better way to fix #5158 than changing the ref.

Levy-Tal:
I also can confirm that the env var doesn't work after upgrade from v3.5.0 to v3.12.0
Adding @rruzicic and @langecode.

qwerty287 (approved):

PR #5773, opened 2025-11-21
feat: include CI_COMMIT_TAG env in deployment events

henkka:
# Problem

Bitbucket Data Center does not support [`release` events](https://woodpecker-ci.org/docs/next/administration/configuration/forges/overview), forcing workflows to rely on `deployment` events instead. Currently, the `CI_COMMIT_TAG` environment variable is not populated for deployment events, which prevents our teams using Woodpecker CI from successfully running tag-based deployment pipelines.

# Solution

Updated the logic to apply for `deployment` as well to ensure that `CI_COMMIT_TAG` is correctly populated when the pipeline is triggered by a `deployment` event for tags, enabling tag-based deployments for Bitbucket Data Center forge users.


qwerty287:
~~Thanks, but at least if the deployment does not come from a tag, it should not include this var.~~

Edit, sorry I see that seems to be the case already.

qwerty287 (approved):

PR #5790, opened 2025-11-28
fix: updateRepoPermissions to cleanup old permissions

henkka:
# Problem

Currently, the repository permission synchronization process is effective only for adding access, not removing it. When a user loses access to a repository on the forge, the system fails to remove the corresponding entry from the perms table. This results in stale permissions persisting in the database, allowing users to retain CI access to repositories they should no longer be able to view or interact with.

# Solution

Reworked the `updateRepoPermissions` logic to perform a full reconciliation of user permissions. In addition to upserting new permissions, it now compares the fetch results against stored data and actively deletes permissions for repositories that are no longer present on the forge. This ensures that local access rights accurately reflect the current state of the user's permissions on the forge.

### DIFF
--- a/.woodpecker/test.yaml
+++ b/.woodpecker/test.yaml
@@ -73,7 +73,7 @@ steps:
   lint-license-header:
     image: *golang_image
     commands:
-      - go install github.com/google/addlicense@latest # cspell:words addlicense
+      - make install-addlicense # cspell:words addlicense
       - 'addlicense -check -ignore "vendor/**" **/*.go'
     when: *when
 
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ vendor: ## Update the vendor directory
 	go mod tidy
 	go mod vendor
 
-format: install-tools ## Format source code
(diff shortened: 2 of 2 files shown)


qwerty287:
Looks good to me - but it will likely take even more time when you login (#5665)… Can you check the tests?

henkka:
> Looks good to me - but it will likely take even more time when you login (#5665)… 

I'd say really marginally as we're only doing DB calls. I wasn't aware of the mentioned PR, but I really like the async feature flag implemented there👍 

> Can you check the tests?

Oops, fixed tests and added new test for this case.

qwerty287:
Is it actually necessary to get the current repos? Couldn't you do a `NOT IN`?

henkka:
> Couldn't you do a `NOT IN`?

good point! Refactored the whole implementation with that in mind

qwerty287:
The test still seems to fail. Otherwise it looks good to me

henkka:
Great! The mock expectation has been updated now

PR #5809, opened 2025-12-01
Allow packagers to set WebUI root to custom path

6543:
this will allow packages to distributors the server binary and the webui as separate components.

image

see https://github.com/NixOS/nixpkgs/pull/465460#issuecomment-3593889098


web/web_external.go
```diff
@@ -0,0 +1,54 @@
+// Copyright 2023 Woodpecker Authors
```
6543:
```suggestion
// Copyright 2025 Woodpecker Authors
```

6543:
https://woodpecker-ci-woodpecker-pr-5809.surge.sh/docs/next/development/packaging

docs/docs/92-development/10-packaging.md
```diff
@@ -0,0 +1,18 @@
+# Packaging
+
+If you repackage it, we encourage to build from source, witch requires internet connection.
```
qwerty287:
```suggestion
If you repackage it, we encourage to build from source, which requires internet connection.
```

docs/docs/92-development/10-packaging.md
```diff
@@ -0,0 +1,18 @@
...
+
+We also for offline builds provide a tarball with all dependencies vendored and WebUI prebuild
```
qwerty287:
```suggestion
For offline builds we also provide a tarball with all dependencies vendored and web UI prebuilt
```

docs/docs/92-development/10-packaging.md
```diff
@@ -0,0 +1,18 @@
...
+on the [release page](https://github.com/woodpecker-ci/woodpecker/releases).
+
+## Distribute WebUI on own directory
```
qwerty287:
```suggestion
## Distribute web UI in own directory
```

docs/docs/92-development/10-packaging.md
```diff
@@ -0,0 +1,18 @@
...
+
+If you want to not embed the WebUI into the binary, you can compile a custom WebUI root path into the binary.
```
qwerty287:
```suggestion
If you do not want to embed the web UI into the binary, you can compile a custom web UI root path into the binary.
```

docs/docs/92-development/10-packaging.md
```diff
@@ -0,0 +1,18 @@
+# Packaging
```
anbraten, line 1: Would not add those super specific details to the docs as they are likely become outdated / we would need to support this a kind of stable way and it would mainly confuse 99% of users. Would suggest to add the relevant parts to the code instead.

6543: well i thought if i should document it or not ... and so i in the end did document it but i think if it does hurt us we can always remove it

xoxys: Only 18 lines? We are not able to keep it updated? IMO its good to document it.

docs/docs/92-development/10-packaging.md
```diff
@@ -0,0 +1,18 @@
...
+For offline builds we also provide a tarball with all dependencies vendored and web UI prebuilt
+on the [release page](https://github.com/woodpecker-ci/woodpecker/releases).
+
+## Distribute web UI in own directory
+
+If you do not want to embed the web UI into the binary, you can compile a custom web UI root path into the binary.
+
+Add `external_web` into the tags and use `-X go.woodpecker-ci.org/woodpecker/v3/web.webUIRoot=/some/path` build flag to set custom path.
+
+example: <!-- cspell:ignore webui -->
```
xoxys:
```suggestion
Example: <!-- cspell:ignore webui -->
```

docs/docs/92-development/10-packaging.md
```diff
@@ -0,0 +1,18 @@
+# Packaging
+
+If you repackage it, we encourage to build from source, which requires internet connection.
+
+For offline builds we also provide a tarball with all dependencies vendored and web UI prebuilt
+on the [release page](https://github.com/woodpecker-ci/woodpecker/releases).
+
+## Distribute web UI in own directory
+
+If you do not want to embed the web UI into the binary, you can compile a custom web UI root path into the binary.
```
xoxys:
```suggestion
If you do not want to embed the web UI in the binary, you can compile a custom root path for the web UI into the binary.
```

docs/docs/92-development/10-packaging.md
```diff
@@ -0,0 +1,18 @@
...
+
+Add `external_web` into the tags and use `-X go.woodpecker-ci.org/woodpecker/v3/web.webUIRoot=/some/path` build flag to set custom path.
```
xoxys:
```suggestion
Add `external_web` to the tags and use the build flag `-X go.woodpecker-ci.org/woodpecker/v3/web.webUIRoot=/some/path` to set a custom path.
```

docs/docs/92-development/10-packaging.md
```diff
@@ -0,0 +1,18 @@
+# Packaging
+
+If you repackage it, we encourage to build from source, which requires internet connection.
+
+For offline builds we also provide a tarball with all dependencies vendored and web UI prebuilt
+on the [release page](https://github.com/woodpecker-ci/woodpecker/releases).
```
xoxys:
```suggestion
For offline builds, we also offer a tarball with all vendored dependencies and a pre-built web UI
on the [release page](https://github.com/woodpecker-ci/woodpecker/releases).
```

xoxys (approved):
Some wording nits.

docs/docs/92-development/10-packaging.md
lafriks: Maybe it should be better in git not in docs?

xoxys: I dont get why we try that often to hide information/make it hard to find it.

anbraten: TLDR: Please lets not spent too much time on this discussion and just put them next to the code which documents such things as well. 

> Only 18 lines? We are not able to keep it updated? IMO its good to document it.

Outdated docs are more confusing then helping IMO. In the end you need to look into the code for such specific details again if outdated, therefore I favor documenting such things it in code directly instead. 

> I dont get why we try that often to hide information/make it hard to find it.

- mainly to not overload 99% of users with such specific implementation details (why would a normal "advanced" user care about setting this, compiling from source is quite a special case already, using something else to go:embed is nothing someone apart from the nixos build cared before?)
- one result from such details is the docs search will index them as well, bloating the results with pretty unimportant details
- for an end user the docs are the source of truth for how to work with WP

xoxys: No, that's why we have a dedicated developer section in the docs? If we argue this way, we can drop it entirely.

> TLDR: Please lets not spent too much time on this discussion and just put them next to the code which documents such things as well.

Don't get me wrong but looks like there is the need for a discussion. Otherwise, I would say, "Let's not spent too much time on this discussion and just add it to the existing dev docs" 🤷

qwerty287:
Lgtm from my side as well here. I'd also keep it in the docs.

PR #5810, opened 2025-12-01
Makefile: add target to generate man pages

6543:
Add code go generate man pages

image

### DIFF
--- a/cmd/agent/man.go
+++ b/cmd/agent/man.go
@@ -40,5 +40,5 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	fmt.Printf(md)
+	fmt.Print(md)
 }
--- a/cmd/server/man.go
+++ b/cmd/server/man.go
@@ -31,5 +31,5 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	fmt.Printf(md)
+	fmt.Print(md)
 }
--- a/cmd/cli/man.go
+++ b/cmd/cli/man.go
@@ -28,5 +28,5 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	fmt.Printf(md)
+	fmt.Print(md)
 }


cmd/server/man.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2018 Drone.IO Inc.
```
6543:
```suggestion
// Copyright 2025 Woodpecker Authors
```

cmd/agent/man.go
```diff
@@ -0,0 +1,44 @@
...
+	local.New(),
+}
+
+func main() {
+	app := core.GenApp(backends)
+	md, err := docs.ToMan(app)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Printf(md)
```
qwerty287: Why `Printf` and not `Print`?

6543: well yes you are right ... my fingers just added the f because of doing so too often

qwerty287:
Just tried it, and it does not include the env vars in the man page. I know that this is how man pages look like, but this way we kind of change the way how you would configure your instance.

6543:
hmm good question how to include the env vars...

6543:
@qwerty287 created https://github.com/urfave/cli-docs/issues/21

anbraten:
How do you plan to share man pages / who is requesting them. Currently the docs are our main point of source for configuration related details, now we would have to provide all of those information as part of the cli args / flags as well.

6543:
I'll package the man page for nixos ... so as woodpecker-ci project we dont have to do anything and we should stick to our website :)

anbraten:
Thanks. (Knowing the usecase first hand helps other validating them or suggesting alternatives. Especially as we have to prevent running into a feature creep at some places)

6543:
this patch here i could also directly maintain at nixos but it's better to have the codehooks to generate docs e.g. manpages upstream as to conflict less

(re revactor to move the app declaration out of the main func was the important part)

PR #5815, opened 2025-12-02
fix: Reintroduce stop all pods on workflow

MartinSchmidt:
As part of the headless services #5764 I removed a few to many lines when destroying the workflow, so all running pods is no longer stopped when halting the workflow, I reintroduce the required lines in the PR.


MartinSchmidt:
@xoxys  and @qwerty287 I introduced a bug in the k8s backend i didn't notice before, was part of the last cleanup changes. Here is a fix for it, this must be included before 3.13, or #5764 should be rolled back, since it introduces a bug.

PR #5820, opened 2025-12-04
Allow untrusted repo to still drop network for steps

6543:
as it is an undocumented feature this might change / move into a more generalized backend network option at some point ...

but i think we could at least support it as proposed in this patch meanwhile

### DIFF
--- a/pipeline/frontend/yaml/linter/linter_test.go
+++ b/pipeline/frontend/yaml/linter/linter_test.go
@@ -87,16 +87,6 @@ steps:
     <<: *base-step
     image: golang:latest
 `,
-	}, {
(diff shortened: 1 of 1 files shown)


6543:
`.woodpecker.yaml`:
```yaml
skip_clone: true

steps:
  ping ok:
    image: bash
    commands: ping -c 2 193.99.144.80

  ping none:
    image: bash
    commands: ping -c 2 193.99.144.80
    network_mode: none
```

image

6543:
i had to remove the testcase as it would fail even if it is right, because we would test against config not allowed in jsonschema

Issue #5140, opened 2025-04-30
Cannot delete repository if corresponding VCS repository has been deleted.


flyingfishflash:
### Component

server

### Describe the bug

v3.5.2

I am trying to delete a repository for which the corresponding VCS repo has been deleted.

When I go to repositories -> {repository} -> actions -> delete repository

The server responds with a http 500 error.

it logs this to the console:

`
Apr 30 08:37:51 nas woodpecker-ci-app[5742]: 12:37PM ERR error="Error #01: The target couldn't be found.\n" ip=xxx.xxx.xxx.xxx latency=64.931237 method=DELETE path=/api/repos/3 status=500 user-agent="Mozilla/5.0 (X11; Linux x86_64; rv:138.0) Gecko/20100101 Firefox/138.0"`

### Steps to reproduce

delete a repository in your vcs that also has a corresponding repo record in woodpecker, then try and delete the woodpecker repo record.

### Expected behavior

When deleting a repository, if the corresponding VCS repo is not found, this should be handled effectively.

woodpecker repo records should be deleted even if VCS repo is not present.

### System Info

```shell
v3.5.2
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

ZUOLI11111111:
Please Assign Me.

tuxmainy:
Confirmed. I deleted a github fork and then tried to remove the repository in woodpecker:
```json
{"level":"error","error":"Error #01: GET https://api.github.com/repos/tuxmainy/ti-monitoring/hooks: 404 Not Found []\n","ip":"I USE IPv6","latency":182.020017,"method":"DELETE","path":"/api/repos/8","status":500,"time":"2025-11-02T12:29:24Z","user-agent":"Mozilla/5.0 (X11; Linux x86_64; rv:144.0) Gecko/20100101 Firefox/144.0","time":"2025-11-02T12:29:24Z"}
```

[...]

PR #5822, opened 2025-12-04
ignore hook removal when repo is already deleted on github

tuxmainy:
Hi,
this fixes an "internal error" when trying to delete a repository in woodpecker which is already deleted on github.

Other forges might also be affected but I think it has to be fixed in every forge implementation itself because every forge will response differently.

see issue #5140 

regards
Daniel

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,7 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/yaronf/httpsign v0.4.1
 	github.com/zalando/go-keyring v0.2.6
-	gitlab.com/gitlab-org/api/client-go v1.5.0
+	gitlab.com/gitlab-org/api/client-go v1.7.0
 	go.uber.org/multierr v1.11.0
 	golang.org/x/crypto v0.45.0
 	golang.org/x/net v0.47.0
--- a/go.sum
+++ b/go.sum
@@ -599,8 +599,8 @@ github.com/zalando/go-keyring v0.2.6 h1:r7Yc3+H+Ux0+M72zacZoItR3UDxeWfKTcabvkI8u
 github.com/zalando/go-keyring v0.2.6/go.mod h1:2TCrxYrbUNYfNS/Kgy/LSrkSQzZ5UPVH85RwfczwvcI=
 github.com/zenazn/goji v0.9.0/go.mod h1:7S9M489iMyHBNxwZnk9/EHS098H4/F6TATF2mIxtB1Q=
 github.com/ziutek/mymysql v1.5.4/go.mod h1:LMSpPZ6DbqWFxNCHW77HeMg9I646SAhApZ/wKdgO/C0=
-gitlab.com/gitlab-org/api/client-go v1.5.0 h1:kEHoqgNywPXV+I0oORiaOqy4hnWuqbUyPnl0tAF7anM=
-gitlab.com/gitlab-org/api/client-go v1.5.0/go.mod h1:RQfw64j1FE+KMZUAKsi1ZOOvwbWxHn9SkyZg+IAvjk4=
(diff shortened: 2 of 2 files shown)


qwerty287:
Thanks!

For consistency reasons, it would be nice to have this for all forges and not just github. Can you try to check it out as fas as possible? If not, we can also take a look of course. But I'd rather not merge this with only fixing the issue for github.

tuxmainy:
> For consistency reasons, it would be nice to have this for all forges and not just github. Can you try to check it out as fas as possible? If not, we can also take a look of course. But I'd rather not merge this with only fixing the issue for github.

Yeah, that would be nice. But I wonder how to do this? Every forge might have another response if the repo is already deleted. Also the error interface only has a String, thus you cannot detect specific error for all repos.

Any thoughts?

tuxmainy:
Maybe just listing all repos and check for the one you want to delete prior actual deleting? But this might give problems if somehow just the visibility is lost.

Edit: Like this?
```diff
diff --git a/server/api/repo.go b/server/api/repo.go
index be109a043..a810fbc34 100644
--- a/server/api/repo.go
+++ b/server/api/repo.go
@@ -34,6 +34,7 @@ import (
        "go.woodpecker-ci.org/woodpecker/v3/server/store"
        "go.woodpecker-ci.org/woodpecker/v3/server/store/types"
        "go.woodpecker-ci.org/woodpecker/v3/shared/token"
+       "go.woodpecker-ci.org/woodpecker/v3/shared/utils"
 )
 
 // PostRepo
@@ -457,10 +458,29 @@ func DeleteRepo(c *gin.Context) {
                return
        }
 
-       if err := _forge.Deactivate(c, user, repo, server.Config.Server.WebhookHost); err != nil {
+       // request all repos from forge
+       _repos, err := utils.Paginate(func(page int) ([]*model.Repo, error) {
+               return _forge.Repos(c, user, &model.ListOptions{
+                       Page:    page,
+                       PerPage: perPage,
+               })
+       }, maxPage)
+       if err != nil {
                _ = c.AbortWithError(http.StatusInternalServerError, err)
                return
        }
+       // lookup repo which we want to delete
+       // when found, remove hooks
+       for _, _repo := range _repos {
+               // is this suitable enough for a compare?
+               if _repo.ForgeRemoteID == repo.ForgeRemoteID {
+                       if err := _forge.Deactivate(c, user, repo, server.Config.Server.WebhookHost); err != nil {
+                               _ = c.AbortWithError(http.StatusInternalServerError, err)
+                               return
+                       }
+                       break
+               }
+       }
 
        if remove {
                if err := _store.DeleteRepo(repo); err != nil {
```

qwerty287:
No, I'd go with the approach you used already, for github: error handling inside the forge. But this needs to be added to the other forges as well. As you said a general solution won't work.

tuxmainy:
> No, I'd go with the approach you used already, for github: error handling inside the forge. But this needs to be added to the other forges as well. As you said a general solution won't work.

I have added the same logic to gitea. But I fear I am done here. I don't have the possibility to test other forges as well and I don't like to commit code which I haven't tested before :(

lafriks:
Would be better to use short variable names, like `errr, ok := ...` ( ok is common use case to check cast result etc)

lafriks:
Would be better to use short variable names, like `errr, ok := ...` ( ok is common use case to check cast result etc)

tuxmainy:
> Would be better to use short variable names, like `errr, ok := ...` ( ok is common use case to check cast result etc)

Thanks,  also `make test` should work now. Please keep an eye on my changes as I am not familiar with golang.

qwerty287:
Thanks, the gitea code looks fine for me as well. It's not optimal with the type but the only idea I have is a simple bool var `had404` which is set by the pagination func to `true` once a 404 occured. Then we can skip the error check at all after the pagination.

And there are more forges: Forgejo (should be basically the same as gitea), GitLab, Bitbucket and Bitbucket DC.

PR #5834, opened 2025-12-06
Implement SourceHut forge backend

ddevault:
This is an initial proof-of-concept. Will add remarks inline to highlight discussion points.


server/forge/sourcehut/helper.go
```diff
@@ -0,0 +1,80 @@
...
+	commit, ok := update.New.Value.(*git.Commit)
+	if !ok {
+		return nil
+	}
+
+	forgeUrl := fmt.Sprintf("%s/%s/%s/commit/%s",
+		c.gitURL, from.Repository.Owner.CanonicalName, from.Repository.Name,
+		update.New.Id)
+
+	return &model.Pipeline{
```
ddevault, line 68: I could add changed files here, but would need to parse the unified diff to do so. Would it be desirable to add a dependency for that?

server/forge/sourcehut/sourcehut.go
```diff
@@ -0,0 +1,510 @@
...
+	}
+	if user.Repository == nil {
+		return nil, nil
+	}
+	return c.toRepo(user.Repository, ver), nil
+}
+
+func (c *SourceHut) Repos(ctx context.Context, u *model.User, p *model.ListOptions) ([]*model.Repo, error) {
+	// TODO: Pagination on SourceHut does not work well with Woodpecker's
+	// higher-level internals (and doesn't work well at all, to be frank)
```
ddevault, line 255: SourceHut's API pagination leaves something to be desired, but one quality that needs to be considered is that it uses opaque strings for cursors, rather than integer page numbers, which requires higher level refactoring in Woodpecker.

server/forge/sourcehut/sourcehut.go
```diff
@@ -0,0 +1,510 @@
...
+		if err != nil {
+			resp.Body.Close()
+			return nil, err
+		}
+		resp.Body.Close()
+
+		entries = append(entries, &forge_types.FileMeta{
+			Name: ent.Name,
+			Data: data,
+		})
```
ddevault, line 361: This interface design requires fetching the full contents of every file in a directory, including ones that may or may not be applicable to the workflow. Would be nice to separate these steps in the interface.

server/forge/sourcehut/sourcehut.go
```diff
@@ -0,0 +1,510 @@
...
+	owner, name, _ := strings.Cut(string(r.ForgeRemoteID), "/")
+	user, _, err := git.GetRepo(client, ctx, owner[1:], name)
+	repo := user.Repository
+
+	_, err = git.RegisterPushWebhook(client, ctx, repo.Id, gitWebhookPayload, link)
+	if err != nil {
+		return err
+	}
+
+	// XXX: Ideally we would store the webhook ID here
```
ddevault, line 392: It would be nice if forge backends could store arbitrary metadata in the database for, for instance, the repo table. This would allow us to store the webhook ID and simplify the webhook deletion process later.

server/forge/sourcehut/sourcehut.go
```diff
@@ -0,0 +1,510 @@
...
+
+	target := user.Repository.Reference.Follow
+	return &model.Commit{
+		SHA:      target.Id,
+		ForgeURL: fmt.Sprintf("%s/%s/%s/commit/%s", c.gitURL, owner, name, target.Id),
+	}, nil
+}
+
+func (c *SourceHut) PullRequests(ctx context.Context, u *model.User, r *model.Repo, p *model.ListOptions) ([]*model.PullRequest, error) {
+	return nil, nil // TODO
```
ddevault, line 456: SourceHut does not have "pull requests". Instead, we use patches sent to a mailing list. By associating patches with a git repository, we can use features like status indicators, but right now there is no means of getting this information into the forge backend because it would require a separate UI for configuring which mailing list applies to a git repository.

This leads into another question, which is that the woodpecker frontend makes a lot of assumptions that the forge will use GitHub-style workflows like pull requests. The list of applicable events, and terminology like "pull request", would ideally reflect the capabilities/design of the underlying forge.

server/forge/sourcehut/sourcehut.go
```diff
@@ -0,0 +1,510 @@
...
+	account, err := meta.FetchUser(client, ctx, org[1:])
+	if err != nil {
+		return nil, err
+	}
+	return &model.Org{
+		Name:    account.CanonicalName,
+		IsUser:  true,
+		Private: false,
+	}, nil
+}
```
ddevault, line 511: SourceHut does not (currently) support orgs. I did my best here to shim in the necessary code to account for that, but some extra attention here would be great.

6543:
uuuuh :)

ddevault:
> uuuuh :)

Yes?

6543:
= nice will look into it once i have time :)
& happy to hear from sourcehut, had nice chat with sh guys at fosdem :)

ddevault:
Oh, great :) I remember meeting you at FOSDEM, glad to be working together on this now.

server/forge/sourcehut/sourcehut.go
qwerty287: While this is right, changing this would probably add much complexity to the codebase. Having separate event handling for every forge would no longer allow us to let the UI define these labels, icons, etc. Also quite relevant features like the pull request gating is getting harder. Sure, it's possible, but I'm not sure if it's the right way to go (but I also don't know how complex that would be exactly).

About your first paragraph, can you elaborate some more here? Cause I don't know if I get that correctly. I never used sourcehut. Would it be possible to use the patches just (or mostly) like pull requests? Yes, there probably wouldn't be events like "pr closed" or "pr metadata updated". So the question mainly is if sourcehut can send a webhook linked to a repo if a patch comes in and if this patch can be cloned directly. If yes, this should work I think, if not, that would be hard to implement at woodpeckers side. And what do you mean by "separate UI"? A separate UI in woodpecker?

server/forge/sourcehut/sourcehut.go
qwerty287: Looks fine so far. Bitbucket also doesn't I think.

server/forge/sourcehut/helper.go
qwerty287: How much would that change the binary size? If it's not much we can add it I think.

qwerty287 (commented):
Thanks! Added some comments to some of your points

server/forge/sourcehut/sourcehut.go
ddevault: We can send a webhook indeed when a patch comes in, but it's not necessarily *linked* to a repository. A single mailing list can have several associated repositories within it, or vice-versa. We would need a separate UI to associate a repository (the resource known to woodpecker) with a specific mailing list to process patches from, which would basically involve a UI very similar to the one for picking a repository from a list of repositories present on the forge, only you would be picking out a mailing list instead.

Applying the patch for testing is not as simple as cloning it directly and checking out a magic reference like GitHub does, though. You need a separate step which basically amounts to `curl <patch url> | git am -3`.

As for the UI being able to accommodate different events and such for different forges, I agree that it will add complexity, though in this case we could probably just treat "patch" as a synonym for "pull request". But I think that it's a bit shortsighted to avoid having forge-specific UI tweaks on the basis that it adds complexity to the codebase; on the other hand it adds constraints to how well woodpecker can fit into each of the forges it supports. Currently all of the forges supported are GitHub-like, but they do still have differences which would probably be accounted for in Woodpecker, and it's pretty limiting to suppose that Woodpecker will only be able to support GitHub-like forges.

server/forge/sourcehut/helper.go
ddevault: Not much, probably. I think this one would do the trick:

https://godocs.io/github.com/bluekeyes/go-gitdiff/gitdiff

ddevault:
I think the CI failures are not entirely related to my work? Is that the case?

qwerty287:
They are related, the CI doesn't run `go generate` afaik as we commit all the generated files as well.

6543:
Thanks qwerty for step in, i currently did not have mouch time ... and the holydays are packed with dates ... 😓

server/forge/sourcehut/sourcehut.go
6543: in this case we just paginate in the forge adapter itself and return all ...

server/forge/sourcehut/sourcehut.go
6543: yes:
https://github.com/woodpecker-ci/woodpecker/blob/7cef7a8a15c697487d57d087ae80485659eb18db/server/forge/bitbucket/bitbucket.go#L498-L501

6543:
@ddevault I updated https://github.com/woodpecker-ci/woodpecker/blob/main/server/forge/forge.go so to know what the interface expects ... :)

also i addressed some points so you can focus on the core forge adapter ...

_through docs still is a shim and needs to be done_

Issue #368, opened 2021-09-27
Bot / api token as environment variable


anbraten:
### Clear and concise description of the problem

As a developer I want to create plugins which could interact with the woodpecker-api via curl or the official client. For this I would need some token.

### Suggested solution

The needed token to access the woodpecker-ui could automatically be injected into each woodpecker step, so plugins could directly use it. It would be best if the bot-token is only valid for some short duration or the time the pipeline run is being executed and has some limited permissions.

### Additional context

Would be useful for #367 

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.github.io/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

6543:
I would suggest to create temporary auth token for each job with limited scope of access ... so not usig #665 since this has a other usecase ... but we can share code of cource.

Idea: extemd proc structure witb a token field

anbraten:
I would extract the (api)-token from the user and add it to a separate model / table to have one single place to store tokens. If a pipeline gets started we could simply add a new token and remove it again after the pipeline finished.

PR #5835, opened 2025-12-06
Revert "Send configuration as part of the request for external configuration"

anbraten:
Reverts woodpecker-ci/woodpecker#5831

This was intentionally replaced by netrc credentials at some point as those credentials allow to load any kind of files from the forge. Allowing more powerful config extensions than just ".yml converters". We also removed this to skip internal config loading at some point for repos with a config extensions to support repos to not have a config at all (was requested for forks for example) and to avoid the extra forge api call. Also see https://github.com/woodpecker-ci/woodpecker/issues/4341#issuecomment-2688129248

Unfortunately the migration seems to got lost in https://github.com/woodpecker-ci/woodpecker/pull/4682

### DIFF
--- a/docs/versioned_docs/version-3.13/20-usage/72-extensions/40-configuration-extension.md
+++ b/docs/versioned_docs/version-3.13/20-usage/72-extensions/40-configuration-extension.md
@@ -42,10 +42,6 @@ class Request {
   repo: Repo;
   pipeline: Pipeline;
   netrc: Netrc;
-  configuration: {
-    name: string; // filename of the configuration file
-    data: string; // content of the configuration file
-  }[];
 }
 ```
 
@@ -56,12 +52,15 @@ Checkout the following models for more information:
 - [netrc model](https://github.com/woodpecker-ci/woodpecker/blob/main/server/model/netrc.go)
 
 :::tip
-The `netrc` data is pretty powerful as it contains credentials to access the repository. You can use this to clone the repository or even use the forge (Github or Gitlab, ...) API to get more information about the repository.
+The `netrc` data is pretty powerful as it contains credentials to access the repository. You can use this to fetch files or other information (like changed files, issues) from the repository using the forge api or even clone the repository.
 :::
 
 Example request:
 
 ```json
+// Please check the latest structure in the models mentioned above.
+// This example is likely outdated.
+
 {
   "repo": {
     "id": 100,
@@ -123,12 +122,12 @@ Example request:
     "updated_at": 0,
     "verified": false
   },
-  "configs": [
-    {
-      "name": ".woodpecker.yaml",
-      "data": "steps:\n  - name: backend\n    image: alpine\n    commands:\n      - echo \"Hello there from Repo (.woodpecker.yaml)\"\n"
-    }
-  ]
+  "netrc": {
+    "machine": "myforge.com",
+    "login": "myUser",
+    "password": "myPassword",
+    "type": "forge"
+  }
 }
 ```
 


qwerty287:
Don't have time right now to write a detailed comment, please give me one or two days...

qwerty287:
I don't know if this is the way to go. Maybe we could add a switch to allow both possibilities - include configs or not. But this would add more complexity to woodpecker again.

You're right, in the recent discussion I forgot about the more powerful usecase without yaml files. However, if you just want a simple yaml converter, it's way easier if you don't have to parse the netrc.

I thought about some more usecases that would require sending the config - e.g. chaining the config extensions - but I don't think we'll ever allow that, so that argument doesn't really count.

I don't really know what's the better way, requiring the users to include that complexity or doing that directly.

Anyways, however: This change is actually breaking. #5831 fixed how this *should* work according to the docs.

techknowlogick:
IIRC this is why Drone has conversion AND configuration extensions, one can transform the config (incl if it's starlark, jsonnet, etc..), and the other doesn't receive any content and responds with a complete yaml. If you didn't want to make this "breaking" you could consider keeping the functionality, but renaming the extension to "conversion" and having the secondary one without the config being sent be "configuration"

anbraten:
> Anyways, however: This change is actually breaking. https://github.com/woodpecker-ci/woodpecker/pull/5831 fixed how this should work according to the docs.

As it was removed with 3.0.0 and was not working that way it's not breaking (you could not have updated if you expected it to work that way). We even had it in the docs until it got removed and it's in the changelog as well.

qwerty287:
We can merge it if you don't agree to me, I won't block it, but I also think it should be possible to write simple converters without having to re-fetch the config.

anbraten:
Would merge it for now and consider the mentioned in a follow up design of the feature.

6543:
we have to also backport to the docs now...

6543:
PS: i personally see no harm to have the config send next to netrc as the external config service can just ignor ethe old config if it wants to. but it provides a huge benefit for those who want to write an simple and easy config mutation service without having to worry about the forge at all :)

6543:
I think the real issue is: https://github.com/woodpecker-ci/woodpecker/issues/4341#issuecomment-2469875849

>  I've tested that quickly and Woodpecker is trying to parse the file(s) before sending them further.

I did not look into the code jet but I would say if an external config extension is configured we should ignore parse issue and just move the content to the service .... and in case where we want to config something completly different the external service still can fetch via netrc

6543:
> Anyways, however: This change is actually breaking. https://github.com/woodpecker-ci/woodpecker/pull/5831 fixed how this should work according to the docs.

well thats a clear yes and no ... as the info never existed and as such only users of the `next` version could see it aktually breaking but if we see the docs as hard source of truth it is breaking

6543 (approved):
lgtm to not hold the release back but we should discusst this more

lafriks:
Imho woodpecker server should provide API with pipeline run specific token auth that would allow to get files from forge etc to abstract away from need to implement forge specific API calls also in config server

6543:
well agree and on the TODO since 2022 (https://github.com/woodpecker-ci/woodpecker/issues/368#issuecomment-1006805144) :laughing:

sloonz:
> PS: i personally see no harm to have the config send next to netrc as the external config service can just ignor ethe old config if it wants to. but it provides a huge benefit for those who want to write an simple and easy config mutation service without having to worry about the forge at all :)

This.

Sorry, if I’m a bit salty, but I’m getting a bit tired of this. Woodpecker 3 broke this without any prior announcement (nor any real justification ?), then there was a discussion about reintroducing it, it was reintroduced in the docs but not in the code (which should mean that it’s fine in principle, shouldn’t it ?), then I made the PR that reintroduced it in the code, it got integrated, Woodpecker 3.13.0 got released, changelog says "that feature is here", I start working on a Woodpecker v3-compatible version of https://github.com/sloonz/woodpecker-signed-env, and now it’s gone again ?

I don’t see why you insist so much on removing that feature. AFAICT (tell me if I’m wrong ?) it doesn’t break anything, it doesn’t introduce any significant amount of complexity in the code, it's how the doc says (well, said, at that point) it should work, and let’s be straight, why the hell would you _not_ do it that way, given that if you don’t provide the old config (that you fetch anyway), every single plugin that just want to do transformations has to re-implement itself the whole "fetch the configuration file from the forge" logic, for _every_ forge.

Which is a significant work that *woodpecker has already done anyway* ? I’m not going to take that massive undertaking in my simple transformation plugin, and I doubt anyone is going to. I’d rather stay on Woodpecker v2. And if for any reason "staying on Woodpecker v2" is no longer an option ? Maintaining a fork of woodpecker re-enabling this feature is orders-of-magnitude simpler than "just use netrc to fetch the configuration yourself".

>> I've tested that quickly and Woodpecker is trying to parse the file(s) before sending them further.
>
> I did not look into the code jet but I would say if an external config extension is configured we should ignore parse issue and just move the content to the service .... and in case where we want to config something completly different the external service still can fetch via netrc

Incorrect, at least at the time of 9294492977a4ae453f36d05c6e25f6c640957779 (or maybe the "ignore parsing errors if configuration plugin is configured" is already here ? I did not look at the code). My configuration plugin can receive a .jsonnet without issue if I point pipeline path to .woodpecker/config.jsonnet. A quick-and-dirty test shows me that sending the resulting yaml works fine, the whole pipeline runs without any issue.

Woodpecker *could* have a woodpecker-jsonnet plugin today, if not for that weird fixation on "configuration plugins cannot receive the configuration fetched by woodpecker". It was literally on my todo-list, just after "make woodpecker-signed-env compatible with woodpecker v3".

qwerty287:
I totally understand you're not amused, and I also have to say sorry for my misunderstanding of the original issue. I thought it's a bug in code, while it actually was a bug in the documentation. The overall communication was really bad, to you as well as internally.

Actually, there is a usecase for not sending the data, which is completely skipping the forge and let everything be handled by the extensions. Then you don't need config files in the repo itself at all. This is not possible right now but we can extend the feature in future releases.

As multiple of us maintainers wrote, it should however still be possible to write simple conversion plugins that can get the configuration.

To fix this issue and solve the core problem, I'll work on an enhancement to support both asap, I'll likely be able to do it tomorrow. Then you can choose for now whether you want to enable sending the configurations, and if they're not sent, we can extend the feature later to support extensions completely without the forge.

sloonz:
No worries. I did not have this use case in mind.

That being said, I don’t understand why implementing this forgeless feature *in the future* requires to stop sending (or gating behind a flag) the configuration from the forge *now* ?

Obviously, if forgeless feature is enabled, then the configuration sent to the configuration service is empty. But that’s just one line added to the documentation, no need for any additional configuration knob ? And it's not even a breaking change, since all existing setups are not forgeless.

Or am I missing something ?

qwerty287:
Depending on the defaults changing this later could be breaking (but I also think it's possible to write it in a non-breaking way...). But I also don't know what exactly @anbraten was imagining

qwerty287:
See https://github.com/woodpecker-ci/woodpecker/pull/5978 for the new feature.

6543:
@sloonz also sorry for my side that it had annoyed you. I decisions still stands because those who do not use `next` can not see anything breaking of what was working in before.

And about your usecase yes we should support that already through we had planed the release and this issue here had hold it back already for days. so that's why it was decided to fix it in the next release cycle

6543:
@sloonz https://github.com/woodpecker-ci/woodpecker/pull/5978 got merged and should be available as next soon

PR #5852, opened 2025-12-11
Add Container Registry credential extension

techknowlogick:
The configuration extension already exists, but this one would allow for dynamic addition of container registry credentials being injected into runs. That way you don't need to set static credentials in the database.

I might work on the secrets or environment variables extensions support next.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
     hooks:
       - id: golangci-lint
   - repo: https://github.com/igorshubovych/markdownlint-cli
-    rev: v0.46.0
+    rev: v0.47.0
     hooks:
       - id: markdownlint
         exclude: '^(docs/versioned_docs/.*|CHANGELOG.md)$'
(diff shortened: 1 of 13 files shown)


lafriks:
Why does the base registry take priority? Imho extension should be priority as they could be more specific

techknowlogick:
> Why does the base registry take priority? Imho extension should be priority as they could be more specific

fair, I've just updated it so extensions take priority (with repo taking precedence over global)

qwerty287:
Thanks! We also don't need the migration itself, model sync happens automatically.
And it seems there is no UI added here?

techknowlogick:
@qwerty287 sounds good. A habit of mine of struct changes in Gitea where any changes get a migration ;)
I'll drop that migration.

As for UI I'm saving that for another PR since I wanted to validate the actual functionality first, and I'm more familiar with the backend code than the UI (so any changes needed I could implement much faster). I figured that for mvp it was fine to go with this route.
If not having UI is a blocker, lmk and I can add it in for this PR:)

qwerty287:
@techknowlogick can you fix the linter? https://ci.woodpecker-ci.org/repos/3780/pipeline/30254/35

LGTM otherwise.

techknowlogick:
thanks @qwerty287, I've just sent lint fix :)

xoxys:
Nice one! But can we add docs? 😬

qwerty287:
Thanks for the work @techknowlogick again! I added UI and docs in #5993.

techknowlogick:
@qwerty287 thanks to you too :)

PR #5854, opened 2025-12-12
feat(kubernetes): add support for pod affinity and anti-affinity configurations

MartinSchmidt:
Introduce support for configuring pod affinity and anti-affinity in Kubernetes. This allows users to specify rules for pod placement based on node labels and relationships with other pods, enhancing scheduling flexibility. 

The implementation includes new command-line flags and updates to the backend options structure to accommodate these configurations.

This allows one to easily target all pods for a specific workflow to run on same node: 

```yaml
      WOODPECKER_BACKEND_K8S_POD_TOLERATIONS: |
        - key: woodpecker-pool
          operator: Exists
          effect: NoSchedule
      WOODPECKER_BACKEND_K8S_POD_NODE_SELECTOR: |
        woodpecker-pool: true
      WOODPECKER_BACKEND_K8S_POD_AFFINITY: |
        podAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector: {}
            matchLabelKeys:
              - woodpecker-ci.org/task-uuid
            topologyKey: "kubernetes.io/hostname"
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector: {}
            mismatchLabelKeys:
            - woodpecker-ci.org/task-uuid
            topologyKey: "kubernetes.io/hostname"
```

Here all pods for one workflow are will be run on a single node, and no other workflow can run on the same node.
This together with karpenter allows on depand nodes to be created and deprovisioned with a pool that has the woodpecker-pool taint.

### DIFF
--- a/pipeline/backend/kubernetes/pod.go
+++ b/pipeline/backend/kubernetes/pod.go
@@ -474,9 +474,13 @@ func toleration(backendToleration Toleration) v1.Toleration {
 }
 
 func affinity(stepAffinity, agentAffinity *v1.Affinity, allowFromStep bool) *v1.Affinity {
-	if allowFromStep && stepAffinity != nil {
-		log.Trace().Msg("using affinity from step backend options")
-		return stepAffinity
+	if stepAffinity != nil {
+		if allowFromStep {
+			log.Trace().Msg("using affinity from step backend options")
+			return stepAffinity
+		} else {
+			log.Debug().Msg("Step affinity is disallowed by instance configuration, ignoring it")
+		}
 	}
 
 	if agentAffinity != nil {
@@ -484,6 +488,7 @@ func affinity(stepAffinity, agentAffinity *v1.Affinity, allowFromStep bool) *v1.
 		return agentAffinity
 	}
 
+	log.Trace().Msg("no affinity configured")
 	return nil
 }
 


MartinSchmidt:
Any change this can make it into 3.13? 
Since we currently are running on my fork, but would like to go back running on upstream.

xoxys:
@henkka as you also actively use Kubernetes would you mind reviewing as well? :)

henkka (approved):
Looks good to me, didn't spot any modifications that would be breaking in any way, nice feature! 👍

pipeline/backend/kubernetes/pod.go
```diff
@@ -472,6 +473,20 @@ func toleration(backendToleration Toleration) v1.Toleration {
 	}
 }
 
+func affinity(stepAffinity, agentAffinity *v1.Affinity, allowFromStep bool) *v1.Affinity {
```
xoxys: We should write at least to debug log in case stepAffinity is set but disallowed similar to https://github.com/MartinSchmidt/woodpecker/blob/23a05686027c50e008a00ea51e250210c89fc44e/pipeline/backend/kubernetes/pod.go#L124-L126

xoxys (commented):
Thanks! LGTM so far, just a nit.

pipeline/backend/kubernetes/pod.go
MartinSchmidt: Good idea, but I changed it so it raises an error, "silent" failures can be a pain to find, and if it users do something that isn't allowed it believe it should fail, not just omit it. This could have adverse effects.
Also updated tests, and added an additional one.

xoxys: In general, I understand your point, but I'm not sure about this behavior. I would prefer to keep the behavior the same as for existing options. We can change this behavior for all "*_ALLOW_FROM_STEP" options in a dedicated PR but this might need discussions and would be breaking for existing options.

MartinSchmidt: Changed it to just the debug comment as your suggestion, then I will create a separate PR for the other thing afterwards.

xoxys: Thanks!

6543:
thanks for adding tests by the way :)

PR #5857, opened 2025-12-13
Implement retry logic in HTTP Send method

myselfghost:
Added retry logic with exponential backoff to the Send method for HTTP requests, improving error handling and resilience against transient failures.

### DIFF
--- a/pipeline/frontend/yaml/types/base/slice_test.go
+++ b/pipeline/frontend/yaml/types/base/slice_test.go
@@ -36,7 +36,7 @@ func TestStringOrSliceYaml(t *testing.T) {
 		assert.Equal(t, StringOrSlice{"bar", "baz"}, s.Foo)
 
 		d, err := yaml.Marshal(&s)
-		assert.Nil(t, err)
+		assert.NoError(t, err)
 
 		s2 := StructStringOrSlice{}
 		assert.NoError(t, yaml.Unmarshal(d, &s2))


qwerty287:
Thanks, but could you explain why this is necessary?

myselfghost:
> Thanks, but could you explain why this is necessary?

In my actual usage, I found that every week some pipelines randomly fail to start due to HTTP request failures, possibly because the HTTP service is unstable. Therefore, retrying HTTP requests is  necessary.

qwerty287:
You're using a config extension?

myselfghost:
> You're using a config extension?

Yes, my entire CI pipeline relies on External Configuration API

qwerty287:
Ok, so for backoffs we use `github.com/cenkalti/backoff/v5`. Can you use it?

myselfghost:
> Ok, so for backoffs we use `github.com/cenkalti/backoff/v5`. Can you use it?

ok , done

qwerty287:
Would it be possible to let the lib handle the waiting etc, so we don't need a for loop etc.?

Similar to how this is done in https://github.com/woodpecker-ci/woodpecker/blob/main/cmd/server/server.go#L96-L104.

myselfghost:
> Would it be possible to let the lib handle the waiting etc, so we don't need a for loop etc.?
> 
> Similar to how this is done in https://github.com/woodpecker-ci/woodpecker/blob/main/cmd/server/server.go#L96-L104.

done

server/services/utils/http.go
```diff
@@ -89,49 +94,164 @@ func NewHTTPClient(privateKey crypto.PrivateKey, allowedHostList string) (*Clien
...
+// Send makes an http request with retry logic.
 func (e *Client) Send(ctx context.Context, method, path string, in, out any) (int, error) {
-	uri, err := url.Parse(path)
-	if err != nil {
-		return 0, err
-	}
+	// Maximum number of retries
+	const maxRetries = 3
+	// Initial backoff duration
+	const initialBackoff = 500 * time.Millisecond
```
qwerty287: Since this is the default value, let's just remove it and use the default of the lib.

server/services/utils/http.go
```diff
@@ -89,49 +94,164 @@ func NewHTTPClient(privateKey crypto.PrivateKey, allowedHostList string) (*Clien
...
+	// Maximum backoff interval
+	const maxBackoffInterval = 5 * time.Second
```
qwerty287: Also this should not be necessary with `maxTries`.

server/services/utils/http.go
myselfghost: done

server/services/utils/http.go
myselfghost: done

lafriks:
Does it need result type at all, can't it just return int?

myselfghost:
> Does it need result type at all, can't it just return int?

Dear maintainers, if there are any details that still need to be adjusted, please feel free to make the changes directly. @qwerty287 @lafriks

qwerty287:
@lafriks changed that, do you want to check it out again?

qwerty287:
@woodpecker-ci/maintainers as I modified quite some parts of this, does somebody else wants to review?

qwerty287:
Not sure how - we can of course send requests to non-existing servers, but then we somehow have to catch that and count the tries. If the server itself is reachable it won't retry… Do you have an idea?

anbraten:
You could start an http server in the test with a handler that returns 500 the first x times and then a config. Similar to https://github.com/woodpecker-ci/woodpecker/blob/98929ba3912ec8deb06bf345cc1f752e980cd031/server/services/config/combined_test.go#L187

qwerty287:
Ah, sorry, I thought any status code is not retried. Yes, this works.

qwerty287:
Test added in https://github.com/woodpecker-ci/woodpecker/pull/5857/changes/a9e49a14cea50b556e85eaa63cdd84114bdcf29e

PR #5872, opened 2025-12-16
Update Helm chart installation command

abhiyerra:
Updating helm install docs based on: https://github.com/woodpecker-ci/helm/blob/main/charts/woodpecker/README.md#installation


xoxys:
Thanks. IIRC, the `--version` flag is required. Please also change https://github.com/woodpecker-ci/woodpecker/blob/main/docs/docs/30-administration/05-installation/20-helm-chart.md

PR #5875, opened 2025-12-17
Change linter warning for clone plugin not trusted to error

simen64:
This changes the linter warning for when a clone plugin used isn't in `PLUGINS_TRUSTED_CLONE` to an error, causing the pipeline to stop.

This makes it easier to limit what clone plugins are allowed to clone at all giving more control when wanting to restrict clone plugins.


qwerty287:
Thanks, but I'm against this change.

In it's current implementation, you can use anything for cloning, but have to inject the credentials manually if it's not trusted. I don't see a good reason why this would be bad to support. You get the warning anyways.

simen64:
The reason i want this implemented is because i am making a plugin that verifies git signatures before running the pipeline with a goal of not having trust in the Forge itself. With how it is now anyone with access to the Forge can change the workflow file and swap out my clone plugin for the default one and circumvent the signature checks. The way i would prevent this would be by only setting my clone plugin as trusted, hence why i see this change is fit.

xoxys:
If someone has hijacked access to forge, it's over anyway, right? I don't know if CI systems should be able to cover such scenarios. Is there any other CICD system that handles this scenario? If so, we could check how it's done there to get some more ideas.

However, I'm also against changing the linter to fulfill this scenario. If you have to implement something right now, a configuration extension could work. This way can store your CI configuration elsewhere (e.g. a private S3/file server/whatever) instead of the repo to prevent any changes via the forge.

xoxys:
If someone has hijacked access to forge, it's over anyway, right? I don't know if CI systems should be able to cover such scenarios. Is there any other CICD system that handles this scenario? If so, we could check how it's done there to get some more ideas.

However, I'm also against changing the linter to fulfill this scenario. If you have to implement something right now, a configuration extension could work. This way can store your CI configuration elsewhere (e.g. a private S3/file server/whatever) instead of the repo to prevent any changes via the forge.

simen64:
> If someone has hijacked access to forge, it's over anyway, right?

Unfortunately yes this is how most CI systems are set up, however CI is a really good place to limit damage in the event of Forge compromise e.g, by not releasing or deploying compromised code. And deploying something like signature checks is a good way of doing this, but it relies on not being able to bypass it by writing to the workflow file that's in the Forge.

Maybe we could have a server configuration like `WOODPECKER_STRICT_CLONING`?

xoxys:
Understood. What about the suggestion to use https://woodpecker-ci.org/docs/usage/extensions/configuration-extension?

simen64:
Using the configuration extension can work, but it requires a separate trusted server containing the workflow files and then you need some logic to either compare hashes of the one woodpecker sends to see if its been tampered or just send the one on the server regardless. You would also need a good way to update your workflows. This is a lot more complex, requires more setup, and is difficult to scale.

xoxys:
You have to deal with one or another pain I guess and to put trust to one or another component. For me, using a config extension is the way to go for this use case.

simen64:
Having a way to only let one or more plugins clone means you only need trust in the woodpecker server instead of a whole nother server for the configuration extension. I understand that this may not be wanted as default behavior, but maybe as a server option?

qwerty287:
If, then as per-repo option, not global please (global default could be used).

simen64:
We could have a server config like `WOODPECKER_STRICT_CLONE_REPOS` that would be a list of repos strict cloning would apply?

qwerty287:
No, an option in the repo settings in the UI.

simen64:
Having this option in the UI would counteract trying to remove trust in the Forge, an attacker who has compromised the Forge could just log into the UI, disable strict cloning and then add back the default clone plugin bypassing something like signature checks.

qwerty287:
Well, if an attacker can log in to Woodpecker, they can also modify the trusted clone plugins (those can be configured per repo as well).

simen64:
Yes true, i should have said this earlier, but i also think removing that from the UI as a part of strict cloning would be necessary.

xoxys:
Sorry to say again, but the entire trust model of woodpecker is strictly coupled to the forge. If you really have a scenario with real trust issues to the forge you have to move the configs to a dedicated, isolated, audited and hardened system (by using a config extension).

zc-devs:
> Is there any other CICD system that handles this scenario?

[Drone](https://docs.drone.io/signature/)?

xoxys:
Indeed, some kind of config signing could cover this use case, and we might want to think about it. However, the current approach is IMO the wrong one. For this reason, I'm going to close this PR.

PR #5880, opened 2025-12-18
Add background to status icons

xoxys:
This PR is up for discussion.

In my opinion, the status icons are hard to read in dark mode due to the dark background and the colored icons. To fix the contrast issue, a white background was added to all status icons.

Before:

image
image

After:

image
image

### DIFF
--- a/web/src/components/atomic/SvgIcon.vue
+++ b/web/src/components/atomic/SvgIcon.vue
@@ -1,6 +1,6 @@
 <template>
   <svg fill="currentColor" :width="size" :height="size" viewBox="0 0 24 24">
-    <circle v-if="bgCircle" cx="12" cy="12" :r="9" class="fill-gray-300" />
+    <circle v-if="bgCircle" cx="12" cy="12" :r="9" class="fill-transparent dark:fill-gray-300" />
     <path :d="path" />
   </svg>
 </template>


anbraten:
I get your point. The white seems a bit to bright somehow. Maybe it would be an option to use a lighter red as that seems to have lower brightness compared to the status green.

xoxys:
I can try but increasing the brightness of the red doesnt worked out well in the past :D It looks either pinkish or is too bright in contrast to the bg etc.

xoxys:
Besides that. Do you want to have the white background removed again and try to fix the contrast of the red icons by adjusting the brightness? Because personally I think the white bg should stay either way.

lafriks:
I agree that this looks way much better and readable

lafriks:
Maybe it could be not white but very light gray then to not have too much contrast?

xoxys:
image

xoxys:
@woodpecker-ci/maintainers More opinions? Can we merge it?

xoxys:
Good catch, will fix it.

qwerty287:
It seems the pending icon now looks completely off...

xoxys:
Not that bad IMO howerver feel free to improve or revert the PR

lafriks:
I agree, all other icons look good but this one would be better with the dark middle part imho

Issue #5841, opened 2025-12-07
add simpler repo output format in cli


kouhaidev:
### Clear and concise description of the problem

customizing output format makes it easier to use woodpecker-cli in batch scripts. it would be useful to have a new format replacing the old `--format` argument, but on a per-element basis. this would reduce boilerplate

```
woodpecker-cli repo ls --output 'go-template={{range . }}{{printf "%d %s %s\n" .ID .Name .FullName}}{{end -}}' | grep woodpecker-ci/
1 woodpecker woodpecker-ci/woodpecker
2 woodpecker woodpecker-ci/helm
3 woodpecker woodpecker-ci/infrastructure
```

### Suggested solution

proposal: `go-format` output format will be equivalent to `go-template`, but with an implicit `{{ range . }}` and `{{ end }}` and newlines separating each template output.

```
woodpecker-cli repo ls --output 'go-format={{printf "%d %s %s" .ID .Name .FullName}}' | grep woodpecker-ci/woodpecker
1 woodpecker woodpecker-ci/woodpecker
2 woodpecker woodpecker-ci/helm
3 woodpecker woodpecker-ci/infrastructure
```

### Alternative

no change from #4888

### Validations

- [x] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/versions]
- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

kouhaidev:
this should probably get applied to all other `--output` flags as well.

xoxys:
I dislike this approach. Go template is explixitly for advanced usage and provides flexibility which gets removed with implicit formatting.

qwerty287:
I agree to @xoxys. Other opinions @woodpecker-ci/maintainers ?

xoxys:
Its already implemented as switch-case https://github.com/woodpecker-ci/woodpecker/blob/main/cli/repo/repo.go#L70 We can just add `go-format` as additional supported output format. That would be fine for me.

PR #5882, opened 2025-12-19
CLI: Allow single output template

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/5841

### DIFF
--- a/cli/repo/repo_test.go
+++ b/cli/repo/repo_test.go
@@ -40,6 +40,16 @@ func TestRepoOutput(t *testing.T) {
 			args:     []string{"output", "--output", "go-template={{range . }}{{.Name}} {{.ForgeURL}} {{.Trusted.Network}}{{end}}"},
 			expected: "repo1 git.example.com true\n",
 		},
+		{
+			name:     "go-format output",
+			args:     []string{"output", "--output", "go-format={{.Name}} {{.ForgeURL}} {{.Trusted.Network}}"},
+			expected: "repo1 git.example.com true\n",
+		},
+		{
+			name:    "invalid go-template",
+			args:    []string{"output", "--output", "go-template={{.InvalidField}}"},
+			wantErr: true,
+		},
 	}
 
 	repos := []*woodpecker.Repo{


anbraten:
As there is already a way to achieve this without any unusual workaround, the FR would be a no fix for me. Especially as we would need to add similar functions to all other commands for consistency again.

xoxys:
Can you explain why adding this as additional output format is bad? Yes this would need to be added to all other cli commands. However, we could move this to an output package to avoid code duplication but I dont see why we should not add it at all.

cli/repo/repo.go
```diff
@@ -80,6 +80,20 @@ func repoOutput(c *cli.Command, repos []*woodpecker.Repo, fd ...io.Writer) error
 		if err := tmpl.Execute(out, repos); err != nil {
 			return err
 		}
+	case "go-format":
```
6543, line 83: do we have user documentation somewhere and in the --help menue?

qwerty287: no, there's no docs for this right now at all

6543: well witch means we can change/improve it at any time

xoxys: What has this to do with existing docs? We are talking about an additional optional output format. Nothing changes for existing outputs, documented or not...

qwerty287:
@anbraten are you still against this change? Or can we continue and merge this?

anbraten:
I'm still not quite convinded that it's worth the code as you can already achieve the same formatting. And writing `--output 'go-template={{range . }}{{printf "%d %s %s\n" .ID .Name .FullName}}{{end -}}'` vs `--output 'go-format={{printf "%d %s %s" .ID .Name .FullName}}'` is not really that much of an issue IMO 🤔

qwerty287:
I also don't really think it's a big issue, but I can understand it… I mainly opened this PR to fix the issue, not because I need it. So from my side we can also close the issue as wontfix. But I also don't think adding this hurts.
@xoxys what do you think?

xoxys:
We maintain dead code (e.g. encryption) and many other rarely used code and this one is pretty trivial and easy to maintain, so add it. If you dont want to I can take over to finish it.

qwerty287:
What is to finish here? From my side this was ready to merge

xoxys:
Needs to be added to all commands that support custom outputs already.

qwerty287:
@xoxys added it for the pipelines as well.

xoxys:
Can you please also add the test case to repo_test?

qwerty287:
@anbraten As you're the only one who's against this, can we merge this?

Issue #4422, opened 2024-11-21
Infinite loading on manual run on branch without pipelines


zc-devs:
### Component

server, web-ui

### Describe the bug

Trying to run a pipeline manually on a branch, which doesn't have any pipelines, I get infinite loading and no errors.

### Steps to reproduce

1. Woodpecker + Gitea
2. Create empty repository
3. Enable it in Woodpecker
4. Go to repo, hit "Run pipeline"
5. Select whatever branch you want, which does not contain Woodepecker's configs (`main` in my case)
6. Hit "Run pipeline"
7. See infinite loading and no errors
![Screenshot 2024-11-21 141347](https://github.com/user-attachments/assets/3e3814d5-ae8c-4597-b510-a9398e04a763)
8. See errors in the server logs
```log

{"level":"trace","time":"2024-11-21T11:13:03Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/router/router.go:43","message":"[POST] /api/repos/2/pipelines"}
{"level":"debug","time":"2024-11-21T11:13:03Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/router/middleware/session/repo.go:157","message":"user1 granted &{UserID:1 RepoID:2 Repo:<nil> Pull:true Push:true Admin:true Synced:1732187329 Created:1721072630 Updated:1732187329} permission to wp/test"}
{"level":"trace","time":"2024-11-21T11:13:03Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/services/config/forge.go:98","message":"configFetcher[wp/test]: user did not define own config, following default procedure"}
{"level":"trace","time":"2024-11-21T11:13:03Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/services/config/forge.go:145","message":"fetching .woodpecker/ from forge"}
{"level":"trace","error":"configFetcher: fallback did not find config: configs not found: .woodpecker/, .woodpecker.yaml, .woodpecker.yml","time":"2024-11-21T11:13:03Z","caller":"/woodpecker/src/github.com/woodpecker-ci/woodpecker/server/services/config/forge.go:62","message":"Fetching config files: Attempt #1 failed"}
[...]

PR #5883, opened 2025-12-19
Show warning if there is no workflow to run

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/4422

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -49,7 +49,7 @@
         "value": "Variable value"
       },
       "show_pipelines": "Show pipelines",
-      "no_manual_workflows": "There were no workflows you can run manually, or your filters exclude all of them."
+      "no_manual_workflows": "There were no manually executable workflows, or the filter exclude all of them."
     },
     "deploy_pipeline": {
       "title": "Trigger a deployment for current pipeline #{pipelineId}",


web/src/views/repo/RepoManualPipeline.vue
```diff
@@ -95,12 +95,22 @@ async function triggerManualPipeline() {
 
   emit('close');
 
-  await router.push({
-    name: 'repo-pipeline',
-    params: {
-      pipelineId: pipeline.number,
-    },
-  });
+  if (typeof pipeline == 'string') {
```
anbraten, line 98: Could we check for the specific error here as there are several possible errors?

qwerty287: No, the server returns a 204 without any response (-> empty string).

web/src/assets/locales/en.json
```diff
@@ -48,7 +48,8 @@
         "name": "Variable name",
         "value": "Variable value"
       },
-      "show_pipelines": "Show pipelines"
+      "show_pipelines": "Show pipelines",
+      "no_workflows": "No workflows to run were found."
```
anbraten:
```suggestion
      "no_manual_workflows": "There were no workflows you can run manually, or your filters exclude all of them."
```

web/src/views/repo/RepoManualPipeline.vue
anbraten: I see. 4xx would throw an error and 204 has no content.

web/src/views/repo/RepoManualPipeline.vue
```diff
@@ -95,12 +95,22 @@ async function triggerManualPipeline() {
...
+    // if this is a string there is no workflow to run with the 'manual' event
```
anbraten:
```suggestion
    // if this is a string (http 204) there is no workflow to run with the 'manual' event
```

web/src/views/repo/RepoManualPipeline.vue
qwerty287: Exactly.

web/src/assets/locales/en.json
```diff
@@ -48,7 +48,8 @@
         "name": "Variable name",
         "value": "Variable value"
       },
-      "show_pipelines": "Show pipelines"
+      "show_pipelines": "Show pipelines",
+      "no_manual_workflows": "There were no workflows you can run manually, or your filters exclude all of them."
```
xoxys:
```suggestion
      "no_manual_workflows": "There were no manually executable workflows, or the filter exclude all of them."
```

Wording nit.

xoxys (approved):

6543:
a screenshot would have been nice but i'll see it in my next build probably now ;)

thanks tor makin UX better!

Issue #4361, opened 2024-11-11
Disabling a repo doesn't remove CRON builds


pat-s:
### Component

server

### Describe the bug

When disabling a repo, the `id` in the repos table is set to `0`. However, existing CRON tasks defined in the repo are not altered in the `crons` table.

The server is trying to execute them normally and fails in the logs with
﻿
<img width="728" alt="image" src="https://github.com/user-attachments/assets/16e2236b-0754-4f5b-9fea-905aa3cac4e5">

Whether or not the CRONs should be deleted is debatable (likely not so they continue to work in case the repo gets enabled again). Yet then they should gain a new column which indicates whether they are active or not.

The following DB query can temporarily clean up (= delete) CRON jobs of deactivated repos:

```sql
DELETE FROM crons
WHERE repo_id IN (
    SELECT id
    FROM repos
    WHERE user_id = '0'
);
```

This could also be part of a daily "garbage collection" job.

### Steps to reproduce

1. Define CRON
1. Disable a repo
2. Inspect server logs during execution

### Expected behavior

CRONs are disabled/not run when a repo is disabled.

### System Info

```shell
2.7.3
```

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

PR #5884, opened 2025-12-19
Do not run crons for disabled repos

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/4361

### DIFF
--- a/server/store/datastore/cron.go
+++ b/server/store/datastore/cron.go
@@ -50,7 +50,7 @@ func (s storage) CronDelete(repo *model.Repo, id int64) error {
 // CronListNextExecute returns limited number of jobs with NextExec being less or equal to the provided unix timestamp.
 func (s storage) CronListNextExecute(nextExec, limit int64) ([]*model.Cron, error) {
 	crons := make([]*model.Cron, 0, limit)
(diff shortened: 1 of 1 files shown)


server/store/datastore/cron.go
```diff
@@ -50,7 +50,7 @@ func (s storage) CronDelete(repo *model.Repo, id int64) error {
 // CronListNextExecute returns limited number of jobs with NextExec being less or equal to the provided unix timestamp.
 func (s storage) CronListNextExecute(nextExec, limit int64) ([]*model.Cron, error) {
 	crons := make([]*model.Cron, 0, limit)
-	return crons, s.engine.Where(builder.Lte{"next_exec": nextExec}).Limit(int(limit)).Find(&crons)
+	return crons, s.engine.Join("INNER", "repos", "repos.id = crons.repo_id").Where(builder.Lte{"next_exec": nextExec}).And("repos.active").Limit(int(limit)).Find(&crons)
```
anbraten: Would it be possible to wrap repos.active in sth like builder.True to have it a bit more explicit looking?

qwerty287: There is no `builder.True`, only `builder.Eq{"active": true}` would be possible

anbraten: builder.Eq would be nice

PR #5886, opened 2025-12-19
Migrate to maintained tink-go

anbraten:
based on https://codeberg.org/crowci/crow/pulls/518/commits/4dce49d8a5cfbf51d89c26a06dc7ea0dfc0b37af

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -48,7 +48,8 @@
         "name": "Variable name",
         "value": "Variable value"
       },
-      "show_pipelines": "Show pipelines"
+      "show_pipelines": "Show pipelines",
+      "no_manual_workflows": "There were no manually executable workflows, or the filter exclude all of them."
     },
     "deploy_pipeline": {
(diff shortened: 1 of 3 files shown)


qwerty287 (approved):
Untested from my side, but the code should be unused (and broken) anyways

xoxys:
Does this finally fix encryption? If so there should be an existing issue to link.

qwerty287:
I don't think it does. The issues were in our code iirc (and there is an option to use plain aes which is broken as well I think)

xoxys:
Why bother with migration then?

xoxys:
@anbraten my question still applies

anbraten:
The lib changed owner and the original repository was archived meanwhile noting to migrate to this one.

xoxys:
Thats not the question. The entire integration is broken and useless. That why I asked why migrate instead of dropping it completely and start with a working encryption setup. This code tree exists in wp for a very long time and is broken since the beginning so Im wondering why we maintain dead code.

PR #5890, opened 2025-12-20
Fix: Pagination stops prematurely when fetching repositories with archived repos

phamleduy04:
The `/api/user/repos?all=true` endpoint only returns a subset of repositories instead of all repos that I have (200+).

The `Paginate` utility function incorrectly assumes all pages return the same number of items. It stops pagination when `len(batch) < lenFirstBatch`, which breaks when forge implementations filter results after fetching for archived repos.

solution is to send one additional API call per pagination to confirm no more pages exist

Updated test case to expect the additional API call.

I tested on Github only, didn't test on others forge.


lafriks:
How can a page return less than a full page items if more pages exist? 🤔

phamleduy04:
if the page have archived repository

phamleduy04:
sorry i was away but the repo archived code is defined [here](https://github.com/woodpecker-ci/woodpecker/blob/main/server/forge/github/github.go#L241)

```go
for _, repo := range list {
		if repo.GetArchived() {
			continue
		}
		repos = append(repos, convertRepo(repo))
	}
```

it skiped if the repo is archived, which can result in the second page number of repos less than the original page

lafriks (approved):

qwerty287:
I don't think this is a proper fix: if there's a page with archived repos only, it will have size 0, so the pagination exits.

Better fix would likely be to just move the filtering out of the pagination function and just do this afterwards, so the pagination func always gets the full pages.

qwerty287:
fixed by https://github.com/woodpecker-ci/woodpecker/pull/5924 for now

Issue #3293, opened 2024-01-29
Cron: Add Enabled Option


HerHde:
### Clear and concise description of the problem

As a user, I often want to pause the execution of cron jobs temporarily. Currently, I need to delete them for it, saving the name and cron string somewhere in the meantime.

### Suggested solution

- Add an option to enable/disable a cron job / set it to active/inactive in the job's settings
- Add a pause/resume button in the cron job list (for example with the icons `pause` and `resume` and respective title)

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

PR #5896, opened 2025-12-22
Allow to disable a cron

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/3293


lafriks:
Should it allow setting an empty string cron name?

qwerty287:
Added the check back, thanks. It wasn't possible before as well though because there's the `Validate` func which checks that

lafriks:
Please also add NOT NULL for Enabled column

server/api/cron.go
```diff
@@ -197,26 +198,29 @@ func PatchCron(c *gin.Context) {
...
 		}
 		cron.NextExec = nextExec.Unix()
 	}
-	if in.Name != "" {
-		cron.Name = in.Name
+	if in.Name != nil && *in.Name != "" {
+		cron.Name = *in.Name
+	}
+	if in.Enabled != nil {
+		cron.Enabled = *in.Enabled
```
6543, line 223: if it is reenabled we should calc next exec, else it will runn imideatly

6543: -> b587ec90cd94899f47c1048069c4e76961286a54

6543:
@qwerty287 i added b587ec90cd94899f47c1048069c4e76961286a54 hope thats ok :)

6543 (approved):
lgtm backend :)

qwerty287:
> @qwerty287 i added b587ec90cd94899f47c1048069c4e76961286a54 hope thats ok :)

Oh you're right, missed that. Thanks

lafriks (approved):

Issue #3558, opened 2024-03-26
Add description to secrets


laurentS:
### Clear and concise description of the problem

A repo can use many secrets, some of which may have non-obvious origins. I find it would be useful to be able to add notes to a secret, which could contain any information that the user deems helpful for them. An example use case would be:
secret_name: SOME_API_TOKEN
secret_value: <super_secret_value>
description: "Used by process xxx. Expiry date: ddmmyy, renew by going to ..., include API scope blablabla"

### Suggested solution

This could be a simple plain text field, visible to anyone who has access to create/modify secrets, along with the secret name. The aim is to provide hints/document the secret.
I noticed that gitlab recently added this to their CI, which I find extremely helpful.
In a perfect world, secrets would be generated in code, with proper documentation, but it's not always possible in tiny projects.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

PR #5898, opened 2025-12-22
Allow to add a note to secrets

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/3558

### DIFF
--- a/.woodpecker/docker.yaml
+++ b/.woodpecker/docker.yaml
@@ -2,7 +2,7 @@ variables:
   - &golang_image 'docker.io/golang:1.25'
   - &node_image 'docker.io/node:24-alpine'
   - &xgo_image 'docker.io/techknowlogick/xgo:go-1.25.x'
-  - &buildx_plugin 'docker.io/woodpeckerci/plugin-docker-buildx:6.0.3'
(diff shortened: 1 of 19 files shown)


qwerty287:
@6543 have you checked the UI? (In my tests it was working)

web/src/components/secrets/SecretEdit.vue
```diff
@@ -41,6 +41,10 @@
         <CheckboxesField v-model="innerValue.events" :options="secretEventsOptions" />
       </InputField>
 
+      <InputField v-slot="{ id }" :label="$t('secrets.note')">
```
6543, line 44: works through i would make it more visible somehow

- add id up the page?
- add it as fadeout text into the list?!?
- other ideas

qwerty287: > add id up the page?

Can do this, but I wanted to separate it from the actual secret and would like to keep it like this therefore.

> add it as fadeout text into the list?!?

What do you mean? A tooltip when hovering over the name? That's a good idea.

qwerty287: Added it as tooltip in the list.

6543:
thanks for the UX improvement!

I know there was something we could improve but you narowed it down :tada:

PR #5912, opened 2026-01-02
Fix cron var editing

qwerty287:
regression of https://github.com/woodpecker-ci/woodpecker/pull/5897


xoxys:
If this is a regression please link at least the intoducing PR. I would still like to see more useful PR descriptions in general.

PR #5914, opened 2026-01-04
fix: add `set -e` to commands sh script

CosmicHorrorDev:
tacks `set -e` onto the start of the `sh` script generated from the workflow's `commands` to better match the _'roughly equivalent'_ script described in `commands`' docs

https://github.com/woodpecker-ci/woodpecker/blob/f9df6cdda6e0c0c87a081c21ec6ff60594da91f1/docs/docs/20-usage/20-workflow-syntax.md?plain=1#L155-L159

---

when i was setting up woodpecker CI on one of my repos using codeberg's official instance, i noticed a `/bin/sh: 38: [[: not found` in the logs while the pipeline continued happily chugging along (yes i was trying to use bash syntax. oops)

```text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 6.68s
+ if [[ $FEATURES == "--features=syntect-default-onig" ]]; then
  cargo doc --jobs=4 $FEATURES --no-deps
fi

/bin/sh: 38: [[: not found
+ cargo test --jobs=4 --all-targets
   Compiling stable_deref_trait v1.2.1
```

https://ci.codeberg.org/repos/15883/pipeline/19/10#L204

adding an explicit `set -e` command was enough to get the pipeline to properly fail

```text
+ not-a-command smiley-face
/bin/sh: 38: not-a-command: not found
❌ Exit Code 127
```

https://ci.codeberg.org/repos/15884/pipeline/19/4#L99

## disclaimer

i haven't tested these changes against an actual deployment, so the reasoning was just based on the events detailed above. hopefully everything is right :pray::sweat_smile:


CosmicHorrorDev:
mmm trying to re-collect a late night debugging session, but it looks like this isn't entirely right. i need to look into things more first. sorry for the noise!

---

upd: ugh. looks like Arch symlinks `sh -> bash` by default, so no wonder reproducing locally was being so weird :facepalm: 

and looking over `man set` it looks like there is an explicit exception carved out for the "compound list immediately following the ... if". so it's intended behavior :melting_face:

PR #5922, opened 2026-01-07
fix: send correct argument to rpc call for name/url

jolheiser:
While working on an addon forge I eventually figured out that my plugin was crashing because gob choked on encoding `nil` rather than `[]byte{}` https://github.com/woodpecker-ci/woodpecker/blob/b65424967ac0d5d7e1bd3b08c96cf10a7b2979e8/server/forge/addon/server.go#L46

I've fixed it for both `Name` and `URL` and my plugin stopped crashing after applying this patch to a local test instance.

/cc @qwerty287 since it relates to addon forges 🙂

### DIFF
--- a/server/forge/addon/client.go
+++ b/server/forge/addon/client.go
@@ -67,13 +67,13 @@ type RPC struct {
 
 func (g *RPC) Name() string {
 	var resp string
-	_ = g.client.Call("Plugin.Name", nil, &resp)
+	_ = g.client.Call("Plugin.Name", []byte{}, &resp)
 	return resp
 }
 
 func (g *RPC) URL() string {
 	var resp string
-	_ = g.client.Call("Plugin.URL", nil, &resp)
+	_ = g.client.Call("Plugin.URL", []byte{}, &resp)
 	return resp
 }
 


anbraten (approved):

anbraten:
Could we easily test this somehow?

jolheiser:
Unfortunately the "easiest" way for me to test this was using a [nix vm](https://github.com/jolheiser/gitpecker/blob/48e1801527db80e00240f29fafeb9faae8739a06/flake.nix#L61-L160)

I'm not entirely sure if there's a way to unit test this, but an integration test would be nice to have somehow. 

Might be able to put together a very minimal addon forge and spin it up, but it's not trivial. 

----

Although to note, it likely would have been caught more easily if this error wasn't swallowed, because patching in a Println was what tipped me off to the `encoding/gob` issue.

https://github.com/woodpecker-ci/woodpecker/blob/b65424967ac0d5d7e1bd3b08c96cf10a7b2979e8/server/forge/addon/client.go#L70

Changing the method signature would be a slightly annoying interface change, but is there a way to inject some logging there easily?

qwerty287:
Yes, of course we can add a logging statement there. I can add it in a PR later.
I think when I wrote that I thought that this will never give an error if the wrapped function (i.e. the func of the addon) does not return one. But well this is a kinda stupid assumption thinking about it again ;)

I can also check out unit tests, I actually don't think this should be too complex. We can just reuse one of our real forges (or write a minimal one), and run it in the test suite directly… Not sure if that works but I would expect it. I can also check this out later.

Issue #5902, opened 2025-12-24
Not all user's repositories are synced


beckyeh-ye:
### Component

server, cli, web-ui

### Describe the bug

Not all user's repositories are synced

### Steps to reproduce

When I go to /repos to click Add repository, some repo are not show  on the list.
And I had tried rollback to 3.10.x those repo display again.

Another issue is if the repo with admin permssion more than 100, some repo not show on the list and can't search those repo.  and the web ui haven't a page to click

### Expected behavior

will show all repo when user have admin permission on it and if user have more the 100 repo can change page or use search to find out. 

### System Info

```shell
3.12.x
```

### Additional context

like issue #5667 

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

sloretz:
I see the same issue in both the web UI and CLI using Github. I only see 58 repos.

```console
$ ./woodpecker-cli repo sync | wc -l
58
$ ./woodpecker-cli --version
woodpecker-cli version 3.12.0
```

I see #5638 says it added pagination to the Repos interface. I'm not familiar with this codebase. Is it possible the web-ui and CLI are only requesting one page?

shuhaowu:
Same here, I only see 98 with the `repo sync` command.

lafriks:
I think #5890 should fix this

ericwooley:
In the meantime you can do it with the cli:

go to your github repo, view the source and look for repository_id, grab that number EG: <meta name="octolytics-dimension-repository_id" content="1...7">

Grab the 1...7

Then run `woodpecker-cli repo add 1...7`

PR #5679, opened 2025-10-23
Revert to forge internal implementation of pagination for `Repos()` and `Teams()` for gitea/forgejo

6543:
This partially reverts commit f93ef16fe305adcff711c4a0d370706d5c08ad83.

to close https://github.com/woodpecker-ci/woodpecker/issues/5667

PR #5924, opened 2026-01-08
Revert repos pagination for GH and BB

qwerty287:
just like https://github.com/woodpecker-ci/woodpecker/pull/5679 but for github and bitbucket.
BB DC has this already, and for gitlab it's not necessary as nothing is filtered there.

Also added some simple code modernizations.

replaces https://github.com/woodpecker-ci/woodpecker/pull/5890
closes https://github.com/woodpecker-ci/woodpecker/issues/5902


6543 (approved):
i dislike it that we cant proper use pagination from upstream but as hotfix till we know something better its ok

PR #5939, opened 2026-01-10
Add 3.13 docs

qwerty287:


6543:
before we release 3.13 i want this to get fixed:
image

i try to do so could you wait for a few hours?

qwerty287:
Fine, but is that a new issue?

6543:
> Fine, but is that a new issue?

just discovered -> i was going to fix our "to many requests to docker.io" and adding the registry org wide did produce that error

6543:
-> more context is in the matrix maintainer channel

6543:
@qwerty287 was xorm migration issue ... fixed: https://github.com/woodpecker-ci/woodpecker/issues/5952

Issue #5948, opened 2026-01-10
UI: new contrast makes servies icon look bad


6543:
the new contrast is good ... we just forgot to update the services icon to work propperly:

<img width="557" height="181" alt="Image" src="https://github.com/user-attachments/assets/48c1fdb2-871d-42ce-be70-ffc7f4a03872" />

<img width="277" height="269" alt="Image" src="https://github.com/user-attachments/assets/ecdfd58f-86c7-413e-9e12-53e65c766e54" />

6543:
#5958 only solved it partially:

<img width="236" height="82" alt="Image" src="https://github.com/user-attachments/assets/aa01ec43-e3a1-45e2-870d-42a8e225d3f5" />

<img width="217" height="61" alt="Image" src="https://github.com/user-attachments/assets/4051f6af-92b0-4bcd-9121-211323dacb0d" />

## still broken:

<img width="557" height="181" alt="Image" src="https://github.com/user-attachments/assets/48c1fdb2-871d-42ce-be70-ffc7f4a03872" />

<img width="277" height="269" alt="Image" src="https://github.com/user-attachments/assets/ecdfd58f-86c7-413e-9e12-53e65c766e54" />

xoxys:
Will fix it.

PR #5958, opened 2026-01-11
Apply background only to required status icons

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/5948 introduced in https://github.com/woodpecker-ci/woodpecker/pull/5880

Only apply background to status icons that benefit from it. Other status icons that work well without it, as well as service icons, are ignored now.

image


lafriks (approved):

6543:
screenshots of the service icon would have been nice 😅

6543:
@xoxys thanks for the fix but we did not catch it completly: https://github.com/woodpecker-ci/woodpecker/issues/5948#issuecomment-3734918017

PR #5965, opened 2026-01-12
Update docs for v3.13

6543:
as some bits have changed meanwhile


xoxys:
Thats why we have

>  Merge documentation PR **(shortly before release)**

The release process should be started **after** possible blockers are discussed and once started, finished within a day.

6543:
> Thats why we have
> 
> > Merge documentation PR **(shortly before release)**
> 
> The release process should be started **after** possible blockers are discussed and once started, finished within a day.

ACK

PR #5967, opened 2026-01-13
Use repo-user for api call of cron

anbraten:
We are currently using the cron creator for the forge api calls. 
This has the disadvantage that if:

- the cron creator leaves the repo / permission was removed
- or was deleted 
- or the repo user should be changed (chown call)

the cron pipeline would not work anymore and updating the cron would only work by removing and recreating it.

### DIFF
--- a/docs/docs/20-usage/72-extensions/40-configuration-extension.md
+++ b/docs/docs/20-usage/72-extensions/40-configuration-extension.md
@@ -42,10 +42,6 @@ class Request {
   repo: Repo;
   pipeline: Pipeline;
   netrc: Netrc;
-  configuration: {
-    name: string; // filename of the configuration file
-    data: string; // content of the configuration file
-  }[];
 }
 ```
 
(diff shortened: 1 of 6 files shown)


6543:
@anbraten you have to regen the swagger docs as cron type has changed because of the comment

server/model/cron.go
```diff
@@ -24,7 +24,7 @@ type Cron struct {
 	ID        int64             `json:"id"         xorm:"pk autoincr 'id'"`
 	Name      string            `json:"name"       xorm:"name UNIQUE(s) INDEX"`
 	RepoID    int64             `json:"repo_id"    xorm:"repo_id UNIQUE(s) INDEX"`
-	CreatorID int64             `json:"creator_id" xorm:"creator_id INDEX"`
+	CreatorID int64             `json:"creator_id" xorm:"creator_id INDEX"` // TODO: drop with next major version
```
lafriks, line 27: I think it still can be useful to show who created the cronjob, just don't use it for token

PR #5977, opened 2026-01-15
Per-Workflow and Per-Workflow-Step badge generation

### LINKED ISSUE #2373: Per-pipeline badge and UI tab

We're using multiple pipelines to test different parts of a project, and would like to have badges showing us the state of each part

### Suggested solution

Provide per-pipeline badges and UI tabs

### Alternative

_No response_

### Additional context

_No response_

### Validations

TumbleOwlee:
This PR introduces support of the query parameters `workflow` and `step` to be used to select a specific workflow and step when generating the badges.

Closes #2373 

### Solution

The support of workflow and step selection is added to `GetBadge(..)`  by extracting the query parameters and using `WorkflowGetTree(pipeline)` to select the workflow and the `children` of the selected workflow to select the specified step. Afterwards the name of the workflow and step are used as the badge title. In case only the workflow is specified the name of the workflow is used. If a step is also given, the badge title is given as `<workflow>: <step>`.

Additionally the UI is extended to include the input fields for the workflow and step.
Image

### Drawback

The solution requires the call to `_store.WorkflowGetTree(..)` to retrieve the workflow and step information.

### Testing

I updated the test case of the badge generation and also deployed the patched woodpecker server on my personal server. 

### Remaining Issues

- [ ] Support longer titles in badges to support workflow and workflow-step badge titles.
    - I'm currently not sure how to tackle it. 
image

- [ ] Add missing translations
- [ ] Add tests for workflow-step-badge generation. 
    - Currently not sure how to mock test data for a pipeline.


TumbleOwlee:
Hi @qwerty287, if you can take a look any time, that would be perfect. If you by chance have an idea how to tackle the badge width issue and test issue, I'm open to any suggestions. Thanks in advance.

server/api/badge.go
```diff
@@ -73,6 +73,9 @@ func GetBadge(c *gin.Context) {
 		branch = repo.Branch
 	}
 
+	name := "unknown"
+	status := model.StatusDeclined
```
qwerty287: Remove this here and directly set it in l. 105.

server/api/badge.go
```diff
@@ -99,11 +102,44 @@ func GetBadge(c *gin.Context) {
 			log.Warn().Err(err).Msg("could not get last pipeline for badge")
 		}
 		pipeline = nil
+	} else {
+		name = "pipeline"
+		status = pipeline.Status
 	}
```
qwerty287, line 106: Like this here:

```suggestion
	}

	name = "pipeline"
	status = pipeline.Status
```

web/src/assets/locales/de.json
```diff
@@ -354,7 +354,9 @@
                 "type_html": "HTML",
                 "type_markdown": "Markdown",
                 "type_url": "URL",
-                "events": "Events"
+                "events": "Events",
+                "workflow": "Workflow",
```
qwerty287: Do not change any language except English in PRs. Translations are done in weblate.

server/api/badge.go
```diff
@@ -99,11 +102,44 @@ func GetBadge(c *gin.Context) {
...
 
 	// we serve an SVG, so set content type appropriately.
 	c.Writer.Header().Set("Content-Type", "image/svg+xml")
-	c.String(http.StatusOK, badges.Generate(pipeline))
+
+	// Allow workflow (and step) specific badges
+	workflowName := c.Query("workflow")
+	if len(workflowName) != 0 {
+		name = workflowName
```
qwerty287, line 114: I think it should give an error if it's not found. And not just silently set the status to declined.

TumbleOwlee:
Thanks for your feedback. While keeping your feedback in mind, I prepared some changes. Additionally I tried to look up some sources to get an idea how the label width could be determined dynamically. Based on my research, I'm currently not sure if the possible result would justify the introduced complexity for simple badge generation because I not only have to calculate the width of the label based on the used font, but also have to keep the width in mind when modifying the coordinates in the SVG x-positions of the labels.

Because on that I want to propose another solution. Instead of calculating the width of the badges dynamically, we just provide a second wider badge option that's only used if workflow and step is selected. They look like the current badges. I just doubled the width of the label area. The drawback remains the fixed width of the normal and wider badges that would enforce a label length limit, the user has to keep in mind to prevent unreadable badge labels.
```diff
diff --git a/server/api/badge.go b/server/api/badge.go
index 72735ed7a..606b645b7 100644
--- a/server/api/badge.go
+++ b/server/api/badge.go
@@ -98,11 +98,12 @@ func GetBadge(c *gin.Context) {
                if !errors.Is(err, types.RecordNotExist) {
                        log.Warn().Err(err).Msg("could not get last pipeline for badge")
                }
-               pipeline = nil
+               c.String(http.StatusOK, badges.Generate("pipeline", nil, false))
+               return
        }

        name := "pipeline"
-       status := pipeline.Status
+       useLargeBadge := false

        // we serve an SVG, so set content type appropriately.
        c.Writer.Header().Set("Content-Type", "image/svg+xml")
@@ -111,32 +112,34 @@ func GetBadge(c *gin.Context) {
        workflowName := c.Query("workflow")
        if len(workflowName) != 0 {
                name = workflowName
-               status = model.StatusDeclined
-
                workflows, err := _store.WorkflowGetTree(pipeline)
                if err == nil {
                        for _, wf := range workflows {
                                if wf.Name == workflowName {
                                        stepName := c.Query("step")
                                        if len(stepName) == 0 {
-                                               status = wf.State
-                                       } else {
-                                               // If step is explicitly requested
-                                               name = name + ": " + stepName
-                                               for _, s := range wf.Children {
-                                                       if s.Name == stepName {
-                                                               status = s.State
-                                                               break
-                                                       }
+                                               c.String(http.StatusOK, badges.Generate(name, &wf.State, useLargeBadge))
+                                               return
+                                       }
+                                       // If step is explicitly requested
+                                       name = name + ": " + stepName
+                                       useLargeBadge = true
+                                       for _, s := range wf.Children {
+                                               if s.Name == stepName {
+                                                       c.String(http.StatusOK, badges.Generate(name, &s.State, useLargeBadge))
+                                                       return
                                                }
                                        }
                                        break
                                }
                        }
                }
+       } else if pipeline != nil {
+               c.String(http.StatusOK, badges.Generate(name, &pipeline.Status, useLargeBadge))
+               return
        }

-       c.String(http.StatusOK, badges.Generate(name, status))
+       c.String(http.StatusOK, badges.Generate(name, nil, useLargeBadge))
 }

 // GetCC
diff --git a/server/badges/badges.go b/server/badges/badges.go
index f64b2592d..86efddc93 100644
--- a/server/badges/badges.go
+++ b/server/badges/badges.go
@@ -28,6 +28,12 @@ var (
        badgeStarted = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="102" height="20" role="img" aria-label="pipeline: started"><title>pipeline: started</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="102" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="53" height="20" fill="#555"/><rect x="53" width="49" height="20" fill="#dfb317"/><rect width="102" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="275" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="430">pipeline</text><text x="275" y="140" transform="scale(.1)" fill="#fff" textLength="430">pipeline</text><text aria-hidden="true" x="765" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="390">started</text><text x="765" y="140" transform="scale(.1)" fill="#fff" textLength="390">started</text></g></svg>`
        badgeError   = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="90" height="20" role="img" aria-label="pipeline: error"><title>pipeline: error</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="90" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="53" height="20" fill="#555"/><rect x="53" width="37" height="20" fill="#9f9f9f"/><rect width="90" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="275" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="430">pipeline</text><text x="275" y="140" transform="scale(.1)" fill="#fff" textLength="430">pipeline</text><text aria-hidden="true" x="705" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="270">error</text><text x="705" y="140" transform="scale(.1)" fill="#fff" textLength="270">error</text></g></svg>`
        badgeNone    = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="90" height="20" role="img" aria-label="pipeline: none"><title>pipeline: none</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="90" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="53" height="20" fill="#555"/><rect x="53" width="37" height="20" fill="#9f9f9f"/><rect width="90" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="275" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="430">pipeline</text><text x="275" y="140" transform="scale(.1)" fill="#fff" textLength="430">pipeline</text><text aria-hidden="true" x="705" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="270">none</text><text x="705" y="140" transform="scale(.1)" fill="#fff" textLength="270">none</text></g></svg>`
+
+       largeBadgeSuccess = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="159" height="20" role="img" aria-label="pipeline: success"><title>pipeline: success</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="159" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="106" height="20" fill="#555"/><rect x="106" width="53" height="20" fill="#44cc11"/><rect width="212" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="560" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="860">pipeline</text><text x="560" y="140" transform="scale(.1)" fill="#fff" textLength="860">pipeline</text><text aria-hidden="true" x="1320" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="430">success</text><text x="1320" y="140" transform="scale(.1)" fill="#fff" textLength="430">success</text></g></svg>`
+       largeBadgeFailure = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="159" height="20" role="img" aria-label="pipeline: failure"><title>pipeline</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="159" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="106" height="20" fill="#555"/><rect x="106" width="53" height="20" fill="#e05d44"/><rect width="212" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="560" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="860">pipeline</text><text x="560" y="140" transform="scale(.1)" fill="#fff" textLength="860">pipeline</text><text aria-hidden="true" x="1320" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="350">failure</text><text x="1320" y="140" transform="scale(.1)" fill="#fff" textLength="350">failure</text></g></svg>`
+       largeBadgeStarted = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="159" height="20" role="img" aria-label="pipeline: started"><title>pipeline</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="159" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="106" height="20" fill="#555"/><rect x="106" width="53" height="20" fill="#dfb317"/><rect width="212" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="560" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="860">pipeline</text><text x="560" y="140" transform="scale(.1)" fill="#fff" textLength="860">pipeline</text><text aria-hidden="true" x="1320" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="390">started</text><text x="1320" y="140" transform="scale(.1)" fill="#fff" textLength="390">started</text></g></svg>`
+       largeBadgeError   = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="143" height="20" role="img" aria-label="pipeline: error"><title>pipeline</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="143" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="106" height="20" fill="#555"/><rect x="106" width="37" height="20" fill="#9f9f9f"/><rect width="212" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="560" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="860">pipeline</text><text x="560" y="140" transform="scale(.1)" fill="#fff" textLength="860">pipeline</text><text aria-hidden="true" x="1250" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="270">error</text><text x="1250" y="140" transform="scale(.1)" fill="#fff" textLength="270">error</text></g></svg>`
+       largeBadgeNone    = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="143" height="20" role="img" aria-label="pipeline: none"><title>pipeline</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="143" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="106" height="20" fill="#555"/><rect x="106" width="37" height="20" fill="#9f9f9f"/><rect width="212" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="560" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="860">pipeline</text><text x="560" y="140" transform="scale(.1)" fill="#fff" textLength="860">pipeline</text><text aria-hidden="true" x="1250" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="270">none</text><text x="1250" y="140" transform="scale(.1)" fill="#fff" textLength="270">none</text></g></svg>`
 )

 // Replace placeholder with specific name.
@@ -36,17 +42,40 @@ func SetBadgeName(badge, name string) string {
 }

 // Generate an SVG badge based on a pipeline.
-func Generate(name string, status model.StatusValue) string {
-       switch status {
-       case model.StatusSuccess:
-               return SetBadgeName(badgeSuccess, name)
-       case model.StatusFailure:
-               return SetBadgeName(badgeFailure, name)
-       case model.StatusError, model.StatusKilled:
-               return SetBadgeName(badgeError, name)
-       case model.StatusPending, model.StatusRunning:
-               return SetBadgeName(badgeStarted, name)
-       default:
-               return SetBadgeName(badgeNone, name)
+func Generate(name string, status *model.StatusValue, useLargeBadge bool) string {
+       if useLargeBadge {
+               if status == nil {
+                       return SetBadgeName(largeBadgeNone, name)
+               }
+
+               switch *status {
+               case model.StatusSuccess:
+                       return SetBadgeName(largeBadgeSuccess, name)
+               case model.StatusFailure:
+                       return SetBadgeName(largeBadgeFailure, name)
+               case model.StatusError, model.StatusKilled:
+                       return SetBadgeName(largeBadgeError, name)
+               case model.StatusPending, model.StatusRunning:
+                       return SetBadgeName(largeBadgeStarted, name)
+               default:
+                       return SetBadgeName(largeBadgeNone, name)
+               }
+       } else {
+               if status == nil {
+                       return SetBadgeName(badgeNone, name)
+               }
+
+               switch *status {
+               case model.StatusSuccess:
+                       return SetBadgeName(badgeSuccess, name)
+               case model.StatusFailure:
+                       return SetBadgeName(badgeFailure, name)
+               case model.StatusError, model.StatusKilled:
+                       return SetBadgeName(badgeError, name)
+               case model.StatusPending, model.StatusRunning:
+                       return SetBadgeName(badgeStarted, name)
+               default:
+                       return SetBadgeName(badgeNone, name)
+               }
        }
 }
```

Since Go doesn't support an equivalent of the ternary operator `?`, I multiplied the switch-case. If any more idiomatic equivalent exists, please let me know.

With the diff above, the badge generation would produce these outputs:
* No workflow, no step: [status-badge](https://github-ci.code-ape.dev/repos/1)
* Workflow, no step: [status-badge](https://github-ci.code-ape.dev/repos/1)
* Worknow and step: [status-badge](https://github-ci.code-ape.dev/repos/1)
* Unknown workflow: [status-badge](https://github-ci.code-ape.dev/repos/1)

I can also increase/decrease the label width area if requested.

Would this be a good direction to follow or should I do more research into the dynamic label width calculation and work out a solution based on that?

qwerty287:
The width is still fixed in the large ones, right? Because then we might solve it for most labels, but as workflow and step names can be of any length, it's easy to have a bad one again. Dynamic width would be the best...

TumbleOwlee:
I hope this goes in the right direction. I reworked the badge generation. I added the necessary steps to generate the badges dynamically based on the required width for the labels. To retrieve the width for each glyph I used the `DejaVuSans.ttf` since it's used in the badge labels. I couldn't find an easier/cleaner way to do it, so if this should be changed, I will take additional time to research another option.

With this approach the labels can be as long as the user likes it to be. Here some examples:

* Short label: [status-badge](https://github-ci.code-ape.dev/repos/1)
  * As you can see here, I enforced a minimal width of the label to prevent really short badges. In these cases the spacing of the characters will increase.
* Long label: [status-badge](https://github-ci.code-ape.dev/repos/1)
* Very long label: [status-badge](https://github-ci.code-ape.dev/repos/1)
  * Should a maximum limit be applied? Currently only a minimum is enforced.

Please keep in mind, these examples are given by my own server and unfortunately I don't have any workflows in my repository with such long names. Therefore the status is `none` for these examples.

qwerty287:
Nice - without looking at the code this looks good already. I don't think we need a minimum at all for the width. 

Maybe you could also check out some lib like https://github.com/narqo/go-badge to generate this? So we don't have to reinvent the wheel…

TumbleOwlee:
Thanks again for your quick feedback. I checked out https://github.com/narqo/go-badge and https://github.com/essentialkaos/go-badge, too. But the results are ... not as expected. For demonstration purposes I generated badges for names with different character count.

**narqo/go-badges**:
image

**essentialkaos/go-badge**:
image

It looks like both can't handle long labels. Using the same generation for long status labels, I get this result. First 4 badges use `narqo/go-badge`, last 4 use `essentialkaos/go-badge`.

image

The libraries can handle longer status labels more easily. I have to check if other libraries exist that can handle longer labels or if I just miss something when using these libraries.

server/api/badge.go
TumbleOwlee: Previously, in l. 101 the pipeline was set to `nil` if in error was returned by `_store.GetPipelineBadge(..)`. This would result in the `none` badge at the end. For now, I used the same approach but using the `status` reference. If it's `nil`, the `none` badge is generated. Should this return no badge and another HTTP status code?

Since I added the use of the `narqo/go-badge` library, I added error handling if the badge generation fails. In that case, the `InternalServerError` is returned. Should this be another status code?

server/badges/badges_test.go
```diff
@@ -22,13 +22,42 @@ import (
...
+	badgeStarted = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="111" height="20"><linearGradient id="smooth" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><mask id="round"><rect width="111" height="20" rx="3" fill="#fff"/></mask><g mask="url(#round)"><rect width="57" height="20" fill="#555"/><rect x="57" width="54" height="20" fill="#dfb317"/><rect width="111" height="20" fill="url(#smooth)"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11"><text x="29.5" y="15" fill="#010101" fill-opacity=".3">pipeline</text><text x="29.5" y="14">pipeline</text><text x="83" y="15" fill="#010101" fill-opacity=".3">started</text><text x="83" y="14">started</text></g></svg>`
```
TumbleOwlee: Since the badges are generated dynamically, I generated them once and added them for the checks in the test case.

server/badges/badges.go
```diff
@@ -14,33 +14,59 @@
...
+	case model.StatusError, model.StatusKilled:
+		return badgeStatusError, badge.Color(badgeColorError)
 	default:
-		return badgeNone
+		return badgeStatusNone, badge.Color(badgeColorNone)
+	}
+}
+
+// Generate an SVG badge based on a pipeline.
+func Generate(name string, status *model.StatusValue) (string, error) {
```
TumbleOwlee, line 52: Since the badge generation by `narqo/go-badge` can return an error, I changed the signature to return the badge and the possible error. Additionally the error is handled when calling `Generate(..)`. Is that correctly handled?

TumbleOwlee:
Maybe you got the notification for my last comment that I deleted. I was just dumb and messed it up. 
As of now I added the use of `narqo/go-badge` for simple badge generation. I also updated my instance, so the badges in my last comment that link to my instance are generated by the library now. This also allowed the removal of a lot of my previously added generation code.

server/api/badge.go
qwerty287: Ah ok, then let's keep it how it's now. I don't like it but if handled it like this already…

server/badges/badges.go
qwerty287: Yes, that's correct.

server/badges/badges_test.go
```diff
@@ -22,13 +22,42 @@ import (
...
-	assert.Equal(t, badgeNone, Generate(nil))
-	assert.Equal(t, badgeSuccess, Generate(&model.Pipeline{Status: model.StatusSuccess}))
-	assert.Equal(t, badgeFailure, Generate(&model.Pipeline{Status: model.StatusFailure}))
-	assert.Equal(t, badgeError, Generate(&model.Pipeline{Status: model.StatusError}))
-	assert.Equal(t, badgeError, Generate(&model.Pipeline{Status: model.StatusKilled}))
-	assert.Equal(t, badgeStarted, Generate(&model.Pipeline{Status: model.StatusPending}))
-	assert.Equal(t, badgeStarted, Generate(&model.Pipeline{Status: model.StatusRunning}))
+	status := model.StatusDeclined
+	badge, err := Generate("pipeline", &status)
+	assert.Nil(t, err)
```
qwerty287: There's a NoError func for this:

```suggestion
	assert.NoError(t, err)
```

server/badges/DejaVuSans.ttf
qwerty287, line 1: This file is is not necessary anymore, right?

go.mod
```diff
@@ -171,6 +172,7 @@ require (
 	github.com/muesli/ansi v0.0.0-20230316100256-276c6243b2f6 // indirect
 	github.com/muesli/cancelreader v0.2.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/narqo/go-badge v0.0.0-20230821190521-c9a75c019a59 // indirect
```
qwerty287: Why is this `indirect`? Can you run `go mod tidy` to fix this?

server/badges/badges_test.go
TumbleOwlee: Perfect, thanks for the hint. Replaced the calls to `assert.Nil` with `assert.NoError`.

server/badges/DejaVuSans.ttf
TumbleOwlee: Yes, sorry. Forgot to remove it - now removed.

go.mod
TumbleOwlee: Yes, my bad. Forgot to run it before pushing. Done.

TumbleOwlee:
I applied the requested changes. If anything else comes to mind, please let me know and I try to provide the changes as quick as possible.

qwerty287:
Thanks! Looks good to me so far, haven't tested yet though. 

@woodpecker-ci/maintainers somebody else maybe should take a look as well

qwerty287:
@TumbleOwlee can you check out the image lib? It seems to have a security issue: https://ci.woodpecker-ci.org/repos/3780/pipeline/30874/22#L41

Maybe the badge library is not that nice, it's unmaintained for 3 years… :(

And the linter fails as well: https://ci.woodpecker-ci.org/repos/3780/pipeline/30874/35

TumbleOwlee:
> @TumbleOwlee can you check out the image lib? It seems to have a security issue: https://ci.woodpecker-ci.org/repos/3780/pipeline/30874/22#L41

I can take a look later.
 
> Maybe the badge library is not that nice, it's unmaintained for 3 years… :(

Maybe use https://github.com/essentialkaos/go-badge instead? It looks like it would require to add the `.ttf` file again. Last release is from last year. But as of now it only has 22 stars, so no idea what the future will bring regarding maintenance. 

Or I try to get in contact with the maintainer of `narqo/go-badge`.

> And the linter fails as well: https://ci.woodpecker-ci.org/repos/3780/pipeline/30874/35

I will fix it later when I'm back home. I should have made sure to run it myself to prevent it...

xoxys:
The implementation in github.com/narqo/go-badge is quite basic, I would vote to copy it into wp code.

TumbleOwlee:
@qwerty287 @xoxys As I had to update the dependencies to fix the security issues and you both voted for the copy of the `narqo/go-badge` library into the woodpecker repository, I just did that. Since the library hasn't received an update in the past 3 years, I had to apply minor changes to prevent security issues reported by `trivy` and linting issues:

* Updated the dependencies 
* Minor rework of the initialization of the badge drawer to prevent issues with lint because of use of `panic`
* Minor update of the tests to use `assert.NoError` and `assert.Equal`

If I did anything wrong or if I should apply additional changes, please let me know. 

I already run `make lint` and `trivy fs --skip-dirs web/,docs/ .`  locally with no remaining errors.

xoxys:
@TumbleOwlee thanks for the effort. Can you check the remaining linter issued? `server/badges/fonts/Bitstream Vera License.txt` might need an exclusion rule.

TumbleOwlee:
> @TumbleOwlee thanks for the effort. Can you check the remaining linter issued? `server/badges/fonts/Bitstream Vera License.txt` might need an exclusion rule.

Sure, I will take a look this evening.

TumbleOwlee:
@xoxys I fixed the lint issues. 

The only issue remaining is https://ci.woodpecker-ci.org/repos/3780/pipeline/30889/23 and I'm not sure how to fix it correctly. I've never really worked with `npm`, `pnpm` and so on. So I just tried it with `pnpm update lodash`, but that did nothing. So I just removed the lines mentioning `lodash@4.17.21`. Running `make build-docs`, `make generate-docs` worked without any issue. 
Additionally, I checked if I just have to remove the file and run `pnpm install --no-frozen-lockfile`. That worked but the content of the `pnpm-lock.yaml` changed a lot. So not sure if that's the correct use.

Is the dependency `lodash@4.17.21` not used as of now and can simply be removed in the lock file?

TumbleOwlee:
I will have to take another look at the spellcheck. Not sure why nothing was reported locally by running the same command as specified in the pipeline file, but now the same command fails in the pipleine...

qwerty287:
This securitycheck is not related I think. Just ignore it.

But could you remove the unused stuff from the go-badge lib? And adding a license header to these files would be good as well, check that from the go-badge license. Similar to https://github.com/woodpecker-ci/woodpecker/blob/main/server/services/utils/hostmatcher/hostmatcher.go#L1

Sorry if I'm opening this again, but would it be possible to not embed the full font? And instead somehow only store the char widths? Because we shouldn't need most of the font itself - just the chars. (Wait with changing that till we finished discussion on the point)

qwerty287:
> I will have to take another look at the spellcheck. Not sure why nothing was reported locally by running the same command as specified in the pipeline file, but now the same command fails in the pipleine...

That's from the file name "Bitstream Vera License"

TumbleOwlee:
> But could you remove the unused stuff from the go-badge lib? And adding a license header to these files would be good as well, check that from the go-badge license. Similar to https://github.com/woodpecker-ci/woodpecker/blob/main/server/services/utils/hostmatcher/hostmatcher.go#L1

Would it be something like 
```
// Copyright 2023 The narqo/go-badge Authors. All rights reserved.
// SPDX-License-Identifier: MIT.
```
and for the files with additional changes something like this?
```
// Copyright 2023 The narqo/go-badge Authors. All rights reserved.
// SPDX-License-Identifier: MIT.
```
See: https://github.com/woodpecker-ci/woodpecker/blob/main/server/ccmenu/cc.go

> Sorry if I'm opening this again, but would it be possible to not embed the full font? And instead somehow only store the char widths? Because we shouldn't need most of the font itself - just the chars. (Wait with changing that till we finished discussion on the point)

> That's from the file name "Bitstream Vera License"

Sure, I'll wait before tackling this. If the plan will be to extract the glyph advance width and just store the map of glyphs and their advance width, the issue with the filename is also automatically solved. I'll wait until hearing back from you. Thanks in advance for all the time you take.

**Edit:** Just out of curiosity, I already looked into the freetype library to extract the necessary advance widths of all glyphs of any `.ttf` file, and to generate the necessary go map. So if you decide to embed the advance widths directly instead of the `.ttf` file, the preparation is in place and I could work on that.
Besides that and independent of your decision, maybe we have to switch the font to support more glyphs? Currently the badge generation will fail to calculate the correct width if many cyrillic characters are used, since the `vera.ttf` doesn't provide them (only 268 glyphs). I'm not sure if we have to keep that in mind or not. The `DejaVuSans.ttf`, I used previously, contains 6188 glyphs in total.

qwerty287:
@xoxys what do you think about just storing the widths instead of the full font?

xoxys:
I don't have a strong opinion. To me this sounds like a micro-optimization, storing the entire font vs just the widths should have no real impact neither on the server or client side. The only downside of embedding the entire font (regardless of the chosen font) is the increased binary size, but we talk about a few kb, do I miss something? The only real issue could be the license concerns of Vera TTF but if we replace it with e.g. DejaVuSans this is solved also.

qwerty287:
Ok, then I'd say let's keep it how it's now, but just replace the vera font with dejavu @TumbleOwlee

TumbleOwlee:
@qwerty287 Excuse the short delay. 

I applied the following changes:
* Replace Vera with DejaVuSans font
* Add license headers
* Remove unused function
* Fixed linting issues

The only remaining issue is a CVE in `./docs`: 
```
Report Summary

┌────────────────┬──────┬─────────────────┬─────────┐
│     Target     │ Type │ Vulnerabilities │ Secrets │
├────────────────┼──────┼─────────────────┼─────────┤
│ pnpm-lock.yaml │ pnpm │        1        │    -    │
└────────────────┴──────┴─────────────────┴─────────┘
Legend:
- '-': Not scanned
- '0': Clean (no security findings detected)


pnpm-lock.yaml (pnpm)
=====================
Total: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 1, CRITICAL: 0)

┌─────────────────┬────────────────┬──────────┬────────┬───────────────────┬───────────────┬─────────────────────────────────────────────────────────────┐
│     Library     │ Vulnerability  │ Severity │ Status │ Installed Version │ Fixed Version │                            Title                            │
├─────────────────┼────────────────┼──────────┼────────┼───────────────────┼───────────────┼─────────────────────────────────────────────────────────────┤
│ fast-xml-parser │ CVE-2026-25128 │ HIGH     │ fixed  │ 4.5.3             │ 5.3.4         │ fast-xml-parser: fast-xml-parser has RangeError DoS Numeric │
│                 │                │          │        │                   │               │ Entities Bug                                                │
│                 │                │          │        │                   │               │ https://avd.aquasec.com/nvd/cve-2026-25128                  │
└─────────────────┴────────────────┴──────────┴────────┴───────────────────┴───────────────┴─────────────────────────────────────────────────────────────┘
```
Since that isn't part of the `server` module, I skipped it for now. Let me know if I should also fix this issue.

xoxys:
The security issue can be ignored and is unrelated to your PR.

server/badges/color.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2026 Woodpecker Authors
```
qwerty287: Please remove that line. That's how we handle it with other similar files as well

```suggestion
```

.ecrc
```diff
@@ -12,6 +12,7 @@
     "server/store/datastore/feed.go",
     "cmd/server/openapi/docs.go",
     "_test.go",
-    "Makefile"
+    "Makefile",
+    "server/badges/fonts/Bitstream Vera License.txt"
```
qwerty287:
```suggestion
    "Makefile"
```

go.mod
```diff
@@ -128,6 +128,7 @@ require (
 	github.com/goccy/go-json v0.10.5 // indirect
 	github.com/goccy/go-yaml v1.18.0 // indirect
 	github.com/godbus/dbus/v5 v5.2.2 // indirect
+	github.com/golang/freetype v0.0.0-20170609003504-e2365dfdc4a0 // indirect
```
qwerty287: Do we still need this? Can you run `go mod tidy` again?

server/badges/utils.go
```diff
@@ -0,0 +1,15 @@
...
+
+package badges
+
+import (
+	"regexp"
+	"strings"
+)
+
+// https://github.com/badges/shields/blob/b6be37d277b64a90bde98ca10446c9d433a56681/badge-maker/lib/xml.js#L7
+func stripXMLWhitespace(xml string) string {
```
qwerty287: This method is only used once in `server/badges/style.go`, and there with a static content. Can you "pre-execute" this method and just put the resulting string into the `flatTemplate` var instead of executing this at runtime?

server/badges/color.go
```diff
@@ -0,0 +1,47 @@
...
+	ColorYellow      = Color("Yellow")
+	ColorYellowGreen = Color("YellowGreen")
+	ColorOrange      = Color("Orange")
+	ColorRed         = Color("Red")
+	ColorBlue        = Color("Blue")
+	ColorGrey        = Color("Grey")
+	ColorGray        = Color("Gray")
+	ColorLightGrey   = Color("LightGrey")
+	ColorLightGray   = Color("LightGray")
+)
```
qwerty287, line 15: Can we convert this directly to the color codes and remove all the other stuff?
I.e. put here:

```suggestion
const (
	ColorLightGreen  = "#44cc11"
	ColorGreen       = "#97ca00"
	ColorYellow      = "#dfb317"
	ColorYellowGreen = "#a4a61d"
	// ...
)
```

And I think we should only keep the colors that we also use.

server/badges/color.go
TumbleOwlee: Removed the line and kept only the `narqo/go-badge Authors` line. To separate the woodpecker and go-badge code cleanly, I moved the drawer handling from `badges.go` to `drawer.go` and only added the `narqo/go-badge Authors` line in `drawer.go`.

.ecrc
TumbleOwlee: Done.

go.mod
TumbleOwlee: Done. It's now direct instead of indirect, because it is required to parse the font file.

server/badges/utils.go
TumbleOwlee: Done

server/badges/color.go
TumbleOwlee: Done. I removed all unused colors and also removed the `Light` prefix - we use the `ColorLightGreen` but since we drop all other colors including `ColorGreen`, I renamed `ColorLightGreen => ColorGreen`

TumbleOwlee:
> Sorry that I have so much to criticize here ;) I hope that's the last time. If you don't want to continue I can also fix those myself.

5 comments are nothing at all. I appreciate you taking your time and providing important feedback. I have a really good experience contributing to this project. Thank you.

If something else comes up, please let me now. If you want to apply any changes on your own, please feel free to do it - especially if you want to make immediate progress, since I try my best, but may not be able to apply the changes immediately.

qwerty287:
Yes thanks, now the code looks good :) I'll give it a test later today and will merge then if everything's working. Thanks again for the contribution!

qwerty287 (approved):

server/badges/drawer.go
```diff
@@ -0,0 +1,130 @@
...
+
+// cspell:words Verdana
+
+import (
+	"bytes"
+	"html/template"
+	"io"
+	"sync"
+
+	"github.com/golang/freetype/truetype"
```
6543, line 14:
```go
	"github.com/golang/freetype/truetype"
```

it is archived :/

we should find an alternative

TumbleOwlee: First, sorry for that, it wasn't my intention to introduce an outdated dependency. I missed that. 

As the alternatives, I found, haven't received updates in quite some time and the ones with updates in the last 2 years aren't popular, it may be an option to get back to the idea to drop the font and only use a pre-generated map of rune widths ([comment](https://github.com/woodpecker-ci/woodpecker/pull/5977#issuecomment-3791022227)). I have already tested it in the past ([comment](https://github.com/woodpecker-ci/woodpecker/pull/5977#issuecomment-3791279601)). If that's an option, I could implement a draft. In total it would drop the dependencies to `golang.org/x/image` and `github.com/golang/freetype`,

qwerty287: If that's not too much effort for you, that would be awesome, yes! Also thought about that already but didn't do it.

qwerty287: If that's not too much effort for you, that would be awesome, yes! Also thought about that already but didn't do it.

xoxys: as an alternative use go std packages

```golang
import (
	"golang.org/x/image/font"
	"golang.org/x/image/font/opentype"
	"golang.org/x/image/font/sfnt"
)

func mustNewFontDrawer(size, dpi float64) (*font.Drawer, error) {
	f, err := sfnt.Parse(fonts.DejaVuSans)
	if err != nil {
		return nil, err
	}

	face, err := opentype.NewFace(f, &opentype.FaceOptions{
		Size:    size,
		DPI:     dpi,
		Hinting: font.HintingFull,
	})
	if err != nil {
		return nil, err
	}

	return &font.Drawer{
		Face: face,
	}, nil
}
```

PR #5978, opened 2026-01-16
Support exclusive config extensions

qwerty287:
Fix our issues in #5835 

Add a new setting "Exclusive" to the config extension.
If a config extension is marked as exclusive, the extension will be the only one that's called. The forge and a possibly configured global extension is skipped. If the global configuration is exclusive, the forge will be skipped at all. If a non-exclusive repo-level extension is configured, then it will still call both extensions, but not the forge.

### DIFF
--- a/.woodpecker/test.yaml
+++ b/.woodpecker/test.yaml
@@ -127,7 +127,7 @@ steps:
       - test
       - sqlite
     pull: true
-    image: docker.io/woodpeckerci/plugin-codecov:2.2.1
+    image: docker.io/woodpeckerci/plugin-codecov:2.2.2
     settings:
       files:
         - agent-coverage.out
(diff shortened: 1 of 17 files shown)


docs/docs/20-usage/72-extensions/40-configuration-extension.md
```diff
@@ -122,12 +126,12 @@ Example request:
     "updated_at": 0,
     "verified": false
   },
-  "netrc": {
-    "machine": "myforge.com",
-    "login": "myUser",
-    "password": "myPassword",
-    "type": "forge"
-  }
+  "configs": [
```
6543:
```suggestion
  "configuration": [
```

docs/src/pages/migrations.md
```diff
@@ -187,7 +187,6 @@ The Webhook tokens have been changed for enhanced security and therefore existin
   Image pull secrets must now be set explicitly via env var `WOODPECKER_BACKEND_K8S_PULL_SECRET_NAMES` ([#4005](https://github.com/woodpecker-ci/woodpecker/pull/4005))
 
 - Webhook signatures now use the `rfc9421` protocol
-- Replaced `configs` object by `netrc` in external configuration APIs
```
anbraten: Would keep it as it is breaking for all version from 3.0 til the one we merge this MR

6543:
Thanks :rocket:

PR #5983, opened 2026-01-17
Remove deactivated secrets type from container definition

6543:
just rm


qwerty287 (approved):

qwerty287:
There still is a test case for this

Issue #4385, opened 2024-11-16
Steps with the same name


zc-devs:
### Component

other

### Describe the bug

Trying to use steps with the same name, pipeline fails or works wrong.

https://github.com/woodpecker-ci/woodpecker/pull/3411#issuecomment-2194567897:
> be aware that steps now can have the same name

### Steps to reproduce

1. Make pipeline
```yaml
skip_clone: true
steps:
  notification:
    depends_on:
      - build
    image: alpine
    commands:
      - echo 'Alright'
  build:
    image: alpine
    commands:
      - echo 'Build'
  build:
    image: alpine
    commands:
      - echo 'Test'
  package:
    image: alpine
    commands:
      - echo 'Package'
```
2. Get error in editor
![Screenshot 2024-11-16 1](https://github.com/user-attachments/assets/5a7a174d-134f-439b-a831-571ab1fda98c)
3. Run pipeline
4. Get error in UI
![Screenshot 2024-11-16 2](https://github.com/user-attachments/assets/420f965c-6216-4e87-88fa-ef2ed36c4f2e)

### Expected behavior

- pipeline fails with error like `step names must be unique`
- or pipeline runs two steps with the same name

### System Info

```shell
Woodpecker `next-75017ac7ca`, Gitea `1.22.3`, Postgres `16`, Kubernetes `v1.30.6+k3s1`
```

### Additional context

https://github.com/woodpecker-ci/woodpecker/pull/3411#issuecomment-2194567897
https://github.com/woodpecker-ci/woodpecker/pull/3411#discussion_r1844941371

I'll attach the logs

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

zc-devs:
> mkdir: can't create directory '/woodpecker/src/': Permission denied

First pipeline on `2.7.1` works as second on `next`:

![Screenshot 2024-11-17 1](https://github.com/user-attachments/assets/e84c69f8-e2d6-4fe4-bd81-ab61ed432243)

PR #5988, opened 2026-01-18
Prefer map syntax for services

6543:
as services must have unique names as they are used for there host names, and services start all in parallel anyway, we should prefer map syntax.

because the map syntax in yaml represents how they work best:
https://github.com/woodpecker-ci/woodpecker/blob/4e0b91fbd23436c49ef647f945ac59ebaab4da0a/.woodpecker/securityscan.yaml#L34-L41
https://github.com/woodpecker-ci/woodpecker/blob/4e0b91fbd23436c49ef647f945ac59ebaab4da0a/.woodpecker/test.yaml#L144-L159

this change also makes sure re-serialize works just fine :)


6543:
this pr also to be used in https://codeberg.org/lafriks/woodpecker-pipeline-transform/pulls/33#issuecomment-9962346

pipeline/frontend/yaml/linter/linter_test.go
```diff
@@ -181,11 +182,15 @@ func TestLintErrors(t *testing.T) {
 			from: "steps: { build: { image: golang }, publish: { image: golang, depends_on: [ binary ] } }",
 			want: "One or more of the specified dependencies do not exist",
 		},
+		{
+			from: "{ when: { event: push }, steps: [ { name: test, image: docker } ], services: [ { name: redis, image: redis } ] }",
+			want: "[bad_habit] Services should be defined as map not as list",
```
6543:
```suggestion
			want: "[\"[bad_habit] Services should be defined as map not as list\"]",
```

qwerty287:
I don't like this change. It would be nice if steps and services have a consistent syntax.

I personally think we should just forbid duplicated names (e.g. via the linter), for both steps and services.

6543:
I removed all unrelated fixes and test enhancements into #5991

lafriks:
> I don't like this change. It would be nice if steps and services have a consistent syntax.
> 
> I personally think we should just forbid duplicated names (e.g. via the linter), for both steps and services.

I agree, the linter should check for duplicate names, I also prefer list not a map syntax

6543:
I removed the warning and added an check if it is a list and was declared twice

6543:
also we support duplicated step names and it has no conflict so why should we now block it?

qwerty287:
Because it's not working properly: https://github.com/woodpecker-ci/woodpecker/issues/4385

(Didn't try myself, but since that issue was opened nothing was changed to fix it)

6543:
might have missed that issue but it should be solved since worker api for steps changed to uuid for identifyers

qwerty287:
I think this issue was opened after we changed that. But I would also have to try that first.

qwerty287:
I just tested it again, and the issue is the `depends_on`. See https://ci.codeberg.org/repos/16087/pipeline/3.

The test-1 pipeline has a `depends_on` to the duplicated step `build`. And then only one build step runs. If you remove the depends_on it works (test-2 pipeline).

6543:
well depends_on is a fundamental feature ... forgot about that conflict ... so in this case as soon as it is used we need to lint for that.

i have a look what is more complex (denai of it allways or just at the dag case)

pipeline/frontend/yaml/types/container_map.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+// ContainerMap contains collection of containers.
+type ContainerMap struct {
+	ContainerMap map[string]*Container
+	Duplicated   []string
```
qwerty287, line 26: Do we have to put this into the parsing code? I think the linter should check separately for duplicated entries and for parsing just use the `ContainerList`

PR #5993, opened 2026-01-20
Add Container Registry credential extension

qwerty287:
From https://github.com/woodpecker-ci/woodpecker/pull/5852

Added UI and docs.

### DIFF
--- a/cli/repo/cron/cron_add.go
+++ b/cli/repo/cron/cron_add.go
@@ -47,6 +47,11 @@ var cronCreateCmd = &cli.Command{
 			Usage:    "cron schedule",
 			Required: true,
 		},
+		&cli.BoolFlag{
+			Name:  "enabled",
+			Usage: "whether cron is enabled",
+			Value: true,
+		},
 		common.FormatFlag(tmplCronList, true),
 	},
 }
@@ -58,6 +63,7 @@ func cronCreate(ctx context.Context, c *cli.Command) error {
 		schedule         = c.String("schedule")
 		repoIDOrFullName = c.String("repository")
 		format           = c.String("format") + "\n"
+		enabled          = c.Bool("enabled")
 	)
 	if repoIDOrFullName == "" {
 		repoIDOrFullName = c.Args().First()
@@ -77,6 +83,7 @@ func cronCreate(ctx context.Context, c *cli.Command) error {
 		Name:     cronName,
 		Branch:   branch,
 		Schedule: schedule,
+		Enabled:  enabled,
 	}
 	cron, err = client.CronCreate(repoID, cron)
 	if err != nil {
--- a/cli/repo/cron/cron_update.go
+++ b/cli/repo/cron/cron_update.go
@@ -50,6 +50,11 @@ var cronUpdateCmd = &cli.Command{
 			Name:  "schedule",
 			Usage: "cron schedule",
 		},
+		&cli.BoolFlag{
+			Name:  "enabled",
+			Usage: "whether cron is enabled",
+			Value: true,
+		},
 		common.FormatFlag(tmplCronList, true),
 	},
 }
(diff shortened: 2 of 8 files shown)


docs/docs/20-usage/72-extensions/50-registry-extension.md
```diff
@@ -0,0 +1,145 @@
...
+- [repo model](https://github.com/woodpecker-ci/woodpecker/blob/main/server/model/repo.go)
+- [pipeline model](https://github.com/woodpecker-ci/woodpecker/blob/main/server/model/pipeline.go)
+
+Example request:
+
+```json
+// Please check the latest structure in the models mentioned above.
+// This example is likely outdated.
+
+{
```
anbraten, line 54: Was thinking for the config extension already if we could simply add those responses to the openapi docs as object as well.

server/pipeline/items_test.go
```diff
@@ -138,7 +139,7 @@ steps:
 
 	mockManager.On("EnvironmentService").Return(nil, nil)
 
-	pipelineItems, err := parsePipeline(forge, store, pipeline, user, repo, yamls, envs)
+	pipelineItems, err := parsePipeline(context.Background(), forge, store, pipeline, user, repo, yamls, envs)
```
anbraten:
```suggestion
	pipelineItems, err := parsePipeline(t.Context(), forge, store, pipeline, user, repo, yamls, envs)
```

server/services/registry/with_extension.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+type withExtension struct {
+	base      Service
+	extension *httpExtension
+}
+
+// NewWithExtension returns a registry service that combines a base service with an HTTP extension.
+// The extension is called during RegistryListPipeline to fetch additional registry credentials.
+func NewWithExtension(base Service, extension *httpExtension) Service {
```
anbraten, line 33: combined should be doing the same thing

qwerty287: No, because it's the other way round:
Combined uses the base service for everything, while the other services are read only. And the base one has priority.
Here it has to use the read only services first. Of course this can be implemented with some extra `bool`switch, should I do this? Not sure how easy this will be in combined though.

docs/docs/20-usage/72-extensions/50-registry-extension.md
qwerty287: Can we move that to another PR?

server/services/registry/with_extension.go
anbraten: Couldn't you achieve this by just swapping the registries passed to combined already 🤔

docs/docs/20-usage/72-extensions/50-registry-extension.md
anbraten: Yes

server/services/registry/with_extension.go
qwerty287: No because combined has a list of readonly services (that are not prioritized) and only one full service (that is prioritized).

Here we need one readonly that's prioritized.

qwerty287:
@woodpecker-ci/maintainers could I get a review here? :)

server/services/registry/with_extension.go
```diff
@@ -0,0 +1,135 @@
...
+	"go.woodpecker-ci.org/woodpecker/v3/server/model"
+)
+
+type withExtension struct {
+	base      Service
+	extension *httpExtension
+}
+
+// NewWithExtension returns a registry service that combines a base service with an HTTP extension.
+// The extension is called during RegistryListPipeline to fetch additional registry credentials.
```
6543:
```suggestion
// The extension is called during RegistryListPipeline to fetch additional registry credentials and
// the extension registries taking priority.
```

server/services/registry/with_extension.go
6543: I also looked into the Cobo implementation... but as long as we dont refactor it to specify an option to who is priorizing what ... we cant use it and so this simple aproach is fine by me :)

PR #5999, opened 2026-01-22
Render MD in pipeline titles

qwerty287:
After: Screenshot 2026-01-22 at 08-57-43 Aktivitäten · woodpecker-ci_woodpecker · Woodpecker

Before: Screenshot 2026-01-22 at 08-57-48 Aktivitäten · woodpecker-ci_woodpecker · Woodpecker


qwerty287:
Oh this breaks the `…` shortening: 
grafik

Somebody knows how we can fix that?

xoxys:
I dont think there is an easy way.

qwerty287:
@xoxys fixed with an easy fix. marked supports inline parsing which makes it working: https://marked.js.org/using_advanced#inline

xoxys:
Nice one! Can you share a screenshot how it will look like with a shortened header that uses code blocks?

qwerty287:
grafik

The heading is not rendered in inline mode.

Issue #2875, opened 2023-11-26
Fix pipeline cancelling


qwerty287:
### Component

server, agent

### Describe the bug

This is mainly a summary issue of https://github.com/woodpecker-ci/woodpecker/issues/833, https://github.com/woodpecker-ci/woodpecker/issues/2062, https://github.com/woodpecker-ci/woodpecker/issues/2911, #4349, https://github.com/woodpecker-ci/woodpecker/issues/2197

I've been trying to debug this without real success.

I've been using the local backend, and can do the following observations:
- cancel pipeline while running: completely broken. The commands are finished, the step is marked as success, the pipeline too (https://github.com/woodpecker-ci/woodpecker/issues/2911)
-  cancelling a pending pipeline seems to work for me

On ci.woodpecker-ci.org, I can see (uses docker backend):
- cancel pending pipeline, agent is available: the pipeline starts anyways (this probably is #2062)
- cancel running pipeline: works in general, but new status is failing but should be killed

zc-devs:
Woodpecker 2.1.1, Kubernetes.

- cancel pending pipeline: removed from queue, released resources, `killed` pipeline status, `skipped` step status
[Screenshot 2024-01-08 1](https://github.com/woodpecker-ci/woodpecker/assets/127358482/b094237d-4bca-4361-bcbe-de631b2d466d)

- cancel running pipeline: removed from queue, released resources, `success` pipeline and step statuses
[Screenshot 2024-01-08 2](https://github.com/woodpecker-ci/woodpecker/assets/127358482/32a928cd-8381-4a82-80c2-0d1d2a50ef91)

zc-devs:
https://github.com/woodpecker-ci/woodpecker/issues/2253#issuecomment-2076542998

fernandrone:
I've got a related issue, which is somewhat worrisome. 

I was able to reproduce the original buck on a 2.3.0 installation with Kubernetes backend. I've observed it's inconsistent: sometimes cancelling will correctly show the running step as killed/cancelled and mark the pipeline as canceled. The last step to run will show "Oh no, we got some errors! Canceled" (remaining steps in the same workflow will show as grey, with the message "This step has been canceled.").  Sometimes, it will show the last step to run as successful instead (and remaining steps in the same workflow will also show as grey, with the message "This step has been canceled.").

[...]

Issue #2192, opened 2023-08-10
Pipeline never completes, timeout doesn't kick


strk:
### Component

server

### Describe the bug

We have jobs that have been running for over 16h and are still considered in progress, like this:
https://woodie.osgeo.org/repos/30/pipeline/777

The timeout is set to 180 minutes (3 hours)

### System Info

```shell
source	"https://github.com/woodpecker-ci/woodpecker"
version	"1.0.1"
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

strk:
We found that the problem was a non-silent infinite loop. Is the woodpecker timeout only effective during silent periods ?
https://trac.osgeo.org/postgis/ticket/5473

Issue #1870, opened 2023-06-18
when.platform does filter based on server not agent platform string


6543:
we might need to figure out how to compile pipelines based on that ...

I would suggest that this filter does only apply if root.platform or root.lables.platform is set ... so that you can do something like this:

```yaml
matrix:
    PLATFORMS:
        - linux/amd64
        - linux/arm

platform: ${PLATFORMS}

steps:
    hello:
        image: golang
        commands:
            - echo hello woodpecker
            - sleep 3s
            - echo build things ...

    arm-only:
        image: bash
        commands: echo from arm only
        when:
            platform: linux/arm
```

PR #6011, opened 2026-01-25
Fix pipeline cancellation status handling and step state synchronization

### LINKED ISSUE #833: Cancle a Running Pipeline update status as failed

### LINKED ISSUE #2062: canceled pipelines get marked as passed (workflows stay cancled)
version: v1.0.0-rc1

when the task queue is not empty and agents are bussy, as son as there is a free slot the canceled pipeline gets marked as passed.

e.g. https://ci.woodpecker-ci.org/repos/3780/pipeline/6895

my edjucated guess is, that the task queue handler has the pipeline still in it when a pipeline gets canceled, and when it checks that entry it sees that there is nothing to do anymore and mark it green.

so there are two ways we should ensure: `canceled pipelines should not get updated as passed by the task-queue-handler` and `pipeline gets removed from task queue on cancel`

### LINKED ISSUE #2911: Cancle workflow on local backend make it as successfull
os: windows
shell: cmd.exe

### LINKED ISSUE #3848: Exit-codes are not used by server
The state send to the server contains an exit-code for the workflow. However as workflows have no exit-code they get simply ignored and are only considered by logging:

https://github.com/woodpecker-ci/woodpecker/blob/2bda19024ece0bf0deab5244fb8799f653d5adfc/agent/runner.go#L159-L175

In case of a canceled pipeline we even seem to remove the error: https://github.com/woodpecker-ci/woodpecker/blob/2bda19024ece0bf0deab5244fb8799f653d5adfc/agent/runner.go#L168

sounds to be related to #2197

### LINKED ISSUE #4349: Cancelled pipelines marked as failed
### Component

server

### Describe the bug

Sometimes pipeline cancellations still result in errors that will flag the pipeline as failed instead of cancelled:

image

### Steps to reproduce

Have not found a reliable way to reproduce this specific case.

### Expected behavior

Pipeline should be marked as cancelled instead of failed.

### System Info

```shell
next-bf5405b6cc
```

### Additional context

_No response_

### Validations

6543:
### Summary

Fixes critical issues with pipeline status updates when pipelines are cancelled or workflows are killed by the agent.
The agent's gRPC client error handling during cancellation was causing incorrect status propagation, leading to cancelled pipelines being marked as either "failed" or "success" instead of "killed".

### Problems Fixed

1. **Cancelled running pipelines show incorrect status**: When cancelling a running pipeline, the gRPC Wait() error causes the pipeline to be marked as "failed" instead of "killed"
2. **Status updates overwrite cancellation state**: Cancelled pipelines initially show correct "killed" status but get asynchronously updated to "success" 
3. **Implicit cancellation signaling**: Use explicit canceled response property instead of relying on indirect signaling (like exit codes or context errors) for cancellations
4. **Pending pipelines start after cancellation**: Race condition allows cancelled pending pipelines to begin execution when agents become available

### WebUI Screenshot showing working canceled steps & workflow
image

### Changes

- Add explicit `Canceled` field to gRPC protocol (WorkflowState and StepState) - bumps protocol version to 15
- Update agent's Wait() to return cancellation status explicitly instead of relying on errors
- Properly handle gRPC client cancellation using `context.WithCancelCause()` to distinguish cancellation reasons
- Prevent status updates from overwriting cancellation state in step status handling
- Ensure cancelled workflows remain in "killed" state throughout their lifecycle
- Add context error checking in local backend's WaitStep for cancelled workflows
- Fix defer placement in logger and tracer to ensure proper cleanup
- Improve step status state machine to handle all transitions correctly (pending→running→success/failure/killed)
- Update backend State to include Started timestamp for accurate timing
- Fix queue cleanup to handle already-deleted tasks gracefully

### Testing

Tested on both local- and docker-backend

read more at #2875

---

close #833
close #3848
close #2062
close #2911
close #4349

---

block https://github.com/woodpecker-ci/woodpecker/pull/6056
block https://github.com/woodpecker-ci/woodpecker/pull/6039

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,7 +57,7 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/yaronf/httpsign v0.4.1
 	github.com/zalando/go-keyring v0.2.6
-	gitlab.com/gitlab-org/api/client-go v1.26.0
+	gitlab.com/gitlab-org/api/client-go v1.27.0
 	go.uber.org/multierr v1.11.0
 	golang.org/x/crypto v0.47.0
 	golang.org/x/image v0.35.0
--- a/go.sum
+++ b/go.sum
@@ -599,8 +599,8 @@ github.com/zalando/go-keyring v0.2.6 h1:r7Yc3+H+Ux0+M72zacZoItR3UDxeWfKTcabvkI8u
 github.com/zalando/go-keyring v0.2.6/go.mod h1:2TCrxYrbUNYfNS/Kgy/LSrkSQzZ5UPVH85RwfczwvcI=
 github.com/zenazn/goji v0.9.0/go.mod h1:7S9M489iMyHBNxwZnk9/EHS098H4/F6TATF2mIxtB1Q=
 github.com/ziutek/mymysql v1.5.4/go.mod h1:LMSpPZ6DbqWFxNCHW77HeMg9I646SAhApZ/wKdgO/C0=
-gitlab.com/gitlab-org/api/client-go v1.26.0 h1:xmcSIsaBtMk1Ep5NsY69hcRo9CXID0EuGM0x+b7kbc0=
-gitlab.com/gitlab-org/api/client-go v1.26.0/go.mod h1:r060AandE8Md/L5oKdUVjljL8YQprOAxKzUnpqWqP3A=
+gitlab.com/gitlab-org/api/client-go v1.27.0 h1:DwmP8515dUrMYBE9OzxvCIji2CbILgUd1oie2etLUY4=
+gitlab.com/gitlab-org/api/client-go v1.27.0/go.mod h1:r060AandE8Md/L5oKdUVjljL8YQprOAxKzUnpqWqP3A=
 go.opentelemetry.io/auto/sdk v1.2.1 h1:jXsnJ4Lmnqd11kwkBV2LgLoFMZKizbCi5fNZ/ipaZ64=
 go.opentelemetry.io/auto/sdk v1.2.1/go.mod h1:KRTj+aOaElaLi+wW1kO/DZRXwkF4C5xPbEe3ZiIhN7Y=
 go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.54.0 h1:TT4fX+nBOA/+LUkobKGW1ydGcn+G3vRw9+g5HwCphpk=
--- a/server/store/datastore/pipeline.go
+++ b/server/store/datastore/pipeline.go
@@ -15,8 +15,11 @@
 package datastore
 
 import (
+	"context"
+	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff/v5"
 	"xorm.io/builder"
 	"xorm.io/xorm"
 
@@ -128,47 +131,80 @@ func (s storage) GetPipelineCount() (int64, error) {
 	return s.engine.Count(new(model.Pipeline))
 }
 
+// CreatePipeline creates a new pipeline with retry logic for unique constraint errors.
 func (s storage) CreatePipeline(pipeline *model.Pipeline, stepList ...*model.Step) error {
-	sess := s.engine.NewSession()
-	defer sess.Close()
-	if err := sess.Begin(); err != nil {
-		return err
-	}
+	// Maximum number of retries
+	const maxRetries = 3
 
-	repoExist, err := sess.Where("id = ?", pipeline.RepoID).Exist(&model.Repo{})
-	if err != nil {
-		return err
-	}
+	// Create backoff configuration
+	exponentialBackoff := backoff.NewExponentialBackOff()
 
-	if !repoExist {
-		return ErrorRepoNotExist{RepoID: pipeline.RepoID}
-	}
-
-	// calc pipeline number
-	var number int64
-	if _, err := sess.Select("MAX(number)").
-		Table(new(model.Pipeline)).
-		Where("repo_id = ?", pipeline.RepoID).
-		Get(&number); err != nil {
-		return err
-	}
-	pipeline.Number = number + 1
-
-	pipeline.Created = time.Now().UTC().Unix()
-	// only Insert set auto created ID back to object
-	if _, err := sess.Insert(pipeline); err != nil {
-		return err
-	}
-
-	for i := range stepList {
-		stepList[i].PipelineID = pipeline.ID
-		// only Insert set auto created ID back to object
-		if _, err := sess.Insert(stepList[i]); err != nil {
-			return err
+	// Execute with backoff retry
+	_, err := backoff.Retry(context.Background(), func() (struct{}, error) {
+		sess := s.engine.NewSession()
+		defer sess.Close()
+		if err := sess.Begin(); err != nil {
+			return struct{}{}, err
 		}
+
+		repoExist, err := sess.Where("id = ?", pipeline.RepoID).Exist(&model.Repo{})
+		if err != nil {
+			return struct{}{}, err
+		}
+
+		if !repoExist {
+			return struct{}{}, ErrorRepoNotExist{RepoID: pipeline.RepoID}
+		}
+
+		// calc pipeline number
+		var number int64
+		if _, err := sess.Select("MAX(number)").
+			Table(new(model.Pipeline)).
+			Where("repo_id = ?", pipeline.RepoID).
+			Get(&number); err != nil {
+			return struct{}{}, err
+		}
+		pipeline.Number = number + 1
+
+		pipeline.Created = time.Now().UTC().Unix()
+		// only Insert set auto created ID back to object
+		if _, err := sess.Insert(pipeline); err != nil {
+			if isUniqueConstraintError(err) {
+				return struct{}{}, err
+			}
+			return struct{}{}, backoff.Permanent(err)
+		}
+
+		for i := range stepList {
+			stepList[i].PipelineID = pipeline.ID
+			// only Insert set auto created ID back to object
+			if _, err := sess.Insert(stepList[i]); err != nil {
+				if isUniqueConstraintError(err) {
+					return struct{}{}, err
+				}
+				return struct{}{}, backoff.Permanent(err)
+			}
+		}
+
+		return struct{}{}, sess.Commit()
+	}, backoff.WithBackOff(exponentialBackoff), backoff.WithMaxTries(maxRetries))
+
+	return err
+}
+
+// isUniqueConstraintError checks if an error is a unique constraint violation error.
+func isUniqueConstraintError(err error) bool {
+	if err == nil {
+		return false
 	}
 
-	return sess.Commit()
+	errStr := err.Error()
+	// Check for common unique constraint error patterns across different databases
+	return strings.Contains(errStr, "duplicate key") ||
+		strings.Contains(errStr, "Duplicate entry") ||
+		strings.Contains(errStr, "UNIQUE constraint failed") ||
+		strings.Contains(errStr, "unique constraint") ||
+		strings.Contains(errStr, "UNIQUE violation")
 }
 
 func (s storage) UpdatePipeline(pipeline *model.Pipeline) error {


6543:
also when Wait() does not return err the step does not get cancled in the background either till it finished ?!?
-> we need special case where wait err is nil but ctx is cancled handled propperly

6543:
related work https://github.com/woodpecker-ci/woodpecker/pull/3850

qwerty287:
That's nothing new…

There's a summary issue for the whole pipeline canceling: https://github.com/woodpecker-ci/woodpecker/issues/2875

6543:
https://github.com/6543-forks/woodpecker/blob/1b74c2baa6dd8b67577510d3dd85246ffd24d60c/agent/runner.go#L102-L113

@qwerty287 I'll work on that now ... will take some time i guess to get through all of it :)

ghost:
@6543 What do you think about addressing this as part of this pull request, given that there are already several PRs today touching the same area?

If you’d prefer, I can also split this out into a separate PR to keep the scope tighter.
Related issue: #1870

ghost:
Issue #2192 is partially related to pull requests #6020 and #6021, but it is not yet finalized. I will be creating a PR that can close the issue.

6543:
well https://github.com/woodpecker-ci/woodpecker/pull/6021 was a good cleanup in terms of code comments and the context is now right ... so i merged it as such

6543:
I think https://github.com/woodpecker-ci/woodpecker/pull/6020 will be obsolete once we address it properly.

6543:
yes mos errors related to status/state handeling come form the legacy code from drone or side efects while woodpecker did diverge form that original codes intention and there was no cleanup made ... jet

6543:
well cancle state is now propergated propperly:

image
image

pipeline/pipeline.go
```diff
@@ -90,6 +90,7 @@ func (r *Runtime) MakeLogger() zerolog.Logger {
 }
 
 // Run starts the execution of a workflow and waits for it to complete.
+// TODO: here the cancle error is not propagated!
```
6543:
```suggestion
```

server/pipeline/workflow_status.go
```diff
@@ -42,5 +42,8 @@ func UpdateWorkflowStatusToDone(store store.Store, workflow model.Workflow, stat
 	if workflow.Error != "" {
 		workflow.State = model.StatusFailure
 	}
+	if workflow.State == model.StatusKilled || state.Canceled {
```
6543:
```suggestion
	if state.Canceled {
```

6543:
the agent listens to the server via Wait and thats the only connection we have to signal stuff from server down to agent in a push way and it only works once ... I have to thing if there are other cases where we might want to push things from server down to agent if so we should add a new interface that does that and send cancel over that else we should alter the wait just to unique signal that workfow got cancled when ended ...

6543:
... anywai this pull is good enouth to fix one longstanding issue right now

ghost:
> well https://github.com/woodpecker-ci/woodpecker/pull/6021 was a good cleanup in terms of code comments and the context is now right ... so i merged it as such

Ok

ghost:
> ... anywai this pull is good enouth to fix one longstanding issue right now

I agree

agent/runner.go
```diff
@@ -143,9 +145,13 @@ func (r *Runner) Run(runnerCtx, shutdownCtx context.Context) error {
 	state := rpc.WorkflowState{
 		Started: time.Now().Unix(),
 	}
+
+	// Initialize workflow on the server
```
6543:
```suggestion
```

6543:
looks like now that we have it working we have an UI glitch:
image

agent/rpc/client_grpc.go
```diff
@@ -177,8 +178,13 @@ func (c *client) Wait(ctx context.Context, workflowID string) (err error) {
 			// non-fatal errors
 			log.Warn().Err(err).Msgf("grpc error: wait(): code: %v", status.Code(err))
 		default:
-			log.Error().Err(err).Msgf("grpc error: wait(): code: %v", status.Code(err))
-			return err
+			switch {
+			case strings.Contains(err.Error(), "desc = queue: task canceled"):
+				return pipeline.ErrCancel
```
6543: we should make that signal explicit and not relay on an queue error msg

rpc/proto/woodpecker.proto
```diff
@@ -48,12 +48,14 @@ message StepState {
...
 
 message WorkflowState {
-  int64  started = 4;
-  int64  finished = 5;
-  string error = 6;
+  int64  started = 1;
+  int64  finished = 2;
+  string error = 3;
+  bool   canceled = 4;
 }
```
6543, line 59: needs more thought: should we have enum so we can signal who has canceled ... if the server canceled it or the agent e.g. because of an agent shouldown

rpc/peer.go
```diff
@@ -67,37 +69,41 @@ type (
 )
 
 // Peer defines a peer-to-peer connection.
+// TODO: suffix Next, Wait, Init, Done and Extend with "Workflow" for docu purpose
```
6543: this needs more docu!

rpc/proto/woodpecker.proto
6543: similar to https://github.com/woodpecker-ci/woodpecker/issues/6024 ... or handled by it ...

rpc/peer.go
ghost:
```golang
// Peer defines the RPC interface between an agent and the Woodpecker server.
//
// It is responsible for coordinating workflow execution, including:
//   - Fetching workflows from the queue
//   - Reporting workflow and step state transitions
//   - Handling workflow completion and cancellation
//
// Method naming intentionally omits explicit "Workflow" suffixes to keep the API concise.
// The workflow or step context is defined by the parameters of each method.
```

6543: > //   - Handling workflow completion and cancellation

duplicate to aboth

we should instead mention how Wait blocks and how ... but tjat is subject to change

also we should prefux the interface

6543:
## TODOs:

services dont report back when they are finished:
image

similar issue for detached steps:
image

because both are just droped into the background:
https://github.com/woodpecker-ci/woodpecker/blob/8b6c814666244d2bc0619aa4c636803b08631db8/pipeline/pipeline.go#L276-L279

rpc/proto/woodpecker.proto
6543: nah dont takkle to much at once

6543:
moved last todo out as it can be handled in a next pull: #6033

pipeline/pipeline.go
```diff
@@ -58,7 +58,8 @@ type Runtime struct {
 	engine  backend.Backend
 	started int64
 
-	ctx    context.Context
+	ctx context.Context
```
6543, line 65: TODO: add comment when to use wna whatvto expect and when not to use

rpc/peer.go
```diff
@@ -15,89 +15,286 @@
...
+	//
+	// Returns:
+	//   - nil on success
+	//   - error if communication fails or server rejects the state
 	Done(c context.Context, workflowID string, state WorkflowState) error
 
-	// Extend extends the workflow deadline
+	// Extend extends the execution deadline for the workflow with the given ID.
+	//
+	// Workflows have a timeout (specified in Workflow.Timeout from Next()). Agents should
```
6543: bad wording

server/rpc/server.go
```diff
@@ -104,29 +109,43 @@ func (s *WoodpeckerServer) Update(c context.Context, req *proto.UpdateRequest) (
...
-func (s *WoodpeckerServer) Wait(c context.Context, req *proto.WaitRequest) (*proto.Empty, error) {
-	res := new(proto.Empty)
-	err := s.peer.Wait(c, req.GetId())
-	return res, err
+// Wait blocks until the workflow is complete.
+// Also signals via err if workflow got canceled.
+func (s *WoodpeckerServer) Wait(c context.Context, req *proto.WaitRequest) (*proto.WaitResponse, error) {
+	res := new(proto.WaitResponse)
+	res.Canceled = false
+	if err := s.peer.Wait(c, req.GetId()); err != nil {
```
6543: move err detection one caller up the stack

pipeline/backend/types/state.go
```diff
@@ -16,11 +16,14 @@ package types
...
 // State defines a container state.
 type State struct {
+	// Unix start time
+	Started int64 `json:"started"`
 	// Container exit code
 	ExitCode int `json:"exit_code"`
 	// Container exited, true or false
 	Exited bool `json:"exited"`
 	// Container is oom killed, true or false
+	// TODO: well known errors as string enums into ./errors.go
```
6543:
```suggestion
	// TODO (6024): well known errors as string enum into ./errors.go
```

server/rpc/rpc.go
```diff
@@ -53,6 +53,7 @@ type RPC struct {
 }
 
 // Next blocks until it provides the next workflow to execute.
+// TODO: on server shutdown release waiting agents
```
6543:
```suggestion
// TODO (6038): Server does not release waiting agents on graceful shutdown
```

rpc/peer.go
6543: c0a7a6903a187f99056d52179b733e21a5b07d8e

6543:
image

works :rock:

agent/runner.go
```diff
@@ -51,6 +50,7 @@ func NewRunner(workEngine rpc.Peer, f rpc.Filter, h string, state *State, backen
 	}
 }
 
+// Run executes an workflow via an backend, tracks its state and reports back the state to the server.
```
qwerty287:
```suggestion
// Run executes a workflow using a backend, tracks its state and reports the state back to the server.
```

agent/runner.go
```diff
@@ -143,9 +141,12 @@ func (r *Runner) Run(runnerCtx, shutdownCtx context.Context) error {
 	state := rpc.WorkflowState{
 		Started: time.Now().Unix(),
 	}
+
 	if err := r.client.Init(runnerCtx, workflow.ID, state); err != nil {
-		logger.Error().Err(err).Msg("workflow initialization failed")
-		// TODO: should we return here?
+		logger.Error().Err(err).Msg("signaling workflow initialization to server failed")
+		// This should never happen, still it did so lets clean up and end
```
qwerty287: This comment is weird ;) what should this mean? "it should never happen still it did"?

pipeline/pipeline.go
```diff
@@ -160,8 +165,10 @@ func (r *Runtime) traceStep(processState *backend.State, err error, step *backen
 		return nil
 	}
 
+	// this is nil if the step was not started jet
```
qwerty287:
```suggestion
	// this is nil if the step was not started yet
```

And I would move the comment into the if branch

server/pipeline/cancel.go
```diff
@@ -40,24 +40,13 @@ func Cancel(ctx context.Context, _forge forge.Forge, store store.Store, repo *mo
...
-		workflowsToCancel []string
-		workflowsToEvict  []string
-	)
-	for _, workflow := range workflows {
-		if workflow.State == model.StatusRunning {
-			workflowsToCancel = append(workflowsToCancel, fmt.Sprint(workflow.ID))
-		}
-		if workflow.State == model.StatusPending {
-			workflowsToEvict = append(workflowsToEvict, fmt.Sprint(workflow.ID))
+	var workflowsToCancel []string
```
qwerty287, line 43: Is that still working with the difference between canceled and skipped status? (Workflows/steps that never started should be skipped, if they started they should be canceled)

qwerty287 (commented):
I'm not really into this code, but so far it seems quite good

pipeline/pipeline.go
6543: i dissagree as this would in my opinion then comment on `processState = new(backend.State)`

but happy to get an argument proving otherwise

server/pipeline/cancel.go
6543: yes as that is handled below:
https://github.com/woodpecker-ci/woodpecker/blob/39acb2a4be8f133b8ff011458dc84475795987ce/server/pipeline/cancel.go#L67-L82

6543: the code changed here only affects the queue

(can also be observed in screenshot)

agent/runner.go
6543: -> 39ce00c03b8865e57674a84940483b0dd38800c2

agent/runner.go
```diff
@@ -90,34 +90,32 @@ func (r *Runner) Run(runnerCtx, shutdownCtx context.Context) error {
...
+		logger.Debug().Msg("start listening for server side cancel signal")
 
-			// Propagate cancellation to pipeline + backend
-			cancel()
+		if canceled, err := r.client.Wait(workflowCtx, workflow.ID); err != nil {
+			logger.Error().Err(err).Msg("server returned unexpected err while waiting for workflow to finish run")
+			cancelWorkflowCtx(err)
 		} else {
+			if canceled {
+				logger.Debug().Err(err).Msg("server side cancel signal  received")
```
lafriks:
```suggestion
				logger.Debug().Err(err).Msg("server side cancel signal received")
```

rpc/proto/woodpecker.proto
```diff
@@ -48,12 +48,14 @@ message StepState {
...
   int32  exit_code = 5;
   string error = 6;
+  bool   canceled = 7;
 }
 
 message WorkflowState {
-  int64  started = 4;
-  int64  finished = 5;
-  string error = 6;
+  int64  started = 1;
```
lafriks, line 55: Is it ok that this will be a breaking change in grpc protocol?

lafriks:
Apart from that question code LGTM

rpc/proto/woodpecker.proto
6543: yes we have the version api that checks for that witch we are not alowed to change beside bumping the integer **in gRPC** for that

and on the same version, its fine if we break and cleanup

pipeline/pipeline.go
6543: thanks for the fix btw.

... after thinking a bit and as you mentioned mentioned it so it might be best de describe this behaviour in tje function description itselfe

6543: @qwerty287 i did b69f68f3526a6cd1ebebc88c04363b9ab2db4d02 should be more clear now

6543:
@qwerty287 :point_up: updated the code comment and code to make it clear and more err resistant

qwerty287 (approved):
Seems fine to me, but untested from my side (I guess you did that…) and I'm also not totally into this code. But as @lafriks comment is solved I guess it's fine

6543:
deployed on our ci: https://ci.woodpecker-ci.org/repos/6106/pipeline/1778/4

qwerty287:
@6543 there seems to be another issue: https://ci.woodpecker-ci.org/repos/3780/pipeline/31596

I canceled that manually but it's showing as succeeded

6543:
hmm that's another issue unrelated to agent rpc but queue implementation ... would you mind creating an issue?

PR #6020, opened 2026-01-26
pipeline: always destroy step on cancellation

Pnkcaht:
### What was happening
When a pipeline step was canceled (e.g. manual cancel or agent shutdown), the step execution would exit early, but **`DestroyStep` was not guaranteed to run**.  
This could leave backend resources (such as Docker containers) running, resulting in **orphaned steps after cancellation**.

### What this PR changes
This PR ensures that **`DestroyStep` is always executed**, regardless of how the step finishes.

- Guarantees cleanup on success, failure, or cancellation
- Executes `DestroyStep` via `defer` in the step execution path
- Uses a non-cancelable context to avoid cleanup being interrupted
- Prevents orphaned backend resources

### Why this approach
Cancellation may stop waiting on a step, but it should **never skip cleanup**.  
Ensuring `DestroyStep` always runs makes step lifecycle handling safer and aligns with expected CI behavior.

This also closes the current gap where cleanup depended on higher-level execution flow.

### Implementation overview
```mermaid
sequenceDiagram
    participant Runtime as Pipeline Runtime
    participant Engine as Backend Engine
    participant Ctx as Step Context
    participant Step as Backend Step

    Runtime->>Engine: StartStep
    Engine->>Step: start

    alt Step canceled
        Ctx-->>Runtime: ctx.Done()
        Runtime->>Engine: DestroyStep (background context)
    else Step finishes normally
        Step-->>Engine: exit
        Runtime->>Engine: DestroyStep
    end

    Engine-->>Runtime: cleanup completed
```

### DIFF
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -256,12 +256,18 @@ func (r *Runtime) exec(step *backend.Step) (*backend.State, error) {
 	}
 
 	// Always ensure the step is destroyed, regardless of how this function exits
-	// (success, error, or context cancellation).
+	// (success, error, or workflow cancellation).
 	//
-	// We intentionally use context.Background() to make sure cleanup is not
-	// interrupted by cancellation.
+	// Cleanup should respect the workflow lifecycle:
+	// - While the workflow is running, we use the workflow context.
+	// - If the workflow context was canceled, we fall back to a bounded shutdown
+	//   context to ensure cleanup can still complete without breaking services.
 	defer func() {
-		if err := r.engine.DestroyStep(context.Background(), step, r.taskUUID); err != nil {
+		ctx := r.ctx
+		if ctx.Err() != nil {
+			ctx = GetShutdownCtx()
+		}
+		if err := r.engine.DestroyStep(ctx, step, r.taskUUID); err != nil {
 			if r.logger != nil {
 				logger := r.MakeLogger()
 				logger.Error().Err(err).Msg("failed to destroy step")


ghost:
Solved here @6543  I ensured that DestroyStep is always executed, even on cancellation/shutdown, preventing orphaned steps/containers.

image

pipeline/pipeline.go
```diff
@@ -250,10 +250,25 @@ func (r *Runtime) execAll(steps []*backend.Step) <-chan error {
...
 	if err := r.engine.StartStep(r.ctx, step, r.taskUUID); err != nil {
 		return nil, err
 	}
 
+	// Always ensure the step is destroyed, regardless of how this function exits
+	// (success, error, or context cancellation).
+	//
+	// We intentionally use context.Background() to make sure cleanup is not
+	// interrupted by cancellation.
+	defer func() {
```
6543, line 265: detatched steps will get destroyed imediatly

ghost: This is intentional. Cleanup must outlive the execution context to avoid orphaned detached steps

6543: this breaks services!

ghost: I will change

ghost:
@6543 The change has been made, what do you think?

image

6543:
fix included in https://github.com/woodpecker-ci/woodpecker/pull/6011

PR #6022, opened 2026-01-26
fix(yaml): report error when workflow is skipped or has no stages

### LINKED ISSUE #5917: Repair failed after forge user was disabled
### Component

server

### Describe the bug

My user was disabled on our forge (bitbucket datacenter) over the holidays by mistake, this led to all our pipelines on the repositories I onboarded to stop, since the access was revoked.

Another admin then tried to repair them, but this also failed, because my users still exists in woodpecker, but fails later in the repair when it tries to refresh my locked token https://github.com/woodpecker-ci/woodpecker/blob/de0179ca4d6769d5a900998ed090f7ffe7d7f685/server/api/repo.go#L677

### Steps to reproduce

1. Onboard a repository.
2. Disable the user on the forge.
3. Now repair repository

### Expected behavior

1. That the repository is correctly fixed on a repair.
2. That since we run single tenant, this shouldn't be tied to a user, since employees come and go. (might be a new seperate discussion)

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-b7b33db98b"}
```

### Additional context

_No response_

### Validations

### LINKED ISSUE #5968: yaml is invalid on server while valid with cli tool
### Component

server

### Describe the bug

I had a certain yaml config for my workflow that was:
- correctly linted by online parsers
- deemed valid by the woodpecker cli

However when using this config on the server it fails with the error:
yaml: line [someline]: did not find expected key

### Steps to reproduce

1. Install server 3.12
2. Install cli 3.12
3. Use following yaml:
```yml
check-commit-msg:
    image: node:alpine
    when:
      event: push
      branch: master
    commands:
      - "npm install -g @commitlint/cli @commitlint/config-conventional"
      - "echo \"module.exports = {extends: ['@commitlint/config-conventional']}\" > commitlint.config.js"
      - "git log -1 --pretty=%B | npx commitlint"
```

### Expected behavior

Either:
1. the yaml to work
2. the CLI to report the same error

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.12.0"}
```

### Additional context

_No response_

### Validations

Pnkcaht:
## What was happening

When a pipeline YAML resulted in a workflow being skipped or producing no runnable stages, the workflow was silently dropped during pipeline creation.

Specifically:

- when conditions that did not match caused the workflow to be skipped
- Workflows that compiled successfully but resulted in zero stages were ignored
- No pipeline error was reported to the user
- The pipeline could end up empty or filtered without explanation

This resulted in:

- Pipelines not running with no clear reason
- Confusing UX (pipeline filtered or empty without visible errors)
- Difficult debugging of YAML configuration issues

## Related issue

Closes #5968
Closes #5917 

## What this PR changes

This PR ensures that workflows are no longer silently skipped during pipeline creation.

- Reports a pipeline error when a workflow is skipped due to when conditions
- Reports a pipeline error when a workflow produces no runnable stages
- Prevents workflows from being dropped without feedback
- Makes YAML and configuration issues visible in the pipeline error list

## Why this approach

During pipeline creation, the step builder is responsible for deciding whether a workflow should be executed.

Previously, certain conditions (when mismatch or empty compilation result) caused workflows to be discarded by returning nil without registering an error.
This made the pipeline behavior opaque to users.

By explicitly reporting these cases as pipeline errors:

- Users get immediate feedback about why a workflow did not run
- Pipelines no longer appear empty or silently filtered
- Misconfigured or incomplete YAML is easier to diagnose

This keeps pipeline behavior predictable and transparent.

## Implementation overview 

```mermaid
sequenceDiagram
    participant User as User / Commit
    participant Server as Woodpecker Server
    participant Builder as StepBuilder
    participant Pipeline as Pipeline Model

    User->>Server: Trigger pipeline
    Server->>Builder: Build workflows
    Builder->>Builder: Evaluate when conditions
    alt when does not match
        Builder->>Pipeline: report pipeline error
    else no runnable stages
        Builder->>Pipeline: report pipeline error
    end
    Pipeline-->>Server: visible errors returned
```

### DIFF
--- a/server/pipeline/stepbuilder/step_builder.go
+++ b/server/pipeline/stepbuilder/step_builder.go
@@ -139,7 +139,13 @@ func (b *StepBuilder) genItemForWorkflow(workflow *model.Workflow, axis matrix.A
 	// parse yaml pipeline
 	parsed, err := yaml.ParseString(substituted)
 	if err != nil {
-		return nil, &errorTypes.PipelineError{Message: err.Error(), Type: errorTypes.PipelineErrorTypeCompiler}
+		return nil, multierr.Append(
+			errorsAndWarnings,
+			&errorTypes.PipelineError{
+				Message: err.Error(),
+				Type:    errorTypes.PipelineErrorTypeCompiler,
+			},
+		)
 	}
 
 	// lint pipeline
@@ -170,7 +176,17 @@ func (b *StepBuilder) genItemForWorkflow(workflow *model.Workflow, axis matrix.A
 		log.Debug().Str("pipeline", workflow.Name).Msg(
 			"pipeline config could not be parsed",
 		)
-		return nil, multierr.Append(errorsAndWarnings, err)
+		return nil, multierr.Append(
+			errorsAndWarnings,
+			&errorTypes.PipelineError{
+				Message: fmt.Sprintf(
+					"workflow %s skipped by when condition",
+					workflow.Name,
+				),
+				Type: errorTypes.PipelineErrorTypeCompiler,
+			},
+		)
+
 	}
 
 	ir, err := b.toInternalRepresentation(parsed, environ, workflowMetadata, workflow.ID)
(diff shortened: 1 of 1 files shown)


ghost:
I analyzed issue #5917 and noticed that the changes described there are already covered by this PR.
In practice, this PR addresses the same two problems by reporting skipped workflows and workflows with no runnable stages as warnings instead of silently ignoring them.

PR #6025, opened 2026-01-27
Feat/logs auto follow scroll

### LINKED ISSUE #5971: Make follow logs dynamic
Currently we have two modis, dont follow new logs and follow new logs.

I want a new modi (or replace both) that does following:

If you are scrowled down it will auto follow new logs. but as soon as you scrowl back up it does not follow anymore.

currently what anys me is that i have t switch the modis frequently and if i forget while looking at a lock it gets lost because of the autoscrowl. also if i forget to reenable it i dont notice a log change

Pnkcaht:
## What was happening

The log viewer currently offers two separate modes: auto-follow and manual scroll.

In practice, this causes frequent UX issues:

- When auto-follow is enabled, scrolling up to inspect earlier logs is interrupted as soon as new logs arrive
- Users must manually toggle modes to inspect logs safely
- If auto-follow is left enabled, the view jumps to the bottom and context is lost
- If auto-follow is disabled and forgotten, new log output can go unnoticed

This results in a frustrating experience when inspecting long or fast-updating logs, especially during running pipelines.

## Related issue

Closes #5971

## What this PR changes

This PR simplifies log following behavior by making it scroll-position aware instead of mode-based.

- Logs automatically follow new output **only when the user is already scrolled to the bottom**
- As soon as the user scrolls up, auto-follow is disabled
- When the user scrolls back to the bottom, auto-follow resumes automatically
- No manual toggle or mode switching is required

## Why this approach

Scroll position is the most reliable signal of user intent.

Previously, auto-follow was controlled by explicit modes, which required constant manual interaction and were easy to forget.
By deriving auto-follow state from the scroll position:

- User intent is respected automatically
- Context is preserved when inspecting older logs
- New output is not missed when returning to the bottom
- The UI becomes simpler and more predictable

## Implementation overview

```mermaid
sequenceDiagram
    participant User
    participant UI as Log Viewer
    participant Stream as Log Stream

    Stream->>UI: new log lines
    UI->>UI: check scroll position

    alt user is at bottom
        UI->>UI: autoFollow = true
        UI->>UI: scroll to bottom
    else user scrolled up
        UI->>UI: autoFollow = false
        UI->>UI: preserve scroll position
    end

    User->>UI: scrolls back to bottom
    UI->>UI: autoFollow = true
```

## Tests 

No automated tests were added.


qwerty287:
-> https://github.com/woodpecker-ci/woodpecker/pull/6036

PR #6036, opened 2026-01-28
Dynamic log following

qwerty287:
taken from https://github.com/woodpecker-ci/woodpecker/pull/6025
Closes https://github.com/woodpecker-ci/woodpecker/issues/5971

:warning: please review extra carefully as the base code is from the AI user. It seems valid to me though and is working, I tested it.

(you can test with something like https://github.com/qwerty287/wp-log-autoscroll)


qwerty287:
@woodpecker-ci/maintainers could I get a review here? :)

6543:
works nicely and the code diff is smal :)

6543:
deployed on https://ci.woodpecker-ci.org

xoxys:
Can we please make it configurable? This gets pretty annoying if you really want to debug long logs while the step is still running. Can be the default but please add back a button to disable it... This change will definetly block me to update to the next realease.

PR #6041, opened 2026-01-29
Fix killed status

MartinSchmidt:
After the https://github.com/woodpecker-ci/woodpecker/pull/6011 was merged, this pr has been repurposed, to just add a cancel reason, so one can see why it was cancelled.

As well as a few tests I had added as part of the old proposed fix.

image

### DIFF
--- a/cmd/server/openapi/docs.go
+++ b/cmd/server/openapi/docs.go
@@ -4678,6 +4678,20 @@ const docTemplate = `{
                 }
             }
         },
+        "CancelInfo": {
+            "type": "object",
+            "properties": {
+                "data": {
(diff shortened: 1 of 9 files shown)


MartinSchmidt:
I now see a issue on the topic here: #2875, and there is another PR in progress #6011

6543:
thansk for the contribution :)

yes I hae a pull that fixes the underlaying cause :)

still once that is merged we still can merge your tests ... :)

MartinSchmidt:
Moved to new PR https://github.com/woodpecker-ci/woodpecker/pull/6072

PR #6049, opened 2026-01-31
Document required forge methods

6543:
we should document what methods are allowed to return `types.ErrNotImplemented` and handle them explizite.

This allows new developers and addons to start with a smaller implementation first

### DIFF
--- a/server/api/login.go
+++ b/server/api/login.go
@@ -145,7 +145,7 @@ func HandleAuth(c *gin.Context) {
 				PerPage: perPage,
 			})
 			if errors.Is(terr, forge_types.ErrNotImplemented) {
-				log.Warn().Msg("Could not fetch membership of user as forge adapter did not implement it")
+				log.Debug().Msg("Could not fetch membership of user as forge adapter did not implement it")
 			} else if terr != nil {
 				log.Error().Err(terr).Msgf("cannot verify team membership for %s", userFromForge.Login)
 				c.Redirect(http.StatusSeeOther, server.Config.Server.RootPath+"/login?error=internal_error")
--- a/server/api/repo.go
+++ b/server/api/repo.go
@@ -397,7 +397,7 @@ func GetRepoBranches(c *gin.Context) {
 
 	branches, err := _forge.Branches(c, repoUser, repo, session.Pagination(c))
 	if errors.Is(err, forge_types.ErrNotImplemented) {
-		log.Warn().Msg("Could not fetch repo branch list as forge adapter did not implement it")
+		log.Debug().Msg("Could not fetch repo branch list as forge adapter did not implement it")
 	} else if err != nil {
 		log.Error().Err(err).Msg("failed to load branches")
 		c.String(http.StatusInternalServerError, "failed to load branches: %s", err)
@@ -438,7 +438,7 @@ func GetRepoPullRequests(c *gin.Context) {
 
 	prs, err := _forge.PullRequests(c, repoUser, repo, session.Pagination(c))
 	if errors.Is(err, forge_types.ErrNotImplemented) {
-		log.Warn().Msg("Could not fetch repo pull-request list as forge adapter did not implement it")
+		log.Debug().Msg("Could not fetch repo pull-request list as forge adapter did not implement it")
 	} else if err != nil {
 		_ = c.AbortWithError(http.StatusInternalServerError, err)
 		return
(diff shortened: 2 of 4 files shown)


server/forge/forge.go
```diff
@@ -105,6 +108,9 @@ type Forge interface {
 	// Dir fetches all files in a directory at a specific commit.
 	// Supports pipeline configurations split across multiple files.
 	// Should return files only.
+	//
+	// Errors:
+	//  - Expect types.ErrNotImplemented to be returned
```
qwerty287: Here and for all methods below some part of the sentence is missing. Something like:
```suggestion
	//  - Expect types.ErrNotImplemented to be returned if not supported by the forge
```

server/api/login.go
```diff
@@ -144,7 +144,9 @@ func HandleAuth(c *gin.Context) {
 				Page:    page,
 				PerPage: perPage,
 			})
-			if terr != nil {
+			if errors.Is(terr, forge_types.ErrNotImplemented) {
+				log.Warn().Msg("Could not fetch membership of user as forge adapter did not implement it")
```
qwerty287: I wouldn't use warn here, rather debug

6543: Info?

I still would like to let it be seen in the logs as it is only i cases of addons being used as forge

qwerty287: I not only refer to this log, but to all you added. I would change all to debug.

Note that if you use a forge that doesn't support all features - bitbucket e.g. misses some I think - this will kind of spam you logs every time the forge method is used… That could be solved with debug.

6543: ok changing

6543:
```suggestion
				log.Debug().Msg("Could not fetch membership of user as forge adapter did not implement it")
```

server/api/repo.go
```diff
@@ -396,7 +396,9 @@ func GetRepoBranches(c *gin.Context) {
 	forge.Refresh(c, _forge, _store, repoUser)
 
 	branches, err := _forge.Branches(c, repoUser, repo, session.Pagination(c))
-	if err != nil {
+	if errors.Is(err, forge_types.ErrNotImplemented) {
+		log.Warn().Msg("Could not fetch repo branch list as forge adapter did not implement it")
```
6543:
```suggestion
		log.Debug().Msg("Could not fetch repo branch list as forge adapter did not implement it")
```

server/api/repo.go
```diff
@@ -435,7 +437,9 @@ func GetRepoPullRequests(c *gin.Context) {
 	forge.Refresh(c, _forge, _store, repoUser)
 
 	prs, err := _forge.PullRequests(c, repoUser, repo, session.Pagination(c))
-	if err != nil {
+	if errors.Is(err, forge_types.ErrNotImplemented) {
+		log.Warn().Msg("Could not fetch repo pull-request list as forge adapter did not implement it")
```
6543:
```suggestion
		log.Debug().Msg("Could not fetch repo pull-request list as forge adapter did not implement it")
```

server/cache/membership.go
```diff
@@ -56,7 +59,10 @@ func (c *membershipCache) Get(ctx context.Context, _forge forge.Forge, u *model.
 	}
 
 	perm, err := _forge.OrgMembership(ctx, u, org)
-	if err != nil {
+	if errors.Is(err, forge_types.ErrNotImplemented) {
+		log.Warn().Msg("Could not check user org membership as forge adapter did not implement it")
```
6543:
```suggestion
		log.Debug().Msg("Could not check user org membership as forge adapter did not implement it")
```

server/services/config/forge.go
```diff
@@ -146,7 +146,9 @@ func (f *forgeFetcherContext) getFirstAvailableConfig(c context.Context, configs
 			files, err := f.forge.Dir(c, f.user, f.repo, f.pipeline, strings.TrimSuffix(fileOrFolder, "/"))
 			// if folder is not supported we will get a "Not implemented" error and continue
 			if err != nil {
-				if !errors.Is(err, types.ErrNotImplemented) && !errors.Is(err, &types.ErrConfigNotFound{}) {
+				if errors.Is(err, types.ErrNotImplemented) {
+					log.Warn().Msg("Could not fetch config folder as forge adapter did not implement it")
```
6543:
```suggestion
					log.Debug().Msg("Could not fetch config folder as forge adapter did not implement it")
```

PR #6056, opened 2026-02-02
Don't propagate workflow error from agent back to agent

6543:
we currently inject errors from workflows back to the queue. this is needed for things that should run in case an error occur.
But we should not report it back as queue error!

with this we wrap all external errors in an own error type so we can filter the out

## before
image
## after
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,7 +57,7 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/yaronf/httpsign v0.4.1
 	github.com/zalando/go-keyring v0.2.6
-	gitlab.com/gitlab-org/api/client-go v1.27.0
+	gitlab.com/gitlab-org/api/client-go v1.28.0
 	go.uber.org/multierr v1.11.0
 	golang.org/x/crypto v0.47.0
 	golang.org/x/image v0.35.0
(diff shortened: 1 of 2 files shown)


server/rpc/rpc.go
```diff
@@ -118,7 +118,10 @@ func (s *RPC) Wait(c context.Context, workflowID string) (canceled bool, err err
 			// we explicit send a cancel signal
 			return true, nil
 		}
-		// unknown error happened
+		if errors.Is(err, new(queue.ErrExternal)) {
+			// we do not have to give back the error an agent already told us
```
6543: the actual fix of this pull ... we filter out the errors that are not queue related as they are external and already logged/handled....

6543:
updated description with screenshot of logs before and after :)

server/queue/fifo.go
```diff
@@ -134,6 +134,9 @@ func (q *fifo) finished(ids []string, exitStatus model.StatusValue, err error) e
 	q.Lock()
 	defer q.Unlock()
 
+	// it's  an external error so we wrap it
```
6543:
```suggestion
	// it's an external error so we wrap it
```

6543:
moved it into queue implementation only ...

server/queue/fifo.go
```diff
@@ -166,7 +169,9 @@ func (q *fifo) Wait(ctx context.Context, taskID string) error {
 		select {
 		case <-ctx.Done():
 		case <-state.done:
-			return state.error
+			if !errors.Is(state.error, new(ErrExternal)) {
```
6543, line 173: the actual fix of this pull ... we filter out the errors that are not queue related as they are external and already logged/handled....

6543, line 173:
```suggestion
			// only return queue errors and no workflow errors
			if !errors.Is(state.error, new(ErrExternal)) {
```

6543:
ok finally it's  ready for review again :sweat_smile:

6543:
yes it is still handled and visible just not back propagated :D

as i touch that part of the code I'm certain ... still did a test on my side:
image

PR #6059, opened 2026-02-03
Fix OAuth token refresh in webhook handling for Bitbucket and GitHub

stardothosting:
## Summary

Fixes #5590
Fixes #5713

This PR fixes an issue where webhook handling fails with "failure to parse hook" error when the user's OAuth access token has expired. The root cause is that the Bitbucket and GitHub forge implementations make API calls during webhook processing without first refreshing the OAuth token.

## Problem

When a webhook arrives from Bitbucket or GitHub, the `Hook()` function (and its helper functions) make API calls to fetch additional data (changed files, repo info, etc.). These API calls use the stored OAuth access token, which may have expired.

**Before this fix:**
1. Webhook arrives
2. `Hook()` makes API calls with potentially expired token
3. API call fails with "OAuth2 access token expired"
4. Error bubbles up as HTTP 400 "failure to parse hook"
5. `forge.Refresh()` is called later in `PostHook()` - but it's too late

**Example error from logs:**

`failure to parse hook error="OAuth2 access token expired. Use your refresh token to obtain a new access token."`

## Solution

Add `forge.Refresh()` calls before making API calls in the webhook handling code paths. This follows the same pattern already used by:
- Bitbucket Data Center forge (`server/forge/bitbucketdatacenter/bitbucketdatacenter.go`)
- Other code paths like `pipeline.Create()`, `cron.go`, etc.

### Changes

**Bitbucket** (`server/forge/bitbucket/bitbucket.go`):
- Added `forge.Refresh()` in `Hook()` before API calls

**GitHub** (`server/forge/github/github.go`):
- Added `forge.Refresh()` in `loadChangedFilesFromPullRequest()`
- Added `forge.Refresh()` in `getTagCommitSHA()`
- Added `forge.Refresh()` in `loadChangedFilesFromCommits()`

## Testing

- All existing Bitbucket and GitHub forge tests pass
- Tested in production environment with Bitbucket (waited for token expiry, webhook succeeded after fix)

## Notes

- The fix is non-blocking: if the store is unavailable from context, it gracefully skips the refresh
- For GitHub OAuth Apps (traditional), tokens don't expire so `Refresh()` returns early. This fix primarily helps users with GitHub Apps (which have expiring tokens) and all Bitbucket users.

PR #6062, opened 2026-02-03
Refresh token before use for all forges

qwerty287:
https://github.com/woodpecker-ci/woodpecker/pull/6059 missed gitea, forgejo and gitlab. BB DC has this already.

Please make sure to check whether PRs that might affect all forges also fix all forges and not just some.

### DIFF
--- a/web/src/assets/locales/bar.json
+++ b/web/src/assets/locales/bar.json
@@ -41,7 +41,7 @@
             "loading": "Lod grod…",
             "no_logs": "Koane Logs",
             "pipeline": "Pipeline #{pipelineId}",
-            "log_title": "Step-Logs",
+            "log_title": "Schrid-Logs",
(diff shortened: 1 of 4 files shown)


anbraten:
What do you think about changing the forge model to have a general wrapper that checks for all endpoints with a user as arg (those needing a token) if a refresh is necessary?

6543:
works for me still I think anbratens idea is not bad ... :)

so I'll not merge it and just wait for u

qwerty287:
> What do you think about changing the forge model to have a general wrapper that checks for all endpoints with a user as arg (those needing a token) if a refresh is necessary?

This doesn't work here because the `Hook` function does not have access to the repo user directly, it needs to identify the repo first. This is all handled inside the forges (and there is no good way to split it, except having two functions `HookParse` and `HookGetRequiredExtraData` or whatever, and I wouldn't like this).

6543:
ok

PR #6067, opened 2026-02-05
Add retry logic for CreatePipeline with backoff

### LINKED ISSUE #3884: Likely concurrency issue pq: 'duplicate key value violates unique constraint \"UQE_pipelines_s\"'

myselfghost:
closes #3884

### DIFF
--- a/server/store/datastore/pipeline.go
+++ b/server/store/datastore/pipeline.go
@@ -15,8 +15,11 @@
 package datastore
 
 import (
+	"context"
+	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff/v5"
 	"xorm.io/builder"
 	"xorm.io/xorm"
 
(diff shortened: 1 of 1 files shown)


qwerty287:
This doesn't fix the actual issue, I don't think this is a good way to go.

myselfghost:
> This doesn't fix the actual issue, I don't think this is a good way to go.

The reality is that this issue has been shelved for too long. We should take measures to prevent it from happening, so as to ensure stable system usage.

qwerty287:
Yes, but there was a solution using a lock already. That would be better I think (that pr was closed at some point, don't know why. I was not involved there)

myselfghost:
> Yes, but there was a solution using a lock already. That would be better I think (that pr was closed at some point, don't know why. I was not involved there)

Although this solution is not the most elegant approach, it can solve the problem and allow the system to operate normally.

6543 (approved):
I would considder it an hotfix that is selfe contained ... so okish for me

6543:
> Yes, but there was a solution using a lock already. That would be better I think (that pr was closed at some point, don't know why. I was not involved there)

@qwerty287 I tryed a locking mechanism and currently it has still edgecases where some sql dialects dont work propperly or poorly and so I would say this hotfix is good enoth for now and self contained.

so reverting it and using an lock mechanism later is still easy to do...

https://github.com/woodpecker-ci/woodpecker/pull/5534#issuecomment-3371992139 -> not sure when i do dedicate time to it ...

PR #5410, opened 2025-08-12
[WIP] Rework Bitbucket webhook integration

6543:
To also have the new metadata event and improve it's integration this pull updates all  fixtures and add new ones.
also we refactor the parsing ...

# Highlights:
 - [when.path filter](https://woodpecker-ci.org/docs/usage/workflow-syntax#path) support
 - PullMetadata event support
 - proper check for Pull events e.g. dont cause one if you only change from draft to read or reverse

TODO:
- [x] impl
- [ ] tests
  - [x] parse unit tests 
  - [ ] Hook() tests with new api calls
- [ ] Docs
  - [ ] BB
  - [ ] mention to regen webhook for old repos to activate
  
## subworks:
- [ ] https://github.com/woodpecker-ci/woodpecker/pull/6071

PR #6071, opened 2026-02-06
server.store: rename GetPipelineLast to GetPipelineLastByBranch

6543:
picked from https://github.com/woodpecker-ci/woodpecker/pull/5410 to get conflicts resolved faster


qwerty287:
Why do we add a method without using it?

6543:
because the user is in https://github.com/woodpecker-ci/woodpecker/pull/5410 and i need to redo it completly from scratch as rebasing is not possible anymore ...

6543:
what GetPipelineLastByForgeURL allows us is to differentiate for bitbucket between pull events and pull_metadata events ... by comparing if we already got an event from an pull and the commit did not change ... -> metadata else pull_push

as BB does not differentiate between them...

qwerty287:
But can't you add the new func in that PR then? I would like to avoid unused functions in the code. Renaming is fine for me

PR #6103, opened 2026-02-10
Use tabs for indentation in embedded JSON

qwerty287:
My editor (and I guess any other also) is not working properly otherwise.

### DIFF
--- a/server/forge/bitbucket/fixtures/handler.go
+++ b/server/forge/bitbucket/fixtures/handler.go
@@ -220,162 +220,162 @@ const tokenNotFoundPayload = `
 
 const repoPayload = `
 {
-  "full_name": "test_name/repo_name",
-  "scm": "git",
-  "is_private": true
+	"full_name": "test_name/repo_name",
+	"scm": "git",
+	"is_private": true
 }
 `
 
 const permissionRepoPayload = `
 {
-  "full_name": "test_name/%s",
-  "scm": "git",
-  "is_private": true
+	"full_name": "test_name/%s",
+	"scm": "git",
+	"is_private": true
 }
 `
 
 const repoPayloadFromHook = `
 {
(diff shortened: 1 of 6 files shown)


xoxys:
This shouldnt be any problem. Embedded templates should use the common indentation of the target language in case of json this shoulf stay as spaces.

6543:
what IDE do you use?

qwerty287:
I use Kate (KDE's text editor), does this work with different editors?

The editor can't know which file type this if it does not analyze the content itself. and this would also only work for common types.

xoxys:
> I use Kate (KDE's text editor), does this work with different editors?

What does "work" means here? I can open the files in vscode/codium and auto-formatting on safe also works.

qwerty287:
By "work" I mean what happens when editing this embedded JSON data. Kate uses tabs then (obviously because the rest of the file uses tabs), while most lines use spaces. Yes, it works correctly for the Go code in the file. There already were many inconsistencies in the JSON, likely happened due to this issue. I don't think any editor handles this.

xoxys:
Correct you have to manually indent it. Feel free to switch to tabs for this particular case but thats not a global solution. In the autoscaler for example we had the same issue with the embedded cloudinit config but cloudinit is picky and fails on tab indened files. The best approach would be to use `//go:embed` instead of embedded multuiline data.

qwerty287:
For larger files yes, embedding is best, but  having a new file for every 3-line json seems a bit overkill to me. I can change it though if you prefer that way

xoxys:
Also fine for me to go with this PR. Just wanted to say, this will not work in all cases and will not be enforced by linters etc. so we will end up with a mixed result again.

qwerty287:
So can we merge this then?

PR #6039, opened 2026-01-28
Report status for detached steps and services

6543:
image
image
image

 ---
close #6033
resolve #2831

PR #6106, opened 2026-02-11
Document how to ignore failure on services

6543:
and make it an official setting for services

as https://github.com/woodpecker-ci/woodpecker/pull/6039 made services report status back now...

> ... detached steps and services should let a pipeline fail. ...

fix: https://ci.woodpecker-ci.org/repos/3780/pipeline/31546/10

### DIFF
--- a/docs/docs/20-usage/60-services.md
+++ b/docs/docs/20-usage/60-services.md
@@ -37,6 +37,20 @@ services:
       - 51820/udp
 ```
 
+## Stopping
+
+Services once not needed anymore get **SIGTERM** signal. If they don't honor it they will just get killed via **SIGKILL**.
+If you have services that don't gracefully stop and it does not matter, just ignore the failure:
+
+```diff
+ services:
+   - name: database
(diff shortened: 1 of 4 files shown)


docs/docs/20-usage/60-services.md
```diff
@@ -37,6 +37,20 @@ services:
       - 51820/udp
 ```
 
+## Stopping
+
+Services that are no longer needed receive a **SIGTERM** signal. If they don't respond to it, they will be forcibly terminated with **SIGKILL**.
+If you have services that don't shut down gracefully and it doesn't matter, you can simply ignore the failure:
```
xoxys:
```suggestion
## Stopping

Services that are no longer needed receive a **SIGTERM** signal. If they do not respond, they are forcibly terminated with **SIGKILL**.
If there are services that do not shut down properly and this doesn't matter, you can simply ignore the error:
```

xoxys (approved):

PR #6119, opened 2026-02-13
Use priority-based merging to determine pipeline and workflow status

### LINKED ISSUE #6114: Manually cancelled pipeline marked as successful
https://ci.woodpecker-ci.org/repos/3780/pipeline/31596

I canceled that manually but it's showing as succeeded

From https://github.com/woodpecker-ci/woodpecker/pull/6011#issuecomment-3891203387

@6543

qwerty287:
expected to close #6114 

Not tested yet.


qwerty287:
@6543 I think it's the source issue of #6114 but I couldn't try it yet. (if you don't like the approach also fell free to comment)

server/api/badge.go
```diff
@@ -120,7 +120,10 @@ func GetBadge(c *gin.Context) {
 				if wf.Name == workflowName {
 					stepName := c.Query("step")
 					if len(stepName) == 0 {
-						if status == nil || wf.Failing() {
+						if status != nil {
+							merged := status.Merge(wf.State)
```
6543: i dont like to see bussines logic in that ... the Pipeline.Status should have stored that and be returned.
if it is not sufficient enouth we have an bug...

server/model/const.go
```diff
@@ -67,6 +68,30 @@ const (
 	StatusCreated  StatusValue = "created"  // created / internal use only
 )
 
+// list of statuses by their priority. Most important is on top.
+var statusPriorityOrder = []StatusValue{
```
6543: looks like bussines logic is now encoded in models ... i would rather move it to pipeline/state as we have it now :)

server/api/badge.go
qwerty287: No because this logic is just about a single step. And merging is necessary because if you have a matrix pipeline you can have multiple steps with the same name.

So this status is not about the full pipeline, only about some of the workflows/steps

server/model/const.go
6543: we kan still consider it a hotfix and have a refactor afterwards but ifbit's not serious and smal why not refactor now ... 🤔

qwerty287: Would fix that properly here, as the bug itself is quite old I think. It's nothing really critical. I'll try to improve it next week.

qwerty287: @6543 moved to `server/pipeline`

6543 (approved):

PR #6130, opened 2026-02-17
Store workflow/step dependencies in database and update API

lafriks:
Needed for frontend to be able to draw dependency graph for pipeline (workflows and steps)

I wanted to split frontend PR from backend as it still needs to be discussed on how to better implement UI for this but that should not be blocking to merge backend part


pipeline/backend/types/step.go
```diff
@@ -37,6 +37,7 @@ type Step struct {
 	Networks       []Conn            `json:"networks,omitempty"`
 	DNS            []string          `json:"dns,omitempty"`
 	DNSSearch      []string          `json:"dns_search,omitempty"`
+	DependsOn      []string          `json:"depends_on,omitempty"`
```
6543, line 40: why do we need it in backend?

server/pipeline/items.go
```diff
@@ -180,14 +180,15 @@ func setPipelineStepsOnPipeline(pipeline *model.Pipeline, pipelineItems []*stepb
...
-					Type:       model.StepType(step.Type),
+					Name:           step.Name,
+					UUID:           step.UUID,
+					PipelineID:     pipeline.ID,
+					PID:            pidSequence,
+					PPID:           item.Workflow.PID,
+					State:          model.StatusPending,
+					Failure:        step.Failure,
+					Type:           model.StepType(step.Type),
+					DependsOnNames: step.DependsOn,
```
6543: why not name it DependsOn like the rest of the codebase?!?

anbraten:
In my previous tests of showing a DAG in the UI, I have used the yaml configs (already exposed via api) to generate dags.

server/pipeline/items.go
lafriks: Because you need to resolve string DependsOn (step names) to IDs but they are only known when all records are inserted, model has both DopendsOnNames -> string array and DependsOn -> int64 array that will be populated after saving all workflows and steps to database from DependsOnNames and then we can update depends_on column with depends_on step ID's

pipeline/backend/types/step.go
lafriks: because it goes through compiler to backend types and then from backend types it's assigned to model type

lafriks:
> In my previous tests of showing a DAG in the UI, I have used the yaml configs (already exposed via api) to generate dags.

Yes but it has many problems:
* There can be matrix pipelines with matrix values in step names

```yaml
matrix:
  ENV:
    - staging
    - production

steps:
  - name: validate-config-${ENV}
    image: alpine
    commands:
      - echo "validate-config-${ENV}"
    depends_on: []

  - name: fetch-secrets-${ENV}
    image: alpine
    commands:
      - echo "fetch-secrets-${ENV}"
    depends_on: []

  - name: build-manifest-${ENV}
    image: alpine
    commands:
      - echo "build-manifest-${ENV}"
    depends_on:
      - validate-config-${ENV}
      - fetch-secrets-${ENV}
```

so in fronted you would need to expand them and guess workflow matrix axis

* Also similar problem would be with dependencies declared as yaml variables

Saving them to database after compiler is already done correct computations is imho best way to do this

server/pipeline/items.go
6543: can we not do it DependsOn -> string array and DependsOnIDs -> int64 array to be consistent with the rest?

pipeline/backend/types/step.go
6543: what?!

6543:
> because it goes through compiler to backend types and then from backend types it's assigned to model type

image

this is a open todo to be remove by #3967 ... so we should not introduce a new issue

pipeline/backend/types/step.go
6543: I'm looking into making https://github.com/woodpecker-ci/woodpecker/pull/3967 ready and move the compiler from before the queue to after it.

once that is done, i thin we can safely store that info in the frontend without needing to change the backend.

hope you can wait bit longer :sweat:

ecmchow:
May I know what is the status of this PR? Need to refactor after [#3967](https://github.com/woodpecker-ci/woodpecker/pull/3967) is merged/released?  

I've also have a PoC on adding a DAG in the UI, happy to contribute on the UI and this PR would allow proper/much simpler implementation on the front-end side
Screenshot 2026-05-29 at 12 29 26 pm

lafriks:
I was quite busy but will rework it in next two weeks

lafriks:
@6543 now only DAG resolved step names are stored not ID's that simplifies logic, is this better?

pipeline/backend/types/step.go
lafriks: I'm not quite sure what you mean without changing backend as I mentioned earlier compiler returns backend type:

https://github.com/woodpecker-ci/woodpecker/blob/0eedb4dee13864c34fc6d3473cffcc68fdb1a297/pipeline/frontend/yaml/compiler/compiler.go#L119

Issue #6132, opened 2026-02-18
CI_COMMIT_TAG is populated when executing pipeline for untagged commits


pztrn:
### Component

server

### Describe the bug

This unexpected behavior was introduced (probably) in https://github.com/woodpecker-ci/woodpecker/pull/5773 and completely illogical and broke everything on our side.

Our algo: build docker containers with short commit hash as image tag when building untagged code and use CI_COMMIT_TAG when building tagged code.

### Steps to reproduce

1. Default unmodified installation for woodpecker server and agent with Gitea integration.
2. Push something in some branch and do not tag it.
3. See things like:

```
Login Succeeded
+ docker build --pull -t some.address/of/image:${CI_COMMIT_TAG} -f someservice/Dockerfile .
ERROR: invalid tag "some.address/of/image:refs/heads/somebranch": invalid reference format
```

### Expected behavior

```
Login Succeeded
+ docker build --pull -t some.address/of/image:$${CI_COMMIT_SHA:0:10} -f someservice/Dockerfile .
```

### System Info

```shell
Woodpecker CI 3.13.0, was working on 3.11.0 and 3.12.0.
```

### Additional context

Pipeline things for untagged code deploy:

```yaml
    commands:
      - docker build --pull -t some.address/of/image:$${CI_COMMIT_SHA:0:10} -f someservice/Dockerfile .
    when:
      - event: deployment
        evaluate: 'CI_COMMIT_TAG == ""'
```

Pipeline things for tagged code deploy:

```yaml
    commands:
      - docker build --pull -t some.address/of/image:$${CI_COMMIT_TAG} -f someservice/Dockerfile .
    when:
      - event: deployment
        evaluate: 'CI_COMMIT_TAG != ""'
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

foi:
here is a workaround: `'not (CI_COMMIT_REF contains "refs/tags")'` and `'CI_COMMIT_REF contains "refs/tags"'`

PR #6142, opened 2026-02-21
Only set tag env on tags

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/6132

Just check for the ref only instead of the event.


xoxys:
Worth to add tests for this method?

lafriks (approved):

PR #6150, opened 2026-02-22
Support one-shot agent execution mode

cliffmccarthy:
This is an updated version of https://github.com/woodpecker-ci/woodpecker/pull/4019, based on the current main branch.  This supports the use case of having the agent itself be ephemeral, which may be useful in various situations, including use of the local backend.

See commit messages for details and revision history.  The first commit preserves attribution to @theanurin, who wrote most of this change; I just rebased it and made minor revisions.

### DIFF
--- a/cmd/agent/core/agent.go
+++ b/cmd/agent/core/agent.go
@@ -331,8 +331,13 @@ func run(ctx context.Context, c *cli.Command, backends []types.Backend) error {
 		})
 	}
 
-	log.Info().Str("version", version.String()).Str("backend", backendEngine.Name()).Str("platform", engInfo.Platform).Int("parallel workflows", maxWorkflows).Bool("single workflow", singleWorkflow).Msg(
-		"starting Woodpecker agent")
+	log.Info().
+		Str("version", version.String()).
+		Str("backend", backendEngine.Name()).
+		Str("platform", engInfo.Platform).
+		Int("parallel workflows", maxWorkflows).
+		Bool("single workflow", singleWorkflow).
+		Msg("starting Woodpecker agent")
 
 	return serviceWaitingGroup.Wait()
 }


cmd/agent/core/agent.go
```diff
@@ -311,13 +321,25 @@ func run(ctx context.Context, c *cli.Command, backends []types.Backend) error {
...
-		version.String(), backendEngine.Name(), engInfo.Platform, maxWorkflows)
+	if singleWorkflow {
+		log.Info().Msgf(
+			"starting Woodpecker agent with version '%s' and backend '%s' using platform '%s' running in single workflow mode",
+			version.String(), backendEngine.Name(), engInfo.Platform)
+	} else {
+		log.Info().Msgf(
+			"starting Woodpecker agent with version '%s' and backend '%s' using platform '%s' running up to %d pipelines in parallel",
+			version.String(), backendEngine.Name(), engInfo.Platform, maxWorkflows)
+	}
```
qwerty287: Would do something like this (I know it was different before):

```suggestion
	log.Info().Str("version", version.String()).Str("backend", backendEngine.Name()).Str("platform", engInfo.Platform).Int("parallel workflows", maxWorkflows).Bool("single workflow", singleWorkflow).Msg(
		"starting Woodpecker agent")
```

docs/docs/30-administration/10-configuration/30-agent.md
```diff
@@ -148,6 +148,15 @@ Configures the number of parallel workflows.
 
 ---
 
+### WOODPECKER_AGENT_SINGLE_WORKFLOW
+
+- Name: `WOODPECKER_AGENT_SINGLE_WORKFLOW`
+- Default: `false`
+
+Configures the agent to exit(shutdown) after first workflow. When configured, `WOODPECKER_MAX_WORKFLOWS` is forced to 1.
```
qwerty287:
```suggestion
Configures the agent to exit (shutdown) after first workflow. When configured, `WOODPECKER_MAX_WORKFLOWS` is forced to 1.
```

cliffmccarthy:
I built and ran the latest revisions and it looks good from my end.

qwerty287:
Thanks! Wouldn't it be better to directly use the `stopAgentFunc` func instead of `ctxCancel`?

cliffmccarthy:
> Wouldn't it be better to directly use the `stopAgentFunc` func instead of `ctxCancel`?

I'll defer to you on which way it should operate.  I tried changing it to call `stopAgentFunc`, and it appears to work, though I'm not sure I understand the intended flow with `stopAgentFunc`.  When it is called from the single workflow conditional, I think `stopAgentFunc` gets called twice, once from the direct call, and once from the original `defer`.  The definition of `stopAgentFunc` reassigns the symbol to a null function, but I don't think that changes the effect of the `defer`; I think the `defer` binds to the original definition of the function at the time it is deferred, based on my experiments.

qwerty287:
@6543 you introduced this `stopAgentFunc` logic. What do you say?

cliffmccarthy:
I studied `stopAgentFunc` some more and now I'm confident I don't understand it and the code associated with it.  I don't think `stopAgentFunc` needs to be a global, or even a symbol at all; I think it could be deferred as a three-line anonymous function with no argument (because the argument is always `nil`).  Where I get lost is in the globals `shutdownCancelFunc` and `shutdownCtx`.  These seem to be involved in cases where `run()` gets called multiple times.  The first call to `run()` defers a no-op `shutdownCancelFunc`; subsequent runs defer the `shutdownCancelFunc` that was set in the deferred operation of the *previous* call to `run()`.  There's some sort of state machine here with global state between calls; I'm not familiar enough with how these contexts are being used to assess what the purpose of this code flow is, and whether it should be involved in what we do when we decide to exit the program after running a single workflow.  It seems to me that calling `ctxCancel()` is probably the right way to initiate the shutdown, but I can't say that with confidence.

qwerty287:
@cliffmccarthy please have some patience. @6543 will check this out when he has time for it...

cliffmccarthy:
Thanks for following up!  There's no rush from my end; feel free to proceed with this PR and any revisions as you see fit whenever your schedule allows.

qwerty287:
@6543 pinging you again… :)

docs/docs/30-administration/10-configuration/30-agent.md
```diff
@@ -148,6 +148,15 @@ Configures the number of parallel workflows.
 
 ---
 
+### WOODPECKER_AGENT_SINGLE_WORKFLOW
```
qwerty287: We do not add the woodpecker prefix in the headers.

```suggestion
### AGENT_SINGLE_WORKFLOW
```

docs/docs/30-administration/10-configuration/30-agent.md
```diff
@@ -148,6 +148,15 @@ Configures the number of parallel workflows.
 
 ---
 
+### AGENT_SINGLE_WORKFLOW
+
+- Name: `WOODPECKER_AGENT_SINGLE_WORKFLOW`
+- Default: `false`
+
+Configures the agent to exit (shutdown) after first workflow. When configured, `WOODPECKER_MAX_WORKFLOWS` is forced to 1.
```
6543:
```suggestion
Configures the agent to exit (shutdown) after executing one workflow. When configured,
`WOODPECKER_MAX_WORKFLOWS` is forced to 1.

This one-shot mode is useful in ephemeral environments that are provisioned on demand
by external automation — for example, when an autoscaler spins up a dedicated machine. In these setups, the agent starts, executes exactly one workflow, and exits, allowing the environment to be cleanly torn down afterward.
```

PR #6155, opened 2026-02-24
Fix ignoring failures

qwerty287:
regression of #6119

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,7 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/yaronf/httpsign v0.4.1
 	github.com/zalando/go-keyring v0.2.6
-	gitlab.com/gitlab-org/api/client-go v1.41.0
+	gitlab.com/gitlab-org/api/client-go v1.42.0
 	go.uber.org/multierr v1.11.0
 	golang.org/x/crypto v0.48.0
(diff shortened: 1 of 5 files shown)


anbraten (commented):
could you add some tests?

kate-swp should be removed

qwerty287:
sorry. I don't get where those swap files come from. They are only kept as long as there are open, unsaved files… :see_no_evil:

server/pipeline/workflow_status_test.go
```diff
@@ -25,6 +25,56 @@ import (
 	store_mocks "go.woodpecker-ci.org/woodpecker/v3/server/store/mocks"
 )
 
+func TestWorkflowStatus(t *testing.T) {
+	tests := []struct {
+		s []*model.Step
+		e model.StatusValue
+	}{
+		{
+			s: []*model.Step{
```
anbraten, line 34: Thanks for adding some tests. Is this function used for other conversions like pending, running as well. If so it might be nice to have some tests for those as well

qwerty287: Added more tests. You're right, I forgot about this. Now this is only skipped for failing steps

6543 (approved):

6543:
deployed to our instance: https://ci.woodpecker-ci.org/repos/6106/pipeline/1859/4

Issue #6135, opened 2026-02-20
schema.json out of sync with actual schema


zmully:
### Component

server

### Describe the bug

The [schema.json](https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/yaml/linter/schema/schema.json) file appears to be out of sync with the actual workflow schema as defined in the pipeline go code. I'm not sure how this schema file is maintained, so I'm opening up an issue instead of a PR for it. 

I'm trying to use this schema.json to autogenerate Typescript interfaces for our internal tooling via json-schema-to-typescript in order to port our existing Drone pipelines to Woodpecker. 

I've already caught a couple of errors in the schema file during this process (this list is not exhaustive or complete by any means):
- [workflow_when_condition](https://github.com/woodpecker-ci/woodpecker/blob/bdf8fb5b801eacbb88d710e295a377cb5576e6fc/pipeline/frontend/yaml/linter/schema/schema.json#L165) is missing the `status` condition, defined in the go code [here](https://github.com/woodpecker-ci/woodpecker/blob/bdf8fb5b801eacbb88d710e295a377cb5576e6fc/pipeline/frontend/yaml/types/workflow.go#L24) which uses this shared constraint type [here](https://github.com/woodpecker-ci/woodpecker/blob/bdf8fb5b801eacbb88d710e295a377cb5576e6fc/pipeline/frontend/yaml/constraint/constraint.go#L45). The [documentation has this correct](https://woodpecker-ci.org/docs/usage/workflow-syntax#status) even though the schema.json does not.
- The [ref conditional](https://github.com/woodpecker-ci/woodpecker/blob/bdf8fb5b801eacbb88d710e295a377cb5576e6fc/pipeline/frontend/yaml/linter/schema/schema.json#L182) in the schema.json is defined as a string, while the go code has the constraint defined as a [list](https://github.com/woodpecker-ci/woodpecker/blob/bdf8fb5b801eacbb88d710e295a377cb5576e6fc/pipeline/frontend/yaml/constraint/constraint.go#L39) which resolves to this shared [list constraint](https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/yaml/constraint/list.go), so should support `include:` and `exclude:` syntax.
- Similar to the ref conditional, the status conditional in the schema.json is not of type list constraint.

[...]

PR #6156, opened 2026-02-24
Update pipeline config schema

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/6135

Everything else in the issue is correct, and nothing that has to be fixed.

Also changes the parsing for the `status` filter from the constraint list (with includes/excludes) to an allowlist only (as we only have two statuses it's quite useless to have an exclude and this was also not documented).

### DIFF
--- a/pipeline/frontend/yaml/constraint/constraint.go
+++ b/pipeline/frontend/yaml/constraint/constraint.go
@@ -98,7 +98,7 @@ func (when *When) IncludesStatusSuccess() bool {
 		return true
 	}
 	for _, c := range when.Constraints {
-		if len(c.Status) == 0 || slices.Contains(c.Status, statusFailure) {
+		if len(c.Status) == 0 || slices.Contains(c.Status, statusSuccess) {
 			return true
 		}
 	}
--- a/pipeline/frontend/yaml/constraint/constraint_test.go
+++ b/pipeline/frontend/yaml/constraint/constraint_test.go
@@ -32,8 +32,6 @@ func TestConstraintStatusSuccess(t *testing.T) {
 		{conf: "{status: [failure]}", want: false},
 		{conf: "{status: [success]}", want: true},
 		{conf: "{status: [failure, success]}", want: true},
-		{conf: "{status: {exclude: [success], include: [failure]}}", want: false},
-		{conf: "{status: {exclude: [failure], include: [success]}}", want: true},
 	}
 	for _, test := range testdata {
 		c := parseConstraints(t, test.conf)
--- a/pipeline/frontend/yaml/linter/schema/.woodpecker/test-when.yaml
+++ b/pipeline/frontend/yaml/linter/schema/.woodpecker/test-when.yaml
@@ -43,7 +43,10 @@ steps:
     commands:
       - echo "test"
     when:
-      ref: 'refs/tags/v**'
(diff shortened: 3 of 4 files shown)


pipeline/frontend/yaml/constraint/constraint.go
```diff
@@ -77,7 +77,7 @@ func (when *When) Match(metadata metadata.Metadata, global bool, env map[string]
 
 func (when *When) IncludesStatusFailure() bool {
 	for _, c := range when.Constraints {
-		if c.Status.Includes("failure") {
+		if slices.Contains(c.Status, "failure") {
```
6543: can we use a const `statusFailure` for failure ... when we already touch this code...

6543:
```
--- FAIL: TestSchema (0.39s)
    --- FAIL: TestSchema/When (0.02s)
        schema_test.go:135: 
            	Error Trace:	/woodpecker/src/github.com/woodpecker-ci/woodpecker/pipeline/frontend/yaml/linter/schema/schema_test.go:135
            	Error:      	Received unexpected error:
            	            	config not valid
            	Test:       	TestSchema/When
            	Messages:   	Validation failed: [steps: Must validate one and only one schema (oneOf) steps.when-ref: Must validate at least one schema (anyOf) steps.when-ref.when: Must validate one and only one schema (oneOf) steps.when-ref.when.ref: Must validate one and only one schema (oneOf) steps.when-ref.when.ref: steps.when-ref.when.ref must be one of the following: "push", "pull_request", "pull_request_closed", "pull_request_metadata", "tag", "deployment", "cron", "manual", "release"]
FAIL
coverage: 72.2% of statements
FAIL	go.woodpecker-ci.org/woodpecker/v3/pipeline/frontend/yaml/linter/schema	0.435s
```

cc @qwerty287

pipeline/frontend/yaml/constraint/constraint.go
```diff
@@ -42,7 +47,7 @@ type (
 		Platform List                        `yaml:"platform,omitempty"`
 		Branch   List                        `yaml:"branch,omitempty"`
 		Cron     List                        `yaml:"cron,omitempty"`
-		Status   List                        `yaml:"status,omitempty"`
+		Status   []string                    `yaml:"status,omitempty"`
```
anbraten, line 50: This looks like a breaking change.

qwerty287: No, because is wasn't documented before.

anbraten: Okay

pipeline/frontend/yaml/constraint/constraint.go
```diff
@@ -77,7 +82,7 @@ func (when *When) Match(metadata metadata.Metadata, global bool, env map[string]
 
 func (when *When) IncludesStatusFailure() bool {
 	for _, c := range when.Constraints {
```
anbraten, line 84: Looking at this status is handled independently from other when conditions? So

when:
  - branch: main
    status: failure
  - event: tag
    status: success

would not be possible, right?

6543: yes ... that's another issue for another day i would say ...

-> https://github.com/woodpecker-ci/woodpecker/issues/6161

PR #6158, opened 2026-02-24
Allow to cancel on failure

qwerty287:
allow to set `failure: cancel` which cancels the pipeline once the step fails. Still needs testing


qwerty287:
Tested it, seems working (except the unrelated status merging issue)

6543:
i'm not sur e if we should have a `cancle` that only affects the workflow and a `cancle-pipeline` ...

qwerty287:
As we don't have that feature at all right now I'd do that in another PR

PR #6162, opened 2026-02-26
Fix CLI exec with workflow matrix feature, where variables are not substituted.

6543:
`woodpecker-cli exec workflow-with-matrix.yaml` is broken, variables are not substituted.

this sets m.Workflow at the end resets the previous set axis data


6543:
source: https://codeberg.org/saces/woodpecker/commit/130d430d31a3e3170bfee71a86a995769a8b9c97

PR #6166, opened 2026-02-27
Refactor pipeline runtime code

6543:
next step in making code more structured and maintainable

- split pipeline exec into small functions
- fix edge-cases
- have less hidden code-flows
- fix possible [race condition](https://github.com/woodpecker-ci/woodpecker/pull/6272)

image

## Test Coverage Update: `pipeline/runtime`

**Overall: 80.5% → 98.3%** (+17.8pp)

| Function | `main` | `refactor` | Δ |
|---|--:|--:|--:|
| `GetShutdownCtx` | 0.0% | 100.0% | +100.0 |
| `WithDescription` | 0.0% | 100.0% | +100.0 |
| `makeLogger` | 75.0% | 100.0% | +25.0 |
| `Run` | 74.2% | 100.0% | +25.8 |
| `traceStep` | 84.6% | 100.0% | +15.4 |
| `completeStep` | — | 100.0% | ✱ new |
| `executeStep` | — | 100.0% | ✱ new |
| `logStages` | — | 100.0% | ✱ new |
| `runBlockingStep` | — | 100.0% | ✱ new |
| `runDetachedStep` | — | 85.7% | ✱ new |
| `runStage` | — | 100.0% | ✱ new |
| `shouldSkipStep` | — | 100.0% | ✱ new |
| `startStep` | — | 100.0% | ✱ new |
| `traceWorkflowSetupError` | — | 100.0% | ✱ new |
| `validateConfig` | — | 85.7% | ✱ new |

PR #6167, opened 2026-02-27
refactor: pipeline runtime add description and prevent race

6543:
simple changes taken from #6166 to reduce the nit noice

### DIFF
--- a/pipeline/frontend/yaml/types/container_test.go
+++ b/pipeline/frontend/yaml/types/container_test.go
@@ -17,7 +17,6 @@ package types
 import (
 	"testing"
 
-	"github.com/docker/docker/api/types/strslice"
 	"github.com/stretchr/testify/assert"
 	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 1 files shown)


pipeline/runtime/runtime.go
```diff
@@ -27,44 +28,59 @@ import (
...
 func (r *Runtime) MakeLogger() zerolog.Logger {
 	logCtx := log.With()
-	for key, val := range r.Description {
+	for key, val := range r.description {
 		logCtx = logCtx.Str(key, val)
 	}
 	return logCtx.Logger()
 }
+
+func (r *Runtime) getErr() error {
```
anbraten: I could use sth like this for such vars:

```
package main

import (
	"fmt"
	"sync"
)

type Protected[T any] struct {
	mu    sync.RWMutex
	value T
}

func NewProtected[T any](initial T) *Protected[T] {
	return &Protected[T]{value: initial}
}

func (p *Protected[T]) Get() T {
	p.mu.RLock()
	defer p.mu.RUnlock()
	return p.value
}

func (p *Protected[T]) Set(v T) {
	p.mu.Lock()
	defer p.mu.Unlock()
	p.value = v
}

// Update allows read-modify-write atomically
func (p *Protected[T]) Update(fn func(T) T) {
	p.mu.Lock()
	defer p.mu.Unlock()
	p.value = fn(p.value)
}
```

6543: I might as well look if new stuff is in current sync package ...

EDIT: not no such type exist ast

pipeline/runtime/shutdown.go
```diff
@@ -27,6 +27,9 @@ var (
 	shutdownCtxLock sync.Mutex
 )
 
+// GetShutdownCtx returns a context that is valid for shutdownTimeout after the
+// first call. It is used as a fallback cleanup context when the runner context
+// is already cancelled.
```
lafriks:
```suggestion
// is already canceled.
```

6543:
can/should i merge it?!?

lafriks (approved):

Issue #6164, opened 2026-02-26
☂️ Release Woodpecker v3.14


6543:
### Prerequisites

- [x] Prepare docs PR for new version and delete old versions (keep only the last three minor versions for the current major version)
  - [x] Run `make generate` locally to update the automatically generated CLI documentation
  - [x] Copy `docs/docs` to `docs/versioned_docs/version-<version>` and delete old versions
  - [x] Create `docs/versioned_sidebars/version-<version>-sidebars.json` and delete old ones
  - [x] Add new version to `docs/versions.json` and delete old versions
  - [x] Add new version to the version list in `docs/src/pages/versions.md`
- [x] Announce the release in the maintainer chat and ask for pending blockers

### Release

- [x] Test the latest container images to make sure they work as expected
- [x] Update `https://ci.woodpecker.org` to the latest version of `next` and verify that it works as expected
- [x] Merge documentation PR (shortly before release)
- [x] Merge the release PR to start the release pipeline

### Post-release

- [ ] Release the Helm Chart. If renovate has not created the upgrade PR already, manually trigger it from the Dependency Dashboard.
- [x] Announce release in relevant chats and on social media platforms
  - [x] Mastodon (check if already posted from the release pipeline)
  - [x] Bluesky (check if already posted from the release pipeline)
  - [x] Matrix

6543:
docs pull: #6163
release pull: https://github.com/woodpecker-ci/woodpecker/pull/6514

qwerty287:
Before the release https://github.com/woodpecker-ci/woodpecker/issues/6079#issuecomment-3938592218 must be fixed.

xoxys:
Another blocker for me: Its now impossible to see failed pipelines in the UI which is a huge UX issue as every failed pipeline that cancels remaining workflows is marked as skipped now. We should fix or revert the change that causes this.

qwerty287:
@xoxys there's #6176 that should fix it. I didn't have the time yet to test it though

hhamalai:
I suggest checking the https://github.com/woodpecker-ci/woodpecker/issues/6184 related to merged PR https://github.com/woodpecker-ci/woodpecker/pull/6056 before releasing, as this might unwanted effects on some backends (kubernetes).

6543:
fix binary release: https://github.com/woodpecker-ci/woodpecker/pull/6309

6543:
fixing cancle: #6320  (but blocked because of refactor)

[...]

PR #6176, opened 2026-03-01
Fix status merging with skipped pipelines

qwerty287:
Split the `skipped` status into `skipped` (skipped due to condition) and `canceled` (skipped due to cancellation but hasn't started).

Still needs testing.

### DIFF
--- a/cli/exec/exec.go
+++ b/cli/exec/exec.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 
 	"codeberg.org/6543/xyaml"
@@ -59,7 +60,7 @@ var Command = &cli.Command{
 	Usage:     "execute a local pipeline",
 	ArgsUsage: "[path/to/.woodpecker.yaml]",
 	Action:    run,
-	Flags:     utils.MergeSlices(flags, backend_docker.Flags, backend_kubernetes.Flags, backend_local.Flags),
+	Flags:     slices.Concat(flags, backend_docker.Flags, backend_kubernetes.Flags, backend_local.Flags),
 }
 
 var backends = []backend_types.Backend{
--- a/cmd/agent/core/run.go
+++ b/cmd/agent/core/run.go
@@ -17,6 +17,7 @@ package core
 import (
 	"context"
 	"os"
+	"slices"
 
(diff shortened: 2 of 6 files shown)


qwerty287:
Seems working:
successful:
grafik

canceled (with step skipped, but workflow and pipeline is correctly marked as canceled):
grafik

xoxys:
Needs frontend fix to get the failed icon in red again? Or is it intended to be gray in this case?

qwerty287:
It is canceled, then it should be gray.

Before #6011 the canceled pipelines always had a failed state and thus were red, but this was the actual bug.

xoxys:
OK then the handling is IMO still wrong. If a workflow fails and cancels other dependent workflows the entire pipeline should marked as failed, no?

xoxys (changes requested):
I still have questions 🙈

qwerty287:
Not sure if I understand you correctly: If one workflow fails, all the other ones that haven't been started will get status `skipped` (in `workflow_status.go`, line 51). Then the final pipeline should be `failed`. The ones that already run will finish running, and the final status will get failed as well.

If you mean something different, it's also possible that I did a mistake. This full status system and how finishing workflows is handled is really complex and ununderstandable. Might need a big refactoring.

xoxys:
My initial issue I posted is that exactly this case you decribed: One workflow failed, some other workflows were cancelled due to the failed one -> the entire pipeline should marked as failed. However this does not work anymore **but was working before**

grafik

In the secreenshot you can see that `security-scan` failed and `docs` was cancelled due to this failure but the entire pipeline is now marked as skipped and should still be marked as failed.

grafik

The second screenshot shows the exact same pipeline a few weeks earlier (with an older wp version) where it is working correctly.

At least that what I tried to describe in chat and https://github.com/woodpecker-ci/woodpecker/issues/6164#issuecomment-3980835942

qwerty287:
Yes, but isn't this what I fixed here?

Working as well for me (on-fail shouldn't run on fails… this was my first test with `runs_on`, thus the name):
grafik

xoxys:
Ok nice. Then I just got confused by your screenshots :D

qwerty287:
Sorry, I actually haven't tested that before.

I thought your point was about `runs_on: [failure]`… But the fix is the same for both

PR #6183, opened 2026-03-02
Fix `when.status` filter evaluation and add workflow-level support

### LINKED ISSUE #6161: Status filter in when condition list entries not evaluated independently
## Bug

Status filters in `when` condition **lists** are not handled independently per entry.

### Expected behavior

Each `when` list entry should be evaluated as a standalone condition. The following should trigger the step on `failure` for the `main` branch push, **and** on `success` for a tag event:
```yaml
when:
  - branch: main
    status: failure
  - event: tag
    status: success
```

### Actual behavior

The status constraints across all list entries are not isolated — the `IncludesStatusSuccess` / `IncludesStatusFailure` checks in `constraint.go` iterate over `when.Constraints` as a whole rather than scoping each status check to its own entry, making the above config impossible to use correctly.

### Root cause

In [`constraint.go#L78-L101`](https://github.com/woodpecker-ci/woodpecker/blob/69c934873d2de720fbd006e6c2445efb7cb6125c/pipeline/frontend/yaml/constraint/constraint.go#L78-L101), `IncludesStatusFailure` and `IncludesStatusSuccess` loop over all constraints rather than respecting the per-entry scope.

 _Originally posted by @anbraten in [#6156](https://github.com/woodpecker-ci/woodpecker/pull/6156/changes#r2855126032)_

https://github.com/woodpecker-ci/woodpecker/blob/69c934873d2de720fbd006e6c2445efb7cb6125c/pipeline/frontend/yaml/constraint/constraint.go#L78-L101

qwerty287:
fix https://github.com/woodpecker-ci/woodpecker/issues/6161

And also allow `when.status` at workflow level. `runs_on` is deprecated then.


server/pipeline/stepbuilder/step_builder.go
6543:
```suggestion
		RunsOn:    parsed.RunsOn, // Depricated. TODO remove in next major.
```

6543:
@qwerty287 hope you don't mind ... i refactored two things :)

-> we often have the usecase of filter unique slices so now there is a generic helper
-> use the official golang deprecation syntax so IDE shows it too

qwerty287:
It's ok for me, but I don't think we need that `UniqSlice`… I mean it won't break if someone sets `status: [failure, failure]`. Wouldn't add extra code for this, but I won't block this of course.

6543:
I know we have more parts in our codebase where we manually use a map to filter uniq stuff and we can with this slowly adopt it more and more.

I also plan to upstream into golang an slices.Filter or slices.Uniq depending on what is already worked on ...

6543 (approved):

pipeline/frontend/yaml/linter/schema/schema.json
```diff
@@ -47,6 +47,7 @@
     },
     "runs_on": {
       "type": "array",
+      "description": "Deprecation: use `when.status` instead. Read more: https://woodpecker-ci.org/docs/usage/workflows#flow-control",
```
6543:
```suggestion
      "description": "Deprecated: use `when.status` instead. Read more: https://woodpecker-ci.org/docs/usage/workflows#flow-control",
```

6543:
@qwerty287 finisched it ... could you look over the new/changed parts and lgtm them ,,, or give feedback :) ?

pipeline/frontend/yaml/linter/linter.go
```diff
@@ -301,14 +301,23 @@ func (l *Linter) lintSchema(config *WorkflowConfig) error {
 	return linterErr
 }
 
-func (l *Linter) lintDeprecations(config *WorkflowConfig) (err error) {
+func (l *Linter) lintDeprecations(config *WorkflowConfig) error {
 	parsed := new(types.Workflow)
-	err = xyaml.Unmarshal([]byte(config.RawConfig), parsed)
-	if err != nil {
-		return err
+	err := xyaml.Unmarshal([]byte(config.RawConfig), parsed)
```
qwerty287, line 306: Why did you change this?
If the unmarshalling fails it should directly return I think

6543: uh right :swan:

qwerty287:
Seems fine for me now, besides that I still don't like the `UniqSlice`.

6543:
in this case as it's your pull I'll just move that out and then merge it

6543:
the current queue impl. should tollerate duplicates. but still i dont like that impl. details of the parser leak out to other parts.

as it can be done later I move this into another pull/issue/discusion for later

6543:
deployed: https://ci.woodpecker-ci.org/repos/6106/pipeline/1877

xoxys:
@woodpecker-ci/maintainers this is a breaking change. Working CI pipelines now fail:

image

Please mark it as such and remember to release a major version bump next if this was not a mistake.

6543:
it schould be written non breaking :/

so this is a bug of this implementation

qwerty287:
Oh sorry, this shouldn't fail indeed, only warn. Will check it out later

PR #6186, opened 2026-03-02
Prevent leaking goroutines on canceled steps

hhamalai:
I've been debugging case where Kubernetes backend agents occasionally are OOMKilled. During the investigation noticed an increasing amount of goroutines on the agent processes which turned out to be multiple goroutines hanging on:
```
runtime.gopark(proc.go:461)
runtime.chanrecv(chan.go:667)
runtime.chanrecv1(chan.go:509)
go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/kubernetes.(*kube).WaitStep(kubernetes.go:325)
go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime.(*Runtime).exec(executor.go:261)
go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime.(*Runtime).execAll.func1.1(executor.go:174)
go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime.(*Runtime).execAll.func1.2(executor.go:200)
runtime.goexit(asm_arm64.s:1268)
go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime.(*Runtime).execAll.func1(executor.go:199)
```

My analysis is that on cancel nothing fires the `finished` channel, as pod is deleted i.e. not updated (there's a registed handler for pod update events). There was already a comment about adding cancellation handler for ctx.Done, so I think this is the proper way (instead of adding event handler for pod deletion).

### DIFF
--- a/docs/docs/20-usage/20-workflow-syntax.md
+++ b/docs/docs/20-usage/20-workflow-syntax.md
@@ -350,6 +350,18 @@ There are use cases for executing steps on failure, such as sending notification
 +      - status: [ success, failure ]
 ```
 
+The filter is aware of the other filters. If you want to run on failures if the event is `tag`, but if it's a `pull_request`, run it on both success and failure:
+
+```diff
+ when:
++  - event: tag
++    status: [ failure ]
++  - event: pull_request
++    status: [ success, failure ]
+```
+
+If there's no matching filter at all or all matching filters don't have set `status`, it will use the default, which means it runs on success only. In the example above this will happen if the event is neither `tag` nor `pull_request`.
+
 #### `platform`
 
 :::note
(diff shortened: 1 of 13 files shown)


6543:
```
golangci-lint run
pipeline/backend/kubernetes/kubernetes_test.go:235:17: use of `context.WithCancel` forbidden by pattern `context\.WithCancel$` (forbidigo)
	ctx, cancel := context.WithCancel(context.Background())
	               ^
pipeline/backend/kubernetes/kubernetes_test.go:284:19: use of `context.WithCancel` forbidden by pattern `context\.WithCancel$` (forbidigo)
			ctx, cancel := context.WithCancel(context.Background())
			               ^
2 issues:
```

you can use context.WithCancelCause instead

PR #6196, opened 2026-03-05
Fix message on gitlab tag event

### LINKED ISSUE #6100: Event tag don't have ci commit message
When I push tag to git and a pipeline start to run with event tag I see on the build name '-' instead the ci commit message I have for the tag

Why?

qwerty287:
fixes #6100


qwerty287:
Fixes the base issue, i.e. the behaviour is then similar to github. But I agree to the issue opener that tag pipelines should use the tag title, but I wouldn't do that here because github also uses the head commit message.

Issue #2976, opened 2023-12-20
Add pagination on repo pipelines view


runephilosof-karnovgroup:
### Clear and concise description of the problem

It is not possible to see list of older pipelines for a repo

### Suggested solution

Add pagination on https://ci.woodpecker-ci.org/repos/3780

### Alternative

Add a load more button.
Although, I prefer pagination, because then you can share a link to the page and you will see the same view.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

anbraten:
I've create a `Load more` button component some time ago, could be worth using it in general: #2977

yekibud:
This would be _very_ helpful.  I frequently find myself wondering what happened to a pipeline, only to find it's no longer available to inspect in the UI or CLI.  I'll add it to my personal contributor list to look into, but unlikely to get to it for a while.

PR #6200, opened 2026-03-05
Add "Load more" pagination to pipeline list

savv:
Addresses #2976

Demo of what it looks like:

image

### DIFF
--- a/web/src/components/repo/pipeline/PipelineList.vue
+++ b/web/src/components/repo/pipeline/PipelineList.vue
@@ -16,7 +16,7 @@
       <span class="text-wp-text-100">{{ $t('repo.pipeline.no_pipelines') }}</span>
     </Panel>
     <div v-if="hasMore && !loading" class="flex justify-center">
-      <Button :text="$t('repo.pipeline.load_more')" @click="$emit('load-more')" />
+      <Button :text="$t('repo.pipeline.load_more')" @click="$emit('loadMore')" />
     </div>
   </div>
 </template>
@@ -35,6 +35,6 @@ defineProps<{
 }>();
 
 defineEmits<{
-  (event: 'load-more'): void;
+  (event: 'loadMore'): void;
 }>();
 </script>


qwerty287:
Thanks!

We are using infinite scroll for pagination though:

https://github.com/woodpecker-ci/woodpecker/blob/739114bf3a177dc0e266badf40096aa0cfc041f3/web/src/compositions/usePaginate.ts#L25

Would be nice to use this instead of a load more button.

savv:
@qwerty287 I gave this a try, but understand this logic a bit less well (which is why I had opted for something simpler). What do you think?

qwerty287:
Thanks, yes. The issue is that we load the pipelines centrally in `web/src/views/repo/RepoWrapper.vue` and just pass it using `provide`/`require` to the pipelines list. I don't have a good solution how we can easily implement that with pagination. If you have an idea - just tell, otherwise I'll try to check this out more detailed again later

savv:
> Thanks, yes. The issue is that we load the pipelines centrally in `web/src/views/repo/RepoWrapper.vue` and just pass it using `provide`/`require` to the pipelines list. I don't have a good solution how we can easily implement that with pagination. If you have an idea - just tell, otherwise I'll try to check this out more detailed again later

@qwerty287 Sorry, I can't find a cleaner way either and I'm not even 100% sure I understand everything. If you have a proposal how data should flow I could give it a crack.

I should also say that Woodpecker has been amazing for us so far (so amazing, that I was initially shocked that this feature is missing). With the aggressive caching that we were able to do on our own CI instance, we got builds down from 5+ min to ~20 seconds. Thanks for maintaining!

qwerty287:
Ok I looked at the code again, and I actually don't think it's easy to make this working right now in a clean way.

What should be done better is esp. how the filtering works. Because if you right now view pipelines for a single branch/PR, it will take the ones the frontend has loaded already and filters it in the frontend. This is really bad style. It should use server-side filtering. For branches, this would be available. For PRs not, and it isn't possibly to implement that cleanly because right now we compare the PRs based on the ref, but every forge uses a different ref for PRs. After #4626 it should be easier to implement the server-side filtering for pull requests.

When we have server-side filtering, we can move the pipeline loading from the repo wrapper to the individual list views, and implement the pagination there.

This would be the clean solution I'd prefer, but it likely won't be doable in the next time.

@woodpecker-ci/maintainers What do you think about this?

xoxys:
I also tried to implement the infinite scroll without success. Having at leadt one working solution for pagination is better than nothing. Would vote to add it if someone as a better approach this can be replaced later.

6543:
agree, refactoring and imprufing can be done later on too :)

qwerty287:
Then please add a TODO comment so we don't forget about it

savv:
> Then please add a TODO comment so we don't forget about it

I added a TODO in RepoPipelines

6543:
could you please the linter?
```
/woodpecker/src/github.com/woodpecker-ci/woodpecker/web/src/components/repo/pipeline/PipelineList.vue
  19:67  error  Custom event name 'load-more' must be camelCase  vue/custom-event-name-casing
```

savv:
Done. I tested this again and it works on our local CI.

PR #6203, opened 2026-03-06
Bitbucket DC: resolve annotated tag SHA to commit SHA before posting build status

### LINKED ISSUE #6202: Bitbucket Data Center: annotated tags don't report build status to the correct commit
### Component

server

### Describe the bug

When a repository uses annotated tags (as opposed to lightweight tags), Bitbucket Data Center's push webhook sends the tag object SHA as ToHash, not the SHA of the commit the tag is pointing to. Woodpecker uses this SHA to post build statuses back to Bitbucket, which fails silently, because the tag object SHA is not a commit.

### Steps to reproduce

Create a pipeline triggered on tag events
Push an annotated tag (git tag -a v1.0 -m "release" && git push --tags)
Pipeline runs but no build status appears on the commit in Bitbucket
Expected behavior:
Build status is displayed on Bitbucket.

### Expected behavior

Build status is displayed on Bitbucket.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-3786077c07"}
```

### Additional context

_No response_

### Validations

dccdis:
Fixes #6202

### DIFF
--- a/server/forge/bitbucketdatacenter/bitbucketdatacenter.go
+++ b/server/forge/bitbucketdatacenter/bitbucketdatacenter.go
@@ -580,7 +580,7 @@ func (c *client) updatePipelineFromCommits(ctx context.Context, u *model.User, r
 		return nil, fmt.Errorf("unable to read commit: %w", err)
 	}
 
-	// Fix for https://github.com/woodpecker-ci/woodpecker/issues/6202:
+	// In Bitbucket Data Center, when using annotated tags, the webhook's ToHash is the tag object SHA, not the actual commit SHA.
 	// Update p.Commit so that build statuses are posted to the correct commit SHA.
 	if p.Event == model.EventTag && commit.ID != "" && commit.ID != p.Commit {
 		p.Commit = commit.ID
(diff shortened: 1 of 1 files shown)


qwerty287:
Thanks! The sha is also cloneable? Because cloning uses the sha by default.

dccdis:
As far as I can tell the bug only affects status reporting. Otherwise the pipeline appears to work perfectly (code clones, runs, passes/fails) both before and after applying the proposed fix.

server/forge/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -579,6 +579,14 @@ func (c *client) updatePipelineFromCommits(ctx context.Context, u *model.User, r
 	if err != nil {
 		return nil, fmt.Errorf("unable to read commit: %w", err)
 	}
+
+	// Fix for https://github.com/woodpecker-ci/woodpecker/issues/6202:
```
lafriks: Maybe add not reference to GitHub but as a description that bitbucket in annotated tags sends tag sha not commit tag so that is clear when reading code why this is here

qwerty287 (approved):

Issue #6079, opened 2026-02-06
Debian package broken


Signum:
### Component

server

### Describe the bug

The package does not create the woodpecker user, woodpecker group or woodpecker home directory at /var/lib/woodpecker.

The package links to the Docs URL https://woodpecker-ci.org/docs/administration/server-config which is 404. This URL is both mentioned in the metadata of the package and in woodpecker-server.env.example

### Steps to reproduce

- Get the DEB package for version 3.13.0.
- Install it.
- systemctl enable --now woodpecker-server
- journalctl -fu woodpecker-server

### Expected behavior

The DEB package provides a working installation.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.13.0"}
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

qwerty287:
This isn't finished yet. The user and group are still not auto-created

qwerty287:
Now it's completely broken. My PR was actually not ready to merge.

https://ci.woodpecker-ci.org/repos/3780/pipeline/31773/15#L48

PR #6205, opened 2026-03-06
Fix license headers

qwerty287:

### DIFF
--- a/web/src/assets/locales/pt.json
+++ b/web/src/assets/locales/pt.json
@@ -251,7 +251,8 @@
                 "declined": "recusado",
                 "error": "erro",
                 "failure": "falha",
-                "killed": "finalizado"
+                "killed": "finalizado",
+                "canceled": "cancelada"
             },
             "we_got_some_errors": "Oh não, ocorreu um erro!",
             "step_not_started": "Essa etapa ainda não foi iniciada.",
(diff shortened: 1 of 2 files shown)


lafriks:
I would prefer if we would use just SPDX identifier:

```
// Copyright 2026 Woodpecker CI Authors.
// SPDX-License-Identifier: Apache-2.0
```

qwerty287:
We can do from my side, but first I'd like to get the linter working.

It's just ignoring that there's a file not having a header… https://ci.woodpecker-ci.org/repos/3780/pipeline/32099/3#L7

It works if I run `addlicense` locally. Does anyone has an idea what's the issue with it?

cmd/cli/app.go
```diff
@@ -1,17 +1,3 @@
-// Copyright 2021 Woodpecker Authors
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
```
6543, line 3: ?

EDIT: ah it's the test that should fail the ci but does not atm...

qwerty287: Yes. And I don't know why... addlicense fails locally for me.

qwerty287:
Finally managed to fix this. https://ci.woodpecker-ci.org/repos/3780/pipeline/32376/3 fails as it should.
Reason: bash doesn't support ** wildcards by default, needs to be enabled.

qwerty287:
@woodpecker-ci/maintainers could I get a review here? The lint step is fixed.

6543:
nice work!!!

6543:
```
VERSION_NUMBER=0.0.0 nfpm package --config ./nfpm/agent.yaml --target dist --packager deb
glob failed: : no matching files
make: *** [Makefile:340: bundle-agent] Error 1
```

?!? related?

qwerty287:
No: https://github.com/woodpecker-ci/woodpecker/issues/6079#issuecomment-3938592218

6543:
oh no :/ ... ok

PR #6209, opened 2026-03-07
fix(db): replace lib/pq with pgx for connection pooler compatibility

### LINKED ISSUE #6097: (pgbouncer issue) Error fetching repository list. pq: bind message supplies 4 parameters, but prepared statement "" requires 1
### Component

server

### Describe the bug

```
curl 'https://ci.redacted.dev/api/user/repos?' \
  -H 'accept: */*' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'cache-control: no-cache' \
  -b '...' \
  -H 'dnt: 1' \
  -H 'pragma: no-cache' \
  -H 'priority: u=1, i' \
  -H 'referer: https://ci.redacted.dev/repos' \
  -H 'sec-ch-ua: "Not(A:Brand";v="8", "Chromium";v="144", "Google Chrome";v="144"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: same-origin' \
  -H 'sec-gpc: 1' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/144.0.0.0 Safari/537.36'
```

Error fetching repository list. pq: bind message supplies 4 parameters, but prepared statement "" requires 1

### Steps to reproduce

try to load the UI

### Expected behavior

API call does not fail with this error

### System Info

```shell
{
  "latest": "v3.13.0",
  "next": "next-b4708c895e",
  "rc": "v3.13.0"
}
```

### Additional context

_No response_

### Validations

brandonros:
## Summary

Fixes #6097

Replace `github.com/lib/pq` with `github.com/jackc/pgx/v5/stdlib` as the PostgreSQL driver.

`lib/pq` uses PostgreSQL's **extended query protocol** with unnamed prepared statements for all parameterized queries. This is incompatible with connection poolers (pgBouncer, pgcat) in **transaction pooling mode** — the pooler can route different clients' Parse/Bind messages to the same server connection, causing:

```
pq: bind message supplies 4 parameters, but prepared statement "" requires 1
```

`pgx` supports **simple query protocol** via the DSN parameter `default_query_exec_mode=simple_protocol`, which avoids prepared statements entirely and is fully compatible with transaction pooling.

## Changes

- **`server/store/datastore/init.go`** / **`init_cgo.go`** — Swap `lib/pq` import for `pgx/v5/stdlib`
- **`server/store/datastore/engine.go`** — Map `"postgres"` driver name to `"pgx"` (pgx/stdlib registers under `"pgx"`)
- **`server/store/datastore/migration/migration_test.go`** — Update imports and driver names
- **`go.mod`** / **`go.sum`** — Add `pgx/v5`, remove direct `lib/pq` dependency

## Why pgx

- Drop-in replacement via `pgx/stdlib` (implements `database/sql` interface — XORM works without changes)
- Supports `default_query_exec_mode=simple_protocol` to avoid prepared statements
- `lib/pq` is in maintenance mode; its author [recommends migrating to pgx](https://github.com/lib/pq#status)
- Other Go projects (Gitea, etc.) have made the same migration for the same reason

## Impact

- **Users connecting directly to PostgreSQL (no pooler):** no behavior change
- **Users with session pooling mode:** no behavior change
- **Users with transaction pooling mode (pgBouncer, pgcat):** append `?default_query_exec_mode=simple_protocol` to their `WOODPECKER_DATABASE_DATASOURCE` to resolve the error

## Performance note

Simple protocol sends parameters as text (not binary) and cannot reuse prepared statement query plans. For Woodpecker's workload (simple CRUD queries), the difference is negligible.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,11 +33,11 @@ require (
 	github.com/google/go-github/v84 v84.0.0
 	github.com/hashicorp/go-hclog v1.6.3
 	github.com/hashicorp/go-plugin v1.7.0
+	github.com/jackc/pgx/v5 v5.8.0
 	github.com/jellydator/ttlcache/v3 v3.4.0
 	github.com/joho/godotenv v1.5.1
 	github.com/json-iterator/go v1.1.12
 	github.com/kinbiko/jsonassert v1.2.0
-	github.com/lib/pq v1.11.2
 	github.com/mattn/go-sqlite3 v1.14.34
 	github.com/migueleliasweb/go-github-mock v1.5.0
 	github.com/moby/term v0.5.2
@@ -142,6 +142,9 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.8 // indirect
 	github.com/hashicorp/go-version v1.7.0 // indirect
 	github.com/hashicorp/yamux v0.1.2 // indirect
+	github.com/jackc/pgpassfile v1.0.0 // indirect
+	github.com/jackc/pgservicefile v0.0.0-20240606120523-5a60cdf6a761 // indirect
+	github.com/jackc/puddle/v2 v2.2.2 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/julienschmidt/httprouter v1.3.0 // indirect
 	github.com/klauspost/cpuid/v2 v2.3.0 // indirect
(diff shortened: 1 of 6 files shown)


qwerty287:
@xoxys what are the reason why you're against this? I don't see any downsides right now… (Besides that the tests are failing)

xoxys:
My main point is blindly replacing a fundamental library to fix a kinda very rare corner case without knowing the side-effects/limitations of another lib. If we are confident enough please go on.

However, this can also be fixed by the users of the pooler thingy as discussed in the issue and we dont even know if the new library really fixes the issue (do we?), this is only clamed by the AI, the same AI that is talking bullshit about the current lib is unmaintained and recommends migration 🤷‍♂️

brandonros:
fair. I can close for now, hopefully somebody else will learn in a quicker fashion a solution to the problem I faced. thanks for checking it out.

PR #6228, opened 2026-03-10
Enable crons if created via CLI

qwerty287:
Crons were not enabled if created via CLI. fixes https://github.com/woodpecker-ci/woodpecker/discussions/6218#discussioncomment-16070162

Also improve the error message if a cron with the same name exists already. Previously a 500 was returned, now a 409 Conflict.

### DIFF
--- a/server/api/cron.go
+++ b/server/api/cron.go
@@ -155,7 +155,7 @@ func PostCron(c *gin.Context) {
 	}
 
 	if err := _store.CronCreate(cron); err != nil {
-		if errors.Is(types.ErrUniqueExists, err) {
+		if errors.Is(err, types.ErrUniqueExists) {
 			c.String(http.StatusConflict, "cron with this exists for this repo already")
 		} else {
 			c.String(http.StatusInternalServerError, "Error inserting cron %q. %s", in.Name, err)


server/api/cron.go
```diff
@@ -152,6 +152,16 @@ func PostCron(c *gin.Context) {
 		}
 	}
 
+	nameExists, err := _store.CronExists(repo, in.Name)
```
lafriks: Can't we just check returned error below from CronCreate and return correct error there?

qwerty287: Can try, but I would prefer the explicit way

qwerty287: @woodpecker-ci/maintainers other opinions?

lafriks: This check is not prone to race condition but of course in this case this would be rear so I would not be blocking also this way

qwerty287: @lafriks changed it. It's not that easy though because there's no specific error type and the message differs between the db engines.

server/store/datastore/cron.go
```diff
@@ -25,6 +28,9 @@ func (s storage) CronCreate(cron *model.Cron) error {
 		return err
 	}
 	_, err := s.engine.Insert(cron)
+	if err != nil && (strings.HasPrefix(err.Error(), "UNIQUE constraint failed") || strings.HasPrefix(err.Error(), "pq: duplicate key value violates unique constraint") || strings.Contains(err.Error(), "Duplicate entry")) {
+		return errors.New("cron with this name exists already for this repo")
```
6543: i think similar to not exist error we could add an error already exist ...
add tests to all supported databases and warap all inserts with that so we can use errors.Is and dont need to relay on string checks outside of storage module

qwerty287: Like [3a7ee59](https://github.com/woodpecker-ci/woodpecker/pull/6228/commits/3a7ee593e6bf69d6761ba315a88b36a3ff89e122)? If not feel free to push to my branch

server/api/cron.go
```diff
@@ -153,7 +155,11 @@ func PostCron(c *gin.Context) {
 	}
 
 	if err := _store.CronCreate(cron); err != nil {
-		c.String(http.StatusInternalServerError, "Error inserting cron %q. %s", in.Name, err)
+		if errors.Is(types.UniqueExists, err) {
```
6543:
```suggestion
		if errors.Is(types.ErrUniqueExists, err) {
```

server/store/datastore/cron.go
```diff
@@ -15,16 +15,22 @@
...
+	"go.woodpecker-ci.org/woodpecker/v3/server/store/types"
 )
 
 func (s storage) CronCreate(cron *model.Cron) error {
 	if err := cron.Validate(); err != nil {
 		return err
 	}
 	_, err := s.engine.Insert(cron)
+	if err != nil && (strings.HasPrefix(err.Error(), "UNIQUE constraint failed") || strings.HasPrefix(err.Error(), "pq: duplicate key value violates unique constraint") || strings.Contains(err.Error(), "Duplicate entry")) {
+		return types.UniqueExists
```
6543:
```suggestion
		return types.ErrUniqueExists
```

server/store/datastore/cron_test.go
```diff
@@ -33,7 +34,7 @@ func TestCronCreate(t *testing.T) {
 	assert.NotEqualValues(t, 0, cron1.ID)
 
 	// cannot insert cron job with same repoID and title
-	assert.Error(t, store.CronCreate(cron1))
+	assert.ErrorIs(t, types.UniqueExists, store.CronCreate(cron1))
```
6543:
```suggestion
	assert.ErrorIs(t, types.ErrUniqueExists, store.CronCreate(cron1))
```

server/store/types/errors.go
```diff
@@ -14,6 +14,11 @@
...
 
-import "database/sql"
+import (
+	"database/sql"
+	"errors"
+)
 
 var RecordNotExist = sql.ErrNoRows
+
+var UniqueExists = errors.New("unique constraint failed")
```
6543:
```suggestion
var ErrUniqueExists = errors.New("unique constraint failed")
```

6543:
I have a followup that does `wrapInsert` like wrapGet etc ... but it will colide with your sol currently .... will ack this and then streamline it in a new pull :rock:

server/api/cron.go
```diff
@@ -153,7 +155,11 @@ func PostCron(c *gin.Context) {
 	}
 
 	if err := _store.CronCreate(cron); err != nil {
-		c.String(http.StatusInternalServerError, "Error inserting cron %q. %s", in.Name, err)
+		if errors.Is(types.ErrUniqueExists, err) {
```
6543:
```suggestion
		if errors.Is(err, types.ErrUniqueExists) {
```

6543:
@qwerty287 i did https://github.com/woodpecker-ci/woodpecker/pull/6259

PR #6250, opened 2026-03-16
Add server config for maximum log lines shown in web UI

confusedsushi:
I frequently find it annoying to not be able to see the whole pipeline log. I know that there is already an issue which requests to implement lazy loading for the log, #2653. That would be indeed the best solution. However that might takes a while.

This PR instead makes it possible to configure the maximum number of lines to show in a pipeline log. This allows to tweak the lines to show to whatever fits best for any individual instance.


6543 (approved):
lgtm as workaround but we should fix it via https://github.com/woodpecker-ci/woodpecker/issues/2653 long term

PR #6251, opened 2026-03-16
Allow disabling isolated home directory for local agents

confusedsushi:
Make it possible to disable the isolated home created by the local agent. To do so an environment variable `WOODPECKER_BACKEND_LOCAL_ISOLATED_HOME` is introduced. Set it to `false` to make the local agent not set the environment variables `HOME` and `USERPROFILE`.

This is needed for a Windows local agent. For a Windows environment these two variables are far from being enough, there are multiple other variables, also there is the registry and other stuff. Also for me, the whole point of a local agent is to use the local environment, what ever it is.

### DIFF
--- a/pipeline/backend/local/flags.go
+++ b/pipeline/backend/local/flags.go
@@ -31,7 +31,7 @@ var Flags = []cli.Flag{
 	&cli.BoolFlag{
 		Sources: cli.EnvVars("WOODPECKER_BACKEND_LOCAL_ISOLATED_HOME"),
 		Name:    "backend-local-isolated-home",
-		Usage:   "set HOME, USERPROFILE and other variables to an isolated directory",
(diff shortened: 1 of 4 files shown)


pipeline/backend/local/flags.go
```diff
@@ -28,4 +28,10 @@ var Flags = []cli.Flag{
 		DefaultText: "system temporary directory",
 		Value:       os.TempDir(),
 	},
+	&cli.BoolFlag{
+		Sources: cli.EnvVars("WOODPECKER_BACKEND_LOCAL_ISOLATED_HOME"),
+		Name:    "backend-local-isolated-home",
+		Usage:   "set HOME, USERPROFILE and other variables to an isolated directory",
```
6543:
```suggestion
		Usage:   "set HOME, USERPROFILE and other variables to an isolated directory, if false we ignore netrc",
```

6543 (approved):
lgtm from my side but i want another one to look at it too

PR #6252, opened 2026-03-16
Add external secret extension implementation

### LINKED ISSUE #929: Hashicorp vault as secret storage integration

Hi there, that would be awesome if secrets would be pullable from Hashicorp Vault, by integration like [this](https://docs.gitlab.com/ee/ci/examples/authenticating-with-hashicorp-vault/)

### Suggested solution

by this https://docs.gitlab.com/ee/ci/secrets/index.html#configure-your-vault-server
and that https://docs.gitlab.com/ee/ci/secrets/index.html#use-vault-secrets-in-a-ci-job

### Alternative

wrapping stuff in bash, that's sad and nasty, because woodpecker does not treat it as concealed string, and secrets gets leaked
[import this onto woodpecker plugin system](https://github.com/drone/drone-vault)

### Additional context

https://www.hashicorp.com/resources/securing-your-ci-pipeline-with-vault

### Validations

mofr93:
## Add external secret extension support

Adds support for fetching secrets from an external HTTP service, mirroring the existing config service extension pattern (`WOODPECKER_CONFIG_EXTENSION_ENDPOINT`).

### New environment variables

- `WOODPECKER_SECRET_EXTENSION_ENDPOINT` - URL of the external secret service
- `WOODPECKER_SECRET_EXTENSION_NETRC` - Send netrc if needed
### Per-repo override

Repos can set their own `secret_extension_endpoint`

### How it works

The external service receives a signed POST request with `{repo, pipeline, netrc}` and returns `[]*model.Secret`. Requests use the same ed25519-signed HTTP client and `WOODPECKER_EXTENSIONS_ALLOWED_HOSTS` allowlist as the config extension.

- **200 OK** — use returned secrets
- **204 No Content** — no external secrets (DB secrets still used if not exclusive)

When not exclusive, DB and external secrets are merged, with external secrets overriding DB secrets by name.

closes #929

### DIFF
--- a/server/services/manager.go
+++ b/server/services/manager.go
@@ -87,7 +87,7 @@ func NewManager(c *cli.Command, store store.Store, setupForge SetupForge) (Manag
 		signaturePrivateKey: signaturePrivateKey,
 		signaturePublicKey:  signaturePublicKey,
 		store:               store,
-		secret:              setupSecretService(store),
+		secret:              setupSecretService(store, c.String("secret-extension-endpoint"), client, c.Bool("secret-extension-netrc")),
 		registry:            setupRegistryService(store, c.String("docker-config"), c.String("registry-extension-endpoint"), client),
 		config:              configService,
 		environment:         environment.Parse(c.StringSlice("environment")),


qwerty287:
Thanks! Please check out the tests and linters ;)

I also don't think we need the exclusive setting here. This is only required for the config because you can then move that fully to the extension. But as we always have the DB, it's fine to always query the secrets from there as well I think.

mofr93:
Okay, that sounds fine. I will update the PR. Thanks

6543:
also have a look at https://woodpecker-ci.org/docs/next/usage/extensions and add a documentation for it :)

location of this docs is at https://github.com/woodpecker-ci/woodpecker/tree/main/docs/docs/20-usage/72-extensions

web/src/assets/locales/en.json
```diff
@@ -535,6 +535,9 @@
   "registry_extension_endpoint": "Registry extension endpoint",
   "config_extension_exclusive": "Exclusive",
   "config_extension_exclusive_desc": "If enabled, will skip all other ways of configuration fetching, including the forge.",
+  "secret_extension_endpoint": "Secret extension endpoint",
+  "secret_extension_exclusive": "Exclusive",
```
6543: i wonder if we need exclusive here... ?!?

as i would just make it like the registry extension, that will priorize external and so overwrite internal if exist
repo config fetching is an edgecase where exclusive makes sense as config fetching has more implications

6543: ah @**qwerty287** already mentioned it anyway :)

mofr93:
Hey guys, Changes made to accommodate your requests. Also streamlined more with the registry extension. docs etc. :)

server/services/secret/combined_test.go
```diff
@@ -0,0 +1,182 @@
...
+			},
+			expectedNames:  []string{"db-secret"},
+			expectedValues: map[string]string{"db-secret": "db-value"},
+			expectedError:  false,
+		},
+	}
+
+	pubEd25519Key, privEd25519Key, err := ed25519.GenerateKey(rand.Reader)
+	if err != nil {
+		t.Fatal("can't generate ed25519 key pair")
```
qwerty287: For required checks, you can use the `require` package from testify, just like `assert`


```suggestion
	require.NoError(t, err, "can't generate ed25519 keypair")
```

Affects all the other `t.Fatal` uses here as well.

server/services/secret/http.go
```diff
@@ -0,0 +1,62 @@
...
+
+type httpExtension struct {
+	endpoint string
+	client   *utils.Client
+}
+
+type secretRequestStructure struct {
+	Repo     *model.Repo     `json:"repo"`
+	Pipeline *model.Pipeline `json:"pipeline"`
+	Netrc    *model.Netrc    `json:"netrc"`
```
qwerty287: I wouldn't send netrc data. That's only helpful to communicate with the forge, but a secret extension doesn't need this I think.

server/services/secret/http.go
```diff
@@ -0,0 +1,62 @@
...
+
+// SecretListPipeline fetches secrets from an external HTTP extension.
+func (h *httpExtension) SecretListPipeline(repo *model.Repo, pipeline *model.Pipeline, netrc *model.Netrc) ([]*model.Secret, error) {
+	body := secretRequestStructure{
+		Repo:     repo,
+		Pipeline: pipeline,
+		Netrc:    netrc,
+	}
+
+	var response []*model.Secret
```
qwerty287: I'd prefer to use the same structure like registries and config extension. I.e. wrap the secrets array in an object in the `secrets` key, just like e.g. https://woodpecker-ci.org/docs/next/usage/extensions/registry-extension#response

server/services/secret/http.go
mofr93: I was actually thinking to use this for querying the forge (eg. Github) for configs related to secrets. An example could be a yaml file with a list of secrets i want the extension to collect. What do you think?

qwerty287: I'd prefer to let the user configure that directly in the extension instead of letting woodpecker send this. But @6543 seems to agree to you, I also won't block this if there are more people in favour of this

6543: well i like the idea of being opt in as it is an unessesary expousure if not needed!

mofr93: In my opinion I prefer to have the Netrc credentials parsed along instead of managing credentials. In the config extension I use it to read files from the forge in the same way.

6543: @mofr93 that's not what i ment!

request: like the exclusive bool flag ... but different func ... have a flag that if optionally enabled pass netrc alongside else {}

PS: beside that this pull is in  good shape and likely to be merged for next release :tada:

mofr93: @6543 I didn't see you comment before now :) 

Should this be a global flag for all extensions or only secret extension?

6543: i would say extension config flag so you can have it globaly disabled while a repo secret extension has it enabled...

mofr93: So something like this.. if i understand :) 

* WOODPECKER_SECRET_SERVICE_NETRC=false (default)
Do not send Netrc.
Do not allow repo to override via gui.

* WOODPECKER_SECRET_SERVICE_NETRC=true
Send Netrc
Allow repo to override via gui

qwerty287: I'm still not fully convinced, sorry. If we add this flag this also affects the config and the registry extension and should at least be extended to them (which can also be done separately of course).

mofr93: @qwerty287 I agree if this flag is added it should be added for config and registry extensions aswell (but yes, separately like you say). 

I just like the idea of being able to opt-in for Netrc as it removes the burden of managing forge credentials in extensions.

qwerty287: Ok, then just go on and add this flag. We can add it afterwards to config and registries.

mofr93: Added WOODPECKER_SECRET_SERVICE_NETRC.

```
WOODPECKER_SECRET_SERVICE_NETRC=true
WOODPECKER_SECRET_SERVICE_ENDPOINT=https://example.com/secrets
```
Netrc will be send to the globally configured endpoint.
This also allows opt-in or opt-out via repo extension settings.

`WOODPECKER_SECRET_SERVICE_NETRC=false` (default). Do not send any NETRC for both global and repo. Do not allow repo override.

We could possibly grey out the box in the repo-level gui?

server/services/manager.go
```diff
@@ -101,7 +103,12 @@ func (m *manager) SignaturePublicKey() crypto.PublicKey {
 	return m.signaturePublicKey
 }
 
-func (m *manager) SecretServiceFromRepo(_ *model.Repo) secret.Service {
+func (m *manager) SecretServiceFromRepo(repo *model.Repo) secret.Service {
+	if repo.SecretExtensionEndpoint != "" {
+		includeNetrc := m.secretServiceNetrc && repo.SecretExtensionNetrc
```
qwerty287: The env var should only be valid for the global extension. Repo extensions should be configured completely separated. That's how the exclusive setting is handled as well.

```suggestion
		includeNetrc := repo.SecretExtensionNetrc
```

mofr93: This should disable sending Netrc server-wide. The repo owner will be able to just toggle it regardless. Is that a good idea?

qwerty287: I think so, yes. The netec data are the credentials of the repo owner (of the user that enabled the repo), so they should be able to decide whether they trust the extensions they use.

mofr93: Okay, makes sense. I will adjust with the change

docs/docs/20-usage/72-extensions/55-secret-extension.md
```diff
@@ -0,0 +1,166 @@
...
+The extension should respond with a JSON object containing a `secrets` array.
+If the extension wants to keep the existing secrets without adding any, it can respond with HTTP status `204 No Content`.
+
+```ts
+class Response {
+  secrets: {
+    name: string; // the secret name, matched by from_secret in pipeline config
+    value: string; // the secret value
+    images?: string[]; // optional: restrict to specific container images
+    events?: string[]; // optional: restrict to specific pipeline events
```
qwerty287, line 145: Did you test this? I would not be sure whether this works. I would expect if this is omitted it won't be available on any event. An empty array here will have the same effect as a secret where you disable all events.

For images it is indeed working like this, if it's empty it will be available for all plugins. Note that setting the image filter also limits to plugins, maybe you could change the comment form `specific container images` to `specific plugins`.

cmd/server/flags.go
```diff
@@ -288,6 +288,16 @@ var flags = append([]cli.Flag{
 		Name:    "registry-service-endpoint",
 		Usage:   "url used for calling registry service endpoint",
 	},
+	&cli.StringFlag{
+		Sources: cli.EnvVars("WOODPECKER_SECRET_SERVICE_ENDPOINT"),
```
qwerty287: I'd prefer for both vars using "extension" instead of "service", that's the correct term. I updated that in https://github.com/woodpecker-ci/woodpecker/pull/6326 for the other ones as well (maybe you wait till this one is merged so you're sure that PR is accepted…).

```suggestion
		Sources: cli.EnvVars("WOODPECKER_SECRET_EXTENSION_ENDPOINT"),
```

And, could you add them to https://woodpecker-ci.org/docs/next/administration/configuration/server#config_service_exclusive? Registries are missing there as will, also fixed in the PR linked above.

docs/docs/20-usage/72-extensions/55-secret-extension.md
qwerty287: Empty events is actually even disallowed: https://github.com/woodpecker-ci/woodpecker/blob/main/server/model/secret.go#L102

mofr93: I might have missed something. will check and fix asap. I'm also using the [woodpecker-ci/example-extensions](https://github.com/woodpecker-ci/example-extensions) for testing this, so nice that you're updating it.

mofr93: I dont get this error.

Removing or leaving `Events` empty works for some reason

```golang
secrets.Secrets = append(secrets.Secrets, &model.Secret{
	Name:   "MY_SECRET",
	Value:  "xxx",
	Images: []string{"mofrtest:latest", "ubuntu"},
	//Events: []model.WebhookEvent{},
})
```

qwerty287: Yes, you won't get the error because you don't run the `Validate()` function.

I also checked again, you were right. The important line is https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/yaml/compiler/compiler.go#L70

Sorry then for the noise.

cmd/server/flags.go
qwerty287: The PR is merged, could you update the env vars?

mofr93: Done.  are we not renaming `secret-service-endpoint`, but only the env var?

qwerty287: You're right. I forgot about this...

qwerty287: -> #6329

docs/docs/20-usage/72-extensions/55-secret-extension.md
```diff
@@ -0,0 +1,166 @@
...
+
+:::info
+The `netrc` field is only included in the request when the global `WOODPECKER_SECRET_EXTENSION_NETRC` is set to `true` (default: `false`) or the per-repo "Send netrc credentials" is checked.
+:::
+
+```ts
+class Request {
+  repo: Repo;
+  pipeline: Pipeline;
+  netrc?: Netrc; // only included when netrc sending is enabled (see below)
```
qwerty287: The paragraph is above ;)

```suggestion
  netrc?: Netrc; // only included when netrc sending is enabled (see above)
```

qwerty287:
Besides those two points it looks good for me now. Thanks again :)

server/services/secret/combined_test.go
```diff
@@ -0,0 +1,182 @@
+// Copyright 2025 Woodpecker Authors
```
qwerty287:
```suggestion
// Copyright 2026 Woodpecker Authors
```

server/services/secret/http.go
```diff
@@ -0,0 +1,69 @@
...
+	body := secretRequestStructure{
+		Repo:     repo,
+		Pipeline: pipeline,
+	}
+	if h.includeNetrc {
+		body.Netrc = netrc
+	}
+
+	response := new(secretResponseStructure)
+	status, err := h.client.Send(context.Background(), net_http.MethodPost, h.endpoint, body, response)
```
qwerty287: Please add a new argument for the context and do not create a new backgroud one. Similar to the http extension of the registries.

mofr93: would be great if you can add the things you find. Thanks :)

cmd/server/flags.go
mofr93: Let me know if you change this or you want me to do it :)

server/services/secret/combined_test.go
```diff
@@ -0,0 +1,182 @@
...
+			for i := range secrets {
+				secretNames[i] = secrets[i].Name
+			}
+			assert.ElementsMatch(t, tt.expectedNames, secretNames, "expected some other secrets")
+
+			for _, s := range secrets {
+				if expected, ok := tt.expectedValues[s.Name]; ok {
+					assert.Equal(t, expected, s.Value, "unexpected value for secret %s", s.Name)
+				}
+			}
```
qwerty287: I'd use `assert.ElementsMatch` only and just put the full secret models into one `expected` field.

So do not separately check names and values but just directly verify everything. Once my registry adaptation is finished, I can link you that PR so you can see what I mean

edit: will fix myself later

6543:
we had some renames to make the extention feature more clear, could you please resolve the conflict :)

PR #6254, opened 2026-03-17
fix(bitbucketdatacenter): resolve tag commit before fetching config

### LINKED ISSUE #6247: Bitbucket DC: tag events are ignored if no matching pipeline exists in the default branch
### Component

server

### Describe the bug

When using Woodpecker CI with Bitbucket Data Center, tag events are not triggering pipelines if the repository's default branch does not contains a pipeline that match the event.

Woodpecker receives the webhook event (repo:refs_changed for a tag creation), but no pipeline is created.
The event appears in the logs, however the build is silently skipped.

If the repository default branch is changed to a branch that contains a matching pipeline configuration (when: event: tag), the tag pipelines start working immediately.

This suggests that Woodpecker only evaluates the .woodpecker.yml from the repository’s default branch when handling Bitbucket webhook events. As a result, tag events may be ignored even if a valid pipeline exists in another branch.

This behavior is confusing because the webhook event is received but no pipeline is created, and no clear log message explains why the event was skipped.

### Steps to reproduce

1.	Configure a repository in `Woodpecker CI` connected to `Bitbucket DC`
2. 	Set the repository **default branch** to main.
3. Ensure that the `main` branch contains a .woodpecker.yaml that **does not include a pipeline matching the tag event**.
4. Create another branch (for example `ci/cicd`) that **does contain a pipeline for tag events**.
5. Push a tag
6. Observe that Woodpecker receives the webhook event but **no pipeline is created**.
7. Change the repository **default branch** to `ci/cicd`.
8. Push another tag 
9. Observe that the pipeline is now triggered successfully.

### Expected behavior

Woodpecker should evaluate the pipeline configuration based on the commit or branch related to the event, rather than only the repository’s default branch.

If a valid .woodpecker.yml exists in the branch where the event originates (or the commit pointed to by the tag), the pipeline should be triggered.

Alternatively, if Woodpecker intentionally evaluates the configuration only from the default branch, it should clearly log that the event was skipped because no matching pipeline was found in the default branch configuration.

Currently, the webhook event is received but the pipeline is silently skipped, which makes debugging difficult.

### System Info

```shell
Version Info: `v3.13.0`
Git provider: `Bitbucket Data Center`
Webhook event: `repo:refs_changed (tag creation)`
```

### Additional context

During debugging we observed the following behavior:
	•	If the repository **default branch is set to main**, tag pipelines only trigger when the pipeline configuration for the tag event exists in main.
	•	If the **default branch is changed to another branch** (for example ci/cicd) that contains a matching pipeline configuration, tag pipelines start working immediately.
	•	Woodpecker logs show that the tag event is received, but no pipeline is created when the default branch does not contain a matching pipeline configuration.

This behavior may be specific to Bitbucket webhook handling, since the webhook payload does not explicitly include the branch associated with the tag event.

### Validations

BillionClaw:
Annotated tags in Bitbucket Data Center use the tag object SHA as ToHash in webhooks, not the commit SHA. When a tag event arrived, the config fetcher tried to read &#96;.woodpecker.yml&#96; using the tag SHA, which failed because tag objects don't contain file content.

This change adds an explicit resolution step in the Hook handler that converts the tag SHA to the actual commit SHA before the config is fetched. Tag events now work regardless of which branch holds the pipeline config.

Fixes #6247


qwerty287:
Thanks! Is the commit used to post the statuses still correct? I.e. the fix from https://github.com/woodpecker-ci/woodpecker/pull/6203

6543:
@dccdis as this bot is unlikely to return feedback ... but it is still a posibilety to have a working fix ... might wana look into it?

dccdis:
Cannot reproduce the bug on `next-a7739a2bb7`

PR #6255, opened 2026-03-17
Sort glossary items alphabetically

paulovitorbal:
Sorted the glossary items from A to Z, as defined by the word glossary: 

`an alphabetical list, with meanings, of the words or phrases in a text that are difficult to understand`  - Oxford Dictionary

The text were keep the same, I only changed the order of the items on the list.


qwerty287 (approved):
Thanks!

PR #6259, opened 2026-03-18
Refactor so storage detects if Insert fails because of unique constraint

6543:
a followup of #6228 that focus on make server storage aware if failed inserts because of unique constraint by introducing new error type

also we rename the old `RecordNotExist` to `ErrRecordNotExist` to satisfy the error lint requirements that our linter currently is missing somehow

### DIFF
--- a/server/store/datastore/helper.go
+++ b/server/store/datastore/helper.go
@@ -53,14 +53,12 @@ func wrapDelete(c int64, err error) error {
 
 func wrapInsert(c int64, err error) error {
 	if err != nil {
-		switch errMsg := err.Error(); {
-		case strings.HasPrefix(errMsg, "UNIQUE constraint failed"),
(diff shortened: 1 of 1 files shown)


server/store/datastore/helper.go
```diff
@@ -51,6 +51,21 @@ func wrapDelete(c int64, err error) error {
 	return nil
 }
 
+func wrapInsert(c int64, err error) error {
+	if err != nil {
+		switch errMsg := err.Error(); {
```
qwerty287: Why do you use a switch here instead of a simple if-else? There's no `else if` branch here so I think switch makes it less readable.

6543: addressed -> 44019ecb986ac82e7805982523091613cb75aac6

Issue #6231, opened 2026-03-11
WOODPECKER_ADMIN does not seem to work


SamMousa:
### Component

server

### Describe the bug

I'm only just getting started, so this could be a documentation issue as well.
As a test I've set up woodpecker locally using docker compose (behind ssl reverse proxy).
Everything works, I can login with github etc.

However I was not marked as an admin, even though I've configured WOODPACKER_ADMIN env var. I even tried to use the same casing (found in the woodpecker CLI after logging in). 

For now I worked around it by extracting the sqlite db and marking myself as admin directly.

### Steps to reproduce

Use this docker compose configuration:
```
networks:
  devproxy:
    name: devproxy2
    external: true
services:
  woodpecker-server:
    networks:
      - devproxy
      - default
    labels:
      traefik.enable: "true"
      traefik.http.routers.woodpecker.rule: "Host(`woodpecker.test`)"
      traefik.http.services.woodpecker.loadbalancer.server.port: "8000"
    image: woodpeckerci/woodpecker-server:v3
    volumes:
      - woodpecker-server-data:/var/lib/woodpecker/
    environment:
      - WOODPECKER_OPEN=false
      - WOODPECKER_HOST=https://woodpecker.test
      - WOODPECKER_GITHUB=true
      - WOODPECKER_GITHUB_CLIENT=YYYY
      - WOODPECKER_GITHUB_SECRET=XXXXX
      - WOODPECKER_ADMIN=[[YOUR USERNAME]]
      - WOODPECKER_AGENT_SECRET=SYSTEM1

  woodpecker-agent:
    image: woodpeckerci/woodpecker-agent:v3
    command: agent
    restart: always
    depends_on:
      - woodpecker-server
    volumes:
      - woodpecker-agent-config:/etc/woodpecker
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      - WOODPECKER_SERVER=woodpecker-server:9000
      - WOODPECKER_AGENT_SECRET=SYSTEM1
volumes:
  woodpecker-server-data:
  woodpecker-agent-config:
  woodpecker-cli-config:
```

Then run `docker compose up`, log in and confirm that you are not admin...

### Expected behavior

I'd love to see some log output that at least tells me the `WOODPACKER_ADMIN` was parsed.
Ideally when such a user logs in for the first time I'd also see something in the logs about making them admin (which would not have helped in this case since it didn't make me admin)

### System Info

```shell
3.13, docker 29.3.0
[...]

PR #6261, opened 2026-03-19
Compare admins case-insensitively

qwerty287:
try to address something that might lead to confusion (https://github.com/woodpecker-ci/woodpecker/issues/6231)


lafriks:
Lint failure

PR #6262, opened 2026-03-19
If agent cancels a workflow we need to report that back to the queue too

6543:
regression of https://github.com/woodpecker-ci/woodpecker/issues/6011

### DIFF
--- a/server/rpc/rpc.go
+++ b/server/rpc/rpc.go
@@ -358,7 +358,11 @@ func (s *RPC) Done(c context.Context, strWorkflowID string, state rpc.WorkflowSt
 			queueErr = s.queue.Done(c, strWorkflowID, workflow.State)
 		}
 	} else {
-		queueErr = s.queue.Done(c, strWorkflowID, model.StatusCanceled)
+		if workflow.Started > 0 {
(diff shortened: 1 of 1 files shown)


server/rpc/rpc.go
```diff
@@ -350,16 +350,18 @@ func (s *RPC) Done(c context.Context, strWorkflowID string, state rpc.WorkflowSt
...
 		if workflow.Failing() {
 			queueErr = s.queue.Error(c, strWorkflowID, fmt.Errorf("workflow finished with error %s", state.Error))
 		} else {
 			queueErr = s.queue.Done(c, strWorkflowID, workflow.State)
 		}
-		if queueErr != nil {
-			logger.Error().Err(queueErr).Msg("queue.Done: cannot ack workflow")
-		}
+	} else {
+		queueErr = s.queue.Done(c, strWorkflowID, model.StatusCanceled)
```
qwerty287: I guess this should be status killed.

Killed means canceled while running already. Canceled means canceled while the step was still pending.

6543: <del>this here reports back not the workflow but the step state...</del> nop it is workfow
to distinkt that i need to add some more checks

good catch!

6543: done

PR #6263, opened 2026-03-19
WIP: Detached or service report err back

6543:
followup of  #6166
blocked by #6166

- [x] now services dont get cancled anymore and wait indefinetly ... fix it!
- [ ] well :point_up: fixed ... but now when services are cancled because workflow end has reached they report bac cancled ... witch marks the workflow cancled :laughing:


6543:
ported tests to  https://github.com/woodpecker-ci/woodpecker/pull/6471

PR #6267, opened 2026-03-20
Have distinct differense between skipped and canceled status

### LINKED ISSUE #6264: WebUI try to load logs of skipped step
The current webui impl. try to load logs for skipped stpes:

Image

version: current main

6543:
regression of https://github.com/woodpecker-ci/woodpecker/pull/6176

fix #6264

image

### DIFF
--- a/cmd/server/openapi/docs.go
+++ b/cmd/server/openapi/docs.go
@@ -5557,11 +5557,11 @@ const docTemplate = `{
                 "StatusKilled": "killed by user",
                 "StatusPending": "pending to be executed",
                 "StatusRunning": "currently running",
-                "StatusSkipped": "skipped as another step failed",
+                "StatusSkipped": "skipped as per condition of current workflow failed/success state",
                 "StatusSuccess": "successfully finished"
             },
             "x-enum-descriptions": [
-                "skipped as another step failed",
+                "skipped as per condition of current workflow failed/success state",
                 "pending to be executed",
                 "currently running",
                 "successfully finished",


web/src/assets/locales/en.json
```diff
@@ -224,6 +224,7 @@
         "cancel": "Cancel",
         "restart": "Restart",
         "canceled": "This step has been canceled.",
+        "skipped": "This step has been skipped as per condition of current workflow state.",
```
6543: I think a shorter txt would be nice but not sure how to frace it without losing infos

6543: `Conditionally skipped` ? what do you thing @qwerty287

server/model/const.go
```diff
@@ -55,7 +55,7 @@ func (s WebhookEvent) Validate() error {
 type StatusValue string //	@name	StatusValue
 
 const (
-	StatusSkipped  StatusValue = "skipped"  // skipped as another step failed
+	StatusSkipped  StatusValue = "skipped"  // skipped as per condition of current workflow state
```
6543:
```suggestion
	StatusSkipped  StatusValue = "skipped"  // skipped as per condition of current workflow failed state
```

web/src/assets/locales/en.json
qwerty287: Just say `This step has been skipped` is enough I think.

server/model/const.go
```diff
@@ -55,7 +55,7 @@ func (s WebhookEvent) Validate() error {
 type StatusValue string //	@name	StatusValue
 
 const (
-	StatusSkipped  StatusValue = "skipped"  // skipped as another step failed
+	StatusSkipped  StatusValue = "skipped"  // skipped as per condition of current workflow failed state
```
qwerty287: It can also be skipped on success right? With `when.status: [failure]`

web/src/assets/locales/en.json
```diff
@@ -258,6 +259,7 @@
         "running": "running",
         "started": "started",
         "skipped": "skipped",
+        "canceled": "canceled",
```
qwerty287, line 262: I intentionally left that "skipped" because the step actually was skipped, but due to cancellation.

server/model/const.go
6543: yes ...

web/src/assets/locales/en.json
6543:
```suggestion
        "skipped": "This step has been skipped.",
```

web/src/assets/locales/en.json
6543: well as you added a distinct cancle state we could argue that it was cancled

we already differ between cancled and killed, with your argument we should also not mark it as killed but cancled !?

i think not hiding info to the user is a god thing in this case

server/model/const.go
6543:
```suggestion
	StatusSkipped  StatusValue = "skipped"  // skipped as per condition of current workflow failed/success state
```

6543:
> I still think we somehow have to make clear what "cancelled" means,but that's likely much harder

agree on that ... my only idea is to add a link to [read more] in the text that links to the docs ... and explain it in the docs or so

anbraten:
Does the user even care about the difference of skipped / canceled / killed? I would like to know was my step started and has it finished? Was it successful or not? Most of the details are clear from the context in the UI. Has a duration and some logs => step was at least started even if it never finished. Having to understand the specific difference of a status makes it rather complicated for a user to know what happened (seems most maintainers including me run into the same issue).
 
Actually I am wondering why we show skipped steps at all. We don't show skipped workflows / steps from (depends_on) 🤔 Is it even really helpful? Probably not.

qwerty287:
If a workflow is skipped due to the status condition it's shown as well.

PR #6271, opened 2026-03-20
Only add compatibility environment variables for drone-ci to plugins

6543:
so we don't cluster our env vars...

taken from #6166


qwerty287:
Is that breaking? I.e. do we mention that in the docs for regular steps?

6543:
it's explizit mentioned in the migration docs for breaking at verion 0.15.0 (https://woodpecker-ci.org/migrations#0150)

we just still had it in the implementation but docs mention it to move away since a long time and new docs dont mention it at all

PR #6276, opened 2026-03-20
Server return conflict status when stale repo causes duplicate insert

### LINKED ISSUE #6266: 500 error when re-adding a repository with the same name after deletion
### Component

server

### Describe the bug

When a Gitea repository is deleted and recreated with the same name, attempting to add it to Woodpecker returns HTTP 500.

Workaround: find the old repository entry in Woodpecker and delete it manually - after that the repository can be added successfully.

### Steps to reproduce

1. Create a Gitea repository
2. Add it to Woodpecker
3. Delete the Gitea repository
4. Create a new Gitea repository with the same name
5. Try to add it to Woodpecker → HTTP 500

### Expected behavior

The user receives a meaningful error message instead of HTTP 500. Ideally, Woodpecker handles this case automatically (e.g. detects the stale record and removes it).

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.13.0"}
```

### Additional context

_No response_

### Validations

api2062:
## Summary
- return `409 Conflict` instead of `500` when repository activation hits a duplicate insert
- surface an actionable message telling users to remove the stale repository entry before re-adding
- add a regression test for `PostRepo` duplicate-insert behavior

## Testing
- not run locally in this environment (clone/setup commands stalled), but added focused unit test coverage for the new branch

Fixes #6266


qwerty287:
@api2062 why did you close this? I just tested this and it seems working.

qwerty287:
Thanks, test failure is unrelated.

PR #6282, opened 2026-03-21
Add opengrep plugin

hnb2:
Good morning,

I would like to submit a [new plugin](https://github.com/KalvadTech/woodpecker-opengrep) we developed for adding opengrep.

Thanks,

### DIFF
--- a/docs/plugins/woodpecker-plugins/plugins.json
+++ b/docs/plugins/woodpecker-plugins/plugins.json
@@ -312,7 +312,7 @@
     },
     {
       "name": "Opengrep",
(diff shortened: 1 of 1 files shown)


qwerty287:
Thanks! However, your docs file is missing the metadata header: see https://woodpecker-ci.org/docs/usage/plugins/creating-plugins#metadata or for an example https://codeberg.org/woodpecker-plugins/s3/src/branch/main/docs.md?display=source#L1-L10

hnb2:
Thank you for the feedback, I updated our docs.md with the metadata and updated the json file to point to it.

qwerty287 (approved):

PR #6295, opened 2026-03-22
Report skipped step state as soon as it's determined

6543:
currently we mark all steps that are pending after a workflow finished as skipped.

this is wrong!

as we know when we decide if a step is skipped we now just signal that back the server.

this pull also rewords the pipeline/state.State to match our terminology

### DIFF
--- a/agent/tracer.go
+++ b/agent/tracer.go
@@ -36,24 +36,24 @@ func (r *Runner) createTracer(ctxMeta context.Context, uploads *sync.WaitGroup, 
 		defer uploads.Done()
 
 		stepLogger := logger.With().
-			Str("image", state.Pipeline.Step.Image).
+			Str("image", state.CurrStep.Image).
 			Str("workflow_id", workflow.ID).
-			Err(state.Process.Error).
-			Int("exit_code", state.Process.ExitCode).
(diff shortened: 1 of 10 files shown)


agent/tracer.go
```diff
@@ -36,24 +36,24 @@ func (r *Runner) createTracer(ctxMeta context.Context, uploads *sync.WaitGroup,
...
-			StepUUID: state.Pipeline.Step.UUID,
-			Exited:   state.Process.Exited,
-			ExitCode: state.Process.ExitCode,
-			Started:  state.Process.Started,
-			Canceled: errors.Is(state.Process.Error, pipeline_errors.ErrCancel),
+			StepUUID: state.CurrStep.UUID,
+			Exited:   state.CurrStepState.Exited,
+			ExitCode: state.CurrStepState.ExitCode,
+			Started:  state.CurrStepState.Started,
+			Canceled: errors.Is(state.CurrStepState.Error, pipeline_errors.ErrCancel) || state.CurrStepState.Skipped,
```
6543: there it is

lafriks (approved):
Apart from lint failures LGTM

6543:
regression is fixed with #6300 (also code simplified and more tests)

PR #6298, opened 2026-03-22
feat(ui): show forge origin in admin users panel

### LINKED ISSUE #5678: [UX] in admin panel Users show forge
Image

there should be an indicator of what user belongs to witch forge ... 
... this is also missing in the detail view :/

mendarb:
Closes #5678

## Summary

Adds a forge indicator to the admin users list and detail view, so admins can see which forge (GitHub, Gitea, GitLab, Forgejo, Bitbucket, etc.) each user belongs to.

## Changes

- **Admin users list**: Added a `Badge` next to each user showing the forge type (e.g. "Forge: GitHub")
- **User detail/edit view**: Added a read-only forge field when editing an existing user
- **Data fetching**: Forges are loaded on mount and cached in a `Map<forgeId, Forge>` for efficient lookup
- **i18n**: Added `admin.settings.users.forge` key for the "Forge" label

## How it works

The component fetches all configured forges via `getForges()` on mount, builds a lookup map by ID, and uses `user.forge_id` to resolve and display the forge type name for each user.

## Test plan

- [ ] Open the admin panel > Users page
- [ ] Verify each user shows a forge badge (e.g. "Forge: GitHub") in the list
- [ ] Click edit on a user — verify the forge field appears as read-only
- [ ] Verify the page works correctly with multiple forges configured
- [ ] Verify no errors when a user has no forge_id set

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -406,7 +406,8 @@
           "placeholder": "User is an admin"
         },
         "delete_user": "Delete user",
-        "edit_user": "Edit user"
+        "edit_user": "Edit user",
+        "forge": "Forge"
       },
       "orgs": {
         "orgs": "Organizations",
--- a/web/src/views/admin/AdminUsers.vue
+++ b/web/src/views/admin/AdminUsers.vue
@@ -18,12 +18,13 @@
       >
         <img v-if="user.avatar_url" class="h-6 rounded-md" :src="user.avatar_url" />
         <span>{{ user.login }}</span>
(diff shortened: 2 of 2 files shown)


qwerty287:
The label should be the URL I think, similar to the login screen

lafriks:
Also imho forge name/type/url that's needed for display should come together in user field struct loaded as join do that it does not generate unneeded requests, that would be especially suboptimal for lists

qwerty287:
On the login page we do it like this as well. But I agree it's not optimal https://github.com/woodpecker-ci/woodpecker/blob/main/web/src/views/Login.vue#L97

6543:
screenshot(s) would be nice

PR #6300, opened 2026-03-22
Fix report skipped status

6543:
image

i first suspected my logic change as initially used the calcled as signal for skiped steps.
but i just forgot to move the struct change alongside grpc stuff ...

... now tested locally multible times to not mess up again

regression of #6295


6543:
thx!

Issue #5346, opened 2025-07-22
Kubernetes: de-privileging the clone step


dhess:
### Clear and concise description of the problem

Hi, thanks for this project! The Kubernetes integration appears to be much better thought out than most other CI systems we've used.

Unfortunately, as far as we can tell, cloning a repo using the Woodpecker-provided `netrc` credentials isn't possible without root privileges, short of marking the project as "Trusted:Security", which is probably worse overall, from a security perspective:

* The default clone step runs as the root user, so the runner namespace must be privileged. On security-conscious Kubernetes distributions like Talos, namespaces are not privileged by default, so the runner namespace needs the `pod-security.kubernetes.io/enforce: privileged` label on these distributions. In our clusters, we prefer to add this label only when the app absolutely requires elevated privileges; e.g., because it needs to use host networking.
* Because the default clone step runs as the root user, the otherwise very handy `WOODPECKER_BACKEND_K8S_SECCTX_NONROOT` option is incompatible with it. The default clone step pod will always fail to start if this option is set.
* As far as I can tell, the only reason that Woodpecker runs the default clone step as the root user is to set the filesystem permissions on the workspace directory. However, unless I'm missing something, that could easily be addressed by using the `fsGroupChangePolicy` pod security context setting, without any need for root privileges.

Initially, #4151 had support for `fsGroupChangePolicy` (e.g., https://github.com/woodpecker-ci/woodpecker/pull/4151/commits/3c7e071a56713c4176e713c8572a94fc21b0b7bc), but that appears to have been removed at some stage before the merge, though it's not clear to me why.

### Suggested solution

Unless there's some other reason why the default clone step needs to run as root, adding support for `fsGroupChangePolicy` in the Kubernetes `backend_options` and de-privileging the default clone step would be a major security posture improvement for the Kubernetes backend.

### Run the clone step manually

Because the `woodpeckerci/plugin-git` plugin is trusted, Woodpecker provides it with the `netrc` creds when run as a plugin. Therefore, we tried running a manual clone step, first like this:

```
skip_clone: true

steps:
  - name: clone
    image: quay.io/woodpeckerci/plugin-git:2.6.5
    backend_options:
      kubernetes:
        securityContext:
          runAsUser: 405 # `guest` user in Alpine
          runAsGroup: 100 # `users` group in Alpine
          fsGroup: 100
          privileged: false
          runAsNonRoot: true
          seccompProfile:
            type: RuntimeDefault
```

[...]

PR #6307, opened 2026-03-23
Kubernetes: Support allowPrivilegeEscalation and capabilities backend_options

Aex12:
With this PR changes, we can finally run workflows in a restricted namespace in the Kubernetes backend.

Related to #5346. I wouldn't say it closes it, but improves the situation by providing means to run in a fully unprivileged namespace.

Also, I'm not sure how should I backport this to v3.13.x, I can't find the release/v3.13 branch

### DIFF
--- a/pipeline/backend/kubernetes/backend_options.go
+++ b/pipeline/backend/kubernetes/backend_options.go
@@ -66,14 +66,16 @@ const (
 )
 
 type SecurityContext struct {
-	Privileged          *bool                                `mapstructure:"privileged"`
-	RunAsNonRoot        *bool                                `mapstructure:"runAsNonRoot"`
-	RunAsUser           *int64                               `mapstructure:"runAsUser"`
-	RunAsGroup          *int64                               `mapstructure:"runAsGroup"`
-	FSGroup             *int64                               `mapstructure:"fsGroup"`
-	FsGroupChangePolicy *kube_core_v1.PodFSGroupChangePolicy `mapstructure:"fsGroupChangePolicy"`
-	SeccompProfile      *SecProfile                          `mapstructure:"seccompProfile"`
-	ApparmorProfile     *SecProfile                          `mapstructure:"apparmorProfile"`
+	Privileged               *bool                                `mapstructure:"privileged"`
+	RunAsNonRoot             *bool                                `mapstructure:"runAsNonRoot"`
+	RunAsUser                *int64                               `mapstructure:"runAsUser"`
+	RunAsGroup               *int64                               `mapstructure:"runAsGroup"`
+	FSGroup                  *int64                               `mapstructure:"fsGroup"`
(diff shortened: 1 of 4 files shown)


pipeline/backend/kubernetes/backend_options.go
```diff
@@ -83,6 +85,13 @@ type SecProfile struct {
 
 type SecProfileType string
 
+type Capabilities struct {
+	Add  []Capability `mapstructure:"add"`
+	Drop []Capability `mapstructure:"drop"`
+}
+
+type Capability string
```
qwerty287: Would just remove that type and directly use strings.

qwerty287 (commented):
Thanks! Is it possible to use these options to do something dangerous that could compromise the agent? Right now I think you can use them in any pipeline. (Sorry I don't have k8s experience...)

Backport is not required.

Aex12:
> Thanks! Is it possible to use these options to do something dangerous that could compromise the agent? Right now I think you can use them in any pipeline. (Sorry I don't have k8s experience...)
> 
> Backport is not required.

@qwerty287 Currently, most kubernetes distributions run pods with `allowPrivilegeEscalation: true` and without dropping the default runtime Linux capabilities unless explicitly configured otherwise. That means workflows today already run with a relatively permissive security context in Kubernetes.

This PR allows users to tighten the workflow pod security context, so they can run in namespaces enforcing the [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/).

That said, we could change this so that `allowPrivilegeEscalation` can only be set to false, this would allow workflow pods to run in restricted namespaces, while avoiding potential harmful behavior if on the future the default value is changed by some container runtime.

Regarding `capabilities`, while it's true it [defaults to the the default set of capabilities granted by the container runtime](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#securitycontext-v1-core), Kubernetes documents it [can also be used to add non-default capabilities](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-capabilities-for-a-container). We could restrict this even further by only allowing to drop capabilities.

I think this changes make sense, I'll submit a new commit with them.

Aex12:
Sorry @qwerty287, I forgot to relaunch the tests after my last commit. I just added a new commit that fixes the test.

SecurityContext is being set because privileged steps set the SecurityContext.Privileged field to true unless set otherwise. This was the existing behavior prior to my changes, which is preserved.

qwerty287:
Code itself seems fine, @woodpecker-ci/maintainers could someone check out the k8s impact?

xoxys:
@Aex12 can you rebase once again? Implementation looks good to me. @henkka @MartinSchmidt as you contributed multiple times to the K8s backend and seem to user it on a daily basis you might want to take a look as well :)

Aex12:
Hi @xoxys, just rebased from main and squashed my commits.

xoxys (approved):

6543 (approved):

PR #6310, opened 2026-03-23
Kubernetes: add instance-wide default and enforced Security Context

Aex12:
This PR depends on PR #6307 being merged first.

### Description
Currently, when using the Kubernetes backend, there is no mechanism for instance owners to define an instance-wide Security Context that applies to all CI workloads running in the cluster.

This poses several problems:
- Repository maintainers are responsible for configuring their workflows to adhere to Pod Security Standards when running workloads in an unprivileged namespace.
- Workflow files are cluttered with securityContext options, which could be defined globally (as this PR does)
- If the namespace where workload pods are run is privileged, individual repositories can make their workflows run as root, posing potential security issues for the rest of the cluster.
- If the namespace where workload pods are run is unprivileged and a repository fails to set the correct securityContext for their workflow, it'll never execute.

This PR tries to fix this by allowing instance owners to define two new variables on the agent:
- `WOODPECKER_BACKEND_K8S_DEFAULT_SECCTX`: This is the securityContext that will be applied by default to all CI workloads running in the cluster, and it can be overridden by workflows.
- `WOODPECKER_BACKEND_K8S_ENFORCED_SECCTX`: This is the securityContext that will be applied to all CI workloads, it cannot be overridden by individual workflows.

An example configuration of this two variables that will allow running the ci workloads in an unprivileged namespace is:
```sh
WOODPECKER_BACKEND_K8S_DEFAULT_SECCTX='{"runAsUser":1000,"runAsGroup":1000,"fsGroup":1000,"fsGroupChangePolicy": "OnRootMismatch"}'
WOODPECKER_BACKEND_K8S_ENFORCED_SECCTX='{"privileged":false,"runAsNonRoot":true,"allowPrivilegeEscalation":false,"seccompProfile": {"type": "RuntimeDefault"}, "capabilities": {"drop": ["ALL"]}}'
```

### Possible deprecations
This PR preserves the functionality of the existing `BACKEND_K8S_SECCTX_NONROOT` variable for backward compatibility purposes, however with this PR changes we can achieve the same functionality by setting
```sh
WOODPECKER_BACKEND_K8S_ENFORCED_SECCTX='{"runAsNonRoot":true}'
```
With the added advantage of it being more flexible for instance owners to configure a wider range of enforced security context options.

### About the environment variable names
I decided to go with SECCTX to make it more consistent with the existing `BACKEND_K8S_SECCTX_NONROOT`, however I think it'll be more appropiate to use `SECURITY_CONTEXT`, to make the intent clear. What is your opinion about this?

### Added documentation
This PR also includes a small guide that explains how to run CI workloads in unprivileged namespaces using the added functionality. Currently the guide includes a workaround for the unprivileged clone step as mentioned in #5346. If PR #6312 gets merged, we can safely remove that from the guide, as it'll become unnecessary.

### DIFF
--- a/docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
+++ b/docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
@@ -359,6 +359,26 @@ backend_options:
 The feature requires Kubernetes v1.30 or above.
 :::
 
+You can set `allowPrivilegeEscalation` to `false` to prevent a container from gaining more privileges than its parent process.
+
+```yaml
+backend_options:
+  kubernetes:
+    securityContext:
+      allowPrivilegeEscalation: false
+```
+
+You can also drop [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) from a container. Adding capabilities is not allowed.
+
+```yaml
+backend_options:
+  kubernetes:
+    securityContext:
+      capabilities:
+        drop:
+          - ALL
+```
+
 ### Annotations and labels
 
 You can specify arbitrary [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) and [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) to be set on the Pod definition for a given workflow step using the following configuration:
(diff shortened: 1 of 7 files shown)


6543:
https://github.com/woodpecker-ci/woodpecker/pull/6307 got merged ...

PR #6316, opened 2026-03-24
Show Woodpecker version on pipeline details

### LINKED ISSUE #4204: Show WP version pipeline was run on
Sometimes it could be helpful to be able to get Woodpecker version the pipeline was run on.

Sys info

#4202

Sim-hu:
Stores the server version at pipeline creation time so that historical pipelines show the version they were actually run on, rather than the current server version.

Changes:
- Add `version` field to the `Pipeline` model (xorm auto-syncs the new column)
- Set `version` in both `Create()` and `Restart()` pipeline flows
- Display the version in the debug Tab
- Add TypeScript type and i18n string

image

Closes #4204

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -272,7 +272,7 @@
       "we_got_some_errors": "Oh no, an error occurred!",
       "duration": "Pipeline duration: {duration}",
       "created": "Created: {created}",
-      "version": "This pipeline ran on Woodpecker {version}.",
+      "version": "The Woodpecker version this pipeline was executed on.",
       "version_header": "Woodpecker version",
       "cancel_info": {
         "superseded_by": "Superseded by #{pipelineId}",


web/src/lib/api/types/pipeline.ts
```diff
@@ -95,6 +95,8 @@ export interface Pipeline {
   changed_files?: string[];
 
   cancel_info: CancelInfo;
+
+  version?: string;
```
qwerty287: Would make this required without `?` and for the old pipelines just show an empty string

Sim-hu: Good point, I'll make it required and show an empty string for older pipelines that don't have the version recorded. Updating now.

qwerty287:
Please run `make generate-openapi`: https://ci.woodpecker-ci.org/repos/3780/pipeline/32731/32

web/src/views/repo/pipeline/PipelineWrapper.vue
```diff
@@ -78,6 +78,10 @@
           <Icon name="duration" />
           <span>{{ duration }}</span>
         </div>
+        <div v-if="pipeline.version" class="flex shrink-0 items-center gap-2" :title="$t('repo.pipeline.version')">
+          <Icon name="tag" />
```
qwerty287: Would add a text as well here that this is the woodpecker version.

web/src/views/repo/pipeline/PipelineWrapper.vue
```diff
@@ -78,6 +78,10 @@
...
+          <span>{{ $t('repo.pipeline.version') }}: {{ pipeline.version }}</span>
```
qwerty287: Please directly use a format string in the locale file instead concatting in html.

Like `Woodpecker version: {version}`

web/src/views/repo/pipeline/PipelineWrapper.vue
```diff
@@ -78,6 +78,10 @@
           <Icon name="duration" />
           <span>{{ duration }}</span>
         </div>
+        <div v-if="pipeline.version" class="flex shrink-0 items-center gap-2" :title="$t('repo.pipeline.version')">
```
qwerty287: This translation now misses the pipeline version. I'd just remove the title.

```suggestion
        <div v-if="pipeline.version" class="flex shrink-0 items-center gap-2">
```

qwerty287:
@Sim-hu could you post a screenshot of this new feature?

Sim-hu:
image

qwerty287:
Thanks. Please move that info to the debug tab.

qwerty287:
Thanks, wouldn't put it into it's own `Panel`. We usually only have one panel.

With a proper header ("Woodpecker version") and then just the version without translated string below is what I'd do.

6543:
old pipelines:
image
new pipelines:
image


PS: screenshot updates are always welcome on UI changes ;)

web/src/assets/locales/en.json
```diff
@@ -272,6 +272,7 @@
       "we_got_some_errors": "Oh no, an error occurred!",
       "duration": "Pipeline duration: {duration}",
       "created": "Created: {created}",
+      "version": "Woodpecker version: {version}",
```
6543:
```suggestion
      "version": "Woodpecker version {version} parsed it.",
```

make it clear that this is the version who parsed the config not the instance version ...

6543: @Sim-hu @qwerty287 do you know any better phrasing?

qwerty287: @6543 I suggest https://github.com/woodpecker-ci/woodpecker/pull/6316#issuecomment-4133808184

6543: well but this still does not highlight it's about the version that recived & parsed the pipeline config...

qwerty287: Ok yes, but then we can also write a short text like `This pipeline ran on Woodpecker {version}.` below the header. Would highlight the version (bold) then.

6543: agree this looks nicer :)

@Sim-hu could you please address this :)

qwerty287:
It now looks like this.
grafik

Some more padding to the top would be good.

And not sure, but do others think the size is fine? Because in my opinion the text is a bit too small.

What do you think about something like this? With a code block.
grafik

Corresponding code would be
```vue
      <InputField v-if="pipeline.version" :label="$t('repo.pipeline.version_header')" class="pt-4">
        <p class="text-wp-text-alt-100 mb-2 text-sm">{{ $t('repo.pipeline.version') }}</p>
        <pre class="code-box">{{ pipeline.version }}</pre>
      </InputField>
```
and you have to change the translation string.

web/src/assets/locales/en.json
Sim-hu: Done! Updated to use a code-box for the version display with the agreed phrasing. Added top padding as well.

web/src/views/repo/pipeline/PipelineDebug.vue
```diff
@@ -8,6 +8,10 @@
       <div class="flex items-center space-x-4">
         <Button :is-loading="isLoading" :text="$t('repo.pipeline.debug.download_metadata')" @click="downloadMetadata" />
       </div>
+      <InputField v-if="pipeline.version" :label="$t('repo.pipeline.version_header')" class="pt-4">
+        <p class="text-wp-text-alt-100 mb-2 text-sm">{{ $t('repo.pipeline.version', { version: pipeline.version }) }}</p>
```
qwerty287:
```suggestion
        <p class="text-wp-text-alt-100 mb-2 text-sm">{{ $t('repo.pipeline.version') }}</p>
```

web/src/assets/locales/en.json
```diff
@@ -272,6 +272,8 @@
       "we_got_some_errors": "Oh no, an error occurred!",
       "duration": "Pipeline duration: {duration}",
       "created": "Created: {created}",
+      "version": "This pipeline ran on Woodpecker {version}.",
```
qwerty287:
```suggestion
      "version": "The Woodpecker version this pipeline was executed on.",
```

6543:
image

6543:
thanks :rocket:

PR #6319, opened 2026-03-24
Fix workflow serialize to omit skip_clone if false

6543:
as skip_clone is false by default we can omit it.

this also prepares the test to use the modern workflow RunsOn to use the when filter.

fixes https://codeberg.org/lafriks/woodpecker-pipeline-transform/issues/35


6543:
@lafriks -> https://codeberg.org/lafriks/woodpecker-pipeline-transform/pulls/38

Issue #6317, opened 2026-03-24
Refactor queue & pubsub plan [WIP]


6543:
# Refactoring Plan: Queue + PubSub Abstraction for Server Clustering

## 1. Problem Statement

Woodpecker CI server currently cannot run in a multi-instance cluster because:

- The queue (`server/queue/fifo.go`) is purely in-memory and monolithic — it mixes task storage, scheduling logic (dependency resolution, `ShouldRun` evaluation, deadline expiry), and wait-signaling (done channels) into one struct.
- The queue is accessed from **three separate places** via a global accessor (`server.Config.Services.Queue`): `server/rpc/rpc.go`, `server/pipeline/queue.go`, and `server/pipeline/cancel.go`. This makes it impossible to replace coherently.
- Pubsub is similarly scattered: `rpc.go` has its own `notify()` and `pipeline/topic.go` has a duplicate `publishToTopic()`.

## 2. Goals

1. **Single scheduler entry point.** All queue and pubsub operations go through one place. No more global accessors.
2. **`ShouldRun()` moves into the scheduler.** In a cluster, multiple RPC layers would each waste claims evaluating this. The scheduler decides what is runnable before anything is handed out.
3. **Separate scheduling business logic** from queue storage into a new `server/scheduler` package.
4. **In-memory stays the default.** Single-server deployments work exactly as before, no extra dependencies.
5. **External backends (Valkey, NATS) become opt-in** via configuration, enabling multi-instance clustering.
6. **Pubsub already an interface** (Phase 1 complete). External backends can be added by implementing `pubsub.PubSub`.
7. **Minimize unnecessary changes.** Reuse `model.Task`, reuse `server/store`. Keep changes scoped to `server/queue`, `server/scheduler`, and the callers that currently use the global accessor.

### Queue access points (the core problem)

| File | What it calls | How |
|------|--------------|-----|
| `server/pipeline/queue.go` | `Queue.PushAtOnce` | global `server.Config.Services.Queue` |
| `server/pipeline/cancel.go` | `Queue.ErrorAtOnce` | global `server.Config.Services.Queue` |
| `server/rpc/rpc.go` | `Queue.Poll`, `Queue.Done`, `Queue.Error`, `Queue.Wait`, `Queue.Extend` | injected via `NewWoodpeckerServer` |

### Pubsub access points (duplicate logic)

| File | What it does |
|------|-------------|
[...]

PR #6320, opened 2026-03-24
Fix pipeline cancel

### LINKED ISSUE #6184: Kubernetes backend continues to execute on cancelled pipeline execution
### Component

agent

### Describe the bug

I noticed plausible regression with how Kubernetes backend schedules and executes steps when user cancel the pipeline. On UI the running step is receiving logs after the cancel is hit, after the step pod completes, following steps are not producing logs to UI, but the step pods are executed on my local kind cluster. 

I bisected the change to 06818ee6adc563c0c48767eac21a67aba313c1ba / https://github.com/woodpecker-ci/woodpecker/pull/6056 . Before that the running step pods were terminated and no further execution were happening.

With the changes from that commit the line in screenshot is evaluated false (state.error == ErrExternal) and nil is returned, and there's not checks for pipeline cancellations.

### Steps to reproduce

With a kubernetes backend
- run a pipeline
- cancel it while a step is running
- observe logs are still produced on Woodpecker UI and further steps are executed on cluster

### Expected behavior

Running step pods should be cancelled, no further execution of the remaining pipeline steps

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"dev"} 
Running from 06818ee6adc563c0c48767eac21a67aba313c1ba
Kubernetes backend with `kind`
```

### Additional context

_No response_

### Validations

6543:
close #6184

### DIFF
--- a/server/queue/fifo.go
+++ b/server/queue/fifo.go
@@ -169,7 +169,14 @@ func (q *fifo) Wait(ctx context.Context, taskID string) error {
 		select {
 		case <-ctx.Done():
 		case <-state.done:
-			return state.error
+			// check if we have a wrapped cancel error and unwrap it
+			if errors.Is(state.error, ErrCancel) {
+				return ErrCancel
+			}
+			// or return queue errors and no workflow errors
+			if !errors.Is(state.error, new(ErrExternal)) {
+				return state.error
+			}
 		}
 	}
 	return nil


pipeline/runtime/workflow.go
```diff
@@ -53,7 +53,7 @@ func (r *Runtime) Run(runnerCtx context.Context) error {
 
 	r.started = time.Now().Unix()
 
-	if err := r.engine.SetupWorkflow(runnerCtx, r.spec, r.taskUUID); err != nil {
+	if err := r.engine.SetupWorkflow(r.ctx, r.spec, r.taskUUID); err != nil {
```
6543: to be able to cancle while workflow is setup

pipeline/backend/docker/docker.go
```diff
@@ -267,6 +267,8 @@ func (e *docker) WaitStep(ctx context.Context, step *backend_types.Step, taskUUI
 		log.Trace().Msgf("ContainerWait returned with resp: %v", resp)
 	case err := <-errC:
 		log.Trace().Msgf("ContainerWait returned with err: %v", err)
+	case <-ctx.Done():
```
6543, line 270: `ContainerWait` should respect the parent context cancelation, but this makes sure we do so in any case

agent/runner.go
```diff
@@ -51,6 +51,8 @@ func NewRunner(workEngine rpc.Peer, f rpc.Filter, h string, state *State, backen
 	}
 }
 
+// TODO: refactor this big function into subfunctions in it's own subpackage
+
```
6543, line 55: unrelated, i can move that out if wanted...

server/queue/fifo.go
```diff
@@ -169,10 +169,7 @@ func (q *fifo) Wait(ctx context.Context, taskID string) error {
 		select {
 		case <-ctx.Done():
 		case <-state.done:
-			// only return queue errors and no workflow errors
-			if !errors.Is(state.error, new(ErrExternal)) {
```
6543, line 173: this is the simple fix ... that lets us cancle again

6543:
:point_up: needs review feedback :)

pipeline/runtime/workflow.go
6543:
```suggestion
	if err := r.engine.SetupWorkflow(r.ctx, r.spec, r.taskUUID); err != nil { //nolint:contextcheck
```

6543:
now with root cause fix:

# cancled
image

# successful failed
image

# sucessfull success
image

6543:
I dislike the current queue interface anywa... I will refactor this away with #6317

PR #6322, opened 2026-03-25
Kubernetes: precreate workingDir as nonroot when required

Aex12:
### Problem
When the working directory is set to a directory that doesn't exists (for example, as `plugin-git` does), kubelet will pre-create it with ownership set to `root:root` and permissions `0755` . This makes pods running as non-root unable to write to it, causing permission errors.

### Solution
Added a `podInitContainer` function that conditionally creates an init container to pre-create the working directory with the correct permissions before the main step container starts.

### Behavior
- If the pod runs as root (`RunAsUser == 0` or unset), no init container is created. Kubelet handles directory creation automatically
- If the working directory matches a volume mount path exactly, no init container is needed. `FSGroupChangePolicy` handles permissions
- An init container is only created when the working directory is nested within a volume mount path
- The init container uses `busybox:stable-musl` with minimal resource limits (5m CPU, 5Mi memory) and drops all capabilities.

### Related issues and PRs
- Solves the error mentioned in https://github.com/woodpecker-ci/woodpecker/issues/5346#issuecomment-3211408746 without requiring a previous step.
- In addition to #6307 and #6310, this will make it easier to run woodpecker ci workloads in a namespace that enforces [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/)


Aex12:
Hi @6543, what do you think about this alternative approach to #6312 for allowing the clone step to run unprivileged?

This approach is more self contained to kubernetes only, so other backends remain unaffected.

This PR doesn't cover the `home` issue mentioned in that PR, but still improves the experience of running woodpecker workloads in unprivileged namespaces in Kubernetes. 

As an example, with this PR changes (in addition to #6322 and #6307) it's possible to run the clone step completely unprivileged by setting the following agent env vars:
```sh
WOODPECKER_BACKEND_K8S_DEFAULT_SECCTX='{"runAsUser":1000,"runAsGroup":1000,"fsGroup":1000,"fsGroupChangePolicy": "OnRootMismatch"}'
WOODPECKER_BACKEND_K8S_ENFORCED_SECCTX='{"privileged":false,"runAsNonRoot":true,"allowPrivilegeEscalation":false,"seccompProfile": {"type": "RuntimeDefault"}, "capabilities": {"drop": ["ALL"]}}'
```
And using this workflow:
```yaml
clone:
  git:
    image: docker.io/woodpeckerci/plugin-git
    settings:
      depth: 0
      home: "/tmp"

steps:
  - name: check-clone-succeeded
    image: alpine
    commands:
      - pwd
      - ls -lah
```
Not as convenient as if `home` was set to `/tmp` by default, as it wouldn't require additional changes to the workflow, but still a step forward in making the Kubernetes backend in unprivileged mode more convenient to use.

pipeline/backend/kubernetes/pod.go
```diff
@@ -284,6 +290,58 @@ func podContainer(step *types.Step, podName, goos string, options BackendOptions
...
+	// if workingDir is not covered by any volume mount, we don't need an init container to precreate it
+	if len(volumeMounts) == 0 {
+		return kube_core_v1.Container{}
+	}
+
+	return kube_core_v1.Container{
+		Name:            "init-" + container.Name,
+		Image:           initContainerImage,
+		ImagePullPolicy: kube_core_v1.PullAlways,
+		Args:            []string{"mkdir", "-p", container.WorkingDir},
```
6543, line 324: I'm thinking if we do have access to the main created container and it's image, and use its property of user to chown it to that ?!? .... or does it work with this already?

pipeline/backend/kubernetes/pod_test.go
```diff
@@ -1209,3 +1209,110 @@ func TestPodAffinityStepOverridesAgent(t *testing.T) {
 	ja := jsonassert.New(t)
 	ja.Assertf(string(podJSON), expected)
 }
+
+func TestInitContainer(t *testing.T) {
+	const expected = `
+	{
+		"name": "init-wp-01he8bebctabr3kgk0qj36d2me-0",
+		"image": "busybox:stable-musl",
+		"imagePullPolicy": "Always",
```
6543, line 1219: why?

pipeline/backend/kubernetes/pod.go
Aex12: The thing is it's not possible to do it in the main container, since the main container workingDir is automatically created by kubelet when it doesn't exists, and kubelet creates it owned by root:root with permissions set to 0755, so a non-root user wont be able to update permissions/ownership of that directory. 

Thats why an init container is needed in this case. The init container overhead is minimal, and it takes less than a second to complete.

pipeline/backend/kubernetes/pod_test.go
Aex12: I decided to go with `imagePullPolicy: Always` and the `stable-musl` tag so that the busybox image in use is always the most up to date. Despite the name, that pull policy will only pull images from the registry when the digest have changed for the image tag, which is not very frequent for busybox (as of writing, last time `stable-musl` tag have been updated was 2 months ago). The image is also very minimal and is only around 900kb of download when it needs to be updated.

We could use a fixed version tag and the `IfNotPresent` policy, but then we'll need to take care of checking for new releases and updating it in the event of a CVE being discovered in the image.

Aex12:
I took the opportunity to rebase from main to make merging easier

xoxys (approved):

6543 (approved):

6543:
@Aex12 thanks for the work ... took us some time to get reviewed :)

PR #6325, opened 2026-03-25
Lint and unify import aliases [replaced]

6543:
we mow make sure to only have one alias in our codebase

replaced by https://github.com/woodpecker-ci/woodpecker/pull/6328

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -164,6 +164,136 @@ linters:
         - '^\s*TODO:'
       capital: true
       period: true
+    importas:
+      no-extra-aliases: true
+      alias:
+        # stdlib
+        - pkg: log
+          alias: std_log
+        - pkg: errors
+          alias: std_errors
+        - pkg: net/http
+          alias: net_http
+
+        # grpc / protobuf
+        - pkg: google.golang.org/grpc/metadata
+          alias: grpc_metadata
+        - pkg: google.golang.org/grpc/credentials
+          alias: grpc_credentials
+        - pkg: google.golang.org/protobuf/proto
+          alias: grpc_proto
+
+        # woodpecker internal
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/types
+          alias: backend_types
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/docker
+          alias: backend_docker
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/kubernetes
+          alias: backend_kubernetes
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/local
+          alias: backend_local
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/pipeline/errors
+          alias: pipeline_errors
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime
+          alias: pipeline_runtime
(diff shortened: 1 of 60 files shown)


.golangci.yaml
```diff
@@ -164,6 +164,136 @@ linters:
...
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/server/pipeline/stepbuilder
+          alias: step_builder
+
+        # mocks
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/server/services/secret/mocks
+          alias: secret_service_mocks
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/server/services/registry/mocks
+          alias: registry_service_mocks
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/server/services/mocks
+          alias: manager_mocks
```
6543, line 232: the aliases are still up for debate i would say :)

.golangci.yaml
```diff
@@ -164,6 +164,136 @@ linters:
...
+    importas:
+      no-extra-aliases: true
+      alias:
+        # stdlib
+        - pkg: log
+          alias: std_log
+        - pkg: errors
+          alias: std_errors
+        - pkg: net/http
+          alias: net_http
```
qwerty287, line 176: I neither like this for errors not for http. Both are often used packages. I'd say they can stay how they are and our own packages have to use clear naming.

.golangci.yaml
```diff
@@ -164,6 +164,136 @@ linters:
...
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/pipeline/frontend/metadata
+          alias: pipeline_metadata
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/pipeline/frontend/yaml/types/base
+          alias: yaml_base_types
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/pipeline/frontend/yaml/types
+          alias: yaml_types
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/server/cron
+          alias: cron_scheduler
+        - pkg: go.woodpecker-ci.org/woodpecker/v3/server/pipeline/stepbuilder
+          alias: step_builder
```
qwerty287, line 224: If we change this I'd also rename the folder and files to `step_builder` (and then we can also drop the rule)

qwerty287:
I personally think you added way to many :laughing: 

I wouldn't strictly define this. I was actually only referring to the casing.

Esp. having custom imports for the backends, all the external docker and k8s packages etc. is not necessary in my opinion. It's clear here which package you mean. Inside the backends: it's the external one, outside: it's the backend.

Also, why do you explicitely add `github.com/cenkalti/backoff/v5` with alias `backoff`? Isn't that the default name already?

I would not rename packages like `github.com/muesli/termenv`, just keep as `termenv` without underscore. Same for `github.com/xeipuuv/gojsonschema`.

And what I will definitely block is `zerolog_log` for `github.com/rs/zerolog/log`. This is our logging package and should be accessed with a short, simple name which is `log`. And this is fine as we also rename the std log package.

In any case, the linter does not seem to work, as it ignores all the zerolog/log renamings that you missed (ex. `agent/runner.go` l. 25).

---

General note: I would try to avoid those aliases as far as possible and only use them if there are conflicts or really unclear names (e.g. all the `types` and `errors` packages).

.golangci.yaml
6543: well step builder will be gone once refactorin it completly

PR #6326, opened 2026-03-25
Rename extension env vars

qwerty287:
Besides `WOODPECKER_CONFIG_SERVICE_ENDPOINT` all changed vars were introduced after the last release, that's why this is not breaking.


6543:
I like the cange, so to not to get confused with our different subsystems :)

PR #6328, opened 2026-03-26
Unify import aliases

qwerty287:
My approach of #6325 

Removed 20 rules, and I'm also open to remove more.

### DIFF
--- a/shared/optional/serialization_json_test.go
+++ b/shared/optional/serialization_json_test.go
@@ -40,10 +40,6 @@ func TestOptionalToJson(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			b, err := json.Marshal(tc.obj)
 			assert.NoError(t, err)
-			assert.EqualValues(t, tc.want, string(b), "gitea json module returned unexpected")
-
-			b, err = json.Marshal(tc.obj)
-			assert.NoError(t, err)
 			assert.EqualValues(t, tc.want, string(b), "std json module returned unexpected")
 		})
 	}
@@ -76,15 +72,10 @@ func TestOptionalFromJson(t *testing.T) {
 	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			var obj1 testSerializationStruct
-			err := json.Unmarshal([]byte(tc.data), &obj1)
+			var obj testSerializationStruct
+			err := json.Unmarshal([]byte(tc.data), &obj)
 			assert.NoError(t, err)
-			assert.EqualValues(t, tc.want, obj1, "gitea json module returned unexpected")
-
-			var obj2 testSerializationStruct
-			err = json.Unmarshal([]byte(tc.data), &obj2)
-			assert.NoError(t, err)
-			assert.EqualValues(t, tc.want, obj2, "std json module returned unexpected")
+			assert.EqualValues(t, tc.want, obj, "std json module returned unexpected")
 		})
 	}
 }


.golangci.yaml
```diff
@@ -164,6 +164,96 @@ linters:
...
+        - pkg: github.com/swaggo/gin-swagger
+          alias: swaggo_gin_swagger
+        - pkg: xorm.io/xorm/log
+          alias: xlog
+        - pkg: github.com/tink-crypto/tink-go/v2/insecurecleartextkeyset
+          alias: insecure_clear_text_keyset
+        - pkg: github.com/migueleliasweb/go-github-mock/src/mock
+          alias: github_mock
+        - pkg: gitlab.com/gitlab-org/api/client-go/v2
+          alias: gitlab
```
qwerty287, line 256: Had to add this rule back because the `swag fmt` command always adds the alias back. That's really weird because it should only format the comments.

pipeline/backend/dummy/dummy.go
```diff
@@ -51,8 +51,8 @@ const (
 	testServiceTimeout = 1 * time.Second
 )
 
-// New returns a dummy backend.
-func New() backend.Backend {
+// New returns a dummy backend_types.
```
6543:
```suggestion
// New returns a dummy backend.
```

pipeline/runtime/runtime.go
```diff
@@ -21,20 +21,20 @@ import (
...
 
-	backend "go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/types"
+	backend_types "go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/types"
 	"go.woodpecker-ci.org/woodpecker/v3/pipeline/logging"
 	"go.woodpecker-ci.org/woodpecker/v3/pipeline/tracing"
 	"go.woodpecker-ci.org/woodpecker/v3/shared/utils"
 )
 
-// Runtime represents a workflow state executed by a specific backend.
+// Runtime represents a workflow state executed by a specific backend_types.
```
6543:
```suggestion
// Runtime represents a workflow state executed by a specific backend.
```

shared/optional/serialization_json_test.go
```diff
@@ -83,7 +82,7 @@ func TestOptionalFromJson(t *testing.T) {
 			assert.EqualValues(t, tc.want, obj1, "gitea json module returned unexpected")
 
 			var obj2 testSerializationStruct
-			err = jsoniter.ConfigCompatibleWithStandardLibrary.Unmarshal([]byte(tc.data), &obj2)
+			err = json.Unmarshal([]byte(tc.data), &obj2)
```
6543: this is intentional, we test against jsoniter and json

shared/optional/serialization_json_test.go
```diff
@@ -43,7 +42,7 @@ func TestOptionalToJson(t *testing.T) {
 			assert.NoError(t, err)
 			assert.EqualValues(t, tc.want, string(b), "gitea json module returned unexpected")
 
-			b, err = jsoniter.ConfigCompatibleWithStandardLibrary.Marshal(tc.obj)
+			b, err = json.Marshal(tc.obj)
```
6543: same here ... either we just drop that testcase or jsoniter must stay...

shared/optional/serialization_json_test.go
qwerty287: OK, didn't know that and I thought this was a mistake.

However jsoniter seems to be deprecated, their repo is archived.

6543: uh ok, in that case we can just remove it, will push to your branch

6543: -> dee5e79e33a20bffde6006ac167ac9d05f55ad0b

6543:
> and I'm also open to remove more.

I think we can do so in smaller refactorings per case bases now :)

at least we dont randomly get new aliases all over the codebase ;)

PR #6329, opened 2026-03-26
Fix extension flag names

qwerty287:
leftover from #6326

### DIFF
--- a/docs/src/pages/migrations.md
+++ b/docs/src/pages/migrations.md
@@ -18,6 +18,10 @@ To enhance the usability of Woodpecker and meet evolving security standards, occ
 
 Extension HTTP calls (as of now the configuration extension) will by default only be allowed to contact external hosts. Set `WOODPECKER_EXTENSIONS_ALLOWED_HOSTS` accordingly to allow additional hosts as needed.
 
+### Internal changes
+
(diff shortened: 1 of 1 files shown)


cmd/server/flags.go
```diff
@@ -275,17 +275,18 @@ var flags = append([]cli.Flag{
 	},
 	&cli.StringFlag{
 		Sources: cli.EnvVars("WOODPECKER_CONFIG_EXTENSION_ENDPOINT", "WOODPECKER_CONFIG_SERVICE_ENDPOINT"), // TODO remove _SERVICE_ var in 4.0.0
-		Name:    "config-service-endpoint",
+		Name:    "config-extension-endpoint",
```
anbraten, line 278: What do you think about adding a new label and section for the changelog that marks "internal breaking" changes. As this for example changes the arg names you can pass to the server command like server --config-extension-endpoint We don't expect users to do so, but it might be nice to warn people having some rather custom integrations. Would be nice for other changes like this as well.

qwerty287: [7d4dbab](https://github.com/woodpecker-ci/woodpecker/pull/6329/commits/7d4dbabec55a553da2b3ba96c1da83e01894cc59)

6543 (approved):

PR #6337, opened 2026-03-28
Set workflow services stuck in running state to finished

samoli:
When a workflow includes service steps, the workflow and pipeline status is permanently stuck as "running" in the database after all steps had finished.

completeChildrenIfParentCompleted was called after UpdateWorkflowStatusToDone, so the status calculation still saw service steps as running. The in-memory step state also wasn't updated after the database update.

The fix moves child completion before the status calculation and syncs the in-memory state so WorkflowStatus sees the finalized step states.

I think this bug was possibly introduced in #6011


samoli:
Ah nice, didn’t see that. Thanks. 

I’ll test again when #6320 is merged.

6543:
ok now the refactoring is merged and i finished https://github.com/woodpecker-ci/woodpecker/pull/6320 ... this works in the same domain but it looks like it solves something different :sweat_smile:

6543:
well #6263 should make it so the final done is only returned once the detached/services did report back there status.

still this here is a nice bugfix, fixing it now and harden the server later on :)

PR #6346, opened 2026-03-30
Separate cron field

qwerty287:
splitted out from https://github.com/woodpecker-ci/woodpecker/pull/4626

Separate the field `cron`.

This also contains many preparations for more from #4626 ex. the separated api models including the deprecated fields.

### DIFF
--- a/server/cron/cron_test.go
+++ b/server/cron/cron_test.go
@@ -66,7 +66,6 @@ func TestCreatePipeline(t *testing.T) {
 		Commit:   "sha1",
 		Event:    "cron",
 		ForgeURL: "https://example.com/sha1",
-		Message:  "test",
 		Ref:      "refs/heads/default",
 		Cron:     "test",
 	}, pipeline)


server/api/pipeline.go
```diff
@@ -78,8 +78,12 @@ func CreatePipeline(c *gin.Context) {
 	pl, err := pipeline.Create(c, _store, repo, tmpPipeline)
 	if err != nil {
 		handlePipelineErr(c, err)
+	}
+
+	if pl != nil {
+		c.JSON(http.StatusOK, pl.ToAPIModel())
 	} else {
-		c.JSON(http.StatusOK, pl)
+		c.Status(http.StatusNoContent)
```
qwerty287, line 86: @6543 to address https://github.com/woodpecker-ci/woodpecker/pull/4626#discussion_r3004873945

The API can also expect a 204, this is already possible (handled in `handlePipelineErr`, line 80)

server/store/datastore/migration/027_add_cron_field.go
```diff
@@ -0,0 +1,74 @@
...
+			return err
+		}
+
+		page := 0
+		oldPipelines := make([]*pipelines, 0, perPage027)
+
+		for {
+			oldPipelines = oldPipelines[:0]
+
+			err := sess.Limit(perPage027, page*perPage027).Where("event = ?", model.EventCron).Cols("id", "sender").Find(&oldPipelines)
```
6543: the migration could be one SQL statement :)

server/cron/cron.go
```diff
@@ -143,7 +143,7 @@ func CreatePipeline(ctx context.Context, store store.Store, cron *model.Cron) (*
 		Branch:              cron.Branch,
 		Message:             cron.Name,
```
6543:
```suggestion
```

server/store/datastore/migration/027_add_cron_field.go
6543:
```
_, err := sess.Exec("UPDATE pipelines SET cron = sender, sender = '' WHERE event = ?", model.EventCron)
return err
```

it should be ... @qwerty287  but i have not tested it myselfe jet :)

qwerty287: Yes that seems working.

Tried with a simple script

```go
package main

import (
	_ "github.com/mattn/go-sqlite3"
	"xorm.io/xorm"
)

func main() {
	xEn, err := xorm.NewEngine("sqlite3", "db.sqlite")
	if err != nil {
		panic(err)
	}
	
	sess := xEn.NewSession()
	
	type pipelines struct {
		ID int64 `xorm:"pk autoincr 'id'"`
		Event string `xorm:"event"`
		Sender string `xorm:"sender"`
		Message string `xorm:"message"`

		// new cron field
		Cron string `xorm:"cron"`
	}

	if err := sess.Sync(new(pipelines)); err != nil {
		panic(err)
	}

	_, err = sess.Exec("UPDATE pipelines SET cron = sender, sender = '', message = '' WHERE event = ?", "cron")
	if err != nil {
		panic(err)
	}
	
}
```

and some test SQLite

6543 (approved):

Issue #1148, opened 2022-09-01
Gitea login endless loop


H4CK3R-01:
### Component

server, web-ui

### Describe the bug

If I try to login into gitea I get redirected to my gitea instance at the first time. There I can authorise woodpecker and then I get redirected back to my woodpecker instance. But again to the login view (/do-login). The next time I try to login the page refreshes only but nothing happens

Another server with the same config but connected to github is working fine.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"0.15.3"}

I'm using docker-compose.
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A join our [Discord Chat Server](https://discord.gg/fcMQqSMXJy) or the [Matrix room](https://matrix.to/#/#woodpecker:matrix.org).

6543:
is it possible for you to test against `next` tag?

H4CK3R-01:
Yes. Unfortunately the same mistake.
```
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-c79d49c8"}
```

H4CK3R-01:
Maybe you have a working docker-compose config for gitea and woodpecker?

6543:
https://woodpecker-ci.org/docs/next/administration/setup#docker-compose and https://woodpecker-ci.org/docs/next/administration/forges/gitea I would say?

H4CK3R-01:
This is exactly the documentation I used. But this does not work :-(

6543:
just to make sure can you post the full config (secrets redacted)?

H4CK3R-01:
My `docker-compose.yml`-file
```
version: '3'

services:
  woodpecker-server:
    image: woodpeckerci/woodpecker-server:next
    restart: unless-stopped
    volumes:
      - woodpecker_server_data:/var/lib/woodpecker/
      - /etc/localtime:/etc/localtime
      - /etc/timezone:/etc/timezone
    environment:
      - WOODPECKER_LOG_LEVEL=${WOODPECKER_LOG_LEVEL}
[...]

PR #6348, opened 2026-03-30
Only redirect after login

qwerty287:
Right now, you're always redirected to the login page if you clicked the log in button once because WP saves the redirect url but it is only removed after successful login. This fixes the issue by not paying attention to the saved url before login.


6543:
@qwerty287 does this also prevent infintie loops in misconfigured setups: https://github.com/woodpecker-ci/woodpecker/issues/1148 ?

qwerty287:
I don't know, never saw such a loop

PR #6349, opened 2026-03-30
Send 404 if logs are not allowed to access

qwerty287:


6543:
wow now precomit has timeouts fetching from github :/

merge by :mage:

PR #6357, opened 2026-03-31
Docker backend: replace docker SDK with moby SDK

qwerty287:
Fixes the securityscan.

Closes #6154


go.mod
```diff
@@ -38,6 +37,8 @@ require (
 	github.com/lib/pq v1.12.1
 	github.com/mattn/go-sqlite3 v1.14.38
 	github.com/migueleliasweb/go-github-mock v1.5.0
+	github.com/moby/moby/api v1.54.0
+	github.com/moby/moby/client v0.3.0
```
6543, line 41: can we use moby v2 as it will come out soon?

qwerty287: Afaik the moby/v2 is only available for the pakcage `github.com/moby/moby/v2`. This package is internal only though and not intended for regular use.

There's no `github.com/moby/moby/client/v2`.

6543: :/

pipeline/backend/docker/convert.go
```diff
@@ -94,7 +94,16 @@ func toHostConfig(step *types.Step, conf *config) *container.HostConfig {
 		config.NetworkMode = container.NetworkMode(step.NetworkMode)
 	}
 	if len(step.DNS) != 0 {
-		config.DNS = step.DNS
+		addrs := make([]netip.Addr, len(step.DNS))
+		for i, dns := range step.DNS {
+			a, err := netip.ParseAddr(dns)
+			if err != nil {
+				log.Error().Err(err).Str("address", dns).Msg("could not parse dns address")
```
6543: I think we should surface this error ... so the workflow fails with that error

6543: I'll add that code :)

6543 (approved):
tested works 🚀

6543:
the new moby interface looks way more usable :)

Issue #6355, opened 2026-03-31
[next-b6632e18e8] Build status is "killed" even though all steps succeeded


lephuongbg:
### Component

server

### Describe the bug

Tried next-b6632e18e8 but there are following issues in screenshot:

<img width="1243" height="478" alt="Image" src="https://github.com/user-attachments/assets/ae6db00f-481f-43b2-8de8-25ed914d71a6" />

### Steps to reproduce

I tested with a barebone yaml file that uses docker dind:

<details>
<summary>config.yaml</summary>

```yaml
clone:
  git:
    image: woodpeckerci/plugin-git
    settings:
      merge-pull-request: true

when:
  - event: pull_request

steps:
  - name: test
    image: docker:cli # use 'docker:<major-version>-cli' or similar in production
    environment:
      DOCKER_HOST: unix:///var/run/dind/docker.socket
    volumes:
      - /var/run/dind:/var/run/dind
    commands:
      - sleep 30
      - docker version

services:
  - name: docker
    image: docker:dind # use 'docker:<major-version>-dind' or similar in production
    privileged: true
    environment:
      DOCKER_HOST: unix:///var/run/dind/docker.socket
    volumes:
      - /var/run/dind:/var/run/dind
```
</details>

### System Info

```shell
{
  "source": "https://github.com/woodpecker-ci/woodpecker",
  "version": "next-b6632e18e8"
}
```

### Additional context

 _Originally posted in [#6164](https://github.com/woodpecker-ci/woodpecker/issues/6164#issuecomment-4160566992)_

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

xoxys:
Cant reproduce yet. Tested on `next-0cc8f4923e` which is one commit behind `b6632e18e8` but the missing commit is irrelevant.

<img width="1041" height="334" alt="Image" src="https://github.com/user-attachments/assets/39c3c9ad-5674-487d-8142-632761427ee1" />

<img width="461" height="1218" alt="Image" src="https://github.com/user-attachments/assets/9093aa3a-42c7-4ea4-a7ef-7481b10a5371" />

[...]

PR #6021, opened 2026-01-26
fix(agent): workflow runner use shutdown context

Pnkcaht:
## What was happening

When a workflow/run was canceled from the UI or API, the server correctly marked the run as Canceled, but the runner machines continued executing job steps.

Specifically:

- The cancel signal was received by the runner
- The workflow state was updated on the server
- Pipeline execution on the runner was not interrupted
- Tests and other long-running steps kept running and consuming runner capacity

This resulted in:

- Wasted compute time
- Runner capacity remaining blocked
- Confusing UX (UI shows Canceled, but jobs keep running)

## Related issue

Closes #5925

## Related PRs

This change complements previous cancellation-related fixes:

- #6018 — fixes container cleanup on step cancellation  
- #6020 — ensures DestroyStep runs on pipeline cancellation

While those PRs focus on step and backend cleanup, this PR ensures that workflow cancellation is properly propagated to the runner execution context.

## What this PR changes

This PR ensures that canceling a workflow immediately stops execution on the runner.

- Propagates server-side cancel events to the runner workflow context
- Cancels the workflow execution context as soon as a cancel signal is received
- Ensures pipeline execution and backend steps are interrupted
- Normalizes cancellation handling so the workflow consistently ends as Canceled

## Why this approach

In Woodpecker, the workflow context is the single source of truth for execution lifecycle.

Previously, receiving a cancel event did not reliably cancel the workflow execution context, allowing pipeline steps to continue running.
By explicitly canceling the workflow context when a cancel signal is received:

- The pipeline runtime receives context.Done()
- Backend implementations (Docker, Kubernetes, etc.) can terminate running steps
- Runner capacity is freed promptly

This aligns runner behavior with what the UI reports and avoids wasted resources.

## Implementation overview

```mermaid
sequenceDiagram
    participant UI as Web UI / API
    participant Server as Woodpecker Server
    participant Runner as Agent Runner
    participant Pipeline as Pipeline Runtime
    participant Backend as Backend Engine

    UI->>Server: Cancel workflow
    Server->>Runner: Cancel signal (Wait)
    Runner->>Runner: cancel workflow context
    Runner->>Pipeline: ctx.Done()
    Pipeline->>Backend: stop execution
    Backend-->>Pipeline: terminate running steps
    Pipeline-->>Runner: execution aborted
    Runner->>Server: report canceled state
```

PR #6361, opened 2026-03-31
Fix workflow being skipped and marked as failed when agent starts before server

6543:
reported at https://github.com/woodpecker-ci/woodpecker/issues/6355#issuecomment-4163701180

When the agent started before the server was available, it retried the connection as expected. However, once the server came up and a workflow was picked up, the pipeline would immediately fail without running any steps — the agent logs showed `workflow context done` firing instantly after `received execution`.

The root cause was a package-level `shutdownCtx` shared across retry iterations. On each failed attempt, `stopAgentFunc` stamped it with a 5-second timeout — starting the clock immediately. By the time the agent successfully connected and received a workflow, `workflowCtx` was derived from this already-expired context, so execution failed before Docker even started a container.

The fix removes the global mutable shutdown context and the `stopAgentFunc` indirection. Instead, `runner.Run()` no longer accepts a `shutdownCtx` parameter — it creates a fresh one locally only when needed for the `Done()` fallback call. The healthcheck server shutdown does the same. This makes the lifetime of each shutdown window explicit and local.

### DIFF
--- a/agent/runner.go
+++ b/agent/runner.go
@@ -33,6 +33,8 @@ import (
 	"go.woodpecker-ci.org/woodpecker/v3/shared/utils"
 )
 
+const shutdownTimeout = time.Second * 5
+
 type Runner struct {
 	client   rpc.Peer
 	filter   rpc.Filter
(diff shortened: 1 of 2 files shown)


xoxys (approved):

6543:
one of the bugs was actually a regression of #6021 ... b.t.w. ... but with this refactoring we did more fixing and preventing :)

PR #6369, opened 2026-04-02
WebUI: remove "lodash" dep

6543:
make our webui smaler by 0.1Mb ... and remove dep that is haunted by security issues and almost replaced by modern browser APIs.

https://developer.mozilla.org/de/docs/Web/API/Window/structuredClone#browser-kompatibilit%C3%A4t


qwerty287:
It seems this broke editing secrets and registries.

Clicking the edit button fails with `DOMException: Proxy object could not be cloned.` (for example in `AdminRegistries.vue`, line 100 for the admin registries)

6543:
-> #6409

PR #6381, opened 2026-04-04
Docker backend should retry to delete volume on "in use" error

### LINKED ISSUE #6360: Docker Backend: Cancled Pipeline does not remove volume
```
4:50PM ERR pipeline/backend/docker/docker.go:373 > could not remove volume 'wp_01kn25ztyhhrry0bwtj2zxrhbz_0_default' error="Error response from daemon: remove wp_01kn25ztyhhrry0bwtj2zxrhbz_0_default: volume is in use - [85e6a1922b240e839b05a556a5dd0f8cc3f7e885e3f01fc1f3801493adb3e716]"
```

Image

Image

version: f121387e61a58db0273388cf00d635f977ce2012

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/6360


xoxys:
Why not use the existing backoff lib we already use?

6543:
@xoxys -> 7c83ce9a2566ce66f42caef33199a5fc1029cca8

not sure if that's cleaner but now the lib is used

xoxys:
Well at least its consistent and the backoff retry is almost always the overall better approach.

pipeline/backend/docker/docker.go
```diff
@@ -339,7 +344,7 @@ func (e *docker) DestroyStep(ctx context.Context, step *backend_types.Step, task
 
 	// we first signal to the container to stop ...
 	if _, err := e.client.ContainerStop(ctx, containerName, client.ContainerStopOptions{
-		Timeout: &containerKillTimeout,
+		Timeout: toRef(containerKillTimeout),
```
xoxys, line 347: Why do we need such a helper?

6543: because of golang, you can not just use reference of const, so you have to use a func that gets the const as copy-by-val and return it's reference

xoxys: How was it working before then?

6543: it was no const but a var, i changed it to const as now the compiler makes sure it can not be changed

in my opinion global vars should be eliminated and moved to const just so that no mistakes do happen...
... if you prevere to remove the helper and have an mutable global var i can revert that back, this is just a smal refactor i included while tuching that code

xoxys (approved):

PR #6388, opened 2026-04-05
feat: woodpecker-cli pipeline log --follow

### LINKED ISSUE #1824: [CLI] pipeline logs follow logs if still running
make `pipeline logs` follow logs if they still run

and let them be addressed by pid

itzwam:
This PR adds event stream support and pipeline logs streaming for running pipelines

close https://github.com/woodpecker-ci/woodpecker/issues/1824


cli/pipeline/log/log_show.go
```diff
@@ -108,5 +128,22 @@ func stepLog(client woodpecker.Client, repoID, number, step int64) error {
...
+	logs, errs, err := client.StreamLogEntries(repoID, number, step)
+	if err != nil {
+		return err
+	}
+
+	for {
+		select {
+		case log := <-logs:
+			fmt.Println(string(log.Data))
+			// fmt.Printf("%d: %s\n", log.Line, string(log.Data))
```
qwerty287: Please remove such lines

cli/pipeline/log/log_show.go
```diff
@@ -108,5 +128,22 @@ func stepLog(client woodpecker.Client, repoID, number, step int64) error {
...
+func streamLog(client woodpecker.Client, repoID, number, step int64) error {
+	logs, errs, err := client.StreamLogEntries(repoID, number, step)
+	if err != nil {
+		return err
+	}
+
+	for {
+		select {
+		case log := <-logs:
+			fmt.Println(string(log.Data))
```
qwerty287, line 140: Is this raw data? Because I think on regular logs it's base64-encoded. Didn't try though

woodpecker-go/woodpecker/repo.go
```diff
@@ -424,6 +426,62 @@ func (c *client) StepLogEntries(repoID, num, step int64) ([]*LogEntry, error) {
...
+					if msgEvent, ok := msg["event"]; ok {
+						switch event := msgEvent; event {
+						case "error":
+							errCh <- fmt.Errorf(msgData)
+							return
+						case "eof":
+							errCh <- nil
+							return
+						default:
+							errCh <- fmt.Errorf("UNKNOWN!%s: %s", msgEvent, msgData)
```
qwerty287:
```suggestion
							errCh <- fmt.Errorf("unknown event %s: %s", msgEvent, msgData)
```

qwerty287 (commented):
Thanks! Added two comments for the first review

cli/pipeline/log/log_show.go
itzwam: golang json already does the job

https://pkg.go.dev/encoding/json
> Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.

qwerty287:
Thanks, tested, works.
However, I think when the step finishes it should exit which does not happen right now - what do you think?

Also please check out the linter https://ci.woodpecker-ci.org/repos/3780/pipeline/33355/36

itzwam:
On my side I got an EOF event then connection closed.
Have u got a différent béhavior ?

qwerty287:
I tried with ci.woodpecker-ci.org and it didn't exit. I'll try again

itzwam:
> I tried with ci.woodpecker-ci.org and it didn't exit. I'll try again

The server I tested with isn't up2date
Will try to investigate with the same endpoint u used

qwerty287:
I tried again, and I think I found the cause: The EOF is sent once the workflow finished, not once the step finished.

itzwam:
wasn't able to start a pipeline or find a running one, will try with a docker instance

seems like there's some things to investigate there
https://github.com/woodpecker-ci/woodpecker/blob/main/server/api/stream.go

cli/pipeline/log/log_show.go
```diff
@@ -32,6 +32,12 @@ var logShowCmd = &cli.Command{
 	Usage:     "show pipeline logs",
 	ArgsUsage: "<repo-id|repo-full-name> <pipeline> [step-number|step-name]",
 	Action:    logShow,
+	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "follow",
```
lafriks, line 37: Short "f" if not already used would be nice

6543:
I think we should wait for https://github.com/woodpecker-ci/woodpecker/pull/6518 to get merged and introduce the streaming api in the go client in it's own pull first ... but I wont block it asap #6518 got decided to be accepted/rejected ...

6543:
note this closes https://github.com/woodpecker-ci/woodpecker/issues/1824 somewhat ... need to look if the autodetection is desired or not ...

PR #6398, opened 2026-04-07
Group logs by command in step logs UI and make them collapsible

lephuongbg:
Currently logs from all command are dumped in the same scroll view, which make it hard to check logs for each command.

This PR does following:

1. Group logs by command and make each section collapsible.

[Screencast_20260403_201617.webm](https://github.com/user-attachments/assets/793227cf-f2bd-4272-b9e9-b1ffa6b2b549)

2. Added buttons to quickly expand all/collapse all log groups.

image

3. When user open a step to check logs, if the step already finished, show log outline at first.

Other notes:

~~Pending issue: Looks like this break auto-scroll at the moment.~~ False positive. I didn't understand that I have to click on the "eye" icon to enable auto-scroll.

Caveat: The code identifies whether a log line is a command by matching it against the pipeline configuration. This might not work 100% of the time, but the PR should already be an improvement over current status quo.

Recreate [PR#6374](https://github.com/woodpecker-ci/woodpecker/pull/6374) since I needed to change the source branch for easier time merging upstream change.


qwerty287:
I think this approach is fine till we have better backend support. I would not parse the configs.

lephuongbg:
@qwerty287 Added some more changes. This is now ready for review.

qwerty287:
Thanks, but please remove all locale changes except English. Translations are managed in weblate.

lephuongbg:
Reverted locale changes as requested. Thanks for your time.

qwerty287:
Thanks. Didn't look at the code yet, but from your screenshots:
1. I think it looks somehow weird that the line number is way more right for the collapsible lines because of the chevron. I don't have a good solution though… Having the chevron right and the number left is bad, moving all lines number more to the right is also not nice. What I could imagine is completely dropping the line numbers for the collapsible lines i.e. the new commands. What do you think? Also @woodpecker-ci/maintainers 
2. I would change the order of the top-left buttons to have the delete icon most right. Separating with the vertical line is a very nice idea.

lephuongbg:
It does look cleaner without the line number, and without the `+ ` prefix. But we will lose the ability to get direct link to the line number. Is that a good compromise?

image

qwerty287:
Yeah I expected such issues. Would be good to keep the line linking

But isn't is possible to just set the `id` HTML element param on the collapsible ones as well with the `L{number}` value?

lephuongbg:
Should be no issue with setting the ID attribute, though the main issue is there is no longer line number to right click and copy link.

Nonetheless, the decision to keep it or not keep it lies with yours as maintainers. I do prefer the cleaner look personally.

qwerty287:
Not being able to copy the line link is fine for me.

@woodpecker-ci/maintainers somebody has objections?

6543:
> not copy line links

of all logs or just the commands that are now bars?!?

qwerty287:
Only the new bars that with the commands that was executed

6543:
I'm fine with that ... as you normally do not link to that... and else you can just link to one line blow

xoxys:
I would like to keep line numbers and ability to link to command lines. Just intend the lines below the collapsible sections to align it with the line numbers again?

lephuongbg:
> Just intend the lines below the collapsible sections to align it with the line numbers again?

FYI, I did try that. The log lines use three-column-grid layout. The first column dynamically change width based on the content (i.e it will change it the last line number is 1-digit, 2-digit, 3-digit and so on), so a fixed indentation is not possible. I tried to use the same grid layout on the new command line but proved quite difficult, as the command line has different styling and also need to use sticky position to stay on top while scrolling).

lephuongbg:
Found a simple solution from how bitbucket pipelines handles it. It is simply to show the log line for each command inside its collapsible log area.

image

qwerty287:
This looks good to me as well. @xoxys @6543 you agree?

xoxys:
Yep works for me 👍

6543 (approved):
works for me :)

xoxys (approved):
Thanks!

xoxys:
@lephuongbg Could you check the build failures?

6543:
@xoxys thats unrelated and on our side :/

qwerty287:
Just noticed this doesn't work with pnpm install: https://ci.woodpecker-ci.org/repos/3780/pipeline/34411/7

6543:
ok should be fixed by https://github.com/woodpecker-ci/woodpecker/pull/6692

this also simplify the logic in some way to relay less on regex and make it able to be unit tested

6543:
also make default collapse user preference: https://github.com/woodpecker-ci/woodpecker/pull/6694

PR #6409, opened 2026-04-08
unwrap obj to clone it

6543:
fix https://github.com/woodpecker-ci/woodpecker/pull/6369#issuecomment-4204638496


6543:
still get errors :/

xoxys (dismissed):

6543:
jep... :/

web/src/lib/utils/index.ts
```diff
@@ -1,7 +1,13 @@
...
 export function debounce<T extends unknown[]>(fn: (...args: T) => void, delay: number): (...args: T) => void {
   let timer: ReturnType<typeof setTimeout>;
   return (...args: T) => {
     clearTimeout(timer);
     timer = setTimeout(fn, delay, ...args);
   };
 }
+
+export function deepClone<T>(value: T): T {
+  return JSON.parse(JSON.stringify(toRaw(value))) as T;
```
6543, line 12: this looks like a hack but all over the internet it is told to be the way to go :/

6543: PS: this only works as our objects are raw data objects and not some that contain functions etc...

lephuongbg: happen to come across this PR and want to suggest using `structuredClone`

https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone

qwerty287: This doesn't work which is the reason why we have this PR.

qwerty287:
For the `showAddRegistry` funcs you can actually use `structedClone`, that works. It's only about the edit function

6543:
well i dont understant why it works or dont in certain cases, i would just stick to our clone func for data for redability

xoxys:
Why not just copy the cloneDeep func from lodash for now?

6543:
it does to much, we just want to copy json object, it can copy objects with functions in it etc...

... this func of lowdash is huge and we dont ned it

https://github.com/lodash/lodash/blob/cb0b9b9212521c08e3eafe7c8cb0af1b42b6649e/lodash.js#L2667

xoxys (approved):
Untested. If it works fine for me. However Ill count this as another micro optimization that caused more trouble then it helped. Personally I dont care about a bigger binary size and even if lodash has sec issues (go stdlib had a lot of critical cves as well in the past months) we are mostly not affected as we dont use the vulnarable code.

PR #6414, opened 2026-04-09
Add connection timeout and graceful shutdown to agent RPC client

### LINKED ISSUE #3971: Shutting agent down shouldn't connect to server when server is not running
### Component

agent

### Describe the bug

Regression #3895 

Trying to exit the agent if the server is not running tries to connect to the server even though it was never connected.

### Steps to reproduce

Run the agent without the server

### Expected behavior

_No response_

### System Info

```shell
main
```

### Additional context

_No response_

### Validations

AhmadNajiKam:
### Problem

When the Woodpecker server becomes unavailable while an agent is running, the agent:
- Retries indefinitely
- Produces excessive warning logs
- Cannot distinguish between temporary vs permanent connection loss

---

### Solution

Introduce connection health tracking and a bounded retry window in the agent gRPC client:

- Track connection state via `connectivity.Ready`
- Start timeout window when connection is lost
- Exit after 2 minutes of continuous disconnection
- Reduce log noise by emitting periodic warnings instead of per-retry errors
- Return `ErrConnectionLost` for explicit handling

---

### Key Changes

- Add `connectionLostAt` tracking to gRPC client
- Introduce `ErrConnectionLost`
- Add `IsConnected()` and timeout logic
- Update retry loops in RPC methods to respect connection state
- Extend `Peer` interface with `IsConnected()`
- Handle graceful shutdown in agent runner

---

### Behavior Changes

| Scenario | Before | After |
|----------|--------|-------|
| Server disconnects | Infinite retries | Exit after 2 minutes |
| Logging | Frequent warnings | Periodic warnings |
| Shutdown | Never | Graceful |

---

### Notes

- Timeout is currently hardcoded (`2m`), can be made configurable later
- Existing tests pass
- Mocks updated for interface change

Closes #3971


6543:
https://ci.woodpecker-ci.org/repos/3780/pipeline/33301/31

6543:
we sqash-merge anyway so just merge main into your branch from time to time :)

pipeline/runtime/runtime_test.go
```diff
@@ -196,17 +196,28 @@ func TestWorkflowWithServiceStep(t *testing.T) {
...
 	if assert.Len(t, traces, 5) {
-		assert.EqualValues(t, backend_types.State{}, traces[0].CurrStepState)
-		assert.Greater(t, traces[2].CurrStepState.Started, int64(0))
-		assert.EqualValues(t, backend_types.State{Started: traces[2].CurrStepState.Started, Exited: true}, traces[2].CurrStepState)
-		assert.EqualValues(t, backend_types.State{}, traces[3].CurrStepState)
-		assert.Greater(t, traces[4].CurrStepState.Started, int64(0))
-		assert.EqualValues(t, backend_types.State{Started: traces[4].CurrStepState.Started, Exited: true}, traces[4].CurrStepState)
-
-		assert.Greater(t, traces[4].Workflow.Started, int64(0))
+		dbStarted := findStartedTrace(traces, "db")
```
6543: unrelated ... just submit another pull :)

agent/rpc/client_grpc.go
```diff
@@ -66,6 +75,23 @@ func (c *client) Close() error {
 	return c.conn.Close()
 }
 
+func (c *client) IsConnected() bool {
+	connected := c.conn.GetState() == connectivity.Ready
```
6543: I wonder if we should also check for `Idle`  as IsConnected is checked before a request is handled ,,,

AhmadNajiKam: Yeah that's a really sharp catch I didn't think about long inactivity :)
I just reviewed the gRPC docs and I will fix it right now

pipeline/runtime/runtime_test.go
AhmadNajiKam: Sure thing I will

agent/rpc/client_grpc.go
AhmadNajiKam: Now it's done, and I moved the unrelated commit to another branch and will submit a PR with it soon.

6543 (approved):
good work :)

cmd/agent/core/agent.go
```diff
@@ -288,6 +291,11 @@ func run(ctx context.Context, c *cli.Command, backends []types.Backend) error {
 
 				log.Debug().Msg("polling new workflow")
 				if err := runner.Run(agentCtx); err != nil {
+					if errors.Is(err, agent_rpc.ErrConnectionLost) {
```
lafriks, line 294: If I read this code right it will shutdown agent if connection is lost even if no shutdown was requested? 🤔

6543: yes ...

@AhmadNajiKam do you mind make the timeout now configurable in the next pull and have an "infinite" option?

AhmadNajiKam: Absolutely, I'll do it today or tomorrow (since I have a hackathon running right now)

6543: @AhmadNajiKam 👀

6543: -> https://github.com/woodpecker-ci/woodpecker/pull/6470

PR #6415, opened 2026-04-09
fix second overflow build error

6543:
last fix was done by looking at the CI output and just verify lint/build locally

this time i verifyed the release build too ... should work now :/

PS: now I'm not so confident anymore so please review :sweat: 

followup to https://github.com/woodpecker-ci/woodpecker/pull/6413


6543:
regression caused by https://github.com/woodpecker-ci/woodpecker/pull/6411

lafriks (approved):

PR #6424, opened 2026-04-11
Fix erroneous linter error for plugin privileges

packrat386:
Currently if you use one of the plugins from the list that were formerly privileged by default and you explicitly configure that step with `privileged: true` you still get a linter error prompting you to add those to WOODPECKER_PLUGINS_PRIVILEGED. As far as I can tell this error message is incorrect because WOODPECKER_PLUGINS_PRIVILEGED is just a way to make certain images function the same as `privileged: true` by default.


pipeline/frontend/yaml/linter/linter.go
```diff
@@ -206,6 +206,12 @@ func (l *Linter) lintPrivilegedPlugins(config *WorkflowConfig, c *types.Containe
 	// lint for conflicts of https://github.com/woodpecker-ci/woodpecker/pull/3918
 	if utils.MatchImage(c.Image, "plugins/docker", "plugins/gcr", "plugins/ecr", "woodpeckerci/plugin-docker-buildx") {
```
qwerty287:
```suggestion
 	if utils.MatchImage(c.Image, "plugins/docker", "plugins/gcr", "plugins/ecr", "woodpeckerci/plugin-docker-buildx") && !c.Privileged {
```

qwerty287 (commented):
Thanks! Would change the if clauses like this

pipeline/frontend/yaml/linter/linter.go
packrat386: done

qwerty287 (approved):

PR #6432, opened 2026-04-13
Remove excessive dev/test log outputs

hhamalai:
With the 3.14.0-rc.0, my server output is flooded with dev debug printlns like
```
mat true <nil> [success failure]
mat true <nil> [success failure]
mat true <nil> [success failure]
mat true <nil> [success failure]
mat true <nil> [success failure]
mat true <nil> [success]
mat true <nil> [success failure]
mat true <nil> [success]
mat true <nil> [failure]
...
```


6543 (approved):

6543:
sorry CI was not accessible and the change looked simple ... i missed that the parsing just happens twice and one is now for nothing and our build fails because it detects that

-> https://github.com/woodpecker-ci/woodpecker/pull/6436

PR #6435, opened 2026-04-13
Kubernetes: detached steps are no services

### LINKED ISSUE #5124: `Detach` action doesn't play nicely with Kubernetes backend
### Component

agent

### Describe the bug

When using the Kubernetes backend, and then running a pipeline that requests for `detach`:
```yaml
steps:
  server:
    image: node:22-alpine
    detach: true
    commands:
      - cd /woodpecker/playwright-tests/server-work/server
      - npm run db:reset
      - exec npm run dev
  ```

The following error shows up in the logs:
```json
{"level":"info","time":"2025-04-25T20:50:02Z","message":"starting Woodpecker agent with version '3.5.2' and backend 'kubernetes' using platform 'linux/amd64' running up to 1 pipelines in parallel"}                                                                      
{"level":"error","error":"rpc error: code = Unknown desc = workflow finished with error Service \"wp-svc-01jsqbepk0ykj9emkhf5mwxeak-server\" is invalid: spec.ports: Required value","time":"2025-04-25T20:50:33Z","message":"grpc error: wait(): code: Unknown"}          
{"level":"warn","repo":"AtvikSecurity/pentracker","pipeline":"959","workflow_id":"1968","error":"rpc error: code = Unknown desc = workflow finished with error Service \"wp-svc-01jsqbepk0ykj9emkhf5mwxeak-server\" is invalid: spec.ports: Required value","time":"2025-04-25T20:50:33Z","message":"cancel signal received"}
```

Then appending the `ports` value to the pipeline:
```yaml
steps:
  server:
    image: node:22-alpine
    detach: true
    commands:
      - cd /woodpecker/playwright-tests/server-work/server
      - npm run db:reset
      - npx ts-node src/scripts/create-test-user.ts
      - exec npm run dev
    ports:
      - 5001
```

The following linter error shows up:
Image

So, definitely a chicken and the egg type problem...

### Steps to reproduce

Explained above

### Expected behavior

_No response_

### System Info

```shell
Version 3.5.2 for both the server and agent.
```

### Additional context

_No response_

### Validations

6543:
the implementation just assumes, that detached steps are services, witch is wrong

close https://github.com/woodpecker-ci/woodpecker/issues/5124


xoxys:
@MartinSchmidt can you take a look?

xoxys:
Not checked the implementation but how are detached steps now work on kunbernetes. They will be regular pods? Containers in the workflow pod? Do we ensure not not end up with orphanded resources after pipeline exited?

6543:
https://github.com/woodpecker-ci/woodpecker/blob/3058b85cc91f3a74e57efc455a9ae6d6a8f3ee31/pipeline/runtime/workflow.go#L42-L52

https://github.com/woodpecker-ci/woodpecker/blob/3058b85cc91f3a74e57efc455a9ae6d6a8f3ee31/pipeline/backend/kubernetes/kubernetes.go#L473-L484

xoxys (approved):
Code lgtm but untested. Thanks for the refs should work.as expected for detached steps.

PR #6436, opened 2026-04-14
Fix build error and include status calculation

6543:
followup of https://github.com/woodpecker-ci/woodpecker/pull/6432

fix:
image

also if we had no match we still returned true on include success.

this now requires at least one constrain match and then returns true immediately instead of checking the rest


6543:
NOTE: IncludesStatusSuccess now has a logic change, but that should not affect real instances as we have filtered non global matching workflows out before checking on what status it should run

xoxys (approved):
Not tested, code lgtm.

Issue #6372, opened 2026-04-03
[v3.14.0-rc-0] When service exits with non-zero status code, the whole pipeline should not be considered failure


lephuongbg:
### Component

server

### Describe the bug

When service is killed after pipeline finishes running, if the exit code is non-zero, the whole pipeline is considered as failure, even though all steps finish successfully.

Services are sidecar processes like docker dind, database, so its exit code shouldn't matter to the pipeline status.

<img width="980" height="565" alt="Image" src="https://github.com/user-attachments/assets/7d2fe0c7-c6c0-47cc-b624-0a778c5428d6" />

### Steps to reproduce

Minimal yaml file to reproduce the issue:

```yaml
when:
    event: push

steps:
   - name: test
     image: ubuntu
     commands: [sleep 30]

services:
  - name: docker
    image: ubuntu
    commands: ["sleep infinity"]
```

### Expected behavior

Pipeline should succeed when all steps success, regardless of service exit status.

### System Info

```shell
v3.14.0-rc-0
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

lephuongbg:
Found out that I can set `failure: ignore` to workaround this. Though IMO this should be the default value for step that is declared in `services` block.

qwerty287:
I disagree. As you say, ignoring the failure works, but I think a failing service should also let the pipeline fail.

lephuongbg:
For a service to fail **before** pipeline finishes, I would agree. But if the service "fails" due to being killed when pipeline finishes normally, it would be quite unexpected to consider that as failure.

6543:
well woodpecker sends a SIGTERM to services first and wait 5sec to let them shutdown gracefully. most service like databases respect that and stop with exit code 0 -> successfully

only those not respecting SIGTERM are getting killed and those we get an error code. what image do you use that fails?

6543:
as for more context why and how this is happening ...: https://github.com/woodpecker-ci/woodpecker/pull/6039

services and detached steps status where just ignored ... this was not some intentional thing, the runtime implementation was just really messy and got just cleaned up in v1.14.

lephuongbg:
[...]

PR #6445, opened 2026-04-15
Docker/K8s: add config for stop timeout

6543:
address https://github.com/woodpecker-ci/woodpecker/issues/6372#issuecomment-4252443253

### DIFF
--- a/pipeline/backend/docker/config.go
+++ b/pipeline/backend/docker/config.go
@@ -17,6 +17,7 @@ package docker
 import (
 	"fmt"
 	"strings"
+	"time"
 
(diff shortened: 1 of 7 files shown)


qwerty287 (approved):

qwerty287:
if lint and test is fixed lgtm

6543:
this mechanism needs refinment ... we can argue that 5s is to short, but now if you cancle a pipeline or a workflow has finished ... for example mariadb will run for another 3 minutes before exeting successfully :/

@xoxys you proposed 3m ... do you have some reason for that number or can we shrink it to <~20s (**by default**)?

PR #6448, opened 2026-04-15
Add `WOODPECKER_FORCE_IGNORE_SERVICE_FAILURE` config to preserve non-breaking behavior by default

### LINKED ISSUE #6372: [v3.14.0-rc-0] When service exits with non-zero status code, the whole pipeline should not be considered failure

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/6372


qwerty287:
Docs are missing

6543:
Well I added docs, still this option should not be here in the long run...

6543:
@qwerty287 done

qwerty287:
Thanks, is it worth adding this to https://woodpecker-ci.org/migrations#user-facing-migrations?

6543:
i would say no, this should be done when we aktually break it by that pull doing so

qwerty287 (approved):

PR #6452, opened 2026-04-15
pipeline runtime on cancel still should wait for stage to close chan

6543:
taken from #6391

currently we do not wait for stage to finish if workfow gets canceled, but as runStage should also respect the context cancel and could have some cleanup running ... we have to wait for it


6543:
included in https://github.com/woodpecker-ci/woodpecker/pull/6391

PR #6453, opened 2026-04-15
Use refactored pipeline builder in cli exec

6543:
based on https://github.com/woodpecker-ci/woodpecker/pull/3967


lafriks (approved):

6543:
well pull was already ready ... but i decided to add an test ... and forgot that test runs real exec that respect env vars that the ci sets ... so in this case i need to clean env vars before running the test :laughing:

PR #6455, opened 2026-04-16
Fix race condition in event stream channel handling

### LINKED ISSUE #6454: woodpecker server panics & crashes at times
### Component

server

### Describe the bug

```
panic: send on closed channel

goroutine 686109 [running]:
panic: send on closed channel

goroutine 685741 [running]:
```

### Steps to reproduce

```
```

### Expected behavior

_No response_

### System Info

```shell
```

### Additional context

_No response_

### Validations

utafrali:
Fixes #6454

We were getting panics when clients disconnected from SSE feeds because the code was closing the channel while goroutines could still be trying to send on it.

The fix removes the explicit close() calls and refactors the select statements so that context cancellation and channel sends happen atomically. If the context is cancelled, we skip the send. The channel gets garbage collected naturally after all goroutines exit the context.

Added a concurrent stress test to make sure this doesn't happen again.

### DIFF
--- a/server/pubsub/memory/pub_test.go
+++ b/server/pubsub/memory/pub_test.go
@@ -30,11 +30,14 @@ func TestPubsubConcurrentCancel(t *testing.T) {
 	testTopic := map[string]struct{}{"test": {}}
 	broker := New()
 
-	for range 50 {
+	for range 100 {
 		ctx, cancel := context.WithCancelCause(t.Context())
-		ch := make(chan []byte, 1)
+		ch := make(chan []byte) // Unbuffered to force blocking sends
 
+		var wg sync.WaitGroup
+		wg.Add(1)
 		go func() {
+			defer wg.Done()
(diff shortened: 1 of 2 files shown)


server/pubsub/memory/pub_test.go
hhamalai, line 1: Could this test be refactored so that it fails on the current main without your fixes? Tried it to verify if this can be used to demonstrate the issue, but test passes on each run.

utafrali:
Done, pushed the fix.

6543:
now I dont know why there are tow pulls working on the same issue by the same people either authored or reviewed!?!?

-> https://github.com/woodpecker-ci/woodpecker/pull/6456

utafrali:
Ah yeah, I'll close this since #6456 is already on it.

6543:
still thanks for the fix

taylorsilva:
FYI, OP is a bot: https://github.com/concourse/concourse/pull/9532#issuecomment-4256453363

PR #6456, opened 2026-04-16
Fix send on closed channel panic in SSE stream handlers

### LINKED ISSUE #6454: woodpecker server panics & crashes at times

hhamalai:
Instead of explicitly closing channels which leads to concurrency issues with channels being used after close, allow channels to be GC'ed.

Remove the recover block that was previously catching&masking the panics in the LogStreamSSE. EventStreamSSE was missing the same

fixes: #6454


server/api/stream_test.go
```diff
@@ -0,0 +1,75 @@
...
+		topic := map[string]struct{}{pubsub.PublicTopic: {}}
+
+		done := make(chan struct{})
+		go func() {
+			defer close(done)
+			EventStreamSSE(c)
+		}()
+
+		// Let the event handler subscribe
+		time.Sleep(20 * time.Millisecond)
```
utafrali: `time.Sleep(20ms)` for synchronization is fragile. Under a loaded CI runner, the goroutine spawned inside `EventStreamSSE` may not have reached the `Subscribe` call (and registered the receiver in the pubsub map) before the sleep expires, meaning the test publishes messages before the subscription is live and misses the race it is trying to reproduce.

A more reliable approach is to instrument the scheduler with a hook or use a channel to signal readiness. For example, the `memory.Subscribe` registers the receiver before blocking, so you could wrap the broker to notify a ready channel on the first Subscribe call:
```go
type notifyingBroker struct {
    pubsub.PubSub
    subscribed chan struct{}
}
func (n *notifyingBroker) Subscribe(c context.Context, t pubsub.Topics, r pubsub.Receiver) error {
    close(n.subscribed)
    return n.PubSub.Subscribe(c, t, r)
}
```
This removes the timing dependency entirely.

server/api/stream.go
```diff
@@ -249,6 +238,7 @@ func LogStreamSSE(c *gin.Context) {
 			log.Error().Err(err).Msg("tail of logs failed")
 		}
 
+		close(batches)
```
utafrali, line 248: `close(batches)` and `cancel(err)` are called back-to-back without waiting for the inner goroutine to finish draining the channel. When `Tail()` returns naturally (end of a finished pipeline), the inner goroutine may still have buffered log entries to serialize and forward to `logChan`. Calling `cancel(nil)` immediately causes the main select loop to exit via `ctx.Done()` and emit the `eof` event before those trailing entries are delivered to the client.

A wait group would fix this:
```go
var innerDone sync.WaitGroup
innerDone.Add(1)
go func() {
    defer innerDone.Done()
    for entries := range batches { ... }
}()

err := server.Config.Services.Logs.Tail(ctx, step.ID, batches)
close(batches)
innerDone.Wait() // drain before signalling eof
cancel(err)
```
Note that the `Wait()` itself is safe because the inner goroutine's select also watches `ctx.Done()`, so if the context is cancelled from the outside (client disconnect) before `Tail()` returns, the inner goroutine will exit promptly and `Wait()` will not block.

server/api/stream_test.go
```diff
@@ -0,0 +1,75 @@
...
+				})
+			}()
+		}
+
+		// Simulate client disconnect mid-publish.
+		cancel()
+		wg.Wait()
+		<-done
+	}
+}
```
utafrali, line 160: Missing newline at end of file (`\ No newline at end of file` in the diff). `gofmt` and most Go tooling expect a trailing newline.

server/api/stream_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/gin-gonic/gin"
+
+	"go.woodpecker-ci.org/woodpecker/v3/server"
+	"go.woodpecker-ci.org/woodpecker/v3/server/pubsub"
+	"go.woodpecker-ci.org/woodpecker/v3/server/pubsub/memory"
+	"go.woodpecker-ci.org/woodpecker/v3/server/scheduler"
+)
+
+func TestEventStreamSSE_ConcurrentDisconnect(t *testing.T) {
```
utafrali: The test covers the `EventStreamSSE` concurrent-disconnect scenario, which is good. However, `LogStreamSSE` was also fixed in this PR (removal of the `recover` block, new select, adding `close(batches)`). A parallel test for `LogStreamSSE` would give confidence that the nested-goroutine path is also race-free. The two-goroutine structure there (outer goroutine wrapping `Tail`, inner goroutine ranging over `batches`) is harder to reason about and more deserving of a race-detector test.

utafrali (approved):
The core fix is correct and is a genuine improvement: removing `close(eventChan)` / `close(logChan)` from defers eliminates the send-on-closed-channel panics, the updated selects properly unblock on context cancellation (better than the old `default`-fallthrough pattern), and adding `close(batches)` fixes a pre-existing goroutine leak in `LogStreamSSE`. The main concerns are test fragility (timing-based sync) and a potential last-entries-lost race when a log stream ends naturally before the inner goroutine drains its buffer.

6543:
now I dont know why there are tow pulls working on the same issue by the same people either authored or reviewed!?!?

-> https://github.com/woodpecker-ci/woodpecker/pull/6455

hhamalai:
> now I dont know why there are tow pulls working on the same issue by the same people either authored or reviewed!?!?
> 
> -> #6455

How you wish to proceed @6543 I suspected the another PR was agent generated as the unit tests still weren't testing against the right fix, and the further commits introduced more regression. Happy to close this one, as the other PR has a good start.

6543:
well i prevere human interaction, if llms are used is fine as long as it's just a tool, the contributor has to own its mistakes and also understand the code ...

I will look into both pulls and then decide what to take from what ... give me time to do so :)

server/api/stream_test.go
6543: fine by me as is

6543 (approved):
nice work

PR #6461, opened 2026-04-17
Support optional flag in depends_on for workflows and steps

### LINKED ISSUE #3823: Skip steps who depend on an failed step witch ignores failure
```yaml
when:
  - event: push
    branch: main

steps:
  failing-step:
    image: alpine
    commands:
      - printenv
      - exit 1
    failure: ignore
  should-not-start:
    image: alpine
    commands:
      - printenv
      - exit 1
    depends_on: 
      - failing-step
```

we should add a new failure mode: `skip_dependencies` or so ...

so we can make checks and skip things based on that ...

blocked by: #3009

### LINKED ISSUE #6457: Skipped workflows are not considered as succeed like documentation suggest it
### Component

agent

### Describe the bug

Documentation says that a workflow that didn't match its `when` condition is skipped, but considered as succeed for workflows depending on it.
https://woodpecker-ci.org/docs/usage/workflow-syntax#when---global-workflow-conditions

> If all conditions in the when block evaluate to true the workflow is executed, otherwise it is skipped, but treated as successful and other workflows depending on it will still continue.

But during my tests, a workflow depending on a skipped workflow is not evaluated.

### Steps to reproduce

Create 3 workflows:

`.woodpecker/test-a.yml`
```
when:
  - event: manual
    evaluate: 'RUN_A == "true" || RUN_C == "true"'

steps:
  - name: test-a
    image: alpine
    commands:
      - echo "Running test-a"
```

`.woodpecker/test-b.yml`
```
when:
  - event: manual
    evaluate: 'RUN_B == "true"'

steps:
  - name: test-a
    image: alpine
    commands:
      - echo "Running test-b"
```

`.woodpecker/test-c.yml`
```
depends_on:
  - test-a
  - test-b

when:
  - event: manual
    evaluate: 'RUN_C == "true"'

steps:
  - name: test-c
    image: alpine
    commands:
      - echo "Running test-c"
```

If you run with `RUN_C=true`, only `test-a` workflow run.

### Expected behavior

Based on documentation, `test-b` being skipped as it didn't match its `when` condition, it should still be considered as succeed to allow following workflow to run. So `test-c` should run after `test-a`.

### System Info

```shell
Woodpecker 3.13.0
```

### Additional context

The reason I'm trying to have this working is that, my real case condition to run A and B are not the same. And I need C to run only once A and B succeed in case they were triggered.

### Validations

simonmeyerrr:
Fixes https://github.com/woodpecker-ci/woodpecker/issues/6457
Fixes https://github.com/woodpecker-ci/woodpecker/issues/3823

## Summary

Extends `depends_on` to accept objects with `name` and `optional` fields, at both workflow and step level. When `optional: true`, the dependency is silently dropped if the referenced workflow/step is not part of the pipeline (e.g. filtered out by `when` conditions). If present, it is enforced as usual.

This is useful in monorepos where workflows use `when: path` or `when: evaluate` to only run when relevant files change, but a downstream workflow (like deploy) needs to wait for all checks that actually ran.

## Example

```yaml
# .woodpecker/deploy.yaml
depends_on:
  - check-a
  - name: check-b
    optional: true
  - name: check-c
    optional: true

steps:
  - name: deploy
    image: alpine
    commands:
      - echo deploying
```

If `check-b` and `check-c `are filtered out by their when conditions, deploy runs after check-a succeeds. If they are present, deploy waits for all three.

The same syntax works at step level within a workflow.

All existing formats (string, string[]) remain fully backward compatible.

## Changes

- Add `Dependency` struct and `DependsOn` type (base/depends_on.go) with custom YAML marshal/unmarshal supporting string, string array, object array, and mixed array
- Use `DependsOn` type in both `Workflow` and `Container` (step) structs
- At build time, drop optional deps referencing non-existent workflows/steps (no runtime/queue changes needed)
- Update DAG compiler to handle optional step dependencies
- Update JSON schema, linter, and docs
- Add unit tests for the new type (unmarshal, marshal, helpers), DAG compiler (3 step-level cases), and step builder (4 workflow-level cases)
- Add e2e scenarios for both workflow-level and step-level optional dependencies


6543 (dismissed):
I'm against a new top level field and would rather alter tje depends on in a unified way ... so steps also can use it etc...

6543:
e.g. current `depends_on` allows string and strin array.
new could also allow object with name and mode field ...

-> need more consensus by all maintainers first

6543:
looking at https://github.com/woodpecker-ci/woodpecker/issues/3823#issuecomment-2185365592  there is already consensus howbit should look like

qwerty287:
Thanks! I also agree to @6543, would prefer a config like
```yaml
depends_on: 
  - name: previous-step
    optional: true
```

And I think we should also support that on step level.

simonmeyerrr:
Updated per feedback from @6543 and @qwerty287.

The separate `optional_depends_on` field is gone. Instead, `depends_on` now accepts objects alongside strings:

```yaml
depends_on:
  - check-a
  - name: check-b
    optional: true
```

This works at both **workflow level** and **step level**, as suggested. Plain strings and string arrays still work as before.

Implementation follows the approach from [#3823 (comment)](https://github.com/woodpecker-ci/woodpecker/issues/3823#issuecomment-2185365592). The new `DependsOn` type lives in `base/depends_on.go` and is shared by both `Workflow` and `Container`. Optional deps are resolved early (build time for workflows, compile time for steps), so no runtime/queue changes were needed.

6543:
NOTE: I want to get v3.14.0 released before merging it in any case...

pipeline/frontend/yaml/constraint/depends_on.go
```diff
@@ -0,0 +1,164 @@
...
+
+import (
+	"errors"
+	"fmt"
+)
+
+// Dependency represents a single dependency with an optional flag.
+// When optional is true, the dependency is silently dropped if the
+// referenced step or workflow is not present in the pipeline.
+type Dependency struct {
```
6543, line 25: I would put this more into the `constraint` package where When also lives

6543:
we should make this ready ... i wana have this in for next release

6543 (changes requested):
I like how it is ... but this is breaking ... for workflow dag!!

before if dep was missing we where ignoring it .... it looks like

~~need to look into it more~~

looks like it is as before ...

6543 (approved):

6543:
@qwerty287 I think it is ready ... but you as release manager should decide if we should merge it and release it now or in the next release ...

PS: on release docs pull need update ...

qwerty287:
I'll check it tomorrow.

pipeline/frontend/yaml/constraint/depends_on.go
```diff
@@ -0,0 +1,170 @@
+// Copyright 2023 Woodpecker Authors
```
qwerty287:
```suggestion
// Copyright 2026 Woodpecker Authors
```

pipeline/frontend/yaml/constraint/depends_on_test.go
```diff
@@ -0,0 +1,178 @@
+// Copyright 2023 Woodpecker Authors
```
qwerty287:
```suggestion
// Copyright 2026 Woodpecker Authors
```

qwerty287 (approved):

PR #6470, opened 2026-04-19
Make agent reconnect retry timeout configurable

6543:
followup https://github.com/woodpecker-ci/woodpecker/pull/6414#discussion_r3062777602

also fix https://ci.woodpecker-ci.org/repos/3780/pipeline/33572/39


qwerty287:
Just wondering, we don't use the backoff library there right?

6543:
we do use backoff lib:
https://github.com/woodpecker-ci/woodpecker/blob/52ed3f13edbd0f822134914e0a6e279898a0af57/agent/rpc/client_grpc.go#L130
https://github.com/woodpecker-ci/woodpecker/blob/52ed3f13edbd0f822134914e0a6e279898a0af57/agent/rpc/client_grpc.go#L119
https://github.com/woodpecker-ci/woodpecker/blob/52ed3f13edbd0f822134914e0a6e279898a0af57/agent/rpc/client_grpc.go#L96

qwerty287:
But shouldn't it be possible to use their internal logic with https://pkg.go.dev/github.com/cenkalti/backoff/v5#WithMaxElapsedTime or https://pkg.go.dev/github.com/cenkalti/backoff/v5#WithMaxTries?

6543:
I can look into that but that would be a refactoring followup pull :)

qwerty287 (approved):
Is ok for me, but remind that you then likely have to change the env var behavior which is breaking as the backoff lib doesn't support cumulated timeout. It can only limit the number of retries (which is what we usually do, so I'd go with that) or the duration between two tries.

PR is not tested at my side

6543:
The lib should support max timeout period ... but even if not, we just have it in our code. Limiting to featureset of an external dependency is not something we should do if easy to make it ourselfe work.

anyway, I'll look into using the backoff lib  more directly

qwerty287:
You're right, the correct function you should use is https://pkg.go.dev/github.com/cenkalti/backoff/v5#WithMaxElapsedTime I guess

PR #6479, opened 2026-04-21
e2e test wait for grpc server teardown and stop agents

6543:
fix race of https://ci.woodpecker-ci.org/repos/3780/pipeline/33699/39

we also now stop agents gracefully

### DIFF
--- a/agent/rpc/client_grpc.go
+++ b/agent/rpc/client_grpc.go
@@ -70,11 +70,6 @@ func NewGrpcClient(ctx context.Context, conn *grpc.ClientConn) rpc.Peer {
 	return client
 }
 
-func (c *client) Close() error {
-	close(c.logs)
-	return c.conn.Close()
-}
-
 func (c *client) IsConnected() bool {
 	state := c.conn.GetState()
 	connected := state == connectivity.Ready || state == connectivity.Idle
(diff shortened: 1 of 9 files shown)


agent/rpc/client_grpc.go
```diff
@@ -506,9 +501,10 @@ func (c *client) processLogs(ctx context.Context) {
 		bytes = 0
 	}
 
-	// ctx.Done() is covered by the log channel being closed
 	for {
 		select {
+		case <-ctx.Done():
```
6543, line 506: note: send wont work if context is closed ... so we can just return here anyway

e2e/setup/agent.go
```diff
@@ -128,8 +128,8 @@ func StartAgent(ctx context.Context, t *testing.T, grpcAddr string, opts ...Agen
 		Timeout: shortTimeout,
 	})
 
-	authCtx, authCancel := context.WithCancelCause(context.Background())
-	t.Cleanup(func() { authCancel(nil) })
+	agentCtx, agentCancel := context.WithCancelCause(context.Background())
```
qwerty287: Isn't it possible to use

```suggestion
	agentCtx, agentCancel := context.WithCancelCause(t.Context())
```

here?

qwerty287 (approved):

PR #6480, opened 2026-04-21
Cleanup e2e test setup

6543:
the initial e2e test works fine but had quite some boilerplate code and shared code ... we now clean up to make it more maintainable

followup of https://github.com/woodpecker-ci/woodpecker/pull/6391


xoxys (approved):
LGTM but I would suggest to merge after release.

PR #6487, opened 2026-04-23
Add 'entrypoint' property to service in schema

rhafer:
The linter is complaining about 'entrypoint' being disallowed in 'services'. However, apart from the linter warning it seems to work just fine. This adds it to the schema.


6543 (approved):
yes services can use entrypoints.

Just one catch: If you use secrets it's same as steps, if you define an entrypoint it's no longer a plugin :)

PR #6510, opened 2026-04-27
docs: bump uuid to v14.0.0

6543:
fix https://ci.woodpecker-ci.org/repos/3780/pipeline/33846/8


6543:
would be nice if we can get that in ... It starts to anoi me to only see read because of that :laughing: 

and docs build just fine (local tested)


=> it's merge also will trigger to create the changelog for the final release to be created :)

xoxys (approved):

PR #6516, opened 2026-04-28
pipeline runtime: move setting step environment variables into own func and add `CI_PIPELINE_STATUS` back

### LINKED ISSUE #4337: Allow notification plugins to notify about the pipeline status

The env var `CI_PIPELINE_STATUS` war broken already and has always reported `success` but was now removed completely in https://github.com/woodpecker-ci/woodpecker/pull/3846

Right now, I don't see any way to use a notification plugin in pipelines to send build status notifications to an external system.

### Suggested solution

Add back `CI_PIPELINE_STATUS` and populate the correct status. We have the correct pipeline status in the UI/API already, why can't we just populate it to the env?

### Alternative

_No response_

### Additional context

_No response_

### Validations

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/4337


6543:
note: blocked till release

xoxys (approved):
Code LGTM but untested.

PR #6517, opened 2026-04-28
docs: better description for when.status filter

### LINKED ISSUE #6512: when.status missing a list of possible statuses
### Component

other

### Describe the bug

No clear list of the statuses that can be used in `when.status` in ["next" documentation](https://woodpecker-ci.org/docs/next/usage/workflow-syntax) as well as in [0.13.x documentation](https://woodpecker-ci.org/docs/usage/workflow-syntax#status) that is confusing. Are the `succes` and `failure` only possible statuses? If no, then what are the other statuses.

### Steps to reproduce

Check the documentation.

### Expected behavior

The documentation has a clear list of possible statuses that can be used in `when.status` or a two lists if some statuses cannot be used.

### System Info

```shell
The information is absent on both the ["next" documentation](https://woodpecker-ci.org/docs/next/usage/workflow-syntax) as well as on the [0.13.x documentation](https://woodpecker-ci.org/docs/usage/workflow-syntax#status).
```

### Additional context

_No response_

### Validations

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/6512

### DIFF
--- a/docs/docs/20-usage/20-workflow-syntax.md
+++ b/docs/docs/20-usage/20-workflow-syntax.md
@@ -342,7 +342,14 @@ when:
 
 #### `status`
 
-There are use cases for executing steps on failure, such as sending notifications for failed workflow/pipeline. Use the status constraint to execute steps even when the workflow fails:
+By default, steps only run when the workflow has succeeded up to that point,
+which is equivalent to `status: [ success ]`.
+
+The `status` filter lets you override this behavior.
+The only accepted values are `success` and `failure`.
+
+A common use case is executing a step on failure, such as sending notifications for a failed workflow/pipeline.
+To run a step regardless of outcome, list both values:
 
 ```diff
  steps:
--- a/docs/docs/20-usage/20-workflow-syntax.md
+++ b/docs/docs/20-usage/20-workflow-syntax.md
@@ -342,7 +342,7 @@ when:
 
 #### `status`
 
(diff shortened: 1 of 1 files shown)


docs/docs/20-usage/20-workflow-syntax.md
```diff
@@ -342,7 +342,14 @@ when:
 
 #### `status`
 
-There are use cases for executing steps on failure, such as sending notifications for failed workflow/pipeline. Use the status constraint to execute steps even when the workflow fails:
+By default, steps only run when the workflow has succeeded up to that point,<br>
+which is equivalent to `status: [ success ]`.
```
DaniilKl, line 346:
```suggestion
By default, steps only run when the workflow has succeeded up to that point,
which is equivalent to `status: [ success ]`.
```

IMO this `<br>` is surplus here.

6543: this is intentional as else the automatic linebreak will cut of the last 3 chars :)

qwerty287: @6543 what do you mean, how can the automatic break cut off chars?

6543: Image

6543: @qwerty287

qwerty287 (approved):

PR #6518, opened 2026-04-28
server: add websocket based stream api as SSE alternative

### LINKED ISSUE #6511: UI: too many open tabs lock WebUI
If you open to many tabs at some point the UI just freeze and other tabs loading will load infinite.

Only closing the tab that starts the lock will get things going again :/

version: v3.12.0 (oder not tested) ... v3.14.0.rc2

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/6511

This moves the SSE in a subpath and adds WebSocked based API that provides the same streams.
The WebUI now by default uses WebSocked and falls back to SSE if WS upgrade fails.

### if SSE fallback happens, we surface an error:
image


server/api/stream_ws.go
```diff
@@ -0,0 +1,279 @@
...
+		log.Debug().Err(err).Msg("user feed: websocket accept failed")
+		return
+	}
+	// CloseNow on defer guarantees the underlying TCP connection is released
+	// even if the normal close handshake did not complete.
+	defer conn.CloseNow()
+
+	log.Debug().Msg("user feed: websocket connection opened")
+
+	user := session.User(c)
```
6543, line 71: i see dublicated code ...  between SSE and WS implementation 🤔

server/api/stream_ws.go
```diff
@@ -0,0 +1,279 @@
...
+		log.Error().Err(err).Msg("log stream: open failed")
+		_ = conn.Close(websocket.StatusInternalError, "can't open stream")
+		return
+	}
+
+	go func() {
+		batches := make(logging.LogChan, maxQueuedBatchesPerClient)
+
+		var innerDone sync.WaitGroup
+		innerDone.Add(1)
```
6543, line 217: also:  dublicated code ... between SSE and WS

server/router/api.go
```diff
@@ -255,6 +255,16 @@ func apiRoutes(e *gin.RouterGroup) {
 				session.MustPull,
 				api.LogStreamSSE)
 			stream.GET("/events", api.EventStreamSSE)
+
+			// WebSocket variants — same payloads as SSE, but each open
+			// connection only consumes one slot in the browser's per-origin
+			// HTTP/1.1 connection budget shared with regular requests.
+			stream.GET("/logs/:repo_id/:pipeline/:step_id/ws",
```
6543: we should use a prefix not append ...

-> SSE will be `/stream/sse/...` (and keep non sse as backwards till we break...) we might remove SSE 🤔 if WS works well ...
-> WS in `/stream/ws/...`

6543: # this is important

for reverse-proxys that need explizite WS config! (nginx, Apache, ...)

-> that's why i think we should have SSE ass fallback mechanism as it is way more robust

6543:
fallback works:
image

tested with:
```diff
diff --git a/server/api/stream_ws.go b/server/api/stream_ws.go
index a39a44d90..2db7c8d43 100644
--- a/server/api/stream_ws.go
+++ b/server/api/stream_ws.go
@@ -147,6 +147,12 @@ func LogStreamWS(c *gin.Context) {
        _store := store.FromContext(c)
        repo := session.Repo(c)
 
+       select {
+       case <-c.Done():
+               return
+       case <-time.After(time.Minute):
+       }
+
        // Validate parameters BEFORE upgrading. Errors are returned as plain HTTP
        // responses, matching how a fetch() client expects auth/validation failures.
        pipelineNum, err := strconv.ParseInt(c.Param("pipeline"), 10, 64)
```

docs/docs/30-administration/10-configuration/10-server.md
```diff
@@ -218,6 +281,10 @@ woodpecker-agent.example.com {
 }
 ```
 
+Caddy's `reverse_proxy` directive forwards WebSocket upgrade headers
```
6543: docs need more love :)

server/api/stream_ws.go
6543: I looked into code dedupe but it get's undredable fast: https://github.com/6543-forks/woodpecker/pull/5
-> this needs more thoughts...

docs/docs/30-administration/10-configuration/10-server.md
```diff
@@ -218,6 +281,10 @@ woodpecker-agent.example.com {
...
+automatically, so the WebSocket-based live updates (`/stream/ws/`) work with
+the configuration above without any extra directives.
+
```
6543:
```suggestion
```

PR #6523, opened 2026-04-30
fix(web): escape HTML in commit messages to prevent XSS

### LINKED ISSUE #6522: Incorrect escaping of commit messages in web ui activity log / pipeline page
### Component

web-ui

### Describe the bug

Hi,

One of the commit messages in the project I'm working on looks like this:

```
<input type="color" /> → <ColorPicker />
```

which, when it was pushed to gitea and triggered an event in Woodpecker, resulted in Woodpecker web ui showing this:

<details>

<summary>screenshots</summary>

Image

Image

and for `type="password"`:

Image

</details>

Thankfully, `<script>` gets removed and does not trigger this.

These versions were tested:

- 3.14.0-rc.2
- next-c1aea9c4ec

I have no experience with Vue, but if nobody else takes this, maybe I will have a look in the next few days...

### Steps to reproduce

1. register a git repository in Woodpecker
2. create a commit there with some HTML in its message
3. open web ui
4. see said HTML in activity log & pipeline status page

### Expected behavior

HTML should be escaped.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.14.0-rc.2"}
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-c1aea9c4ec"}
```

### Additional context

_No response_

### Validations

wucm667:
Fixes #6522

## Bug
When a commit message contains HTML tags like `<input type="color" />`,
the Woodpecker web UI renders them as actual HTML elements instead of
displaying them as plain text. This is a potential XSS vulnerability
in the activity log and pipeline pages.

## Root Cause
The `RenderMarkdown` component uses `v-html` with DOMPurify configured
with `{ USE_PROFILES: { html: true } }`, which allows safe HTML elements
like `<input>` to be rendered. Commit messages pass through this component
without prior HTML escaping.

## Fix
Added HTML escaping (`<` → `&lt;`, `>` → `&gt;`, etc.) to the `message`
and `title` computed properties in `usePipeline.ts`. This ensures all
special characters are displayed as plain text before reaching the
markdown renderer.

## Verification
- TypeScript type check passes
- Vite build succeeds
- ESLint passes on modified file

### DIFF
--- a/web/src/compositions/usePipeline.ts
+++ b/web/src/compositions/usePipeline.ts
@@ -6,6 +6,7 @@ import { useI18n } from 'vue-i18n';
 import { useDate } from '~/compositions/useDate';
 import { useElapsedTime } from '~/compositions/useElapsedTime';
 import type { Pipeline } from '~/lib/api/types';
+import { escapeHtml } from '~/lib/utils';
 
 const { toLocaleString, timeAgo, prettyDuration } = useDate();
 
@@ -74,15 +75,6 @@ export default (pipeline: Ref<Pipeline | undefined>) => {
 
     return prettyDuration(durationElapsed.value);
   });
-
-  function escapeHtml(text: string): string {
-    return text
-      .replace(/&/g, '&amp;')
-      .replace(/</g, '&lt;')
-      .replace(/>/g, '&gt;')
-      .replace(/"/g, '&quot;')
-      .replace(/'/g, '&#x27;');
-  }
 
   const message = computed(() => emojify(escapeHtml(pipeline.value?.message ?? '')));
   const shortMessage = computed(() => message.value.split('\n')[0]);
--- a/web/src/compositions/usePipeline.ts
+++ b/web/src/compositions/usePipeline.ts
@@ -75,10 +75,19 @@ export default (pipeline: Ref<Pipeline | undefined>) => {
     return prettyDuration(durationElapsed.value);
   });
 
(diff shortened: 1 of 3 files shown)


wucm667:
Hey, I noticed this PR was closed. Just wanted to clarify — I'm a real dev, the bot flagged me probably because I forked a bunch of repos to find good first issues. Not spamming.

This is a legit XSS fix for #6522. Could you take another look? Happy to adjust if needed.

web/src/compositions/usePipeline.ts
```diff
@@ -75,10 +75,19 @@ export default (pipeline: Ref<Pipeline | undefined>) => {
     return prettyDuration(durationElapsed.value);
   });
 
-  const message = computed(() => emojify(pipeline.value?.message ?? ''));
+  function escapeHtml(text: string): string {
```
6543: this helper should go into utils ...

6543:
also some tests might be nice...

6543:
> I'm a real dev

sorry, we just get flooded more and more. happy to work with humans :rocket:

wucm667:
Done. Moved `escapeHtml` to `lib/utils` and added unit tests for it. Let me know if anything else needs adjusting.

6543:
image

6543 (approved):

PR #6526, opened 2026-05-01
fix(k8s): respect WOODPECKER_PLUGINS_PRIVILEGED for Kubernetes backend

### LINKED ISSUE #6484: bug: WOODPECKER_PLUGINS_PRIVILEGED and WOODPECKER_ESCALATE have no effect on Kubernetes backend
### Component

server

### Describe the bug

  `WOODPECKER_PLUGINS_PRIVILEGED` and `WOODPECKER_ESCALATE` have no effect
  when using the Kubernetes backend. Setting or removing these variables does
  not change the behavior of pipeline steps at all.     

### Steps to reproduce

  1. Set `WOODPECKER_PLUGINS_PRIVILEGED=myimage:latest` on the server                                                           
  2. Run a pipeline step using `image: myimage:latest` without `privileged: true` in the YAML
  3. Step runs without privileges regardless of the variable                                                                    

### Expected behavior

  1. Set `WOODPECKER_PLUGINS_PRIVILEGED=myimage:latest` on the server                                                           
  2. Run a pipeline step using `image: myimage:latest` without `privileged: true` in the YAML
  3. Step runs without privileges regardless of the variable                                                                    

### System Info

```shell
- Woodpecker version: v3.13.0
  - Backend: Kubernetes (v1.32.2)
```

## Workaround                                                                                                                 
                  
  The only working approach:
  1. Mark the repo as **Trusted** in the admin UI                                                                               
  2. Add `privileged: true` explicitly to the step YAML
                                                                                                                                
  ## Notes        
          
  Tested by setting the variable, removing it, and running pipelines —
  identical behavior in all cases.                                                                                              

### Validations

wucm667:
Fixes #6484

## Problem

`WOODPECKER_PLUGINS_PRIVILEGED` and `WOODPECKER_ESCALATE` have no effect when using the Kubernetes backend on non-trusted repos. The linter's `lintTrusted` check blocks privileged mode for all plugins on non-trusted repos, regardless of whether they are in the privileged plugins allowlist.

## Root Cause

The `lintTrusted` function in `pipeline/frontend/yaml/linter/linter.go` checks if `c.Privileged` is true and blocks it on non-trusted repos. However, it doesn't consider whether the plugin image is in the `WOODPECKER_PLUGINS_PRIVILEGED` allowlist.

This means that even if a plugin is explicitly configured as privileged via the environment variable, the linter still blocks it unless the repo is marked as trusted.

## Solution

Modified `lintTrusted` to skip the privileged check if the container image matches an entry in the `privilegedPlugins` allowlist. This allows explicitly configured privileged plugins to run on non-trusted repos, matching the expected behavior of `WOODPECKER_PLUGINS_PRIVILEGED`.

## Changes

- `pipeline/frontend/yaml/linter/linter.go`: Updated `lintTrusted` to check if the image is in `privilegedPlugins` before blocking privileged mode

### DIFF
--- a/pipeline/frontend/yaml/linter/linter.go
+++ b/pipeline/frontend/yaml/linter/linter.go
@@ -242,7 +242,9 @@ func (l *Linter) lintTrusted(config *WorkflowConfig, c *types.Container, area st
 	yamlPath := fmt.Sprintf("%s.%s", area, c.Name)
 	errors := []string{}
 	if !l.trusted.Security {
-		if c.Privileged {
+		// Allow plugins in WOODPECKER_PLUGINS_PRIVILEGED to run in privileged mode
+		// even on non-trusted repos, since they are explicitly configured as trusted
+		if c.Privileged && (l.privilegedPlugins == nil || !utils.MatchImageDynamic(c.Image, *l.privilegedPlugins...)) {
 			errors = append(errors, "Insufficient trust level to use `privileged` mode")
 		}
 	}
--- a/web/src/compositions/usePipeline.ts
+++ b/web/src/compositions/usePipeline.ts
@@ -6,6 +6,7 @@ import { useI18n } from 'vue-i18n';
 import { useDate } from '~/compositions/useDate';
 import { useElapsedTime } from '~/compositions/useElapsedTime';
 import type { Pipeline } from '~/lib/api/types';
+import { escapeHtml } from '~/lib/utils';
 
 const { toLocaleString, timeAgo, prettyDuration } = useDate();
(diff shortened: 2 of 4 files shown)


6543:
the linter is not the problem!
setting plugins explizite is desired

6543:
@wucm667 you need to rebase your fix and make a real fix ...

wucm667:
@6543 Thanks for the review. I understand the linter change is not the right approach — closing this PR and will open a separate one for the frontend XSS fix.

PR #6550, opened 2026-05-07
fix(agent): warn when agent and server versions do not match

### LINKED ISSUE #6547: Agents don't check they are on the same version as server
### Component

agent

### Describe the bug

We had some interesting issues because our playbooks updated only the server and not the server from 3.13 to 3.14.
If incompatibilities are expected, agents should not connect/pick up tasks from the server

### Steps to reproduce

1. Install woodpecker server 3.14 and woodpecker 3.13. Even though pipelines will run, the status is not correctly updated and errors appear in the server logs.

### Expected behavior

_No response_

### System Info

```shell
Server 3.14/Agents 3.13
```

### Additional context

_No response_

### Validations

wucm667:
Fixes #6547

## Summary

When a Woodpecker agent connects to a server with a different application version (e.g. server 3.14 with agent 3.13), the mismatch can cause subtle failures such as pipeline status not being updated correctly. Currently there is no version compatibility check.

This change adds version comparison at two points:

- **Agent side** (`cmd/agent/core/agent.go`): after the existing gRPC version check, compare the server's reported `ServerVersion` against the agent's own version
- **Server side** (`server/rpc/rpc.go`): in `RegisterAgent()`, compare the agent-reported version against the server version

Both checks log a **warning** when versions differ (skipped when either side is `"dev"` from local builds). Connections are NOT blocked to allow patch-level compatibility, but users are clearly alerted to the mismatch.

## Changes

| File | Change |
|------|--------|
| `cmd/agent/core/agent.go` | Add version comparison after gRPC version check |
| `server/rpc/rpc.go` | Add version validation in RegisterAgent() |

## Test plan

- [x] `go build ./cmd/agent/... ./server/...` — compiles cleanly
- [x] `go test ./cmd/agent/core/... -count=1` — all tests pass
- [x] `go test ./server/rpc/... -count=1` — all tests pass

## Does this introduce a user-facing change?

Yes — agents and server will now log a warning when their versions don't match, helping users diagnose subtle pipeline status issues caused by version mismatch.


6543 (changes requested):
it is about grpc calls.

The implementation should use grpc metadata to add and check grpc version.

a complete version check is already in place on agent start it only is about agents that keep running while server updates. so Next(), Extend(),... must check while Version() must work anyway ...

johanvdw:
Do I understand correctly that this will just log to docker logs/... ? I think ideally we should show this also on the web interface.

6543:
update the issue with more [info](https://github.com/woodpecker-ci/woodpecker/issues/6547#issuecomment-4396695838)

wucm667:
Thanks @6543 for the detailed guidance — your direction makes sense. I'll rework the PR along these lines:

- Send the server version via gRPC metadata (response header) on every RPC call, instead of only checking once at agent startup / `RegisterAgent`.
- Add a client-side interceptor in the agent that compares versions per call:
  - `Next()` / `Init()` mismatch → confirm via `Version()`, log an error and gracefully shut down.
  - `Extend()` / `Done()` mismatch → log a warning but let the current workflow finish.
  - `Version()` itself stays unaffected.
- On the server side, sanitize the agent-reported version and ignore garbage instead of trusting `info.Version` directly.

@johanvdw — agreed that surfacing this in the web UI would be valuable. I'd like to keep this PR focused on the gRPC-level detection and open a follow-up for the UI side; the agent version is already persisted on the server, so it should be mostly a frontend change.

Converting to draft while I rework.

wucm667:
@6543 thanks, redid it.

Server now puts a `server-version` metadata header on every response, and the agent has a client interceptor that checks it on each call. `Next`/`Init` mismatch re-confirms via `Version()` and cancels the agent ctx for a graceful shutdown; `Extend`/`Done`/`Update`/`Log` etc. just warn so a running workflow can still finish. `Version` is skipped on purpose, and `dev` builds on either side disable the check.

Also added server-side sanitization for the agent-reported version (length cap + char whitelist) before it gets stored.

Tests cover the per-method policy and the sanitizer.

@johanvdw agent version is already in the DB, so a UI badge should be mostly a frontend change. Mind if I do that as a follow-up after this lands?

Will flip back to ready once CI's allowed to run.

johanvdw:
> @johanvdw agent version is already in the DB, so a UI badge should be mostly a frontend change. Mind if I do that as a follow-up after this lands?
> 

Sure, I'm already more than happy that people start working so fast on an issue I reported!

qwerty287:
Couldn't we basically replace the rpc version in https://github.com/woodpecker-ci/woodpecker/blob/main/rpc/proto/version.go#L19 by the woodpecker version? That would make it more strict and fails instead of warning, but I don't think that's bad. We get so many reports where the users just miss to update both so I think we can also fail there.

6543:
note we should stick to grpc version as we had version ranges that are interoperable an so upgrading incremental can be made there and we should not just randombly add more restrictions to prohibit such features

wucm667:
@qwerty287 tempting to just hard-fail, but @6543's point about incremental upgrades is key — the gRPC proto version supports range matching (e.g. server 3.14 + agent 3.13 still interop), and swapping it for the app version would break that.
Current approach covers both: proto version guards compatibility ranges, app version via metadata gives mismatch alerts without killing a running pipeline mid-flight.

qwerty287:
So I won't block this, but I think it's fine to really prevent that users use different server/agent versions.

Yes, there are different versions that can be used together without any issues, but I don't see a good reason to prohibit this. Otherwise we should always increment the grpc version every time we change something so that it really breaks - but as this is tied to various different code changes it's really hard to determine whether a change will break. We could not just checking if the proto files were changed.

Otherwise there's no reason for me to keep the grpc version.

AhmadNajiKam:
suggestion: since both processes communicate through gRPC calls and only that would cause incompatibilities, why not to rely on .proto file checks to ensure the both versions are compatible "enough", like we can add a schema version field to each message or we can compare schema hashes at agent register time to ensure compatibility.

qwerty287:
That doesn't work because the protocol can also change if you don't tough the proto files. For example, we send json data sometimes and when we change the json schema, the proto doesn't change but it's not compatible anymore.

AhmadNajiKam:
@qwerty287 Good point.
question: why don't we use some Schema Registry service? Maybe Buf Schema Registry or Confluent for the data schema we'll use in the messages?

6543:
we have an elegant solution: https://github.com/woodpecker-ci/woodpecker/pull/7028

qwerty287:
@6543 as I wrote in the issue already, the solution in #7028 is different. It only checks the gRPC proto version, not the exact server/agent version

6543:
but that must be fine ... the gRPC version is the server-agent contract check!

PR #6553, opened 2026-05-08
'Login' is not a verb

ncfavier:
Fixes a minor grammar mistake in most people's first interaction with Woodpecker: 'login' is a noun corresponding to the verb 'to log in'. Note that `cli_login_success` and friends are unchanged because there 'login' is used as a noun.

* http://notaverb.com/login
* https://english.stackexchange.com/a/5303

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -1,7 +1,7 @@
 {
   "cancel": "Cancel",
-  "login_to_woodpecker_with": "Login to Woodpecker with",
-  "login": "Login",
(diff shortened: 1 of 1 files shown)


6543 (changes requested):
I'm aganst it as most software use it without space .. what also would go would be `Log-in` but that looks bad

ncfavier:
I don't think that's true, but even if it were, that would not be a good reason to copy the mistake.

qwerty287:
I agree to @ncfavier would go with this proposed versions

lafriks:
Maybe use "Sign in" as that is more commonly used?

PR #6559, opened 2026-05-09
Pin pnpm to v10.33.4

6543:
Upstream node LTS image was updated with a newer Corepack whose default
"known good" pnpm version jumped to v11. Because we had no `packageManager`
field in `package.json`, Corepack fell back to that default and silently
bumped pnpm to v11 without any explicit change on our side.

Our tooling does not support pnpm v11 yet, which broke CI.

The fix pins pnpm explicitly so we no longer rely on Corepack's moving
default. Renovate will keep the pinned version up to date going forward.

Also updates the nix dev env to the same pinned version for consistency.

Fixes https://ci.woodpecker-ci.org/repos/3780/pipeline/34117

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -2,7 +2,7 @@
   "name": "woodpecker",
   "version": "0.0.0",
   "private": true,
-  "packageManager": "pnpm@10",
+  "packageManager": "pnpm@10.33.4",
   "scripts": {
     "start": "cd ../ && make generate-docs && cd docs && docusaurus start",
     "build": "pnpm build:woodpecker-plugins && docusaurus build",
--- a/docs/package.json
+++ b/docs/package.json
@@ -2,6 +2,7 @@
   "name": "woodpecker",
   "version": "0.0.0",
   "private": true,
+  "packageManager": "pnpm@10",
   "scripts": {
     "start": "cd ../ && make generate-docs && cd docs && docusaurus start",
     "build": "pnpm build:woodpecker-plugins && docusaurus build",
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1775662680,
-        "narHash": "sha256-N6F+lC0JNTcj9qP6hneP28xkM62C2ld4kexRv9VuTsg=",
+        "lastModified": 1778351894,
+        "narHash": "sha256-7r2iJchc8Ujmld6pd3nQFXE504Ur1apvZaZBdoA/YD4=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "81df7a1e2eefe78b4727c8f3d66293300bf193d1",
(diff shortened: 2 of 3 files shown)


.woodpecker/binaries.yaml
```diff
@@ -8,7 +8,7 @@ when:
 
 variables:
   - &golang_image 'docker.io/golang:1.26'
-  - &node_image 'docker.io/node:24-alpine'
+  - &node_image 'docker.io/node:24.11-alpine'
```
6543:
```suggestion
  - &node_image 'docker.io/node:24.11.1-alpine'
```

.woodpecker/docker.yaml
```diff
@@ -1,6 +1,6 @@
 variables:
   - &golang_image 'docker.io/golang:1.26'
-  - &node_image 'docker.io/node:24-alpine'
+  - &node_image 'docker.io/node:24.11-alpine'
```
6543:
```suggestion
  - &node_image 'docker.io/node:24.11.1-alpine'
```

.woodpecker/web.yaml
```diff
@@ -5,7 +5,7 @@ when:
       - release/*
 
 variables:
-  - &node_image 'docker.io/node:24-alpine'
+  - &node_image 'docker.io/node:24.11-alpine'
```
6543:
```suggestion
  - &node_image 'docker.io/node:24.11.1-alpine'
```

.woodpecker/docs.yaml
```diff
@@ -1,6 +1,6 @@
 variables:
   - &golang_image 'docker.io/golang:1.26'
-  - &node_image 'docker.io/node:24-alpine'
+  - &node_image 'docker.io/node:24.11-alpine'
```
6543:
```suggestion
  - &node_image 'docker.io/node:24.11.1-alpine'
```

6543:
https://github.com/nodejs/node/commit/f3d50d417cf24b2d695d5e59b44d67d4d983617c

xoxys:
How was this introduced in the first place? And why does pnpm pinning fixes it?

6543:
upstream node just chnaged it's default config to serve pnpm version 11

but our tooling is not setup to support it.

if we tell node to use a version we pinned it runs it witj that

so the symptom is pnpm fails the cause is that pnpm was silently bumped to v11

6543:
... the cure is to make this not use default setting but our pined config witch renovate now can maintian ...

xoxys (approved):

xoxys:
Thanks. Still wondering why an lts node container silently bumps pnpm major

lafriks (approved):

6543:
yes! thats breaking behaviour ...

anyway will update the description now to document it better...

6543:
... updated

6543:
nice now renovate does it's job: https://github.com/woodpecker-ci/woodpecker/pull/6561

PR #6573, opened 2026-05-13
format with gofumpt v0.10.0

6543:
just format, no more no less

-> https://github.com/woodpecker-ci/woodpecker/pull/6558


xoxys (approved):

xoxys:
Should we add a diff check to CI to cover it next time it got updated?

qwerty287:
golangci-lint should do this, but likely their internal gofumpt version is not updated yet

6543:
yes in long run golangci will get the bump too ...

PR #6577, opened 2026-05-13
WebUI: show runtime workflow errors

### LINKED ISSUE #3059: Show workflow error in UI

from https://github.com/woodpecker-ci/woodpecker/issues/2937 and https://github.com/woodpecker-ci/woodpecker/pull/2945 and https://github.com/woodpecker-ci/woodpecker/issues/4365

### Suggested solution

display https://github.com/woodpecker-ci/woodpecker/blob/db45b9032ca73d79d3c4a3f25cb5056b0018b386/server/model/workflow.go#L25 in UI, just like for pipeline and step errors

about UI design: I suggest: If the workflow has an error, don't show the step list, and if you click on the workflow, the error panel is opened instead.

### Alternative

_No response_

### Additional context

_No response_

### Validations

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/3059

image
image
image
image


6543:
image

qwerty287:
Could you explain what the error is in your screenshots? Maybe a "real" example would be nice

Also I think it somehow looks weird to make the logs area smaller (like [here](https://github.com/woodpecker-ci/woodpecker/pull/6577#issuecomment-4445069879)), but I also don't have an idea to do it better

xoxys:
I also have some concerns regarding layout jumps/shifts in case of errors. But would need to test it with some real world examples as well.

lafriks:
If there is work flow errors is there even chance that there will be logs? Imho it would be correct to show just erros and hide log section at all

qwerty287:
I think the error can also happen during cleanup after the steps were run

6543:
examples: workflow setup fails ...
see codeberg chat for example

Screenshot_20260514-112702_Element Classic_1_1_1
Screenshot_20260514-112725_Element Classic_1

6543:
as for the layout, suggestions welcome!

PS: this should normally be hidden anyway

6543:
hey it looks like the layout is the issue ... **any** ideas to itterate on?

qwerty287:
Something that would be nice from my perspective, but it needs way more work:
A specific view for a workflow. There we could view the error and possibly more info about the workflow e.g. configuration file.

Because how you did it, the issue is firstly that you view the error always above the logs so the logs have less space, and secondly it is somehow more shown at a pipeline level right now than at a workflow level. But as said, the workflow level doesn't exist in the UI right now.

qwerty287:
We could also just show the error in the pipeline error page with some note that it's only related to a single workflow if we want to have a quick solution.

I guess most users won't really understand the difference between a pipeline and a workflow error anyways.

6543:
mixing pipeline errors -> compile/build/parse/lint errors and runtime workflow errors is bad UX in my opinion

qwerty287:
Maybe we could introduce two sections there, one "parse errors" and one "runtime errors" where runtime errors are the workflow errors? I think if we have headings it's clear what's what

6543:
@qwerty287 can we merge it now and improve UX later ... this is a real issue to be addressed!

6543:
I'll push a new UI soon ...

6543:
image
image

6543:
image
image

6543:
@qwerty287 used claude fable5 to refactor to your suggestion ... the diff looks good to me and plausible but better be more strict ... with the review

web/src/views/repo/pipeline/Pipeline.vue
```diff
@@ -92,6 +93,16 @@ const stepId = toRef(props, 'stepId');
 
 const defaultStepId = computed(() => pipeline.value?.workflows?.[0].children?.[0].pid ?? null);
 
+// Replace the log view with the error panel for parse errors, or for workflow
+// runtime errors when no step produced logs yet. If steps already ran (e.g.
+// the error happened during cleanup), keep the logs accessible; the errors
+// tab still points to the details.
+const showErrorPanel = computed(
+  () =>
+    pipeline.value?.errors?.some((e) => !e.is_warning) ||
```
qwerty287: Why don't you use the `pipelineHasErrorsToShow` func you defined?

qwerty287:
Thanks, the screenshots look very good to me

web/src/views/repo/pipeline/Pipeline.vue
```diff
@@ -92,6 +93,16 @@ const stepId = toRef(props, 'stepId');
...
+    (workflowsWithErrors(pipeline.value).length > 0 && !anyStepStarted(pipeline.value)),
```
qwerty287: Why is the `!anyStepStarted(pipeline.value)` added here? Shouldn't the tab be shown in any case, no matter if steps ran or not?

web/src/lib/pipeline.test.ts
qwerty287, line 1: I'm not sure, but I do not really think those tests for really simple helper funcs are very helpful. It's a pain to review that in my opinion...

6543: removed

web/src/views/repo/pipeline/Pipeline.vue
6543: because of the second check ... this makes sense

if we use `pipelineHasErrorsToShow` directly, and get a runtime error, but already have logs streamed,
without it they would be inaccessible

6543:
@qwerty287 if you want, i can apply:
```diff
diff --git a/web/src/lib/pipeline.ts b/web/src/lib/pipeline.ts
index 486d26840..e6d779e85 100644
--- a/web/src/lib/pipeline.ts
+++ b/web/src/lib/pipeline.ts
@@ -12,6 +12,10 @@ export function anyStepStarted(pipeline?: Pipeline): boolean {
   );
 }
 
+export function pipelineHasNonWarningErrors(pipeline?: Pipeline): boolean {
+  return pipeline?.errors?.some((e) => !e.is_warning) ?? false;
+}
+
 export function pipelineHasErrorsToShow(pipeline?: Pipeline): boolean {
-  return (pipeline?.errors?.some((e) => !e.is_warning) ?? false) || workflowsWithErrors(pipeline).length > 0;
+  return pipelineHasNonWarningErrors(pipeline) || workflowsWithErrors(pipeline).length > 0;
 }
diff --git a/web/src/views/repo/pipeline/Pipeline.vue b/web/src/views/repo/pipeline/Pipeline.vue
index 9b3df0667..755d08751 100644
--- a/web/src/views/repo/pipeline/Pipeline.vue
+++ b/web/src/views/repo/pipeline/Pipeline.vue
@@ -73,7 +73,7 @@ import { requiredInject } from '~/compositions/useInjectProvide';
 import useNotifications from '~/compositions/useNotifications';
 import { useWPTitle } from '~/compositions/useWPTitle';
 import type { PipelineStep } from '~/lib/api/types';
-import { anyStepStarted, workflowsWithErrors } from '~/lib/pipeline';
+import { anyStepStarted, pipelineHasErrorsToShow, pipelineHasNonWarningErrors } from '~/lib/pipeline';
 
 const props = defineProps<{
   stepId?: string | null;
@@ -97,10 +97,10 @@ const defaultStepId = computed(() => pipeline.value?.workflows?.[0].children?.[0
 // runtime errors when no step produced logs yet. If steps already ran (e.g.
 // the error happened during cleanup), keep the logs accessible; the errors
 // tab still points to the details.
-const showErrorPanel = computed(
-  () =>
-    pipeline.value?.errors?.some((e) => !e.is_warning) ||
-    (workflowsWithErrors(pipeline.value).length > 0 && !anyStepStarted(pipeline.value)),
+const showErrorPanel = computed(() =>
+  anyStepStarted(pipeline.value)
+    ? pipelineHasNonWarningErrors(pipeline.value)
+    : pipelineHasErrorsToShow(pipeline.value),
 );
 
 const selectedStepId = computed({
 ```

web/src/views/repo/pipeline/Pipeline.vue
qwerty287: I see, it's the error panel. Sorry misread that

qwerty287:
Yes I think applying this diff makes it more clear. Thanks

6543:
@qwerty287 done

qwerty287 (approved):
Nice, untested from my side

xoxys:
Code LGTM from my side as well. However how does it look on mobile?

6543:
same as before :)

but here you go:

image
image
image
image

6543:
I think there still might be thing to be refined, but i would say that is for another pull and one must have suggestions anyway ...

qwerty287:
Seems we're a bit verbose at recording workflow errors: Every failing step gives one now

image

https://ci.woodpecker-ci.org/repos/3780/pipeline/36327/errors

6543:
we should fix this in the runtime ... so an ci error is no runtime error

PR #6588, opened 2026-05-16
Add CI_PIPELINE_RERUNS environment variable

### LINKED ISSUE #5876: Add ENV to signall rerun
Normally you want a rerun to replay exactly the environments & settings as the last one ...

but there are edgecases where you want to signall a rerun.

a practicable example would be:
piplime consists of two workflows, one failed and can be rerun any time without issue, the second does do some deployment stuff and MUST not run more than once.

i do have another examole with metadata checks etc but that would need more explaining so i go with the simple example. (ask if interested)

so I **propose** an env var that counts reruns. first run -> non existing
second run `CI_PIPELINE_RERUNS=1` witch is incremented each time a rerun happens

this way we can have filters to check against it ...

a simple bool would have done it too but why not communicate propperly the current state

M31ancholy:
- Add RerunCount field to pipeline model and expose it as CI_PIPELINE_RERUNS environment variable when > 0.
- Add RerunCount field to Pipeline model
- Add database migration for rerun_count column
- Set CI_PIPELINE_RERUNS env var in pipeline metadata
- Add unit and e2e tests

Closes #5876


server/store/datastore/migration/028_add_rerun_count_field.go
```diff
@@ -0,0 +1,39 @@
...
+// limitations under the License.
+
+package migration
+
+import (
+	"src.techknowlogick.com/xormigrate"
+	"xorm.io/xorm"
+)
+
+var addRerunCountField = xormigrate.Migration{
```
6543: new fields dont need migration, will happen automatically :)

6543:
@M31ancholy also wouod be good to resolve conflicts :)

M31ancholy:
I realized that i might also need to update the docs, let me do this

M31ancholy:
> I realized that i might also need to update the docs, let me do this

The changes look huge, but actually I only added a single line to explain CI_PIPELINE_RERUNS. All the other changes were caused by GoLand's Markdown formatter.

6543:
now you just have to address issues pointed out by the ci ...

6543 (approved):

PR #6598, opened 2026-05-17
Deduplicate translations

qwerty287:

### DIFF
--- a/web/src/assets/locales/en.json
+++ b/web/src/assets/locales/en.json
@@ -54,15 +54,7 @@
     "deploy_pipeline": {
       "title": "Trigger a deployment for current pipeline #{pipelineId}",
       "enter_target": "Target environment for deployment",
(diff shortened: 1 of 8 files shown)


6543:
> branch has conflicts

lafriks:
With translation deduplication should be very careful, should probably check all translations if removed translation is not different from replaced one as for example English "disabled" in Latvian language must be translated differently depending on context on what is disabled it coule be either "atspējots" or "atspējota". For these changes  currently I did not see problem at least for Latvian language but could be similarly problematic in other languages

English is quite simple language in this regard as for example in Latvian each noun word has 7 different endings depending on how it's used, adjectives even more depending on noun and usage 😉

xoxys:
I would just keep context scoped translations and live with duplications.

qwerty287:
Ok then I'll close this.

PR #6600, opened 2026-05-17
View warning if admin is configured at env level

qwerty287:
If a user has admin perms because of env config, you can still revoke them but they'll get them back once they sign in.

grafik


server/model/user.go
```diff
@@ -66,6 +66,8 @@ type User struct {
 	// environment variable, this value will be set to true on login.
 	Admin bool `json:"admin,omitempty" xorm:"admin"`
 
+	AdminEnv bool `json:"admin_env,omitempty"`
```
6543: xorm missing

6543: eg. `xorm:"-"` as to not store it

6543 (approved):

PR #6605, opened 2026-05-17
Minor adjustments for API usage example

rfinnie:
- Add a space before the WOODPECKER_TOKEN export. Commands which begin with a space are, in bash, executed but not saved to .bash_history. This is true even of bracketed pastes, so in the example usage, the WOODPECKER_SERVER line will be saved, but not WOODPECKER_TOKEN.
- Replace curl -i with -v. -i displays response headers, but to stdout. -v displays (among others) response headers, but to stderr, so the whole thing can be easily piped to, say, jq. You know what? Let's just add jq to the example. :)
- Move the URL to the end and enclose it in quotes, for easier tinkering and adaptation by the user.
- Remove the commented out curl invocation. /api/user is a pretty safe start to execute as part of a multiline copy/paste by the user.


qwerty287:
Thanks. The `#` was not there to indicate a comment I think, but to indicate the end of the shell prompt. I'd say we either remove this from all examples at this page (including the cli one) or we replace it using `$` or something similar.

I also would not add `jq` there and also would prefer `-i` over `-v` as `-i` is way more easy to read as `-v` also prints request headers which does not help most users for a first start.

rfinnie:
Yeah, I noticed the `#` in the other examples after submitting that. I could go either way, but would lean toward putting it back in for consistency.

Re: `-i`, I would agree on readability if `-i` were sent to stderr, but since it's stdout, things start breaking as soon as you try to do anything with the API result (`|` as mentioned, `>`, etc). I argue `-v` is much more usable because of that, but your call.

qwerty287:
More opinions @woodpecker-ci/maintainers about `-i` vs `-v`? In the end I think it's just an example that should stay simple and you can adopt that to your needs.

6543:
I also thing we should not have `-v` as it is quite verbose (as intended by that flag) and anyone using curl should know about -v anyway ... if he does more with curl ...

qwerty287 (approved):

Issue #1292, opened 2022-10-19
No docs on how to setup using the binary packages


neel:
### Clear and concise description of the problem

No documentation on how to setup the CI using binary packages. The docs only explain how to install it using docker.

### Suggested solution

Provide documentation on how to use binary packages and setup the CI e.g. sample system service scripts with the environment variable and database config.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

rahuldave:
Has anyone tried this? I would like to run woodpecker out of sysctl and am willing to take a swipe at this, is it just a matter of passing environment variables?

neel:
any update ?

neel:
I am trying to use the binaries but `woodpecker-server` crashed while Authorising with Gitea with a long trace

```
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x63 pc=0x7f185082def4]

runtime stack:
runtime.throw({0x1115bd4?, 0x0?})
        /usr/local/go/src/runtime/panic.go:992 +0x71
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_unix.go:802 +0x389

goroutine 68 [syscall]:
runtime.cgocall(0xd17e30, 0xc00051dd98)
        /usr/local/go/src/runtime/cgocall.go:157 +0x5c fp=0xc00051dd70 sp=0xc00051dd38 pc=0x405a1c
net._C2func_getaddrinfo(0xc000367110, 0x0, 0xc000353950, 0xc000489e20)
        _cgo_gotypes.go:94 +0x56 fp=0xc00051dd98 sp=0xc00051dd70 pc=0x691ff6
net.cgoLookupIPCNAME.func1({0xc000367110, 0x0?, 0x0?}, 0xc00030afc0?, 0x69873b?)
        /usr/local/go/src/net/cgo_unix.go:160 +0x9f fp=0xc00051ddf0 sp=0xc00051dd98 pc=0x693d1f
net.cgoLookupIPCNAME({0x10da56c, 0x3}, {0xc00030afc0, 0xe})
[...]

PR #6610, opened 2026-05-18
build: release freebsd/openbsd as binary not container

6543:
we release freebsd/openbsd binaries currently as container image.
but docker does not run well and podman? is an edgecase. so we should focus on a jail based backend.

bsd users now have to either compile wp on there own or extract the binary from the container ... witch defets the purpose of an container if we also have binary releases. so now we just release the binary directly ...


6543:
aww well looks like i have to reconsider oci https://freebsdfoundation.org/blog/freebsd-officially-supported-in-oci-runtime-specification-v1-3/

6543:
# TODOs:
 - [x] test oci image on a freebsd ...
 - [ ] make sure local backemd downloads correct git clone plugun binary
 - [ ] package git clone plugin propperly ...

6543:
image
image

6543:
moved into own issue: https://github.com/woodpecker-ci/woodpecker/issues/6655

qwerty287:
Do you need to add openbsd as platform to xgo if you add it to the makefile?

6543:
you mean for the server an openbsd port?

I would focus on freebsd for now

qwerty287:
But why did you add openbsd to the makefile then?

.woodpecker/docker.yaml
```diff
@@ -3,7 +3,7 @@ variables:
   - &node_image 'docker.io/node:24-alpine'
   - &xgo_image 'docker.io/techknowlogick/xgo:go-1.26.x'
   - &buildx_plugin 'docker.io/woodpeckerci/plugin-docker-buildx:6.1.0'
-  - &platforms_release 'linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/386,linux/amd64,linux/ppc64le,linux/riscv64,linux/s390x,freebsd/arm64,freebsd/amd64,openbsd/arm64,openbsd/amd64'
+  - &platforms_release 'linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/386,linux/amd64,linux/ppc64le,linux/riscv64,linux/s390x,freebsd/arm64,freebsd/amd64'
   - &platforms_server 'linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le,linux/riscv64'
```
6543:
```suggestion
  - &platforms_server 'linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le,linux/riscv64,freebsd/arm64,freebsd/amd64'
```

6543: xgo supports freebsd :)

.woodpecker/binaries.yaml
```diff
@@ -38,7 +38,7 @@ steps:
       - apt install -y tree
       - make cross-compile-server
     environment:
-      PLATFORMS: linux|arm64/v8;linux|amd64;linux|riscv64;windows|amd64
+      PLATFORMS: linux|arm64/v8;linux|amd64;linux|riscv64;windows|amd64;freebsd|amd64
```
6543, line 41:
```suggestion
      PLATFORMS: linux|arm64/v8;linux|amd64;linux|riscv64;windows|amd64;freebsd|amd64;freebsd|arm64
```

6543:
so ... freebsd supports oci images and openbsd dont ...

before we just used oci containers as delivery where peopel could extract binarys .... but since we distribute binarys directly we just should do that ... for openbsd

as for freebsd, it gained oci support so we should widen our coi images to cover it

6543:
> But why did you add openbsd to the makefile then?

i initially was just goint to shift BSD-ish artifacts to be binaries ... but then noticed that in the meantime freebsd now moves towards oci support.

so this pull now moves openbsd out of oci into binarie while widen freebsd support

.woodpecker/docker.yaml
```diff
@@ -3,8 +3,8 @@ variables:
   - &node_image 'docker.io/node:24-alpine'
   - &xgo_image 'docker.io/techknowlogick/xgo:go-1.26.x'
   - &buildx_plugin 'docker.io/woodpeckerci/plugin-docker-buildx:6.1.0'
-  - &platforms_release 'linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/386,linux/amd64,linux/ppc64le,linux/riscv64,linux/s390x,freebsd/arm64,freebsd/amd64,openbsd/arm64,openbsd/amd64'
-  - &platforms_server 'linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le,linux/riscv64'
+  - &platforms_release 'linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/386,linux/amd64,linux/ppc64le,linux/riscv64,linux/s390x,freebsd/arm64,freebsd/amd64'
+  - &platforms_server 'linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le,linux/riscv64,freebsd/arm64,freebsd/amd64'
```
6543:
```suggestion
  - &platforms_server 'linux/arm/v7,linux/arm64/v8,linux/amd64,linux/ppc64le,linux/riscv64,freebsd/arm64'
```

.woodpecker/binaries.yaml
```diff
@@ -38,7 +38,7 @@ steps:
       - apt install -y tree
       - make cross-compile-server
     environment:
-      PLATFORMS: linux|arm64/v8;linux|amd64;linux|riscv64;windows|amd64
+      PLATFORMS: linux|arm64/v8;linux|amd64;linux|riscv64;windows|amd64;freebsd|amd64;freebsd|arm64
```
6543:
```suggestion
      PLATFORMS: linux|arm64/v8;linux|amd64;linux|riscv64;windows|amd64;freebsd|amd64
```

qwerty287:
OK. It really confuses me to have this split up in so many ways, maybe we could document somewhere which components are available on which os and installation method? Would also give benefits to users I think.

6543:
#1292 ?

qwerty287:
No, just a list of supported platforms.

6543:
ok i created: https://github.com/woodpecker-ci/plugin-git/pull/349

qwerty287 (approved):
Thanks, the docs are really nice.

Also just wondering, but is anyone using archs like `386`, `ppc64le`, `s390x`?

6543:
`386` we might remove ...
... `ppc64le` yes we have at least one comunity member that was ative in the past esp. around that architecture ... but i have to search through old issues
... `s390x` we have gov. using us and other old companies ... they run such stuff

PR #6628, opened 2026-05-23
Improve local backend Windows home isolation

### LINKED ISSUE #6513: Local backend: home-directory isolation is incomplete and broken on Windows
The local backend tries to isolate each workflow by giving it a temporary home directory. It does this by creating a temp dir per workflow and exporting it via `HOME` and `USERPROFILE`:

<https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/backend/local/local.go> — see `SetupWorkflow` and `StartStep`, where the only home-related variables set are:

```go
env = append(env, "HOME="+state.homeDir)
env = append(env, "USERPROFILE="+state.homeDir)
env = append(env, "CI_WORKSPACE="+state.workspaceDir)
```

On Linux this is mostly fine: `$HOME` is essentially the only variable a well-behaved program consults, and isolation works as intended.

On Windows it does **not** work. Windows has many independent variables that point at user-scoped locations — `APPDATA`, `LOCALAPPDATA`, `HOMEDRIVE`, `HOMEPATH`, `TEMP`, `TMP`, `USERNAME` — and most Windows applications, including PowerShell, .NET, and anything Microsoft-flavored, never look at `HOME` or `USERPROFILE` at all. They look at `APPDATA` / `LOCALAPPDATA`. So the "isolated" home is leaky: tools happily continue to read and write under the real user's `%APPDATA%` and `%LOCALAPPDATA%` while the workflow runs.

The visible symptoms are:

- Files and folders (e.g. `Microsoft/`, `PowerShell/`, `.dotnet/`, NuGet caches) appear in the **real** workspace or get written outside it, polluting the developer's environment.
- Working-tree dirt that's hard to attribute, since the leak is into the real user profile rather than the temp dir.
- Per-user state from previous runs bleeds into the current run, defeating the stated purpose of the isolation.

`.gitignore`-ing those folders is not an acceptable workaround — a clean working tree is a diagnostic signal, and silently masking it risks hiding real problems.

## Why "just add more env vars" isn't enough

It's tempting to extend the list:

```go
env = append(env, "LOCALAPPDATA="+state.homeDir+`\AppData\Local`)
env = append(env, "APPDATA="+state.homeDir+`\AppData\Roaming`)
```

This narrows some of the leakage but does not fix the underlying problem. A real isolated home on Windows would also have to account for at least:

- `HOMEDRIVE`, `HOMEPATH`
- `TEMP`, `TMP` (per-user temp directories)
- `USERPROFILE` already set, but consumers may also resolve it via `SHGetKnownFolderPath`/`FOLDERID_Profile`, which ignores the env var
- The Windows registry (`HKCU`), which holds per-user settings for countless applications and is not redirected by any environment variable
- Configuration files outside `%APPDATA%` (e.g. `%USERPROFILE%\.something`) that some tools resolve through Win32 APIs rather than env vars

AkashKumar7902:
## What changed

The local backend isolated-home mode now redirects Windows user-profile locations into the workflow home directory:

- `APPDATA` -> `<home>\AppData\Roaming`
- `LOCALAPPDATA` -> `<home>\AppData\Local`
- `TEMP` / `TMP` -> `<home>\AppData\Local\Temp`
- `HOMEDRIVE` / `HOMEPATH` derived from the isolated home path

It also creates the Windows app-data/temp directories during workflow setup when the local backend is running in isolated-home mode on Windows.

## Why

On Windows, many tools ignore `HOME` and read/write user state through `%APPDATA%`, `%LOCALAPPDATA%`, `%TEMP%`, and related profile variables. Redirecting only `HOME` and `USERPROFILE` left local backend workflows leaking state into the real user profile.

Closes #6513

## Validation

- `go test -race -tags test ./pipeline/backend/local -count=1`
- `go test -race -tags test ./pipeline/backend/... -count=1`
- `make test-lib`

Note: the broader command was run with the explicit downloaded Go 1.26 toolchain in this workstation: `GOROOT=$(go env GOROOT) GOTOOLCHAIN=local PATH="$(go env GOROOT)/bin:$PATH" make test-lib`.


pipeline/backend/local/local.go
```diff
@@ -175,6 +178,74 @@ func (e *local) StartStep(ctx context.Context, step *types.Step, taskUUID string
...
+		"HOMEPATH="+homePath,
+		"APPDATA="+e.homePath(state.homeDir, "AppData", "Roaming"),
+		"LOCALAPPDATA="+localAppData,
+		"TEMP="+e.homePath(localAppData, "Temp"),
+		"TMP="+e.homePath(localAppData, "Temp"),
+	)
+}
+
+func (e *local) homePath(base string, elems ...string) string {
+	if e.os != "windows" {
```
6543, line 224: multible gated the same way :/

also this code is just to split

[KISS principe](https://woodpecker-ci.org/docs/development/core-ideas)

also if we move isolation behind extra code file, we can add build flags for linux/windows/mac/...

6543 (changes requested):
ned overhal

PR #6630, opened 2026-05-23
fix cli exec skipped workflow cleanup

### LINKED ISSUE #5419: CLI error "invalid volume name or ID: value is empty" if "when" section is pressent
### Component

cli

### Describe the bug

As soon as I have
```
when:
  - event: push
    branch: "**"
```
in my pipeline, the CLI fails with

```
# build-and-deploy.yaml
12:03PM ERR could not remove volume '' error="invalid volume name or ID: value is empty"
12:03PM ERR could not remove network '' error="invalid network name or ID: value is empty"
```

### Steps to reproduce

1. Install CLI
2. run a pipeline with a "when" section
3. The error occurs

### Expected behavior

Should be executed without error

### System Info

```shell
woodpecker-cli version next-dc7795e64b
Docker Server Version: 28.3.0
```

### Additional context

_No response_

### Validations

AkashKumar7902:
## Summary
- return successfully from `woodpecker-cli exec` when a workflow compiles to no runnable stages
- avoid selecting/setting up a backend for workflows filtered out by `when`
- add a regression test that fails if a skipped workflow reaches backend setup

Closes #5419

## Validation
- reproduced on `upstream/main` with a Docker-backed workflow filtered by `when`, which failed with `invalid volume name or ID: value is empty` and `invalid network name or ID: value is empty`
- `GOROOT=$(go env GOROOT) GOTOOLCHAIN=local PATH="$(go env GOROOT)/bin:$PATH" go test -race -tags test ./cli/exec -run TestExecSkipsFilteredWorkflowWithoutBackendSetup -count=1 -v`
- Docker skipped-workflow smoke test: `go run ./cmd/cli --disable-update-check exec --backend-engine docker --repo-path "$tmpdir" "$tmpdir/.woodpecker/build.yaml"` exits 0 after this fix
- Docker matching-workflow smoke test with `--pipeline-event push --commit-branch main` runs `alpine:3.23` and prints step output
- `GOROOT=$(go env GOROOT) GOTOOLCHAIN=local PATH="$(go env GOROOT)/bin:$PATH" go test -race -tags test ./cmd/cli ./cli/... -count=1`
- `GOROOT=$(go env GOROOT) GOTOOLCHAIN=local PATH="$(go env GOROOT)/bin:$PATH" make test-cli`

Note: local `pre-commit` is unavailable in this environment (`pre-commit: command not found`).

### DIFF
--- a/cli/exec/exec.go
+++ b/cli/exec/exec.go
@@ -304,6 +304,9 @@ func execWithAxis(ctx context.Context, c *cli.Command, file, repoPath string, ax
 	if err != nil {
 		return err
 	}
+	if len(compiled.Stages) == 0 {
+		return nil
+	}
 
 	backendCtx := context.WithValue(ctx, backend_types.CliCommand, c)
 	backendEngine, err := backend.FindBackend(backendCtx, backends, c.String("backend-engine"))
--- a/cli/exec/exec_test.go
+++ b/cli/exec/exec_test.go
@@ -1,0 +1,59 @@
+// Copyright 2026 Woodpecker Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


6543 (changes requested):
this is an UX problem

PR #6632, opened 2026-05-23
docs: clarify skipped workflow dependencies

### LINKED ISSUE #6457: Skipped workflows are not considered as succeed like documentation suggest it

AkashKumar7902:
## Summary
- correct the global workflow `when` docs to say skipped workflows are skipped completely
- clarify that globally skipped workflows do not satisfy `depends_on`
- add a workaround note for optional work that should not block dependent workflows

Closes #6457

## Validation
- `cd docs && pnpm format:check`
- `make spellcheck`
- `GOROOT=$(go env GOROOT) GOTOOLCHAIN=local PATH="$(go env GOROOT)/bin:$PATH" make build-docs`
- `git diff --check`

Note: `make build-docs` succeeds with existing plugin-fetch warnings and the existing Docusaurus warning for the old versioned `2.8` CLI page.

### DIFF
--- a/docs/docs/20-usage/20-workflow-syntax.md
+++ b/docs/docs/20-usage/20-workflow-syntax.md
@@ -757,7 +757,13 @@ skip_clone: true
 
 ## `when` - Global workflow conditions
 
-Woodpecker gives the ability to skip whole workflows ([not just steps](#when---conditional-execution)) based on certain conditions by a `when` block. If all conditions in the `when` block evaluate to true the workflow is executed, otherwise it is skipped, but treated as successful and other workflows depending on it will still continue.
+Woodpecker gives the ability to skip whole workflows ([not just steps](#when---conditional-execution)) based on certain conditions by a `when` block. If all conditions in the `when` block evaluate to true the workflow is executed, otherwise the complete workflow is skipped.
+
+:::info
+A workflow skipped by a global `when` block does not count as a successful dependency for other workflows using `depends_on`. Those dependent workflows are not executed.
+
+If another workflow should continue after optional work is skipped, move the condition to the relevant steps or keep a lightweight step that can complete successfully.
+:::
 
 For more information about the specific filters, take a look at the [step-specific `when` filters](#when---conditional-execution).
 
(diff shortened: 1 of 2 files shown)


qwerty287:
Instead of this docs change I would go with https://github.com/woodpecker-ci/woodpecker/pull/6461 that allows to make dependencies optional.

qwerty287:
#6461 is merged

PR #6635, opened 2026-05-23
Add commit timestamp environment variables

### LINKED ISSUE #5245: Add env `CI_COMMIT_TIMESTAMP`

Is it possible to get an env `CI_COMMIT_TIMESTAMP` like [Gitlab](https://docs.gitlab.com/ci/variables/predefined_variables/)? Maybe `CI_PREV_COMMIT_TIMESTAMP` would be useful too.

Why? Currently i have to add git to get this using `git log -1 --format="%at"`.
I want to keep my plugins small so i want to avoid git there.

### Suggested solution

Add `CI_COMMIT_TIMESTAMP` with iso8601 format.

### Alternative

Add git to all plugins.

### Additional context

_No response_

### Validations

AkashKumar7902:
## Summary
- carry pipeline commit timestamps into runtime metadata
- expose `CI_COMMIT_TIMESTAMP` and `CI_PREV_COMMIT_TIMESTAMP` as UTC RFC3339 values when commit timestamps are available
- document the new environment variables and cover current/previous commit timestamp export in tests

Closes #5245

## Validation
- `go test -race -tags test ./pipeline/frontend/metadata ./server/pipeline/metadata -run 'TestEnviron|TestGetWorkflowMetadata' -count=1 -v`
- `go test -race -tags test ./pipeline/frontend/metadata ./server/pipeline/... -count=1`
- `make test-lib`
- `make test-server`


qwerty287:
Thanks to be consistent with other vars like `CI_PIPELINE_CREATED`, can we use a Unix timestamp?

server/pipeline/metadata/metadata.go
```diff
@@ -140,11 +140,12 @@ func metadataPipelineFromModelPipeline(pipeline *model.Pipeline, includeParent b
...
-			Ref:     pipeline.Ref,
-			Refspec: pipeline.Refspec,
-			Branch:  pipeline.Branch,
-			Message: pipeline.Message,
+			Sha:       pipeline.Commit,
+			Ref:       pipeline.Ref,
+			Refspec:   pipeline.Refspec,
+			Branch:    pipeline.Branch,
+			Message:   pipeline.Message,
+			Timestamp: pipeline.Timestamp,
```
6543, line 148: is pipeline.Timestam equal to commit timestamp?!?

6543:
also cli needs patch to allow setting it

server/pipeline/metadata/metadata.go
qwerty287: It is, but we actually stored that without using it.

qwerty287:
added cli and changed to unix timstamps in #6670

PR #6641, opened 2026-05-25
fix(agent): decode CI_SCRIPT to temp file before execution

### LINKED ISSUE #5450: Entrypoint base64 chunking makes pipeline commands fail randomly
### Component

agent

### Describe the bug

The pipeline fails, because the list of commands is cut at a random place.

I estimate it happens about 10 to 20% of my runs, which is really annoying. I'm using images like `amazoncorretto:24.0.2-alpine3.22` to run a list of 10 to 20 commands (gradle build commands). Some of the pipelines report `/bin/sh: syntax error: unexpected end of file` and some report an error the gradle task cannot be executed because it cannot find the specific task, but multiple. Upon close inspection, it turns out is only tries to execute part of the line, like `./gradlew com` instead of `./gradlew compileJava`. The `com` maps to multiple tasks, so Gradle is not sure what to execute.
Running the same job again, clicking rerun in Woodpecker, often results in a passing build.

In an attempt to debug and fix it myself, I tried to print out the complete script (list of commands) at the beginning of the pipeline, which seems to be fine. Also for the times when the pipeline failed. I updated the entrypoint to put the script into a file first, before executing, instead of pipeing it directly to `/bin/sh`. I use:

```
entrypoint: ["/bin/sh", "-c", "echo $CI_SCRIPT | base64 -d > script.sh && pwd && ls -alh && cat script.sh && /bin/sh -xe script.sh"]
```
I also added some debug items to see files, current directory and the contents of the script. Up to now, it seems the update of the entrypoint **fixes the issue**, but as used to fail randomly, I might just have been lucky up to now.

Key difference between the original entrypoint (listed here https://woodpecker-ci.org/docs/usage/workflow-syntax#entrypoint), `["/bin/sh", "-c", "echo $CI_SCRIPT | base64 -d | /bin/sh -e"]`, is the replacement of the pipe with a pass to the file `script.sh` before execution. Depending on the image used to run the pipeline, the `base64 -d` might do some chunking and put a newline at a random point in the script (sometimes), which triggers execution of only a part of the command, instead of the complete command.

### Steps to reproduce

Hard to reproduce, as it is happens only some of the runs.

### Expected behavior

A passing or failing build 100% of the time.

### System Info

```shell
Running Woodpecker-ci 3.9.0 (also had this behavior using 3.8.0, was hoping it was fixed in 3.9.0, but it isn't).
```

### Additional context

_No response_

### Validations

wyf027:
## Summary

- Change the default POSIX entrypoint to decode `CI_SCRIPT` into a temp file under `/tmp` before executing it with `/bin/sh -e`
- Avoids intermittent command truncation when piping `base64 -d` output directly to the shell (reported in #5450)
- Store script outside the workspace as suggested in the issue discussion

## Test plan

- [x] `go test ./pipeline/backend/common/... ./pipeline/backend/docker/... ./pipeline/backend/kubernetes/...`
- [x] Updated unit tests for docker/kubernetes container config generation

Fixes #5450

### DIFF
--- a/docs/docs/20-usage/20-workflow-syntax.md
+++ b/docs/docs/20-usage/20-workflow-syntax.md
@@ -173,7 +173,7 @@ Only build steps can define commands. You cannot use commands with plugins or se
 
 Allows you to specify the entrypoint for containers. Note that this must be a list of the command and its arguments (e.g. `["/bin/sh", "-c"]`).
 
-If you define [`commands`](#commands), the default entrypoint will be `["/bin/sh", "-c", "echo $CI_SCRIPT | base64 -d | /bin/sh -e"]`.
+If you define [`commands`](#commands), the default entrypoint will be `["/bin/sh", "-c", "script=$(mktemp /tmp/.woodpecker-ci-script.XXXXXX) && echo $CI_SCRIPT | base64 -d > \"$script\" && /bin/sh -e \"$script\"; ret=$?; rm -f \"$script\"; exit $ret"]`.
 You can also use a custom shell with `CI_SCRIPT` (Base64-encoded) if you set `commands`.
 
 ### `environment`
--- a/pipeline/backend/common/script.go
+++ b/pipeline/backend/common/script.go
@@ -18,6 +18,10 @@ import (
 	"encoding/base64"
 )
 
+// PosixEntrypointCommand decodes CI_SCRIPT to a temp file before execution to avoid
+// intermittent truncation when piping base64 output directly to /bin/sh.
+const PosixEntrypointCommand = `script=$(mktemp /tmp/.woodpecker-ci-script.XXXXXX) && echo $CI_SCRIPT | base64 -d > "$script" && /bin/sh -e "$script"; ret=$?; rm -f "$script"; exit $ret`
+
 func GenerateContainerConf(commands []string, osType, workDir string) (env map[string]string, entry []string) {
 	env = make(map[string]string)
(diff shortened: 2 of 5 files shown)


qwerty287:
Are we sure any container has a `/tmp` dir?

6543:
☝️

6543 (changes requested):
blocked till implications are known

exxlr:
> Are we sure any container has a `/tmp` dir?

Mostly but not guaranteed. If you build an image `FROM scratch` and then copy a statically linked binary into it, there will only be your binary, `/dev/`, `/etc/`, `/proc/`, `/sys/`. Even the `/bin/sh` and `mktemp` will not be available.

6543:
one thing that i find strange, with this pull the value of the environment gets even longer ... and that does not cause any issues ... but the argument parsing does ?!?

could there be upstream bugs in the shells?

lafriks:
I have never encountered myself this, if this is the environment/specific image shell case, this could probably be moved to some special server setting that would allow to change default on how to execute shell commands but default imho should stay as it is

wyf027:
Closing this since the maintainer feedback indicates the default execution behavior should stay as-is, and the remaining concerns around `/tmp` availability and shell/environment implications do not have a clear merge path for this approach. Thanks for the review and discussion.

6543:
well i guess so ... I personally would not have a workaround and fix the actual cause - as there must be one

PR #6643, opened 2026-05-25
Add GitHub OAuth App setup hint to docker-compose page

mateuszkamola:
The docker-compose page uses GitHub as its example forge but doesn't mention that a GitHub OAuth App needs to be registered beforehand, or what the callback URL should be. This adds a note pointing to the GitHub forge docs, and a hint that WOODPECKER_HOST is used as the base for the /authorize callback URL.


qwerty287:
Thanks, but you don't necessarily need github. Yes, that's the example, but you can register fir a different forge as well. It would be nice if this could be more generic and covers all forges 🙃

mateuszkamola:
> Thanks, but you don't necessarily need github. Yes, that's the example, but you can register fir a different forge as well. It would be nice if this could be more generic and covers all forges 🙃

I agree. I dropped mentioning Github specifically, left only info about registering OAuth App with the chosen forge first and to take a look at the forge docs. I think with that I would save few hours on setting up woodpecker the first time :)

qwerty287 (approved):

PR #6644, opened 2026-05-25
Allow disabling service workspace volumes in k8s

### LINKED ISSUE #3344: K8S backend new behaivor when storage rwx is `false`

julienvincent:
Currently when using the Kubernetes backend all services are provisioned with the workspace PVC volume mounted. This causes problems when using `ReadWriteOnce` PVC volumes in multi-node clusters as the service and step Pods can be provisioned onto different nodes.

A workaround for this is to use the PodAffinity controls on the agent:

```yaml
env:
  WOODPECKER_BACKEND_K8S_POD_AFFINITY: |
    podAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector: {}
          matchLabelKeys: ['woodpecker-ci.org/task-uuid']
          topologyKey: 'kubernetes.io/hostname'
```

This ensures the step Pods are provisioned onto nodes already containing Pods from the same task. This will always result in co-locating the service Pods with step Pods.

While this kind of works, it means that a single Node now needs to be sized appropriately to host all the service Pods as well as the step Pod which uses them.

A better solution that allows distributing service Pods over all the available nodes is to stop attaching the workspace PVC to service Pods. A similar problem/solution is discussed in #3344, as well as some related discussion in
https://github.com/woodpecker-ci/woodpecker/discussions/5312.

The discussion in #3344 mentions attaching a dedicated PVC to services, but IMO this overly complicates things and adds very little value. I believe that for most use-cases services' won't need to interact with a PVC at all. They are just there to run a database or service dependency.

---

This change introduces two new controls:

- An agent-level environment variable `WOODPECKER_BACKEND_K8S_SERVICE_WORKSPACE_VOLUME` which defaults to `true`.
- A step-level Kubernetes backend option `backend_options.kubernetes.workspaceVolume` which also defaults to `true`.

If either of these are set to `false` then the workspace PVC will not be attached to the Pod.

Closes #3344

### DIFF
--- a/docs/docs/20-usage/72-extensions/index.md
+++ b/docs/docs/20-usage/72-extensions/index.md
@@ -8,6 +8,10 @@ There is currently one type of extension available:
 - [Registry extension](./50-registry-extension.md) to get registry credentials from the extension.
 - [Secret extension](./55-secret-extension.md) to get secrets from an external service.
 
+:::note
+Woodpecker's permission handling is linked to the forge. A user on your forge that has admin access to the repo will also get admin permissions for the repository in Woodpecker and can then change the configured extensions. This could be used to get credentials of the forge user. Make sure you trust the repo admins that can sign in to Woodpecker.
+:::
+
 ## Security
 
 :::warning
--- a/docs/docs/20-usage/75-project-settings.md
+++ b/docs/docs/20-usage/75-project-settings.md
@@ -4,6 +4,10 @@ As the owner of a project in Woodpecker you can change project related settings 
 
 ![project settings](./project-settings.png)
 
+:::note
+Woodpecker's permission handling is linked to the forge. A user on your forge that has admin access to the repo will also get admin permissions for the repository in Woodpecker and can then change the settings here.
+:::
+
 ## Pipeline path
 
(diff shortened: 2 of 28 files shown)


6543:
can we remove the flag, it can just cause issues that are undocumented

julienvincent:
@6543 Are you referring to the `WOODPECKER_BACKEND_K8S_SERVICE_WORKSPACE_VOLUME` environment variable?

Happy to remove, but just so I understand - what issues do you foresee this causing? I thought I did update the documentation for this config.

6543:
as it is a backemd flag, it is an agent flag.

let agent config change pieline execution config that is transparent to the server and git-commited config but can affect workflow to fail is something i dont want to encourage :)

julienvincent:
@6543 Makes sense. I've removed the flag

6543 (approved):

PR #6650, opened 2026-05-26
fix(runtime): run OnFailure steps after workflow cancel

### LINKED ISSUE #6564: when.status: [success, failure] does not fire on killed pipelines — killed is undocumented and easily missed
## Summary

When a pipeline is canceled or killed (by the server, a scaler, or a timeout), its status is set to \`killed\`. Steps that declare \`when: status: [success, failure]\` do not run in this case — which means cleanup and notification steps are silently skipped.

This is non-obvious because:
- \`success\` and \`failure\` are the natural pair users reach for when writing "run this regardless of outcome"
- \`killed\` is not prominently documented as a third pipeline status that \`when.status\` must explicitly include
- There is no warning when a step's \`when.status\` list omits \`killed\`

## Impact

Cleanup steps (delete VMs, release locks, send notifications) are silently skipped when a pipeline is killed. Resources leak and operators don't know the pipeline was canceled.

Observed: a build VM was left running and a deploy marker was never written because the cleanup workflow only declared \`status: [success, failure]\`. The pipeline was killed by a stale-pipeline detector after exceeding a run-time threshold.

## Expected behaviour

Either:

**Option A — Document \`killed\` explicitly.** Add \`killed\` to the \`when.status\` reference docs with a clear note that cleanup steps must include it alongside \`success\` and \`failure\` to handle all terminal states.

**Option B — Treat \`killed\` as \`failure\` for \`when.status\` matching.** A killed pipeline is a failed outcome from the user's perspective. Steps that watch for \`failure\` almost always want to catch \`killed\` too. Breaking the two apart requires users to know an implementation detail.

**Option C — Lint / warn.** If a step has \`depends_on\` and \`when.status\` that includes \`failure\` but not \`killed\`, emit a warning at pipeline parse time.

Option A is the minimum; Option B is the most user-friendly.

## Workaround

Add \`killed\` explicitly to every \`when.status\` list where cleanup matters:

```yaml
when:
  - status: [success, failure, killed]
```

wyf027:
## Summary
Fixes #6564

When a pipeline is canceled/killed mid-run, the runtime now continues with later stages instead of aborting immediately. Steps configured with `when.status: [success, failure]` (`OnFailure: true`) still execute for cleanup.

## Root cause
`Runtime.Run` returned `ErrCancel` as soon as the workflow context was canceled, skipping all subsequent stages—even cleanup steps that should run on failure/cancel.

## Approach
- On cancel, drain the current stage, record `ErrCancel`, and continue remaining stages.
- `shouldSkipStep` already skips success-only steps when an error is set, so normal steps are not affected.

## Test plan
- [x] `go test -tags test ./pipeline/runtime/... -run TestRunContextCanceled -v`
- [x] Added `TestRunContextCanceledRunsOnFailureStep`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pipeline/runtime/workflow.go
+++ b/pipeline/runtime/workflow.go
@@ -66,7 +66,10 @@ func (r *Runtime) Run(runnerCtx context.Context) error {
 		select {
 		case <-r.ctx.Done():
 			<-stageChan
-			return pipeline_errors.ErrCancel
+			if r.err.Get() == nil {
+				r.err.Set(pipeline_errors.ErrCancel)
+			}
+			continue
 		case err := <-stageChan:
 			if err != nil {
 				r.err.Set(err)
--- a/pipeline/runtime/workflow_test.go
+++ b/pipeline/runtime/workflow_test.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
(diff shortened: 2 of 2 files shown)


6543 (changes requested):
this is not the way!

we can discusst to add a special onCancel but onError should stop if you cancel a workfow!
also if we have a onCancel we still need a way to cancel the steps running on cancel as well ...

PR #6669, opened 2026-05-29
fix(kubernetes): services should have clean exit state from WaitStep on context expiry/API server errors

hhamalai:
There's a corner case(s) where the pipeline service on Kubernetes backend is not cleaned up appropriately when the context is cancelled, leading to services being shown as running, even while the service pods are terminated.

Also the DestroyWorkflow changed to handler destroying steps with errgroup like handled out with Docker backend, instead of failing on the first error.

### DIFF
--- a/pipeline/backend/kubernetes/kubernetes.go
+++ b/pipeline/backend/kubernetes/kubernetes.go
@@ -28,6 +28,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/sync/errgroup"
+
 	"github.com/cenkalti/backoff/v5"
 	"github.com/rs/zerolog/log"
 	"github.com/urfave/cli/v3"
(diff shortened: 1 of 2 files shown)


6543:
Does https://github.com/woodpecker-ci/woodpecker/pull/6672 solfe this here?

hhamalai:
> Does #6672 solfe this here?
Likely helps at least to solve the timing related issues, not sure about if it solves all these corner cases.

The usage of error groups in DestroyWorkflow step would unify the logic with kubernetes/docker runners a bit. It's by no mean complete refactoring, but such refactoring would make maintaining these different backends a bit more straightforward in long run?

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -350,7 +352,7 @@ func (e *kube) WaitStep(ctx context.Context, step *types.Step, taskUUID string)
 	select {
 	case <-finished:
 	case <-ctx.Done():
-		return nil, ctx.Err()
+		return &types.State{ExitCode: 0, Exited: true}, nil
```
6543, line 355: TODO: make sure we dont get success state in killed steps by cancle

PR #6670, opened 2026-05-29
Add commit timestamp env vars

### LINKED ISSUE #5245: Add env `CI_COMMIT_TIMESTAMP`

qwerty287:
Closes https://github.com/woodpecker-ci/woodpecker/issues/5245
Supersedes https://github.com/woodpecker-ci/woodpecker/pull/6635


cli/exec/flags.go
```diff
@@ -279,11 +279,16 @@ var flags = []cli.Flag{
 		Usage:   "Set the metadata environment variable \"CI_COMMIT_BRANCH\".",
 		Value:   "main",
 	},
-	&cli.StringFlag{
+	&cli.Int64Flag{
```
6543: ?

6543 (approved):

PR #6671, opened 2026-05-29
Add workflow concurrency limit option

### LINKED ISSUE #1461: Ability to restrict more then 1 build at a time for one project when there are multiple agents

I remember that original CI had a feature called concurrency https://discourse.drone.io/t/how-to-limit-build-concurrency-per-project/3500
This helped a lot to prevent build run when other build already in progress. If you commit faster then old build ends new build starts and sometimes they need to spawn e.g. 2 services which have conflicts. Is there an option to somehow limit build to 1 at a time? I know it could be resolved if we would use 1 agents, but then other projects on same CI instance would stuck.

### Suggested solution

Adding option or any other solution which is not obvious from current doc.

### Alternative

_No response_

### Additional context

_No response_

### Validations

### LINKED ISSUE #3579: Limit concurrent builds per repo and user

It would be very helpful to limit the concurrent builds on per-repository and per-user basis. This would prevent a single repository or user from blocking the whole CI for everyone.

E.g. when pushing a lot of commits at once to a repository, woodpecker will build all of them with as many workers as are available, effectively blocking other users completely.

If there already is such a feature, let me know. I haven't been able to find a workaround for that problem.

### Suggested solution

Add an option for administrators to limit the parallel jobs for each individual user.
Add an option for users in the repository settings to set the maximum number of concurrent builds of that repository

### Alternative

Add a fixed per-user limit of concurrent jobs as environment variable. 

### Additional context

_No response_

### Validations

### LINKED ISSUE #4809: Limit pipeline parallel execution per-branch

Currently there is no way to reliability limit same pipeline with different commits not to be executed in parallel. This can be problematic if pipelines are used to do deployments and canceling running executing pipeline is not an option as that could break deployment.

### Suggested solution

Allow to specify either in pipeline or in repo settings that pipelines for each branch are exclusive/should be queued (similarly how that it is possible to set cancel executing pipelines)

### Additional context

In our use case if canceling helm deploy in k8s can result in broken deployment state

### Validations

- [x] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

simonmeyerrr:
Fixes https://github.com/woodpecker-ci/woodpecker/issues/1461
Fixes https://github.com/woodpecker-ci/woodpecker/issues/4809 (duplicate of the above one)
Fixes https://github.com/woodpecker-ci/woodpecker/issues/3579 (as it is technically possible to set `CI_PIPELINE_AUTHOR` in the group name to have concurrency limit per user, but it's more a way to use the current feature to satisfy the need rather than a feature designed for this specific purpose)

## What

Adds an optional `concurrency` setting to the workflow YAML that limits how many instances of a workflow (or group) run at the same time, without cancelling running pipelines.

Supports both shorthand and full form:
```yaml
concurrency: 1
# or
concurrency:
  limit: 1
  group: deploy
```

## Why

"Cancel previous pipeline" isn't safe for deployments: it can abort an in-flight deploy. This lets users serialize deploy workflows instead: when the limit is reached, further instances queue and start once a slot frees up.

## Behavior

- Queued workflows of the same group start in the order their pipelines were created, not the order they become ready (a slower depends_on can't reorder them)
- The ordering reservation is scoped across pipelines, so it never conflicts with intra-pipeline depends_on (no deadlock)
- Group defaults to the workflow name, scoped per repo; customizable via group
- When unset, scheduling behavior is unchanged (no breaking change)

## Tests / docs

- Queue concurrency + deadlock-regression tests, schema + YAML parse tests
- Docs added under Usage → Workflows
- Tested manually on my end locally with a full docker setup and gitea


server/queue/fifo.go
```diff
@@ -335,6 +343,84 @@ func (q *fifo) assignToWorker() (*list.Element, *worker) {
...
+// pipeline, execution order is already defined by depends_on, and reserving a
+// slot for an earlier-ID workflow could deadlock when it depends on a later-ID
+// workflow that shares the same group (e.g. deploy.yaml, which sorts before
+// and so gets a lower ID than test.yaml, depending on test.yaml). Because
+// dependencies never cross pipelines, restricting the reservation to other
+// pipelines keeps the ordering guarantee while making such deadlocks
+// impossible.
+//
+// Expects the queue to be locked by the caller.
+func (q *fifo) canRunConcurrent(task *model.Task) bool {
```
lafriks, line 364: Shouldn't this function have RW sync checks?

simonmeyerrr: This function is only called from `assignToWorker`, which is only called from `process()` while it holds `q.Lock()`. So the queue is already locked when this runs.
This matches the existing convention in the file (`depsInQueue`, `updateDepStatusInQueue`, `removeFromPendingAndWaiting`). These functions all expect the caller to hold the lock, and the function's doc comment states `Expects the queue to be locked by the caller.`.

server/queue/fifo.go
```diff
@@ -335,6 +343,84 @@ func (q *fifo) assignToWorker() (*list.Element, *worker) {
...
+	for element := q.waitingOnDeps.Front(); element != nil; element = element.Next() {
+		other, _ := element.Value.(*model.Task)
+		countAhead(other)
+	}
+
+	return running+ahead < task.ConcurrencyLimit
+}
+
+// taskInstantiationOrder returns a monotonic value reflecting the order in
+// which a task (workflow) was created. Task IDs are the autoincrement workflow
```
lafriks:
```suggestion
// which a task (workflow) was created. Task IDs are the auto incremented workflow
```

6543:
should concurrency not be limmited by agents?

because concurrency is only a problem on the workflow executor side and depends on what specs that one has

6543 (changes requested):
against workflow based concurrency limit

if that is a problem, one can just use the DAG,
else it is a workflow executor problem and should be configured on that end.

lafriks:
No it's not only executor problem, I would also want to have this example for k8s helm deployments this is currently problem, it could break them if two merges are made close together and previous pipeline is still executing deployment (you can not cancel either as that could also break deployment)

lafriks:
DAG can not solve this as this is about same workflow executing from different events in parallel (ex. two pushes to same branch in close to same time etc) not different workflows from single event

simonmeyerrr:
The issue I'm trying to fix with this feature is concurrent deployments when there is multiple commit/push on main in a short period. It can also be used in other contexts.

This feature exist in all main CI providers:
- Github Actions has `concurrency` parameter
- Jenkins has `disableConcurrentBuilds()`
- GitLab CI has `resource_group` explicitly designed for this use case

Options I have without this feature:
- **Enable previous pipeline cancellation** -> not a good solution as cancelling an in progress deployment can break some stuff
- **Create one agent per deploy with maximum workflows set to 1** -> it becomes complicated when, like in my case, you have 6 apps * 3 environments which would require to create 18 agents and it wouldn't prevent breaking of ordering when a workflow dependency run randomly faster on the second commit
- **Use an external lock system** -> sad to have to setup external stuff when it could be integrated, doesn't prevent breaking ordering, require to increase timeouts, consume running workflow for nothing

simonmeyerrr:
Hello @6543, can I do anything to help getting this feature merged ?
Do you think it has a chance of being part of 3.16.0 or it will be 3.17.0 ?

6543:
asap i come to it, i think we can get it into this release

6543:
a simple e2e test is missing with the dummy backend, i would propose:

a new in go written test, that starts an agent with 6 concurrent slots, have a pipeline definitin with two different workflows A and B, A has concurency 1, B has concurency 2, now shedule 3 rounds of that pipeline. then once all finished, compare start and finish times to make sure they respected it.

6543:
the scheduler will undergo a refactoring, have this tested, will prevent regressions :)

server/pipeline/queue.go
```diff
@@ -46,6 +46,20 @@ func queuePipeline(ctx context.Context, repo *model.Repo, activePipeline *model.
...
 
+		// Set up the concurrency limit if the workflow opted in. The group is
+		// scoped to the repository and defaults to the workflow name so that
+		// only "similar" workflows (same workflow, across pipelines) are
+		// limited against each other. When no limit is set the behavior is
+		// unchanged.
+		if item.ConcurrencyLimit > 0 {
+			group := item.ConcurrencyGroup
+			if group == "" {
+				group = item.Workflow.Name
```
6543: this is a hack, just count `` as group, the problem is that you can easily colide currently

6543:
```go
                if item.ConcurrencyLimit > 0 {
                        task.ConcurrencyLimit = item.ConcurrencyLimit

                        if item.ConcurrencyGroup == "" {
                                task.ConcurrencyGroup = fmt.Sprintf("%d/%s/%s", repo.ID, item.Workflow.Name, group)
                        }
                        task.ConcurrencyGroup = fmt.Sprintf("%d//%s", repo.ID, group)
                }
```

for example

pipeline/frontend/yaml/types/concurrency.go
```diff
@@ -0,0 +1,57 @@
...
+	Limit int `yaml:"limit,omitempty"`
+	// Group identifies which workflows are mutually limited. Workflows that
+	// resolve to the same group within a repository are serialized according
+	// to the limit. When empty it defaults to the workflow name, so different
+	// runs of the same workflow are limited against each other.
+	Group string `yaml:"group,omitempty"`
+}
+
+// UnmarshalYAML implements the Unmarshaler interface.
+func (c *Concurrency) UnmarshalYAML(unmarshal func(any) error) error {
```
6543: re-encode missing!

pipeline/frontend/yaml/types/concurrency.go
```diff
@@ -0,0 +1,69 @@
...
+	// shorthand: `concurrency: <int>`
+	var limit int
+	if err := unmarshal(&limit); err == nil {
```
6543: we could use yaml.Node to determine what it is beforhand - see other implementations

docs/docs/20-usage/25-workflows.md
```diff
@@ -144,3 +144,49 @@ The same syntax works at the step level within a workflow: if a step uses `depen
...
+
+You can also use the shorthand form to only set the limit:
+
+```yaml
+concurrency: 1
+```
+
+### Ordering
+
+Queued workflows of the same group start in the order their pipelines were created, **not** in the order they become ready to run. This matters when a workflow depends on other workflows (via `depends_on`) whose duration varies: even if a later pipeline's checks finish first, its limited workflow will not overtake an earlier pipeline that is still waiting. This guarantees that, for example, deployments happen in commit order.
```
6543, line 176: this we changed, by default if no group set each workfow is it's own group ....

6543:
@simonmeyerrr helped you a bit - only some doc nits are missing ... and I need to look at the fifo implementation.

rest should be fine

6543:
## tested that e2e breaks on **main**:

image

server/queue/fifo.go
```diff
@@ -335,6 +343,84 @@ func (q *fifo) assignToWorker() (*list.Element, *worker) {
...
+		countAhead(other)
+	}
+
+	return running+ahead < task.ConcurrencyLimit
+}
+
+// taskInstantiationOrder returns a monotonic value reflecting the order in
+// which a task (workflow) was created. Task IDs are the auto incremented workflow
+// IDs, so a lower value means an earlier instantiation.
+func taskInstantiationOrder(task *model.Task) int64 {
```
6543: i would not relay on task.ID for ordering!

server/model/task.go
```diff
@@ -34,6 +34,16 @@ type Task struct {
...
 	RepoID       int64                  `json:"repo_id"      xorm:"'repo_id'"`
+	// ConcurrencyLimit is the maximum number of tasks sharing the same
+	// ConcurrencyGroup that may run at once. A value <= 0 means unlimited.
+	ConcurrencyLimit int `json:"concurrency_limit" xorm:"NOT NULL DEFAULT 0 'concurrency_limit'"`
+	// ConcurrencyGroup identifies tasks that are limited against each other.
+	// It is empty when no concurrency limit applies.
+	ConcurrencyGroup string `json:"concurrency_group" xorm:"'concurrency_group'"`
+	// Created is the unix timestamp the task's pipeline was created at. It
+	// defines the queue ordering across pipelines, so the queue never relies
+	// on the task ID for ordering.
```
6543:
```suggestion
	// defines the queue ordering across pipelines.
```

6543 (approved):

6543:
@lafriks i changed some bits ... wana have a look or is it ok for you?

lafriks (approved):

lafriks:
Looks good from my side

pipeline/frontend/yaml/types/concurrency_test.go
```diff
@@ -0,0 +1,174 @@
...
+			assert.Equal(t, tc.expected, string(out))
+		})
+	}
+}
+
+func TestConcurrencyRoundTrip(t *testing.T) {
+	tests := []struct {
+		name        string
+		concurrency Concurrency
+		// wantShorthand asserts whether the marshalled form is the bare
```
lafriks:
```suggestion
		// wantShorthand asserts whether the marshaled form is the bare
```

pipeline/frontend/yaml/types/concurrency.go
```diff
@@ -0,0 +1,93 @@
...
+package types
+
+import (
+	"fmt"
+
+	"go.yaml.in/yaml/v4"
+)
+
+// Concurrency limits how many instances of a workflow may run at the same
+// time. It can be unmarshalled from:
```
lafriks:
```suggestion
// time. It can be unmarshaled from:
```

PR #6672, opened 2026-05-29
fix(kubernetes): retry WaitStep when container terminated state not yet finalized

simonckemper:
## Problem

Kubelet sets `pod.Status.Phase = Succeeded` before finalizing `containerStatuses[0].state.terminated`. When the informer sees the phase change and `WaitStep` calls `Get()`, the container status may still show `Terminated == nil`, causing a hard error:

```
no terminated state found for container wp-XXX/wp-XXX
```

This is a known race in the Kubernetes API server/kubelet eventually-consistent model. The window is normally milliseconds but widens to seconds under load (apiserver latency spikes, ResourceQuota admission storms, node pressure).

## Fix

Wrap the post-informer `Get()` + `Terminated == nil` check in `backoff.Retry` with exponential backoff (200ms initial, 5s max interval, 15s total budget). This mirrors the retry pattern already used for `TailStep` log stream recovery (#5550).

Key design decisions:
- **404 returns `backoff.Permanent`**: no retry needed — the pod was GC'd after the informer already observed completion. Returns `ExitCode=0, Exited=true` (existing behavior from v3.15.0).
- **No container statuses is retriable**: apiserver may return the pod before status is populated.
- **`Terminated == nil` is retriable**: kubelet is eventually consistent; a few hundred ms almost always resolves the race.

## Testing

- `go vet ./pipeline/backend/kubernetes/` passes
- Binary builds cleanly with `CGO_ENABLED=0`
- Deployed in production at RootReal CI (52-service Rust monorepo) as `v3.15.0-waitstep-retry`

## Evidence

This race has caused 15+ false pipeline failures in our CI over the past months. The most affected steps are fast-completing ones (cargo lock guard, audit steps) that finish in 50-75 seconds and hit the GC window with high probability. After deploying the patched agent with this retry, no further `no terminated state found` errors have been observed.

## Related

- #5550 — Added `backoff.Retry` for `TailStep` log streaming (same pattern)
- #6186 — Fixed goroutine leak on cancelled steps (added `ctx.Done()` to WaitStep) 
- #3331 — Fixed agent panic when node terminated during step execution


pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -353,7 +353,33 @@ func (e *kube) WaitStep(ctx context.Context, step *types.Step, taskUUID string)
 		return nil, ctx.Err()
 	}
 
-	pod, err := e.client.CoreV1().Pods(e.config.GetNamespace(step.OrgID)).Get(ctx, podName, kube_meta_v1.GetOptions{})
+	// After the informer signals completion, kubelet may not have finalized
+	// containerStatuses yet (phase=Succeeded before state.terminated is set).
+	// Retry with backoff to allow kubelet to catch up, using the same retry
```
6543:
```suggestion
	// Retry with backoff to allow kubelet to catch up.
```

pipeline/backend/kubernetes/kubernetes.go
```diff
@@ -353,7 +353,33 @@ func (e *kube) WaitStep(ctx context.Context, step *types.Step, taskUUID string)
...
+	// configuration as TailStep for log stream recovery (PR #5550).
```
6543:
```suggestion
```

6543 (approved):

PR #6680, opened 2026-05-30
pipeline: use more explizite error for step dependencies filtered out by its conditions

### LINKED ISSUE #6667: Too generic error message when step is not present in DAG
### Component

web-ui

### Describe the bug

When the step is present in the config but does not run because of a filter and another step depends on this step you only get a generic error message: `step 'two' depends on unknown step 'one'`

### Steps to reproduce

1. Write config with two steps where one depends on the other
2. Make a more restrictive filter for the first one
3. Run the config in a setup where the first one does not run

### Expected behavior

A better error message that clearly describes what went wrong.
Or a tip at the end of the error message encouraging the user to look at the filters.

### System Info

```shell
Codeberg CI version 3.14.1
```

### Additional context

https://ci.codeberg.org/repos/7772/pipeline/3816/errors

### Validations

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/6667

### DIFF
--- a/pipeline/frontend/yaml/compiler/compiler.go
+++ b/pipeline/frontend/yaml/compiler/compiler.go
@@ -15,6 +15,7 @@
 package compiler
 
 import (
+	"errors"
 	"fmt"
 	"maps"
 	"path"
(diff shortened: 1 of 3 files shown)


qwerty287:
The step hasn't necessarily been filtered out, it's also possible that the step just does not exist I think (e.g. due to a typo)

6543:
yes in this case the old error message stays ... should i also change it's text to be more clear?

qwerty287 (approved):
ah sorry, missed that. Is ok then

lafriks (approved):

PR #6681, opened 2026-05-31
Sort completed pipeline workflows by their execution order

### LINKED ISSUE #1480: Sort pipelines in multi pipeline execution by order instead of alphabet

When looking at the details of multi pipeline executions, the flow does not go top to bottom but is "random" (based on the pipeline name), this can sometimes be confusing, especially when all pipelines are unfolded and the running one is outside the current view (a finished one up top, and a couple non-running ones below, while only scrolling down shows the running one).

### Suggested solution

Order pipelines by their execution order.

### Alternative

Pipelines could auto collapse when done and pipelines waiting for other steps could also be collapsed, only opened up once something is running inside, though I believe this would be more effort (but I'm not a web dev)

### Additional context

image

If the two currently collapsed pipelines in the middle are opened up, the running pipeline is below the visible part, I've collapsed them to show what would be below...

I've tried to check for issues, but all I found was steps and the overview being resorted.

### Validations

hhharsha36:
Rearrange the workflows based on the `depends_on` execution order with fallback to `started` timestamp when available.

Falls back to PID for workflows that have not started.

Not an ideal fix since workflows which are yet to start will still be in their PID order, but very useful to analyse executed workflows and completed pipelines.

PS: It is my first time contributing to this project and I am learning, so please let me know if I missed something.

closes #1480

### DIFF
--- a/web/src/components/repo/pipeline/PipelineStepList.vue
+++ b/web/src/components/repo/pipeline/PipelineStepList.vue
@@ -63,7 +63,7 @@
     <div class="relative min-h-0 w-full grow">
       <div class="absolute top-0 right-0 left-0 flex h-full flex-col gap-y-2 md:overflow-y-auto">
         <div
-          v-for="workflow in pipeline.workflows"
+          v-for="workflow in sortedWorkflows"
           :key="workflow.id"
           class="border-wp-background-400 dark:border-wp-background-100 bg-wp-background-200 rounded-md border p-2"
         >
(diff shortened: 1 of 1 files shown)


lafriks:
Imho it would be better to use info from #6130 for this to have stable order even when steps are not executed yet

hhharsha36:
> Imho it would be better to use info from #6130 for this to have stable order even when steps are not executed yet

Sorry for my delayed response since I was occupied. Also, apologies as I was not aware of your PR.

I had a look at your PR and from my understanding, it dynamically constructs the dependency between workflows on the backend via the `DependsOn` key. If thats the case, then this PR is completely compatible and should work out of the box once #6130 is complete.

Let me know your thoughts.

PR #6684, opened 2026-05-31
fix(oauth): prevent Cloudflare 524 during Bitbucket login behind tunnel

tonyxkendo:
## Summary
  Fixes OAuth login failures (Cloudflare Error 524) when Woodpecker is deployed behind Cloudflare Tunnel with Bitbucket Cloud as the forge.
  The login page opened an idle SSE connection to `/api/stream/events` on load. During the OAuth round-trip, the reverse proxy closed that connection after ~30s. In practice this correlated with the
   browser never completing the `/authorize?code=...` callback, and login failed with `invalid_grant` / expired state token errors.
  This PR addresses three contributing factors:
  - **Frontend:** Skip SSE on the login page and other routes where it is not needed; close SSE explicitly before OAuth redirect; keep SSE for anonymous public repo views.
  - **Server:** Shorten SSE keepalive interval from 30s to 15s and send both comment and named `event: ping` data frames so proxies reset idle timers reliably.
  - **Server:** Defer post-login repo permission sync to a background goroutine so the OAuth callback can respond quickly (important for users with many Bitbucket repos).
  ## Root cause
  1. Browser opens SSE at `/api/stream/events` on page load
  2. User clicks login → redirect to Bitbucket
  3. Cloudflare Tunnel closes the idle SSE stream after ~30s
  4. OAuth callback from Bitbucket does not complete successfully
  5. Browser shows Cloudflare Error 524
  Existing SSE keepalives were sent every 30s, which matched the observed proxy timeout window.
  ## Changes
  ### Web UI
  - Route-aware SSE subscription via `shouldSubscribeEvents()` / `setupEvents(router)`
  - No SSE on `guest-only` routes (login)
  - No SSE on `authentication: required` routes without a logged-in user
  - SSE still enabled for anonymous public repo pages
  - `closeEvents()` called before OAuth redirect
  ### Server
  - SSE ping interval: 30s → 15s, using `time.Ticker`
  - Heartbeat frame: `: ping` comment + `event: ping\ndata: {}`
  - Applied to both event and log SSE streams
  - Login: set session cookie and redirect first; sync repo permissions in background
  ## Test plan
  - [ ] Open `/login` → confirm no `/api/stream/events` request in browser Network tab before clicking login
  - [ ] Complete Bitbucket OAuth behind Cloudflare Tunnel → login succeeds, no Error 524
  - [ ] After login, SSE opens and pipeline/repo updates still stream live
  - [ ] View a public repo while logged out → live updates still work
  - [ ] Login with many Bitbucket repos → redirect to home is fast; repos appear after background permission sync
  ### Automated tests
  ```bash
  go test ./server/api/... -short -count=1
  go test ./server/api/... -count=1
  pnpm --dir web typecheck
  pnpm --dir web test --run
  ```
  ## Environment where reproduced
  | Field | Value |
  |-------|--------|
  | Woodpecker version | 3.15.0 |
  | Forge | Bitbucket Cloud |
  | Deployment | Docker Compose |
  | Reverse proxy | Cloudflare Tunnel (free tier) |
  | Bitbucket repos | 100+ |

### DIFF
--- a/server/api/login.go
+++ b/server/api/login.go
@@ -15,6 +15,7 @@
 package api
 
 import (
+	"context"
 	"encoding/base32"
 	"errors"
 	"fmt"
@@ -43,6 +44,25 @@ const (
 	perPage            = 50
 	maxPage            = 10000
 )
+
+// SyncRepoPermissionsFn runs repo permission sync after login.
+// Tests may replace this to run synchronously.
+var SyncRepoPermissionsFn = syncRepoPermissionsAsync
+
+func syncRepoPermissionsAsync(ctx context.Context, user *model.User, _store store.Store, _forge forge.Forge, forgeID int64) {
+	go func() {
+		if err := updateRepoPermissions(ctx, user, _store, _forge, forgeID); err != nil {
+			log.Error().Err(err).Msgf("background repo permission sync failed for user %s", user.Login)
+		}
+	}()
+}
+
+// SyncRepoPermissionsSyncForTesting runs permission sync synchronously in tests.
+func SyncRepoPermissionsSyncForTesting(ctx context.Context, user *model.User, _store store.Store, _forge forge.Forge, forgeID int64) {
+	if err := updateRepoPermissions(ctx, user, _store, _forge, forgeID); err != nil {
+		log.Error().Err(err).Msgf("background repo permission sync failed for user %s", user.Login)
+	}
+}
 
 func HandleAuth(c *gin.Context) {
 	// TODO: check if this is really needed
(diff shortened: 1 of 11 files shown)


6543:
> Server: Defer post-login repo permission sync to a background goroutine so the OAuth callback can respond quickly (important for users with many Bitbucket repos).

can you remove this!

this is worth it's own pull on it's own. beside if i can remember correctly we already have such pulls tjat where just not merged because of tje question to force a sync before user logon because of security or do it in the background -> could give a window of atack ... so at least it should be configurable and background disbled by default.

also this pull do a lot at once, can you move them into many ... luke not having Stream open on non needed pages is worth on its own

6543:
background pull: https://github.com/woodpecker-ci/woodpecker/pull/5665

6543:
pull conflicts

PR #6689, opened 2026-06-01
fix(server): embed IANA timezone database for cron schedules

### LINKED ISSUE #6688: Cron schedules fail with “unknown time zone” in server container
### Component

server

### Describe the bug

Even when:
- TZ=Europe/Berlin is set

### Steps to reproduce

`

### Expected behavior

### System Info

```shell
```

### Additional context

### Validations

toller892:
## Problem

Cron schedules fail to accept any non-UTC timezone in minimal containers (Alpine, scratch) that lack system timezone data.

```
Error inserting cron. schedule could not parsed: unknown time zone Europe/Berlin
```

This affects both `TZ` and `WOODPECKER_TIMEZONE` environment variables.

Fixes #6688

## Root Cause

`time.LoadLocation()` relies on the IANA timezone database being available on the system. Alpine Linux and scratch containers don't include this by default.

## Fix

Import `time/tzdata` to embed the IANA timezone database into the server binary. This is the standard Go approach for ensuring timezone support in minimal environments.

### Changes
- `cmd/server/main.go`: Add `import _ "time/tzdata"`
- `cmd/server/tzdata_test.go`: Regression test verifying timezone loading works

### Trade-offs
- Binary size increases by ~450KB (the IANA timezone database)
- This is the same approach recommended by the [Go documentation](https://pkg.go.dev/time/tzdata)

## Testing

Run the regression test to verify the timezone database is embedded:

```
go test ./cmd/server/ -run TestTimezoneDatabaseAvailable -v
```

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -20,6 +20,11 @@ import (
 	"context"
 	"os"
 
+	// Embed the IANA timezone database so that cron schedule
+	// timezones work in minimal containers (Alpine, scratch)
+	// that lack system timezone data.
+	_ "time/tzdata"
+
 	_ "github.com/joho/godotenv/autoload"
 	"github.com/rs/zerolog/log"
 
--- a/cmd/server/tzdata_test.go
+++ b/cmd/server/tzdata_test.go
@@ -1,0 +1,31 @@
+package main
+
+import (
+	"testing"
+	"time"
+)
+
+// TestTimezoneDatabaseAvailable verifies that the embedded IANA timezone
+// database is available. This ensures cron schedules work in minimal
+// containers (Alpine, scratch) that lack system timezone data.
+// Regression test for https://github.com/woodpecker-ci/woodpecker/issues/6688
+func TestTimezoneDatabaseAvailable(t *testing.T) {
+	timezones := []string{
+		"Europe/Berlin",
+		"America/New_York",
+		"Asia/Tokyo",
+		"Australia/Sydney",
+		"UTC",
+	}
+
+	for _, tz := range timezones {
+		loc, err := time.LoadLocation(tz)
+		if err != nil {
+			t.Errorf("time.LoadLocation(%q) failed: %v", tz, err)
+			continue
+		}
+		if loc == nil {
+			t.Errorf("time.LoadLocation(%q) returned nil location", tz)
+		}
+	}
+}


cmd/server/tzdata_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"testing"
+	"time"
+)
+
+// TestTimezoneDatabaseAvailable verifies that the embedded IANA timezone
+// database is available. This ensures cron schedules work in minimal
+// containers (Alpine, scratch) that lack system timezone data.
+// Regression test for https://github.com/woodpecker-ci/woodpecker/issues/6688
```
6543, line 11: remove the test ... in normal dev/test enfironments we dont have this problem, but in our from scratch container, we do - witch this test never runs in

cmd/server/main.go
```diff
@@ -20,6 +20,11 @@ import (
 	"context"
 	"os"
 
+	// Embed the IANA timezone database so that cron schedule
+	// timezones work in minimal containers (Alpine, scratch)
+	// that lack system timezone data.
```
6543, line 25: we should move this codeblock to the cron package where timezone parsing is happening

6543:
also in this regards we should check if local timezone setting was mounted into the container ... and use that as default - else UTC , not sure if this simple trick does the job already that way ?

qwerty287:
There's the build flag `-tags timetzdata`, instead of embedding the data into all binaries (including non-docker ones), couldn't we just add compile with this tag when doing the xgo cross compilation? The cross-compiled binaries are added into the docker images then.

lafriks:
Why not just add timezone data into docker image?

See sample for scratch:

https://github.com/wntrtech/scratch/blob/main/Dockerfile

qwerty287:
closing this PR as it's the wrong approach. Maybe you could open a PR with your idea @lafriks?

lafriks:
Sure, I could do it this week

greggigon:
> closing this PR as it's the wrong approach. Maybe you could open a PR with your idea @lafriks?

I would not say it's a wrong approach, it's a different approach.
Both have it's pros and cons.

Has this actually been fixed? I still get the error on the latest stable build?
`woodpeckerci/woodpecker-server:v3-alpine`

PR #6692, opened 2026-06-03
Fix group logs by command in step, by detecting commands correctly

### LINKED ISSUE #6691: Group logs by command in step logs UI Fail at pnpm install
regression of https://github.com/woodpecker-ci/woodpecker/pull/6398

... noticed this doesn't work with pnpm install: https://ci.woodpecker-ci.org/repos/3780/pipeline/34411/7

_Originally posted by @qwerty287 in https://github.com/woodpecker-ci/woodpecker/issues/6398#issuecomment-4475058246_

6543:
close https://github.com/woodpecker-ci/woodpecker/issues/6691

we now add a yaml parser, so we dont need to depend on flaky regex. (regex is fine in narrow scope, but the yaml can change and is just user input...)

also we extract the logic into unit testable functions!


6543:
I'll add a yaml parser ...

... maintaining regex is not a good thing in the logn run :/

6543:
# works:
image
image
image

qwerty287:
I'm not sure if I like adding a yaml parser to the frontend. How much does that increase the frontend bundle size?

6543:
# main

image
image

# pull
image
image

# diff

## -> 31kb

6543:
the other option would be to mark it in backend ...

... by extend:
https://github.com/woodpecker-ci/woodpecker/blob/925931726162ea011319095f4d115cdd99490282/woodpecker-go/woodpecker/const.go#L42-L51

and use it ...

qwerty287:
> the other option would be to mark it in backend ...
> 
> ... by extend:
> 
> https://github.com/woodpecker-ci/woodpecker/blob/925931726162ea011319095f4d115cdd99490282/woodpecker-go/woodpecker/const.go#L42-L51
> 
> and use it ...

That would be my preferred way, but I guess it's way harder to do this

6543:
yes i thougt long and wide ... but for now it is not realy doable in backend at all.

yes you could do it but it would be a messy code ... the only place where it could make sence is at the log api and that will remove grouping from streaming loggs ...

6543:
well we do a lot of stuff with yaml so having a parser in the ui is not the worst - i can thing of other things to use if for too... like link on a step in log view to jump to to config definition etc..

qwerty287:
https://github.com/woodpecker-ci/woodpecker/issues/6691#issuecomment-4932136806 is a good idea I think instead of this approach

PR #6696, opened 2026-06-03
Add more tests in areas not well covered

6543:
used latest opus 4.8 to generate tests against our implementations ...

# TODOs
- [x] check myselfe if tests do actually make sense or just validate 1+1=2 ;)


6543:
well the `server/forge/types/errors_test.go` just increases code coverage ... rest is valid ... still not huge improvements just some more coverage of the current implementation.

i would say we can merge it but I'm not sure if it was worth my effort to point claude to get more test coverage ...
... and review it

qwerty287 (approved):

Issue #6708, opened 2026-06-08
runAsUser: 0 silently dropped without privileged: true


nijave:
### Component

agent

### Describe the bug

runAsUser: 0 is silently dropped/ignored when the step doesn't have `privileged: true` set

### Steps to reproduce

```
when:
  - event: pull_request
  - event: push
    branch: main

steps:
  mirror-images:
    # renovate: datasource=docker depName=ghcr.io/regclient/regctl
    image: ghcr.io/regclient/regctl:v0.11.5-alpine
    backend_options:
      kubernetes:
        # regclient image runs as non-root; need root to apk add yq/jq.
        securityContext:
          runAsUser: 0
          runAsGroup: 0
    commands:
      - apk add --no-cache yq jq git
      - sh .ci/mirror-images.sh
```

### Expected behavior

It appears this was originally done for "security" purposes but at the very least I'd expect some sort of log or error somewhere this is happening.

Ideally this configuration is supported--`privileged: true` is a radically different security posture than uid 0 with privileges dropped. It is especially much different when running userns where you're not even real root on the host anyway

### System Info

```shell
https://github.com/nijave/vmubtkube-a/tree/main/woodpecker

3.15.0
```

### Additional context

What is especially odd is I can change to a different Docker image that defaults to run as root and everything works fine. Blocking `runAsUser: 0` doesn't seem to make much sense given I can trivially change the runner image to one that does the same thing in the image manifest (vs k8s manifest)

To create a more reasonable security boundary, `runAsNonRoot: true` should be injected although I still highly disagree this should have any linkage to `securityContext.privileged: true`

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

AhmadNajiKam:
---

###  Bug Reproduction: `runAsUser: 0` silently ignored on both Kubernetes and Docker backends without `privileged: true`

I have successfully verified and reproduced this issue using a minimal test case on a fresh environment. The behavior is identical across both backends: `runAsUser: 0` is dropped entirely by the agent unless the container is explicitly forced into a `privileged` posture.

### 💻 Environment Details

* **Host OS:** Arch Linux (Kernel `7.0.9-arch2-1`)
[...]

PR #6710, opened 2026-06-08
Add k8s userns support

nijave:
sidesteps "what is privileged" debate in #6708

### DIFF
--- a/docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
+++ b/docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
@@ -97,6 +97,29 @@ services:
         workspaceVolume: false
 ```
 
+### User namespaces
+
+`hostUsers` controls whether the Pod uses the host's user namespace. When set to `false`, Kubernetes runs the Pod in a dedicated user namespace where UID 0 inside the container maps to a non-root UID on the host, providing an additional layer of isolation.
+
+See the [Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/) for more information on user namespaces.
+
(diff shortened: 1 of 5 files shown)


qwerty287:
Seems fine from code to me, @woodpecker-ci/maintainers could someone check the impact?

6543:
does this option can have any security impact?

6543:
> The runAsUser, runAsGroup, fsGroup, etc. fields in the pod.spec always refer to the user inside the container. These users will be used for volume mounts (specified in pod.spec.volumes) and therefore the host UID/GID will not have any effect on writes/reads from volumes the pod can mount. In other words, the inodes created/read in volumes mounted by the pod will be the same as if the pod wasn't using user namespaces.

> This way, a pod can easily enable and disable user namespaces (without affecting its volume's file ownerships) and can also share volumes with pods without user namespaces by just setting the appropriate users inside the container (RunAsUser, RunAsGroup, fsGroup, etc.). This applies to any volume the pod can mount, including hostPath (if the pod is allowed to mount hostPath volumes).

6543 (approved):
I wonder if we can make it be default to true or if this affects to much 🤔 

@xoxys

nijave:
> does this option can have any security impact?

yeah, should generally be improvement basically user in pod no longer map to user on host (so you can run root in pod and it's no longer uid 0/host root)

nijave:
> I wonder if we can make it be default to true or if this affects to much 🤔
> 
> @xoxys

imo kind of risky at this point since k8s cluster requires special, optional config to support the feature. Could be useful to have a feature flag/config that turns it on globally across woodpecker tho (so pipeline changes aren't needed)

PR #6712, opened 2026-06-09
More test coverage of `server/api/...`

6543:
last time i told clause to go and find areas to increase code coverage where it makes sense and is not just 1+1=2 tests ...

... this time i pointed it to a specific area (server/api/) and let it only done one topic e.g. cron at a time ... also each step where it normally would make desissions i told it to not decide but ask and let me do it ...

so this is a successor of https://github.com/woodpecker-ci/woodpecker/pull/6696 very strickt guidience ... lets see how that goes

-> i used sonnet4.6 for dev setup and info gathering and opus4.8 to write the tests ...


server/store/datastore/teststore.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// testStoreDriverConfig returns the database driver and connection string used
+// for tests. It defaults to in-memory sqlite but can be overridden via
+// environment variables to run the same tests against mysql/postgres.
+func testStoreDriverConfig() (driver, config string) {
+	driver = "sqlite3"
+	config = ":memory:"
+
+	if os.Getenv("WOODPECKER_DATABASE_DRIVER") != "" {
```
6543: do we want other drivers supported here?!?

server/api/global_secret_test.go
```diff
@@ -0,0 +1,246 @@
...
+		assert.Equal(t, http.StatusBadRequest, rec.Code)
+	})
+
+	t.Run("validation fails on empty value", func(t *testing.T) {
+		in := &model.Secret{Name: "api_token", Events: []model.WebhookEvent{model.EventPush}}
+		c, rec := newGlobalSecretCtx(t, http.MethodPost, in)
+
+		PostGlobalSecret(c)
+
+		// Note: the global handler returns 400 on validation failure, unlike
```
qwerty287: Can we add a todo note to change this in the next major?

6543: i can change it in a followup :)

qwerty287: Changing the status code is a breaking change I think. I would just add a todo comment so we change it in the next major

6543: it is non breaking as it never was documented

qwerty287 (approved):

Issue #6717, opened 2026-06-10
local backend: on WINDOWS: use job object to group and kill commands we spawn


6543:
tldr: what we did for unix with https://github.com/woodpecker-ci/woodpecker/pull/6609 but for windows

long ...:

with https://github.com/woodpecker-ci/woodpecker/pull/6609 we move commands the local backend spawn in a seperate process group that is killed as a whole on cancel. on windows we dont have pgrocess-groups but you can group by JOB_OBJECT ...

https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-jobobject_basic_limit_information

NOTE: windows own libs use own exec implementation as golang std do not support it:

**1. Create job** — `windows.CreateJobObject` wrapper
https://github.com/microsoft/hcsshim/blob/5255d2c0a9496129100dc43302ee3adcb10f696c/internal/jobobject/jobobject.go#L100

**2. Set `KILL_ON_JOB_CLOSE`** — the kernel-kill-on-handle-close flag
https://github.com/microsoft/hcsshim/blob/5255d2c0a9496129100dc43302ee3adcb10f696c/internal/jobobject/limits.go#L67-L75
(flag OR'd at L72)

**3. THE key bit — born-in-job binding.** Attach job to STARTUPINFOEX attr list via `PROC_THREAD_ATTRIBUTE_JOB_LIST`:
https://github.com/microsoft/hcsshim/blob/5255d2c0a9496129100dc43302ee3adcb10f696c/internal/jobobject/jobobject.go#L280-L298

**4. Where it's wired into `CreateProcess`** — their own exec pkg (not os/exec). This is the whole reason it works:
- attr list alloc + `EXTENDED_STARTUPINFO_PRESENT`: https://github.com/microsoft/hcsshim/blob/5255d2c0a9496129100dc43302ee3adcb10f696c/internal/exec/exec.go#L111-L127
- job update into attr list: https://github.com/microsoft/hcsshim/blob/5255d2c0a9496129100dc43302ee3adcb10f696c/internal/exec/exec.go#L162-L165
- attr list → `siEx` → `CreateProcess`: https://github.com/microsoft/hcsshim/blob/5255d2c0a9496129100dc43302ee3adcb10f696c/internal/exec/exec.go#L178-L210
- the `WithJobObject` option: https://github.com/microsoft/hcsshim/blob/5255d2c0a9496129100dc43302ee3adcb10f696c/internal/exec/options.go#L52-L57

[...]

PR #6718, opened 2026-06-10
local backend: on windows use process tree-walk as workaround to kill

6543:
not recommended as it could race and leave sub childs of dead parents dangling, but the best we have atm

workaround for #6717


6543:
I wana test it on real windows with actuall load before i mark this non wip :)

pipeline/backend/local/cmd_win.go
```diff
@@ -19,10 +19,18 @@ package local
...
 	"context"
 	"os/exec"
+	"strconv"
 )
 
 func newCmd(ctx context.Context, binary string, args ...string) *exec.Cmd {
 	cmd := exec.CommandContext(ctx, binary, args...)
 
+	// Non perfect workaround till std exec supports JOB_OBJECT
+	// https://github.com/woodpecker-ci/woodpecker/issues/6717
```
6543:
```suggestion
	// https://github.com/woodpecker-ci/woodpecker/issues/6717 & https://github.com/golang/go/issues/79927
```

6543:
## before patch
image


## with the patch
image

also watched proccesses - they are gone :)

qwerty287 (approved):

Issue #6595, opened 2026-05-17
Feature: auto-retry pipelines killed by agent disconnect (distinguish from intentional cancelPreviousPipelines kills)


amalc:
## Summary

When a CI agent is preempted (e.g. a spot/preemptible VM is reclaimed by the cloud provider), Woodpecker marks the in-flight pipeline as `killed`. There is no automatic retry — the pipeline stays dead and the operator must manually retrigger via the API or UI.

At the same time, `cancelPreviousPipelines` also produces `killed` pipelines — but those are intentional: a newer push superseded the old one and retry would be wasteful/incorrect.

Today there is no way to distinguish the two cases, so Woodpecker cannot selectively retry the unintentional kills.

## Desired behavior

When a pipeline is killed because its agent disconnected (heartbeat timeout / agent restart), Woodpecker should automatically re-queue it for execution on a fresh agent — identical to submitting the pipeline fresh from the same commit.

When a pipeline is killed by `cancelPreviousPipelines`, it should remain dead (no retry).

## Suggested implementation

1. Persist a `kill_reason` on the pipeline: `agent_disconnect` vs `cancelled` (cancelPreviousPipelines) vs `manual` (operator action).
2. On agent-disconnect kill, re-enqueue the pipeline with the same commit/config rather than leaving it in `killed` state.
3. Expose `kill_reason` in the API response and UI so operators can see why a pipeline was killed.

## Context

We run Woodpecker against GCP spot VMs (scale-to-zero fleet). Agent preemptions are infrequent but not rare — a pipeline mid-rsync or mid-compile gets hard-killed when the VM is reclaimed. Today the only recovery is a manual `POST /api/repos/:id/pipelines/:number` retrigger, which requires operator awareness and action.

Our fork addressed a related symptom (`ReconcileOrphaned` over-firing on legitimate pipelines after agent disconnect — Peregrine fork #176), but the underlying auto-retry gap remains.

## Related

- Woodpecker `cancelPreviousPipelines` setting
- Peregrine fork #176 — ReconcileOrphaned kills legitimate in-flight pipelines after agent disconnect

6543:
> When a pipeline is killed because its agent disconnected (heartbeat timeout / agent restart), Woodpecker should automatically re-queue it for execution on a fresh agent — identical to submitting the pipeline fresh from the same commit.

this should already happen ... there is a mechanism in the rpc (Extend) that ackts as hartbeat, if it fails, the queue reschedules it after an timeout...

Issue #884, opened 2022-04-15
Add retry / onerror directive to pipelines


fnetX:
### Clear and concise description of the problem

Sometimes it happens that pipelines fail not because of a fault in the changeset, but because of external circumstances (network failures etc).

Especially for prepare pipelines that just set up an environment and are not supposed to check the actual change, doing an automatic retry could be desirable.

Especially with #767 it would be a problem if a failed prepare pipeline is then cached for a while (probably failed pipelines shouldn't be cached, but still) ...

### Suggested solution

Add a `retry`, or `onerror`, `onfail` directive that can take different values for retry policies, e.g.

- none: default behaviour, no retry
- retry: rebuild the step, optionally with a number specified, e.g. retry(3) = retry 3 times, or retry:2 (only twice)
- fast-retry: only retry if the pipeline failed fast, e.g. due to dependency cloning, but not when it was already running long (ideally with a threshold in seconds)
- skip: report as failure, but run the following pipeline anyway (e.g. to see if the project builds even when the lint step fails)
- ignore: mark the step as failed in the Woodpecker UI, but ignore it for the final report and continue with next pipeline (if all other pipelines succeed, report as successful)

I think this could be a great feature for Woodpecker CI.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

anbraten:
There was a setting added for `failure` https://woodpecker-ci.org/docs/next/usage/pipeline-syntax#failure. The retry logic would probably be best implemented by the script / tool. Exp jest, vitest, playwright have integrated retry logics and for go sth like https://pkg.go.dev/github.com/hashicorp/consul/sdk/testutil/retry could be used.

zc-devs:
It would be nice to have such a feature.
```yaml
# Environments
# SKIP_IMAGE=true|false - skip building image
# SKIP_PUBLISH_QUAY=true|false - skip publishing in Quay
# SKIP_PUBLISH_GHCR=true|false - skip publishing in GitHub

variables:
  - &kaniko_image "gcr.io/kaniko-project/executor:v1.22.0-debug"
[...]

PR #6719, opened 2026-06-10
Automatically retry pipelines that fail due to infrastructure events

fuhrysteve:
## Problem

When a build runs on preemptible / spot infrastructure, a node can be reclaimed mid-step. Kubernetes takes the step's pod down (eviction, or SIGKILL → exit 137) while the agent stays connected and reports the workflow as failed. Because there is no `retry: when: system_failure` primitive (as GitLab and Buildkite have), the pipeline fails exactly like a real test failure and has to be restarted by hand.

This is distinct from an agent disconnect (#6595): here the agent is alive, so the queue's existing `resubmitExpiredPipelines` recovery never triggers. It also differs from a declarative retry directive (#884): the goal is automatic, server-side recovery that needs no per-repo opt-in.

## What this does

Detects when a step failed because the infrastructure took it down (rather than the workload itself failing) and automatically restarts the pipeline, bounded by a configurable budget. A pipeline is retried only when **every** failing step is an infrastructure failure — a single genuine failure disables the retry, so real errors are never masked.

Disabled by default (`WOODPECKER_INFRA_RETRY_MAX_ATTEMPTS=0`).

### Detection

The Kubernetes backend reads the pod's `DisruptionTarget` condition, which Kubernetes sets when it takes a pod down for preemption, eviction, or graceful node shutdown (the pod `Reason` field is used as a fallback). This is more reliable than matching exit code 137 or error strings, which are ambiguous — a workload can legitimately exit 137 on its own.

The condition is captured from informer events the moment it appears, not only from the post-completion pod read, so the signal survives the pod disappearing before a clean terminal status is observable (for example when the node is lost).

Only the Kubernetes backend sets the flag today; other backends always report `false`.

### Flow

1. The Kubernetes backend sets `State.InfraFailure` when the pod carries `DisruptionTarget` (`pipeline/backend/kubernetes`).
2. The flag travels agent → server on `StepState` (new proto field) and is persisted on the step.
3. When a pipeline finishes in a failed state, the server restarts it if every failing step is an infra failure and the attempt budget is not spent (`server/pipeline`, triggered from the `Done` RPC). The restart runs as the repo owner — the same identity cron pipelines use, since an automatic retry has no initiating user.
4. `Pipeline.InfraRetryCount` bounds the chain and is incremented atomically as part of creating the restarted pipeline, so a failed persist can never leave a stale count that retries forever.

## Configuration

`WOODPECKER_INFRA_RETRY_MAX_ATTEMPTS` (default `0`, disabled). Documented in the server configuration reference.

## Testing

Three levels.

**Unit** — the retry decision (infra-only is retried; a genuine failure, an exhausted budget, and non-failure terminal states are not) and the `DisruptionTarget` reader.

**In-process end-to-end** (`make test-e2e`) — a real server and agent over gRPC with the dummy backend, which gains a `STEP_INFRA_FAILURE` knob alongside the existing `STEP_OOM_KILLED` / `STEP_EXIT_CODE` ones. Scenarios assert that an infra-only failure auto-restarts (one retry, correct parent and count), a genuine failure does not, and the budget is a hard ceiling.

**Live cluster** (`make test-kind`, `-tags kind`) — creates a real pod on a `kind` cluster, evicts it through the Eviction API (which sets `DisruptionTarget` exactly as a spot-node preemption does), and asserts the backend's `WaitStep` returns an infra-failure state. The test was confirmed to bite: neutering the detection flips it from pass to fail (the evicted pod terminates with exit 137, but the flag stays `false` without the condition read).

Building the live-cluster test is what surfaced the informer-capture hardening described under Detection — a node-lost pod that vanished before reaching a terminal status would otherwise have been missed.

## Related

- #6595 — auto-retry on agent disconnect. This generalises that idea to any infrastructure-flagged step failure; agent disconnect is just one source.
- #884 — a declarative `retry:` directive could later consume the same `InfraFailure` signal for per-pipeline control.

## Notes for reviewers

- `rpc/proto/woodpecker.pb.go` was regenerated for the new `StepState.infra_failure` field (field 9, additive and wire-compatible). I regenerated with a locally available `protoc`; if the generator provenance comment matters, please regenerate with the project's pinned toolchain.
- The live-cluster test (`make test-kind`, build tag `kind`) needs a real Kubernetes cluster and is not wired into a CI lane — it is a local/manual proof that a real eviction sets `DisruptionTarget` and the backend detects it. The in-process e2e (`make test-e2e`) is the gating test. Happy to add a CI job for it if that's preferred.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,6 +206,9 @@ test-lib: ## Test lib code
 test-e2e: ## Test by running yaml config and compare expected result
 	go test -race -cover -coverpkg=./... -coverprofile e2e-coverage.out -timeout 60s -tags 'test $(TAGS)' ./e2e/...
 
+test-kind: ## Test the Kubernetes backend against a live cluster (needs: kind create cluster --name wp-e2e)
+	go test -timeout 200s -tags kind ./pipeline/backend/kubernetes/...
+
 .PHONY: test
 test: test-agent test-server test-server-datastore test-cli test-lib test-e2e ## Run all tests
 
--- a/agent/rpc/client_grpc.go
+++ b/agent/rpc/client_grpc.go
@@ -316,14 +316,15 @@ func (c *client) Update(ctx context.Context, workflowID string, state rpc.StepSt
 	req := &proto.UpdateRequest{
 		Id: workflowID,
 		State: &proto.StepState{
-			StepUuid: state.StepUUID,
-			Started:  state.Started,
-			Finished: state.Finished,
-			Exited:   state.Exited,
-			ExitCode: int32(state.ExitCode),
-			Error:    state.Error,
-			Canceled: state.Canceled,
-			Skipped:  state.Skipped,
+			StepUuid:     state.StepUUID,
+			Started:      state.Started,
+			Finished:     state.Finished,
+			Exited:       state.Exited,
+			ExitCode:     int32(state.ExitCode),
+			Error:        state.Error,
+			Canceled:     state.Canceled,
+			Skipped:      state.Skipped,
+			InfraFailure: state.InfraFailure,
 		},
 	}
 
--- a/agent/tracer.go
+++ b/agent/tracer.go
@@ -38,12 +38,13 @@ func (r *Runner) createTracer(ctxMeta context.Context, logger zerolog.Logger, wo
 			Logger()
 
 		stepState := rpc.StepState{
-			StepUUID: state.CurrStep.UUID,
-			Exited:   state.CurrStepState.Exited,
-			ExitCode: state.CurrStepState.ExitCode,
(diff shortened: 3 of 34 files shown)


6543 (changes requested):
we already have this retry feature if https://github.com/woodpecker-ci/woodpecker/blob/31dd64fd64dd53e8738a8c54a184eecad74f8fcc/rpc/peer.go#L193 is not called by the agent anymore

else be fore we add all that retry logic into the whole codebase - why not make infra more reliable on the parts where it actually fails

6543:
for your issue, the kubernetis backend should handle that!

PR #6721, opened 2026-06-11
Support Unix Sockets

6543:
If you don't want to go through the network stack, just to let a reverse proxy get to your woodpecker, unix sockets are nice!

this also allows you to e.g. use docker with volumes to share it without ever touching the network stack.

golang and the libs fulli support it, we just have to expose it via our settings, and while doing so i add some error checks and docu for it.

image

### DIFF
--- a/agent/rpc/dial.go
+++ b/agent/rpc/dial.go
@@ -18,6 +18,9 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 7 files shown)


qwerty287 (approved):
lgtm, untested from my side

6543:
i tested sunny codepath and error cases, myselfe :)

JFronny:
Thanks for implementing this!
I successfully set this up on my server, but while doing so, noticed an issue I had to work around:
I did not find any way to configure the permissions the unix socket created by woodpecker-server should get.
Gitea exposes `UNIX_SOCKET_PERMISSION` for this purpose, which makes using it from a reverse proxy a bit easier.
Are there any plans to do something similar here?
I would offer to try to implement this, but I doubt I will be able to get to it anytime soon.

6543:
well just propose this enhancement as issue and of you want try to implement it, once you have an initiall draft, we can guide you :)

PR #6727, opened 2026-06-14
Recover agent from expired gRPC access token

### LINKED ISSUE #4144: agent connection seems broken, rpc access token expired
### Component

server, agent

### Describe the bug

Setup a woordpecker server and agent via 2 docker containers with a shared `WOODPECKER_AGENT_SECRET`
Everything works fine at first, but after a couple of hours the communication between server and agent fails.
The logs from the agent are:
```
 ERR src/agent/rpc/client_grpc.go:461 > grpc error: report_health(): code: Unauthenticated error="rpc error: code = Unauthenticated desc = access token is invalid: invalid token: token has invalid claims: token is expired"
5:40PM ERR src/cmd/agent/core/agent.go:251 > failed to report health error="rpc error: code = Unauthenticated desc = access token is invalid: invalid token: token has invalid claims: token is expired"
```
The server doesn't log anything relevant about it I think.
So since this is so basic, I really hesitated to report that bug, believing I made some mistake, so I actually recreated my real setup with the most basic setup just locally and it is the same problem.

### Steps to reproduce

1. start server with:
```
docker run --rm -it -p 8000:8000 -p 9000:9000 -v ./datatest:/var/lib/woodpecker -e WOODPECKER_HOST=http://localhost:8000 -e WOODPECKER_AGENT_SECRET=ADFI34YAKMIGSNKK55IPCKVVJWNULOBQY2QRIBJ42X527NY7GLTQ===1 -v /var/run/docker.sock:/var/run/docker.sock -e WOODPECKER_LOG_LEVEL=debug -e WOODPECKER_GITEA=true -e WOODPECKER_GITEA_CLIENT=... -e WOODPECKER_GITEA_SECRET=...-e WOODPECKER_GITEA_URL=... -e WOODPECKER_OPEN=true -e WOODPECKER_ADMIN=martin --name woodpecker-server-test woodpeckerci/woodpecker-server:v2.7.1-alpine
```
2. start agent with:
```
docker run --rm -it --network host -e WOODPECKER_SERVER=localhost:9000 -e WOODPECKER_AGENT_SECRET=ADFI34YAKMIGSNKK55IPCKVVJWNULOBQY2QRIBJ42X527NY7GLTQ===1 -v /var/run/docker.sock:/var/run/docker.sock -e WOODPECKER_LOG_LEVEL=debug  --name woodpecker-runner-test woodpeckerci/woodpecker-agent:v2.7.1-alpine
```
3. check it works
4. wait for a day and you will see the error in the agent logs, server doesn't report anything related i think

### Expected behavior

_No response_

### System Info

```shell
* woodpeckerci/woodpecker-server:v2.7.1-alpine
* woodpeckerci/woodpecker-agent:v2.7.1-alpine
```

### Additional context

_No response_

### Validations

OfficialOzzy:
## Summary

Fixes an issue where an agent could get permanently stuck using an expired gRPC access token.

The agent stored its access token as a plain string refreshed only on a 30-minute timer. When the server rejected a token with `codes.Unauthenticated` (typically because it expired), there was no recovery path: the error was classified as **permanent**, so the agent's RPC loops (`next()`, `report_health()`, workflow init/wait, …) kept retrying with the dead token and never recovered, orphaning queued/pending workflows. The token field was also read and written across goroutines without synchronization.

## Changes

- **`agent/rpc/auth_interceptor.go`** — the access token is now guarded by a `sync.RWMutex` (`Token()` / `setToken()`). A new exported `RefreshToken(ctx, staleToken)` forces re-authentication; a `refreshMu` serializes it with the background refresh timer, and the `staleToken` check collapses a burst of concurrent refreshes into a single `Auth` call (no stampede). No token material is logged.
- **`agent/rpc/client_grpc.go`** — new optional `tokenRefresher` wired via `SetAuthRefresher`. `classifyRPCErr` now treats `codes.Unauthenticated` as retryable, and `withAuthRefresh` re-authenticates and lets backoff retry with the fresh token, bounded by `maxAuthRefreshes` so a genuinely bad/revoked secret cannot loop forever (even with an infinite connection-retry timeout). When no refresher is configured the code stays permanent — preserving previous behavior.
- **`cmd/agent/core/agent.go`** — wires the auth interceptor into the client.

## Behavior / config

No public configuration changes. The existing 30-minute refresh timer still handles the normal case; this adds a **reactive** safety net for cases the timer path missed (e.g. server restart during a refresh window).

## Tests

- `TestExpiredTokenRecovery` — over an in-process gRPC server, an agent starting with a rejected (expired) token re-authenticates and the retried RPC succeeds.
- `TestRefreshTokenDedup` — concurrent refreshes sharing a stale token collapse into a single re-auth.
- `TestWithAuthRefresh` — recovery-after-refresh, give-up-after-budget, refresh-failure-stays-retryable, non-auth-errors pass through, nil-refresher stays permanent.
- `TestAuthInterceptorTokenConcurrency` — token access is race-free under `-race`.

All pass under `make test-agent` and `make lint` (golangci-lint) is clean.

Closes #4144

### DIFF
--- a/agent/rpc/auth_interceptor.go
+++ b/agent/rpc/auth_interceptor.go
@@ -16,6 +16,7 @@ package rpc
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/rs/zerolog/log"
@@ -25,8 +26,17 @@ import (
 
 // AuthInterceptor is a client interceptor for authentication.
 type AuthInterceptor struct {
-	authClient  *AuthClient
+	authClient *AuthClient
+
+	// mu guards accessToken, which is read by every outgoing RPC (attachToken)
+	// and written by both the background refresh goroutine and on-demand
+	// RefreshToken calls.
+	mu          sync.RWMutex
 	accessToken string
+
+	// refreshMu serializes re-authentication so that a burst of RPCs failing
+	// with an expired token does not trigger a stampede of Auth calls.
+	refreshMu sync.Mutex
 }
 
 // NewAuthInterceptor returns a new auth interceptor.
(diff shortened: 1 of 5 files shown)


agent/rpc/auth_interceptor.go
```diff
@@ -25,8 +26,17 @@ import (
...
+
+	// mu guards accessToken, which is read by every outgoing RPC (attachToken)
+	// and written by both the background refresh goroutine and on-demand
+	// RefreshToken calls.
+	mu          sync.RWMutex
 	accessToken string
+
+	// refreshMu serializes re-authentication so that a burst of RPCs failing
+	// with an expired token does not trigger a stampede of Auth calls.
+	refreshMu sync.Mutex
```
6543, line 39: why have two mutex, can we just have accessTokenLock ?

6543:
... so the agent's RPC loops (next(), report_health(), workflow init/wait, …) kept retrying with the dead token and never recovered ...

in case the agent got deleted server side, the agent should not retry but exits with error ... we should handle that case propperly too!!

PR #6735, opened 2026-06-15
feat(backend/docker): add per-step resource limits and OOM score adjustment

bclermont:
## Summary

Add support for per-step memory, CPU limits and OOM score adjustment in the Docker backend via `backend_options` in workflow YAML.

## Motivation

The Docker backend currently only supports global resource limits via agent environment variables (`WOODPECKER_BACKEND_DOCKER_LIMIT_MEM`, etc.). There is no way to cap individual heavy steps without affecting all steps globally. This leads to OOM situations where the host server freezes because no step-level ceiling exists.

## Changes

- Extended `BackendOptions` struct with `resources.limits.{memory,cpus}` and `oom_score_adj`
- `toHostConfig` now accepts `BackendOptions` and merges per-step limits over global agent limits
- Added `parseMemory` helper supporting `k/m/g` suffixes
- `oom_score_adj` restricted to **[0, 1000]** by design — steps can only be made more killable, never protected from the OOM killer
- JSON schema updated so the linter validates `backend_options.docker`
- Documentation updated with YAML example and option reference table
- All existing tests pass; new tests added for validation, parsing, and `toHostConfig` merge logic

## Usage

```yaml
steps:
  heavy-build:
    image: golang:1.22
    commands:
      - go build ./...
    backend_options:
      docker:
        resources:
          limits:
            memory: 1g
            cpus: 1.5
        oom_score_adj: 500
```

| Option | Type | Description |
|--------|------|-------------|
| `resources.limits.memory` | string | Max memory (e.g. `512m`, `1g`). Overrides `WOODPECKER_BACKEND_DOCKER_LIMIT_MEM`. |
| `resources.limits.cpus` | float | Max CPU cores (e.g. `0.5`, `1.5`). Converted to NanoCPUs. |
| `oom_score_adj` | int (0–1000) | OOM killer priority. `0` = default, `1000` = kill first. Negative values rejected by design. |


pipeline/backend/docker/convert.go
```diff
@@ -74,16 +74,34 @@ func toContainerName(step *types.Step) string {
...
-func toHostConfig(step *types.Step, conf *config) (*container.HostConfig, error) {
+func toHostConfig(step *types.Step, conf *config, options BackendOptions) (*container.HostConfig, error) {
+	// Memory: per-step overrides global
+	memLimit := conf.resourceLimit.MemLimit
+	if options.Resources.Limits.Memory != "" {
+		parsed, err := parseMemory(options.Resources.Limits.Memory)
+		if err != nil {
+			return nil, err
+		}
+		memLimit = parsed
```
6543, line 85: we need checks, golbal conf should be able to be limited but not widen

this allows to overwrite  conf.resourceLimit.MemLimit to have a greater limit!

bclermont: @6543  should I just change it to never go beyond `WOODPECKER_BACKEND_DOCKER_LIMIT_MEM` ?

6543: yes

PR #6737, opened 2026-06-16
Security updates for js dependencies

qwerty287:


xoxys (approved):

6543 (changes requested):
also tried it - does not work, docusaurus needs to release a new version.

this is a majour bump and breaks api. upstream main has it fixed but not released

xoxys:
Sorry I somehow expected that such PRs are at leased validated to be working already.

qwerty287:
The CI failed anyways, so there was no risk in merging a broken patch

6543:
does pnpm allow to use a upstream commit as version ? then we could upgrade to latest docusaurus commit and things be gone ...

6543:
now docusaurus is not the problem anymore ... but `gray-matter`

i created a upstream issue: https://github.com/facebook/docusaurus/issues/12186

and also pinged the maintainer: https://github.com/jonschlinkert/gray-matter/pull/172#issuecomment-4750761610

second one i do not expect much at this point

6543:
and we got new issues: https://ci.woodpecker-ci.org/repos/3780/pipeline/35250/17

6543 (approved):
we fix the rest and yaml will be once one of the upstream makes a move we do that

qwerty287:
@6543 could remove the overrides for dompurify again? If it also works without in https://github.com/woodpecker-ci/woodpecker/pull/6742

6543:
well this bump would be implzite and only for web :/

next bump this could be changed to insecure again, so i prevere this way

qwerty287:
For web you can put `^3.14.11` here: https://github.com/woodpecker-ci/woodpecker/blob/main/web/package.json#L26

And for docs it's ok to keep for me, but I really would like to try to avoid overrides as far as possible. We never remove them again.

6543:
I try to remove them each time i manually do such fixes ... but atm all overrides are nessesary

6543:
aww just was going to push the same :laughing:

6543 (approved):

6543:
@qwerty287 does it has your lgtm too?

qwerty287:
Yes fine for me now :+1:

PR #6738, opened 2026-06-18
Add Prometheus metrics for pipeline step duration and failures

### LINKED ISSUE #5182: Support metrics with step-level labels to enable fine-grained Prometheus alerting

As a user of Woodpecker, I want to be able to monitor and create alerts based on the error rate and execution frequency of specific steps within my pipelines, so that I can ensure reliability in key stages such as deploy or test.

Currently, the available Prometheus metrics (e.g., woodpecker_pipeline_count, woodpecker_pending_steps, etc.) do not expose enough granularity to identify the performance or failures of individual steps. I’m particularly looking for the ability to filter by step name or label, which is essential to trigger alerts using Prometheus Alertmanager based on issues in critical steps.

We are running Woodpecker version 2.7 and facing limitations when trying to build such alerting due to this lack of visibility.
This feature would unlock more precise observability and alerting for teams operating at scale.

### Suggested solution

In the web-ui or in the metrics endpoint (used by Prometheus), expose new metrics (or expand existing ones) that include a label for the step name.

Example metric proposal:

```
woodpecker_step_duration_seconds{pipeline="xyz", repo="myrepo", step="deploy"} 12.3
woodpecker_step_failures_total{pipeline="xyz", step="test"} 1
```

### Alternative

_No response_

### Additional context

I understand this could raise cardinality concerns, especially in larger installations, so maybe the feature could be optional and configurable via environment variable or config file to enable step-level metrics selectively.

### Validations

M31ancholy:
## Changes 
 - Add `WOODPECKER_STEP_LEVEL_METRICS` server option.
  - Move server metrics collection into the `server/metric`.
  - Export step-level Prometheus metrics when enabled:
    - `woodpecker_step_failures_total{pipeline,repo,step}`
    - `woodpecker_step_duration_seconds{pipeline,repo,step}`
  - Record failed step counts after step status updates.
  - Record step duration after a step exits.
  - Document the new metrics and configuration option in the server configuration docs.
Closes #5182


cmd/server/flags.go
```diff
@@ -129,6 +129,12 @@ var flags = append([]cli.Flag{
 		Usage:   "metrics server address",
 		Value:   "",
 	},
+	&cli.BoolFlag{
```
6543, line 132: i dont see why we should put this behind a flag?

M31ancholy: According to #5182, the reporter explicitly called out the potential cardinality impact of adding step-level metrics. I think this is a valid concern, especially for larger installations, since the combination of `repo`, `pipeline`, and `step` labels could lead to a significant increase in Prometheus time series.

6543: well I would have it turned on by default i this case ...

cmd/server/flags.go
```diff
@@ -129,6 +129,12 @@ var flags = append([]cli.Flag{
...
+		Sources: cli.EnvVars("WOODPECKER_STEP_LEVEL_METRICS"),
+		Name:    "step-level-metrics",
+		Usage:   "enable step-level metrics",
+		Value:   false,
```
6543:
```suggestion
		Value:   true,
```

docs/docs/30-administration/10-configuration/10-server.md
```diff
@@ -658,6 +665,15 @@ Example: `:9001`
 
 ---
 
+### STEP_LEVEL_METRICS
+
+- Name: `WOODPECKER_STEP_LEVEL_METRICS`
+- Default: `false`
```
6543:
```suggestion
- Default: `true`
```

docs/docs/30-administration/10-configuration/10-server.md
```diff
@@ -400,8 +400,16 @@ woodpecker_waiting_steps 0
...
 woodpecker_worker_count 4
+# HELP woodpecker_step_failures_total Total number of pipeline step failures.
+# TYPE woodpecker_step_failures_total counter
+woodpecker_step_failures_total{pipeline="1",repo="woodpecker-ci/woodpecker",step="deploy"} 1
+# HELP woodpecker_step_duration_seconds Step duration in seconds.
+# TYPE woodpecker_step_duration_seconds gauge
+woodpecker_step_duration_seconds{pipeline="1",repo="woodpecker-ci/woodpecker",step="deploy"} 12
 ```
 
+Step-level metrics are only exported when `WOODPECKER_STEP_LEVEL_METRICS` is enabled.
```
6543:
```suggestion
Step-level metrics are exported as long as `WOODPECKER_STEP_LEVEL_METRICS` is not disabled.
```

6543 (approved):

Kernald:
I'm not too familiar with the Woodpecker codebase nor metrics, but from a quick glance, there seems to be a fundamental (albeit fairly easy to fix) issue with this PR: the pipeline _id_ (monotonously increasing) is used as a label, as far as I can tell. This means that the cardinality of labels is going to explode very quickly, even with a small set of repos/pipelines on these repos. It also makes it harder to track metrics from a given pipeline over time. The easy fix there would be to switch to the pipeline _name_ instead - which should be fairly trivial.

Another issue, from what I understand, is that `woodpecker_step_duration_seconds` is using a `gauge` over a `histogram`. This means basing the sampling of that data on the scraping frequency, losing anything in between. A histogram would abstract this, bucketing the values and giving the proper distribution at every scrape, even if a step ran x times between two scrapes, or didn't run in a while.

I'm happy to open a PR to switch to the pipeline name, but I'm not familiar with Go nor the Woodpecker codebase at all, switching to a histogram might be a bit out of my competence here - but it would be nice to have this fixed before 3.16.0 gets released to avoid metrics changing shape on the next release.

M31ancholy:
> I'm not too familiar with the Woodpecker codebase nor metrics, but from a quick glance, there seems to be a fundamental (albeit fairly easy to fix) issue with this PR: the pipeline _id_ (monotonously increasing) is used as a label, as far as I can tell. This means that the cardinality of labels is going to explode very quickly, even with a small set of repos/pipelines on these repos. It also makes it harder to track metrics from a given pipeline over time. The easy fix there would be to switch to the pipeline _name_ instead - which should be fairly trivial.
  > 
  > Another issue, from what I understand, is that `woodpecker_step_duration_seconds` is using a `gauge` over a `histogram`. This means basing the sampling of that data on the scraping frequency, losing anything in between. A histogram would abstract this, bucketing the values and giving the proper distribution at every scrape, even if a step ran x times between two scrapes, or didn't run in a while.
  > 
  > I'm happy to open a PR to switch to the pipeline name, but I'm not familiar with Go nor the Woodpecker codebase at all, switching to a histogram might be a bit out of my competence here - but it would be nice to have this fixed before 3.16.0 gets released to avoid metrics changing shape on the next release.
  Thanks for the detailed review!

On the pipeline name:
You're right — using the pipeline ID as a label will cause cardinality explosion. I initially implemented it based on the issue's suggestion (with the toggle option), but switching to the workflow name（possibly the pipelineName to you because it seems there is no pipeline name） is definitely a better approach. I'll open a PR in a few days to make that change on `woodpecker_step_failures_total`, and I'd appreciate it if you could take another look then.

On gauge vs histogram:
I'll be honest — my original thinking was that a gauge can record the exact value, unlike a histogram which only buckets. I didn't properly consider the scraping issue you mentioned, where data between scrapes gets lost. That's my oversight, sorry about that.

However, now that we're switching to the workflowName, I agree that a histogram for metric `woodpecker_step_duration_seconds` is the better fit. The question is how to handle the bucket division — should we use a fixed set of default buckets for everyone, or allow users to customize the boundaries through configuration? I'd love to hear your thoughts on this.

Kernald:
> but switching to the workflow name (possibly the pipelineName to you because it seems there is no pipeline name)

Sorry, I tend to conflate workflow/pipeline name - yes, workflow name is probably what we'd want to label against.

Regarding the histogram - I don't think I've ever seen a software allowing me to define my own buckets, but that's likely anecdotal. I wonder if there's any public data we could leverage to try and pinpoint a set of bucket sizes that makes sense?

M31ancholy:
> Kernald

When you have a moment, could you take a look at this #6777 ? I've hard-coded the buckets for now. If someone raises an issue about it, we can revisit the approach then.

PR #6744, opened 2026-06-19
docs: deps migrate gray-matter to maintained fork

6543:
do what upstream unreleased https://github.com/facebook/docusaurus/pull/12181 does ...

... as per answer to https://github.com/jonschlinkert/gray-matter/pull/179#issuecomment-4751897979

### DIFF
--- a/docs/pnpm-lock.yaml
+++ b/docs/pnpm-lock.yaml
@@ -11,6 +11,7 @@ overrides:
   dompurify: ^3.4.11
   http-proxy-middleware: ^3.0.6
   undici: ^7.28.0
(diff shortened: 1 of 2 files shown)


6543:
well npm dep hell is just a mess - we wont fully get rid of overloads until the ecosystem changes ... :/

qwerty287 (approved):
> well npm dep hell is just a mess - we wont fully get rid of overloads until the ecosystem changes ... :/

Unfortunately yes. We could actually migrate our docs to a non-js based system like hugo but I guess that would be really much work

6543:
Well as its js is just used to generate static content, no real harm realy can be done once generated ... so it does not matter if JS or Go does the job :)

PR #6753, opened 2026-06-20
Refactor server scheduler collect in new interface

6543:
next step in https://github.com/woodpecker-ci/woodpecker/issues/6317


qwerty287 (approved):
untested

6543:
I'd like to resolve the TODO first :)

6543:
awww :swan: 

the to resolve the TODO, we need to move a lot more code around than i thought - it has to be done but we should do it in it's seperate pull - I finalize this one, so it stays by it's scope

6543:
=> pull is big enouth

qwerty287 (approved):
still seems fine, still untested at my side

PR #6754, opened 2026-06-20
Refactor use store in server scheduler

6543:
before queue had store inserted, now scheduler has it, so we can make more persistent things with it.

this does:
 - [x] mark skipped workfows as such without rpc layer
 - [ ] make code more readable
 - [ ] make code arch more stright forward
 - [ ] clear way for scheduler to back propagate results ... e.g. to forge via states
   - [ ] look into if we can use pub/sub there too :thinking: ... (mulit instance race risistant?)
 - [ ] scheduler should not import rpc
 - [ ] ...

----
followup of https://github.com/woodpecker-ci/woodpecker/pull/6753
step of https://github.com/woodpecker-ci/woodpecker/issues/6317


6543:
I want the scheduler to be a pure state maschine that gets signals from extern e.g. api / agent-rpc and forwards them to forge / agent-rpc ... this is by no means the full refactor, it's a way in the right direction but i need more time to refine it.

as for server-instance-cluster-wide locking stuff, a good candidate is atomic DB swaps etc.. and for signal distribution the pubsub/queue stuff ... it needs more thinkig ...

PR #6759, opened 2026-06-21
Harden agent rpc to check agentID on workflow

6543:
We now assign on Next, right when the agent that got the workflow, the agent ID to the workflow.
Later each call checks that this is still the same.

PS: shout-out to `Yaohui Wang` for finding this edgecase (it is not worth a CVE etc... as the security boundary's stay the same nontheless)

### DIFF
--- a/server/rpc/errors.go
+++ b/server/rpc/errors.go
@@ -22,6 +22,6 @@ var (
 
 	ErrAgentIllegalLogStreaming = errors.New("agent can not append logs to a step that is marked not running")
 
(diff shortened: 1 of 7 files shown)


qwerty287 (approved):

qwerty287:
lint and test failures (I guess mocks have to be regenerated)

6543:
we now need to link step to an workflow too ... there was no store api for it jet ... so now we have that too

6543:
thx e2e test -> catched skip case :sweat_smile:

PR #6771, opened 2026-06-23
Kubernetes: allow custom image when precreating workingDir as nonroot

nati737:
### Problem
When the working directory is set to a directory that doesn't exist (for example, as `plugin-git` does), kubelet will pre-create it with ownership set to `root:root` and permissions `0755`. This makes pods running as non-root unable to write to it, causing permission errors.

### Context
This PR extends [Kubernetes: precreate workingDir as nonroot when required (#6322)](https://github.com/woodpecker-ci/woodpecker/pull/6322#issue-4131798381), which introduced an init container to pre-create and fix permissions on the working directory when needed.
However, the original implementation hardcodes the init container image to `busybox:stable-musl`, which is not suitable for all environments

### Solution
Added support for a configurable init container image used for working directory preparation.
Key changes:
- Introduces configuration option for init container image, using the `WOODPECKER_BACKEND_K8S_PERMISSION_INIT_IMAGE` environment variable.
- Keeps `busybox:stable-musl` as the default.
- Allows overriding the image for on-prem/restricted environments.

### Behavior
Unchanged from the previous implementation, except that the init container image is now configurable via backend configuration.

### Related issues and PRs
Extends [Kubernetes workingDir permission fix (#6322)
](https://github.com/woodpecker-ci/woodpecker/pull/6322#issue-4131798381)


6543 (approved):

6543:
@nati737 hey could you address the CI issues
-> adjust tests
-> fix formating

6543:
I did fix it for you ...

PR #6777, opened 2026-06-24
fix(metrics): reduce step metric cardinality

M31ancholy:
## Summary

- Change `woodpecker_step_duration_seconds` from a gauge to a histogram so step durations are recorded as observations instead of only keeping the latest value.
- Use workflow names instead of pipeline IDs for step metric labels, aligning `step_failures_total` and `step_duration_seconds` with workflow-level metric semantics.
- Record zero-length step durations by allowing `Finished >= Started`.

## Breaking Change

- `woodpecker_step_duration_seconds` is now a Prometheus histogram instead of a gauge.
- Step metric label `pipeline` has been replaced by `workflow` for:
  - `woodpecker_step_failures_total`
  - `woodpecker_step_duration_seconds`

Fixes concerns raised in [#6738 ](https://github.com/woodpecker-ci/woodpecker/pull/6738#issuecomment-4785757978)

### DIFF
--- a/docs/docs/30-administration/10-configuration/10-server.md
+++ b/docs/docs/30-administration/10-configuration/10-server.md
@@ -402,10 +402,21 @@ woodpecker_waiting_steps 0
 woodpecker_worker_count 4
 # HELP woodpecker_step_failures_total Total number of pipeline step failures.
 # TYPE woodpecker_step_failures_total counter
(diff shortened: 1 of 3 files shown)


Kernald:
Looks great, thanks for the quick action!

lafriks (approved):

PR #6779, opened 2026-06-24
Rework pipeline data structure: Commit

6543:
-> start split out from https://github.com/woodpecker-ci/woodpecker/pull/4626

## split out with claude, so rougth review needed ... will only make it ready once i did so myselfe

---

The pipeline model carried the commit identity as four flat columns (commit, message, timestamp, email). Group them into a reusable Commit substruct (sha, message, forge_url, author{name,email}, timestamp), reusing the existing model.Commit type already returned by BranchHead.

All forge producers (gitea, forgejo, github, gitlab, bitbucket, bitbucket datacenter), the cron and manual pipeline builders, and every consumer (file/dir fetch, commit-status, skip-ci, metadata, feed) are updated to read and write the nested struct. Commit reads are nil-safe.

The HTTP API stays backward compatible: APIPipeline and APIFeed still emit the deprecated commit, message, timestamp and author_email fields, populated from the substruct in ToAPIModel, so existing clients and the web UI keep working. The flat message, timestamp and email columns are dropped and the commit column is converted to JSON by migration 028.

This is a self-contained slice carved out of the larger pipeline restructure (#4626) to ease review and upstreaming.


AhmadNajiKam:
---
## **issue (blocking):** 

**Context:** `server/store/datastore/migration/028_add_commit_substruct.go`

**Problem:** Offset pagination loop has no stable sort order, so pages can skip or duplicate rows mid-migration:

The page loop computes its offset purely from the loop counter:

```go
sess.Limit(perPage, page*perPage).Find(&oldPipelines)
```

There's no `OrderBy` here, so the database isn't guaranteed to return a stable, disjoint slice of rows for each `page` value  especially while we're concurrently updating rows on the same table inside the loop. On a quiet table this will often happen to work, which is exactly what makes it risky: it can silently skip rows during a real migration run on a large `pipelines` table and nobody would necessarily notice until a row turns up missing its new `commit` JSON column after the old columns are dropped.

Suggest either:
- adding an explicit `OrderBy("id ASC")` to make the offset deterministic, or
- switching to keyset pagination (`WHERE id > lastID ORDER BY id ASC LIMIT perPage`, tracking `lastID` from the last row of the previous page)  this also avoids relying on the table being a stable snapshot during the scan, since none of the updates here touch `id`.

Given this runs as a one-way data migration that's immediately followed by `dropTableColumns` on the source columns, any rows missed here lose their commit message/timestamp/email permanently with no path to recovery.

---

AhmadNajiKam:
###  issue (blocking)

**Context:** `server/forge/bitbucketdatacenter/bitbucketdatacenter.go`, `updatePipelineFromCommits`

**Problem:** The nil-guard for `p.Commit` is added in the wrong place. It appears *after* the first dereference, not before it:

```go
commit, _, err := bc.Projects.GetCommit(ctx, r.Owner, r.Name, p.Commit.SHA)
if err != nil {
    return nil, fmt.Errorf("unable to read commit: %w", err)
}

// guard added here three lines too late
if p.Commit == nil {
    p.Commit = &model.Commit{}
}
if p.Event == model.EventTag && commit.ID != "" && commit.ID != p.Commit.SHA {
    p.Commit.SHA = commit.ID
    ...
}
```

`p.Commit.SHA` is read on the very first line, before the guard runs. If this function is ever entered with `p.Commit == nil`, it panics immediately on that line the defensive code added three lines later does nothing to prevent it. The fact that a guard was added at all signals the author knew this needed defending; it just landed after the read it was meant to protect, not before it.

Today every Bitbucket Data Center constructor (`convertRepositoryPushEvent`, `convertPullRequestEvent`) happens to populate `Commit` before this runs, so it's not reachable through the current call sites but that's incidental, not enforced, and the very next caller that doesn't guarantee a populated `Commit` turns this into a server crash.

**Suggested fix:** Move the guard above the `GetCommit` call:

```go
if p.Commit == nil {
    p.Commit = &model.Commit{}
}

commit, _, err := bc.Projects.GetCommit(ctx, r.Owner, r.Name, p.Commit.SHA)
if err != nil {
    return nil, fmt.Errorf("unable to read commit: %w", err)
}

if p.Event == model.EventTag && commit.ID != "" && commit.ID != p.Commit.SHA {
    p.Commit.SHA = commit.ID
    ...
}
p.Commit.Message = commit.Message
```

This keeps the existing nil-safe pattern already used correctly elsewhere in this same PR (`gitlab.go`'s `File`/`Dir`/`loadCommitFromSHA` all check `pipeline.Commit == nil` before the first read) this is just one call site where the check landed after the dereference instead of before it.

6543:
well the

> Problem: Offset pagination loop has no stable sort order, so pages can skip or duplicate rows mid-migration

could be a problem ... it should not be, but it wont hurt to be explizite ... -> added it to https://github.com/woodpecker-ci/woodpecker/pull/6796

@AhmadNajiKam keep in mind this pull here is a draft because i used claude to port over just the commit changes ... but we should not trust claude to do it's job correctly! ... so before this is not marked ready i did not have looked into it, so expect bad things to be in this (so not sure if it is worth you effort to review :sweat_smile: )

server/api/pipeline.go
```diff
@@ -90,19 +90,25 @@ func CreatePipeline(c *gin.Context) {
...
+		Event: event,
+		Commit: &model.Commit{
+			SHA:      commit.SHA,
+			Message:  "MANUAL PIPELINE @ " + opts.Branch,
+			ForgeURL: commit.ForgeURL,
+			Author: model.CommitAuthor{
+				Name:  user.Login,
+				Email: user.Email,
+			},
+			Timestamp: time.Now().UTC().Unix(),
```
6543, line 102: we must use commit timestamp

server/cron/cron.go
```diff
@@ -144,11 +144,14 @@ func CreatePipeline(ctx context.Context, store store.Store, cron *model.Cron) (*
 	}
 
 	return repo, &model.Pipeline{
-		Event:               model.EventCron,
-		Commit:              commit.SHA,
+		Event: model.EventCron,
+		Commit: &model.Commit{
+			SHA:       commit.SHA,
+			ForgeURL:  commit.ForgeURL,
+			Timestamp: cron.NextExec,
```
6543, line 151: same

PR #6781, opened 2026-06-25
Server can move pipeline parsing into background on long creation times

6543:
the [fork](https://github.com/quintoandar/woodpecker) had an patch that always made parsing into background,

i can see the idea but for normal cases you want the result to be reported back, so we make a timeout that will decide if sync or async happens and make it configurabel.

esp. interesting if you use config fetching extensions ....


server/api/hook.go
```diff
@@ -203,12 +211,55 @@ func PostHook(c *gin.Context) {
...
+	// avoid the forge timing out and retrying the webhook delivery, we wait only
+	// up to WebhookSyncTimeout for creation to finish. If it completes in time we
+	// respond synchronously with the created pipeline (preserving the old
+	// behavior and API response). If it takes longer we respond 202 Accepted and
+	// let creation finish in the background.
+	//
+	// The background goroutine must not use the gin request context: gin cancels
+	// it once the handler returns. We detach from cancellation (keeping request
+	// values) and apply our own hard cap instead.
+	bgCtx, cancel := context.WithTimeout(context.WithoutCancel(c.Request.Context()), backgroundPipelineCreationTimeout)
```
6543:
```suggestion
	bgCtx, cancel := context.WithTimeout(context.WithoutCancel(context.Background()), backgroundPipelineCreationTimeout)
```

lafriks (approved):

PR #6784, opened 2026-06-25
Allow images for secrets to contain SHA256 for image pinning

### LINKED ISSUE #6783: Pinned image versions with SHA256 sum not accepted by secret plugin restriction
### Component

web-ui, server

### Describe the bug

Saving secret does not work when restricting its access to plugins of format `image@sha256:abcde...12345`.

In principle, the image format is supported by woodpecker CI, although undocumented. E.g., modifying the SHA256 will not execute the step. I guess that it only needs allow this format also in the secret saving.

### Steps to reproduce

Having a workflow with
```yml
    image: codeberg.org/gedankenstuecke/git-pages-woodpecker-plugin@sha256:76091e8189842f8d620fb5d1ab048ed449fe3516521ef0acc7f8c5a608c01e59
```
and wanting to use it with a secret, but restricting only to this version, fails with a red popup message on saving the secret of
```
: Error inserting secret. invalid secret image: image
'codeberg.org/gedankenstuecke/git-pages-woodpecker-plugin@sha256:76091e8189842f8d620fb5d1ab048ed449fe3516521ef0acc7f8c5a608c01e59'
do not match regexp
'^([\w\d\-_\.]+(:\d+)?/)?([\w\d\-_\.][\w\d\-_\.\/]*/)?([\w\d\-_\.]+)(:[\w\d\-_]+)?$
```

(I didn’t try if using the API is affected, too)

### Expected behavior

Save the secret with the restriction.

### System Info

```shell
Using Codeberg’s instance, version 3.15.0
```

### Additional context

_No response_

### Validations

mahlzahn:
Closes #6783

### DIFF
--- a/server/model/secret.go
+++ b/server/model/secret.go
@@ -88,6 +88,7 @@ var validDockerImageString = regexp.MustCompile(
 		`([\w\d\-_\.][\w\d\-_\.\/]*/)?` + // optional url prefix
 		`([\w\d\-_\.]+)` + // image name
 		`(:[\w\d\-_]+)?` + // optional image tag
+		`(@sha256:[a-f\d]{64})?` + // optional sha256 digest
 		`$`,
 )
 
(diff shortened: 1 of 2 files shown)


6543 (approved):

6543:
thanks!

PR #3811, opened 2024-06-21
Support github refresh tokens

anbraten:

Issue #3804, opened 2024-06-19
Unauthorized user tokens preventing builds from running when protected by SAML enforcement


fernandrone:
### Component

agent

### Describe the bug

This is a very intermittent bug that has been hard to track down. We run Woodpecker on GitHub in an [organization protected by organization SAML enforcement](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-saml-single-sign-on-for-your-organization/enforcing-saml-single-sign-on-for-your-organization#about-enforcement-of-saml-sso-for-your-organization).

This has happened twice, once in December 2023, then on the 29th of April. When it does, most of our builds will start to fail with the following error:

```
{"level":"error","error":"GET https://api.github.com/repos/quintoandar/<redacted>/contents/.woodpecker?ref=ced8fe72e731630f9888d10f4aad083071f0b83d:  403 Resource protected by organization SAML enforcement. You must grant your OAuth token access to this organization. []","repo":"quintoandar/<redacted>","user":"<redacted>","time":"2024-04-29T18:17:15Z","message":"could not get folder from forge"}
```

This essentially means that the user that activated the repository (the user Woodpecker is making requests on behalf on to clone the repository and retrieve the woodpecker file) is authenticated but their token has not been granted access to the organization.

We've found out that when this happens, the "fix" was _apparently_ to logout and then login again with the indicated user. After that the builds would work again, presumably because Woodpecker would have generated a new, valid token.

To avoid dealing with issues of users leaving the organization we tend to automate enabling of Woodpecker repositories so they all authenticate on behalf of the same user. Initially we were using a "bot" GitHub account dedicated for this. However, after this bug, we've moved to using an administrador's account, since it's easier to logout/login of said account to fix the problem. Either way the point being that most repositories will be making their requests on behalf of the same user, which avoids situations of "random user has left the organization, breaking random repository number 42" _but_ at the same time makes this issue way more destructive, as when it happens our whole Woodpecker instances stops working.

This seems very similar to this issue https://github.com/woodpecker-ci/woodpecker/discussions/2482 except that:

1. We have already configured Woodpecker as an Oauth App, which should not have expiring tokens.

2. The issue blocking access is always specifically about SAML enforcement; the token _seems_ to be valid as in authenticated, just not authorized to access the organization.

[...]

Issue #4052, opened 2024-08-21
Failure to handle multiple workflows (secondary rate limits and timeouts)


fernandrone:
### Component

server

### Describe the bug

We've been experience two issues with repositories that have a "large" number of workflows: I have a repository with 41 workflows and another with 61. They're certainly outliers... but they exist.

1. [Secondary rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits). They will show up as this on the server:

```
{"level":"error","error":"POST https://api.github.com/repos/<org>/<repo>/statuses/<hash>: 403 You have exceeded a secondary rate limit. Please wait a few minutes before you try again. If you reach out to GitHub Support for help, please include the request ID <id>.","time":"2024-08-21T13:31:15Z","message":"error setting commit status for <org>/<repo>/11082"}
```

If this is resolved in a timely manner (in less than 10 seconds) then the server replies with a 400 and a "failure to parse hook" message. We have observed that this is a relatively common occurrence, repositories with 20 or so workflows will receive "a few" of those daily.

2. A more general issue is timeouts. [GitHub has a 10 second timeout on webhooks](https://docs.github.com/en/webhooks/testing-and-troubleshooting-webhooks/troubleshooting-webhooks#timed-out). If the server takes too long to request all the workflows (wether or not it is rate limited) and doesn't reply within 10 second then the webhook will show as a "504".

From what I've seem both issues are often correlated. A secondary rate limit will likely cause the webhook to timeout. That said _it seems_ that a timeout alone is possible and itself not destructive. If a webhook timeout happens but the Woodpecker server does manage to parse all workflows (even if it takes more than 10 seconds) and process them, the job is created, picked up by an agent and even updates the github UI.

However when secondary rate limits happen everything breaks. The server code doesn't handle rate limits so it just errors out and the pipeline job is not created. From the user's perspective this is a silent error. They make a push but no pipeline is started, so it can be very hard to debug without access to the server logs.

FWIW I'm testing this internally https://github.com/quintoandar/woodpecker/pull/20 which uses a github secondary rate limit library to try and fix the first issue. Looks promising.

[...]

Issue #2851, opened 2023-11-20
[refactor] dont store secrets in task queue


6543:
currently the whole workflow config for the backend is stored in the queue.

**This includes the secrets!**

We should inject them at the server if an agent do pull that workflow out of the queue instead. So we dont have to worry about leaking via e.g. redis.

6543:
first we have to move the logic that we store the complete backend config of a workflow in queue ... :/

6543:
NOTE: also with #3723 we might consider to do so for them too ...

6543:
NOTE: makes WP also more secure as we dont have to trust the queue to handle secrets well ... e.g. leak it in logs/api etc

PR #6806, opened 2026-07-02
Support GitHub App installation tokens for the GitHub forge

slayerjain:
This adds first-class GitHub App support to the GitHub forge: when `WOODPECKER_GITHUB_APP_ID` and `WOODPECKER_GITHUB_APP_PRIVATE_KEY`(`_FILE`) are configured, Woodpecker authenticates repo-scoped server-side operations and clone credentials with **installation access tokens** instead of the repo owner's personal OAuth token. User login stays on the OAuth flow, and repositories the app is not installed on transparently fall back to the current user-token behavior.

## Why

Today every server-side API call (commit statuses, pipeline config fetching, changed-file lookups) and every clone runs on the OAuth token of the user who enabled the repository. For larger organizations this has three structural problems:

- all repos of an org share one user's 5,000 req/h rate limit
- pipelines break when that user leaves the org or loses access (see the SAML variant in #3804, and #2482/#2488/#4079)
- commit statuses appear under a personal account instead of a bot identity

Installation tokens get per-installation rate limits that scale with repository count, survive user departure, and post statuses as the app's bot account. This also allows removing the long-stale docs warning against GitHub Apps (added in #2499, obsolete since #3811 introduced refresh-token support).

## What changed

- `server/forge/github/app.go`: authenticates as the app (RS256 JWT via the already-vendored `golang-jwt/jwt/v5`; client id or app id as issuer, backdated `iat` per GitHub's guidance) and mints installation access tokens through the existing `go-github` dependency — no new module needed. Installation lookups and tokens are cached with validity margins (5 min for API calls, 45 min for clone credentials); "not installed" results are negative-cached and uninstall/suspension invalidates the cache and falls back gracefully.
- Token selection wired into `Netrc` (`x-access-token`), `Status`, `File`/`Dir`, `Branches`, `BranchHead`, `PullRequests` and the webhook changed-files/tag loaders. The loaders skip the repo-owner DB lookup entirely when the app covers the repo. Tokens are installation-wide on purpose so same-org submodule clones keep working.
- Config plumbing: new env vars → `AdditionalOptions["app-id"/"app-private-key"]` (same channel as the bitbucket-dc machine account) → `github.Opts`. No DB migration needed. Private keys are accepted as PEM or base64-encoded PEM, including keys whose newlines were mangled by env files or input fields.
- The private key is **write-only** in the API/UI (like `oauth_client_secret`): responses replace it with an `app-private-key-set` marker, and updates that omit it keep the stored key. Clearing the app id disables app auth again.
- Validation: invalid app credentials are rejected at save time in the forge API (HTTP 400) and fail fast at boot for env-managed forges, instead of breaking every forge operation at runtime.
- New admin endpoint `GET /forges/{id}/app-health` plus a "Test GitHub App" button in the admin forge form so operators can verify the credentials actually work.
- Admin UI fields (app id + multi-line private key), docs rewrite of the GitHub App section (required permissions, webhook note, token-expiry caveat), fixture-based unit tests for the token flow (JWT signature verification, caching, fallback, PEM variants) and API tests for redaction/merge/health.

## Notes

- Webhook handling is unchanged: Woodpecker keeps creating per-repo webhooks, the app's own webhook stays disabled. Org-level auto-enrollment via app installation events (#71) would be a natural follow-up but is out of scope here.
- A token cached before an app is suspended/uninstalled can be served until its validity margin passes (bounded by the 10-minute forge cache); this matches the behavior of `ghinstallation`-based tools (Atlantis, ArgoCD).


slayerjain:
This seems to be working fine in Keploy on-prem CI. Please let me know if any changes needed before merge.

docs/docs/30-administration/10-configuration/12-forges/20-github.md
```diff
@@ -32,6 +28,41 @@ Do not use a "GitHub App" instead of an Oauth2 app as the former will not work c
 After your App has been created, you can generate a client secret.
 Use this one for the `WOODPECKER_GITHUB_SECRET` environment variable.
 
+## GitHub App
+
+In addition to the OAuth login, Woodpecker can authenticate as a [GitHub App](https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/about-creating-github-apps) using installation access tokens for server-side API calls (commit statuses, pipeline configuration fetching, changed-file lookups) and for cloning repositories.
```
6543, line 33: ... for cloning repos

this could have security implications and need to be looked at

docs/docs/30-administration/10-configuration/12-forges/20-github.md
```diff
@@ -32,6 +28,41 @@ Do not use a "GitHub App" instead of an Oauth2 app as the former will not work c
...
+```
+
+For repositories the app is not installed on, Woodpecker falls back to the OAuth token of the user who enabled the repository.
+
+The private key is write-only: the API and the admin UI never return it, and leaving the field empty when updating a forge keeps the stored key.
+After saving, the "Test GitHub App" button in the admin UI (or `GET /api/forges/{id}/app-health`) verifies that the credentials work.
+
+:::note
+Installation access tokens expire after one hour and are scoped to the repositories of the app installation.
+Clone credentials of pipelines that stay queued for a very long time may therefore expire before the clone step runs, and clone access to private repositories outside the installation (e.g. git submodules in another organization) requires falling back to user OAuth tokens by not installing the app on the affected repositories.
```
6543: we should ention https://github.com/woodpecker-ci/woodpecker/issues/2851 as this is an area that will be resolved long term

server/api/forge.go
```diff
@@ -15,16 +15,33 @@
...
 	"go.woodpecker-ci.org/woodpecker/v3/server/model"
 	"go.woodpecker-ci.org/woodpecker/v3/server/router/middleware/session"
 	"go.woodpecker-ci.org/woodpecker/v3/server/store"
 )
 
+// validateForge ensures the forge configuration can actually be constructed,
+// so invalid configurations (e.g. an unparsable GitHub App private key or a
+// private key without an app id) are rejected at save time instead of
+// breaking every operation of the forge afterwards. Only GitHub forges are
+// validated for now: their constructor is side-effect free and rejects
```
6543: helpers are fine but if added we should make them generic

server/api/forge.go
```diff
@@ -126,16 +148,52 @@ func PatchForge(c *gin.Context) {
...
+// restoreSecretOptions copies write-only secret options from the stored
+// forge configuration when an update request omits them or sends them empty,
+// mirroring the OAuthClientSecret semantics. Clearing the GitHub App id
+// drops the stored private key, so app authentication can be disabled again.
+func restoreSecretOptions(forge *model.Forge, oldOptions map[string]any) {
+	if forge.Type != model.ForgeTypeGithub {
+		return
+	}
+	// never persist the redaction marker clients echo back
+	delete(forge.AdditionalOptions, "app-private-key-set")
```
6543: a lot of magic consts ... :/

We need to recheck how to properly handle optional settings ...

server/api/forge.go
```diff
@@ -126,16 +148,52 @@ func PatchForge(c *gin.Context) {
...
+		return
+	}
+	forge.AdditionalOptions["app-private-key"] = oldKey
+}
+
+func hasNonEmptyAppID(options map[string]any) bool {
+	switch v := options["app-id"].(type) {
+	case string:
+		return v != ""
+	case float64: // GitHub app ids are numeric, they may be set as JSON number via the API
```
6543: have not looked at how options are de-/serialixed but my guess is that only one type will be valid

server/forge/setup/setup.go
```diff
@@ -144,6 +145,15 @@ func setupGitHub(forge *model.Forge) (forge.Forge, error) {
 	// get additional config and be false by default
 	mergeRef, _ := forge.AdditionalOptions["merge-ref"].(bool)
 	publicOnly, _ := forge.AdditionalOptions["public-only"].(bool)
+	// GitHub App credentials are optional, without them user OAuth tokens are used for all API calls
+	var appID string
+	switch v := forge.AdditionalOptions["app-id"].(type) {
+	case string:
```
6543: same: cant we ensure string?

server/router/api.go
```diff
@@ -243,6 +243,7 @@ func apiRoutes(e *gin.RouterGroup) {
 		{
 			forgeBase.Use(session.MustAdmin())
 			forgeBase.POST("", api.PostForge)
+			forgeBase.GET("/:forge_id/app-health", api.GetForgeAppHealth)
 			forgeBase.PATCH("/:forge_id", api.PatchForge)
```
6543: we have enouth forge_id paths ... we can make a new group

docs/docs/30-administration/10-configuration/12-forges/20-github.md
slayerjain: Makes sense. I have narrowed the minted token to just the job's repo and with read-only content permissions.

docs/docs/30-administration/10-configuration/12-forges/20-github.md
slayerjain: Done

server/api/forge.go
slayerjain: validateForge now dry-runs setup.Forge() for every type except addons (constructing those starts an external process). This required relaxing setupBitbucketDatacenter's hard error on a missing oauth-enable-project-admin-scope — the admin UI never sets that key, so generic validation would have rejected every UI-saved bitbucket-dc forge; it now defaults to false like the other optional booleans. Side effect worth noting: saving a bitbucket-dc forge without git machine credentials now fails with a 400 at save time instead of breaking at runtime.

server/api/forge.go
slayerjain: created consts and the secret redaction/restore logic is driven by a per-forge-type table.

server/api/forge.go
slayerjain: yes, enforced during write.

server/forge/setup/setup.go
slayerjain: same as above

server/router/api.go
slayerjain: done

docs/docs/30-administration/10-configuration/12-forges/20-github.md
slayerjain: Please resolve this if you think it's fine.

john-tipper:
I think this has the potential to assist with [#4052](https://github.com/woodpecker-ci/woodpecker/issues/4052) too. I'd love to see this merged because I'm hitting rate limits in a monorepo in a large organisation, exactly as @slayerjain mentions.

6543:
couldbyou mere current main into this pulls head branch and resolve conflicts :)

thanks, review still happen just time is limited on our side

slayerjain:
Merged current `main` into the head branch and resolved the conflicts — the PR is mergeable again.

Notes on the resolution, in case it helps review:

- `server/api/forge_test.go` and `server/services/setup_test.go` were add/add conflicts (both sides created them independently). Both sides' tests are kept — main's org subtests were folded into the existing `TestPostForge`/`TestPatchForge` rather than duplicated.
- `TestSetupForgeServiceKeepsOrgs` was adapted to the new `setupForgeService(cmd, store, setupForge)` signature and now runs through the file's `runForgeSetup` helper, so it parses flags the same way production does.
- `server/forge/github/app.go` still referenced `go-github/v89` while the rest of the package moved to v90 on main — this auto-merged cleanly but would not have built. Bumped to v90 and dropped the duplicate requirement; `go.mod`/`go.sum` are now identical to main.

`go build ./...`, `go test -tags test ./...`, `vue-tsc --noEmit` and the web unit tests all pass locally.

john-tipper:
Apart from resolving the conflicts again, what is stopping this PR from being merged please? Would really love to see this merged as we're continuing to hit rate limits because of this issue. @6543 @slayerjain

qwerty287:
Our review capabilities are limited. And big PRs like this one often fall into oblivion, sorry…

slayerjain:
Conflicts resolved, rebuilt on current `main`, green.

Worth flagging: `server/forge/github/app.go` was still on `go-github/v90` after #7099 moved the package to v91 — it merges cleanly, then fails to build.

@john-tipper nothing is blocking this from my side. @qwerty287's answer is the real one: at ~2.7k lines the size is the problem, not any specific objection.

Happy to split it the way #6851 became #7024/#7025: (1) forge-level token minting plus the Netrc/Status/File/Dir plumbing, (2) admin API and UI including the write-only private key, (3) `/forges/{id}/app-health` and its UI. (1) alone is what fixes the rate limiting. Say the word and I'll open them.

PR #6809, opened 2026-07-02
Gate Kubernetes nodeSelector backend step config behind agent config

kanywst:
`nodeSelector` is the only step backend option applied without a gate: the other step-settable pod fields each have a `*AllowFromStep` flag, but a step's `nodeSelector` is merged into the pod unconditionally (`pod.go` `nodeSelector()`), so anyone with push access can pin pipeline pods onto chosen nodes.

Lower impact than #6792 since taints still keep steps off tainted nodes, but on a cluster that separates tenants by node labels a step can target another tenant's nodes.

Gated now behind `WOODPECKER_BACKEND_K8S_POD_NODE_SELECTOR_ALLOW_FROM_STEP` (default false), matching the existing pattern. If this was intentional, feel free to close; the gate just makes the field consistent.

`TestNodeSelectorAllowFromStep` checks the step selector is dropped when the flag is off and applied when on.

```bash
go test ./pipeline/backend/kubernetes/ -run TestNodeSelectorAllowFromStep -count=1 -v
```


6543 (approved):

6543:
@kanywst hi, thanks for the pull!

the spellcheck is failing ... i would suggest `nodeselector` -> `node_selector` or so ... also look at the other ci issues ... if they are addressed this should be good to go :rocket:

6543:
https://ci.woodpecker-ci.org/repos/3780/pipeline/35740/35

pipeline/backend/kubernetes/node_selector_allow_from_step_test.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2024 Woodpecker Authors
```
lafriks:
```suggestion
// Copyright 2026 Woodpecker Authors
```

PR #6822, opened 2026-07-09
frontend/linter: export SchemaDefinition for external reuse

HunterZeroSix:
Exports `SchemaDefinition` from the linter package so it can be reused by extensions.

### DIFF
--- a/pipeline/frontend/yaml/linter/schema/schema.go
+++ b/pipeline/frontend/yaml/linter/schema/schema.go
@@ -28,11 +28,11 @@ import (
 )
 
 //go:embed schema.json
-var schemaDefinition []byte
+var SchemaDefinition []byte
 
 // Lint lints an io.Reader against the Woodpecker `schema.json`.
 func Lint(r io.Reader) ([]gojsonschema.ResultError, error) {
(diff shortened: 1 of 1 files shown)


lafriks:
Would embedded variable have value when importing it as go lib?

6543:
@HunterZeroSix can you tell use the usecase :)

If i would have integrated the json would just have fetched it via
```
https://raw.githubusercontent.com/woodpecker-ci/woodpecker/refs/tags/${version}/pipeline/frontend/yaml/linter/schema/schema.json
```

HunterZeroSix:
I'm building a Woodpecker config extension and want to sanitize my configuration before returning it to the server. Exporting the struct would allow me to avoid an extra network call and keep everything in sync with the server's own types.

My initial goal was to strip out extra keys I added and double-check against the schema, but I’ve realized the Woodpecker server handles that linting anyway. Given that, is this change still something you'd be open to for general extension use, or should I hold off?

HunterZeroSix:
I will close this one.
I think there is no need for it.

6543:
having the linting feature in an own package might be usefull but atm it is buried deep not to be exported ... and exposing the variable will more be a hack than an actual help ...

PR #6824, opened 2026-07-09
Add step type label to step-level Prometheus metrics

healdropper:
woodpecker_step_duration_seconds and woodpecker_step_failures_total now
include a `type` label with values `"step"` or `"service"`, allowing
dashboards and alerting rules to filter out service containers (e.g.
`redis`, `postgres`) from actual pipeline build steps.

## Problem

Service containers (defined under `services:` in a pipeline) run for the
entire duration of the pipeline. This caused them to appear as the
"slowest" or "most failing" steps in Prometheus/Grafana dashboards, even
though they are background dependencies, not build steps.

## Solution

The `type` label is derived from `model.StepType` at metric recording time
in `rpc.go`:
- `type="step"` for commands, plugins, clone and cache steps
- `type="service"` for service containers

This allows users to filter cleanly with `{type="step"}` instead of
maintaining a fragile regex exclusion list.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -163,6 +163,7 @@ require (
 	github.com/julienschmidt/httprouter v1.3.0 // indirect
 	github.com/klauspost/compress v1.18.4 // indirect
 	github.com/klauspost/cpuid/v2 v2.3.0 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/leodido/go-urn v1.4.0 // indirect
 	github.com/lestrrat-go/blackmagic v1.0.4 // indirect
 	github.com/lestrrat-go/dsig v1.0.0 // indirect
--- a/server/rpc/rpc.go
+++ b/server/rpc/rpc.go
@@ -214,7 +214,7 @@ func (s *RPC) Update(c context.Context, strWorkflowID string, state rpc.StepStat
 		(step.State == model.StatusFailure ||
 			step.State == model.StatusKilled ||
 			step.State == model.StatusError) {
-		metric.FailurePipelineStepInfoCount.WithLabelValues(workflow.Name, repo.FullName, step.Name).Inc()
+		metric.FailurePipelineStepInfoCount.WithLabelValues(workflow.Name, repo.FullName, step.Name, string(step.Type)).Inc()
 	}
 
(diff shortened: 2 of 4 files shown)


server/rpc/rpc.go
```diff
@@ -209,12 +209,17 @@ func (s *RPC) Update(c context.Context, strWorkflowID string, state rpc.StepStat
...
+		stepType = "service"
+	}
+
 	if metric.FailurePipelineStepInfoCount != nil &&
 		state.Exited &&
 		(step.State == model.StatusFailure ||
 			step.State == model.StatusKilled ||
 			step.State == model.StatusError) {
-		metric.FailurePipelineStepInfoCount.WithLabelValues(workflow.Name, repo.FullName, step.Name).Inc()
+		metric.FailurePipelineStepInfoCount.WithLabelValues(workflow.Name, repo.FullName, step.Name, stepType).Inc()
```
6543: just use step type directly ... there is more than just service

go.mod
```diff
@@ -163,6 +163,7 @@ require (
 	github.com/julienschmidt/httprouter v1.3.0 // indirect
 	github.com/klauspost/compress v1.18.4 // indirect
 	github.com/klauspost/cpuid/v2 v2.3.0 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
6543, line 166: how did this made it's way in??

healdropper: server/rpc (test) → prometheus/client_golang/prometheus/testutil → kylelemons/godebug/diff

The testutil.ToFloat64 calls in TestRPCUpdateStepTypeMetric use prometheus/testutil, which in turn depends on godebug for diffing metric output in test failures. It was pulled in by the original commit that added those tests (Add step type label to step-level Prometheus metrics)

6543 (approved):

healdropper:
The Trivy finding GO-2026-5932 on golang.org/x/crypto is pre-existing and not introduced by this PR

healdropper:
The TestAgentFindByToken failure is unrelated to this PR, I haven't touched anything in server/store/datastore/. could Comeone take a look at it?

6543:
restarted ci

Issue #6808, opened 2026-07-02
Run time using multiple time sources


joelberkeley:
### Component

web-ui, agent, server

### Describe the bug

In this run https://ci.codeberg.org/repos/14915/pipeline/880
the times start as negative. On Matrix, qwerty287 surmised this was caused by different system times for the agent and server, and that the UI was combining these times to work out time elapsed. They also pointed out it should only be using one source of time to calculate elapsed time (I assume the agent time?).

When I fixed my system time, I think the problem went away (I'm not certain it was using my local runner).

As an aside, this means negative time is printed wrong - it should be `-00:14`, not `-1:-14`

### Steps to reproduce

Can't guarantee it's reproducible.

Start a run with a local agent that has a different system time than the server (my clock was out by ~14s).

### Expected behavior

Elapsed time shown in UI regardless of what the agent system time is (assuming the agent is self-consistent)

### System Info

```shell
Server: 3.16
Agent: 3.14 (just noticed this)


services:
  woodpecker-agent:
    image: woodpeckerci/woodpecker-agent:v3.14
    command: agent
    restart: always
    volumes:
      - woodpecker-agent-config:/etc/woodpecker
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      - WOODPECKER_SERVER=grpc.ci.codeberg.org:443
      - WOODPECKER_AGENT_SECRET=${WOODPECKER_AGENT_SECRET}
      - WOODPECKER_GRPC_SECURE=true
      - WOODPECKER_MAX_WORKFLOWS=8

volumes:
  woodpecker-agent-config:
```

### Additional context

(I didn't read the docs or check the bug is fixed in the next version. I'm filing this at the request of qwerty287. I've also fixed my system time so it wouldn't be a useful test if I updated my agent).

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

qwerty287:
Thanks, likely the issue is:

It uses the start time from the server and the end time from the agent. If they are different, you have wrong (possibly negative) times.

[...]

PR #6825, opened 2026-07-09
fix(web): clamp negative durations to avoid garbled time display

TowyTowy:
## What

Step and pipeline durations occasionally render as garbled values such as `-1:-5` or `-5 seconds` in the UI (see #6808).

A duration is computed from a start and an end timestamp that can come from different clocks:
- the browser (`Date.now()`) for still-running items,
- the server for the step/pipeline start,
- the agent for the step/pipeline finish.

Any skew between these clocks can make the duration negative. `splitDuration` passed that straight through, so `durationAsNumber` produced `-1:-5` and `prettyDuration` produced `-5 seconds`.

## How

Clamp the duration to zero in `splitDuration`, the single helper both formatters funnel through, so the UI always shows a sane value. This also covers running items whose negative offset is pure browser/server clock skew, which can't be corrected server-side.

Adds unit tests for `durationAsNumber`/`prettyDuration`, including the negative (skewed-clock) regression cases (they fail without the clamp).

This addresses the "negative time is printed wrong" part of #6808. The underlying single-time-source question (aligning the server/agent step timestamps) is still being discussed on the issue, so this intentionally only makes the display robust and does not change how the timestamps are recorded.

Refs #6808

### DIFF
--- a/web/src/compositions/useDate.test.ts
+++ b/web/src/compositions/useDate.test.ts
@@ -1,0 +1,38 @@
+import { describe, expect, it } from 'vitest';
+
+import { useDate } from './useDate';
+
+const { durationAsNumber, prettyDuration } = useDate();
+
+describe('useDate', () => {
+  describe('durationAsNumber', () => {
+    it('formats sub-minute durations', () => {
+      expect(durationAsNumber(0)).toBe('00:00');
+      expect(durationAsNumber(5000)).toBe('00:05');
+    });
+
+    it('formats minutes and hours', () => {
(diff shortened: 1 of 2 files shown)


6543:
this is more a workaround that a propper fix

TowyTowy:
Fair enough! I kept it display-side on purpose since the time-source discussion in #6808 hadn't settled — but agreed the real fix is a single authoritative clock. If you land on the server-authoritative approach you suggested there, I'm happy to pick up the implementation.

PR #6828, opened 2026-07-09
Add `fuzz` target to make

6543:
Add `fuzz` to make and add code hooks to do so.
This allows us to scrutinize our implementations ...


pipeline/frontend/yaml/matrix/testdata/fuzz/FuzzParse/39d390d3eac7d5f1
```diff
@@ -0,0 +1,2 @@
+go test fuzz v1
```
6543: rm

6543:
ok last fuzz run finally turned back clean :tada: ...

qwerty287:
Do we keep this updated if the CI never executes the tests? I doubt that

6543:
well as fuzzing is quite time consuming and non deterministic I'm against CI coverage ... I added it to the tasks to be done before release on tje release template ...

to be precice a good fuzzing can take 30min up to 2h ... and in normal case nothing breaks ...

qwerty287:
I know, but it's very likely that someone will forget to update the fuzz tests if changing something and then everything breaks when you want to do the release. Maybe we could limit fuzzing to a few seconds in CI just to make sure it's still working? If it finds an error then we also found a bug in our code… I don't see an issue if it's non-deterministic here

6543:
also fuzzing is not a test and does not add to test coverage ... it's aim is to surface the unknown unknowns ...

qwerty287:
I know, still you have to maintain it. I wouldn't execute it in ci to "test" our code, but to make sure fuzzing is still working.

6543:
ah sure i can add a fuzzing step that does just run few seconds for each fuzze hook

6543:
@qwerty287 fuzz now runs for ~7min

qwerty287:
Thanks, still quite long for a simple ci test but it's OK as our regular tests need way longer anyways. I'll chrck the PR again later

qwerty287 (approved):

PR #6829, opened 2026-07-09
fix(server): record step, workflow and pipeline times from the server clock

### LINKED ISSUE #6808: Run time using multiple time sources

TowyTowy:
The proper fix, as discussed — following the closure of #6825 (the display-side workaround was rightly rejected as "more a workaround than a proper fix").

### Problem (#6808)

A step's **start** time was recorded from the server clock, but its **finish** time was copied from the **agent** clock. When the agent and server clocks are out of sync (the reporter's was ~14s off), durations came out wrong or negative. The same split exists at the workflow and pipeline levels, where all start/finish times come from the agent's `WorkflowState` — and in a multi-agent pipeline, different workflows can even carry timestamps from *different* agent clocks.

As diagnosed by @qwerty287 and suggested by @6543 ("i would trust the server more ... update the statemachine to use the server time for finished too"), this makes the **server the single authoritative clock** for every persisted `Started`/`Finished` timestamp.

### What now uses server time

- **Step `Started`** — `CalcStepStatus` on the pending→running transition (was `state.Started`; already effectively server time since the "started" trace carries a zero value, now unconditionally so).
- **Step `Finished`** — both finish branches and the cancellation branch (was `state.Finished`). *This is the core bug.*
- **Workflow `Started`** — `UpdateWorkflowStatusToRunning` (was `state.Started`, via `Init`).
- **Workflow `Finished`** — `UpdateWorkflowStatusToDone` (was `state.Finished`, via `Done`).
- **Pipeline `Started`** — `Init` handler call-site (was `state.Started`).
- **Pipeline `Finished`** — inherits the now-server-authoritative `workflow.Finished` (no separate change).
- **Killed service/daemon children** — the `Done` handler stamps server time (was `state.Finished`).

### What deliberately stays agent-sourced

- **Relative in-step / log-line timing** — these are relative offsets (per @qwerty287: "the log times are relative times, not timestamps"), so they stay on the agent. No web-UI or log changes.
- The agent still **sends** its `Started`/`Finished` over RPC — they're used only as **presence flags** (`Finished==0` → still running; `Started==0` → skipped), never persisted as values, so they can't be dropped from the protocol.
- **Skipped steps** keep their (near-always-zero) agent `Finished`: a skipped step never ran (`Started==0`), so there's no start/finish pair to skew.

### Open question for maintainers

@6543 also floated a second option in the issue — keep agent times but **sanitize on the server** (clamp drift beyond the gRPC retry timeout, override `end < start`). That subthread ended without resolution. This PR implements the simpler server-authoritative direction; if you prefer the drift-clamping variant instead, say so and I'll re-cut it — the server-time baseline here is a natural foundation to layer clamping onto later.

### Tests

Updated the `server/pipeline` unit tests to assert server-clock behavior and added `TestStepStatusSkewedAgentClock`, a two-phase test that starts a step and then finishes it with a badly skewed agent clock; it fails on `main` (negative duration) and passes here. `go test ./server/pipeline/ ./server/rpc/` passes, `go vet` clean.

Fixes #6808

### DIFF
--- a/server/pipeline/step_status.go
+++ b/server/pipeline/step_status.go
@@ -31,10 +31,23 @@ import (
 func CalcStepStatus(step model.Step, state rpc.StepState) (_ *model.Step, cancelPipelineFromStep bool, _ error) {
 	log.Debug().Str("StepUUID", step.UUID).Msgf("Update step %#v state %#v", step, state)
 
+	// now is the server clock. We use it as the single authoritative source
+	// for the persisted Started and Finished timestamps instead of trusting
+	// the agent-supplied state.Started/state.Finished. Previously the start
+	// time was recorded from the server clock (state.Started is 0 on the
+	// "started" trace) while the finish time was copied from the agent clock,
+	// so clock skew between agent and server produced wrong or even negative
+	// step durations (#6808). The agent-reported timestamps are still used as
+	// presence flags (0 vs non-0) to decide the state transition, but never as
+	// stored values. Relative in-step/log timing stays agent-sourced.
+	now := time.Now().Unix()
+
 	switch step.State {
 	case model.StatusPending:
 		// Handle skip before anything else — skipped steps never started,
-		// so we must not set Started or transition through Running.
+		// so we must not set Started or transition through Running. A skipped
+		// step never ran (Started stays 0), so there is no start/finish pair
+		// that could be skewed; keep the agent value here as-is.
 		if state.Skipped {
 			step.State = model.StatusSkipped
 			if state.Finished != 0 {
@@ -47,17 +60,11 @@ func CalcStepStatus(step model.Step, state rpc.StepState) (_ *model.Step, cancel
 		if state.Finished == 0 {
 			step.State = model.StatusRunning
 		}
-		step.Started = state.Started
-		if step.Started == 0 {
-			step.Started = time.Now().Unix()
-		}
+		step.Started = now
 
 		// Handle direct transition to finished if step setup error happened
 		if state.Exited || state.Error != "" {
-			step.Finished = state.Finished
-			if step.Finished == 0 {
-				step.Finished = time.Now().Unix()
-			}
+			step.Finished = now
 			step.ExitCode = state.ExitCode
 			step.Error = state.Error
 
@@ -75,10 +82,7 @@ func CalcStepStatus(step model.Step, state rpc.StepState) (_ *model.Step, cancel
 	case model.StatusRunning:
 		// Already running, check if it finished
 		if state.Exited || state.Error != "" {
-			step.Finished = state.Finished
-			if step.Finished == 0 {
-				step.Finished = time.Now().Unix()
-			}
+			step.Finished = now
 			step.ExitCode = state.ExitCode
 			step.Error = state.Error
 
@@ -101,7 +105,7 @@ func CalcStepStatus(step model.Step, state rpc.StepState) (_ *model.Step, cancel
 	if state.Canceled && step.State != model.StatusKilled {
 		step.State = model.StatusKilled
 		if step.Finished == 0 {
-			step.Finished = time.Now().Unix()
+			step.Finished = now
 		}
 	}
 
--- a/server/pipeline/step_status_test.go
+++ b/server/pipeline/step_status_test.go
@@ -16,9 +16,11 @@ package pipeline
 
 import (
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 
 	"go.woodpecker-ci.org/woodpecker/v3/pipeline"
 	"go.woodpecker-ci.org/woodpecker/v3/rpc"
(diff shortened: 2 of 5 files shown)


qwerty287:
Thanks. For discussion: Can we remove the started field from the RPC communication then completely?

TowyTowy:
Good question — after this PR the server reads exactly one of the two. `StepState.Started` is never read server-side anymore (it's only set in `agent/tracer.go`), so it could be dropped from the step RPC outright. `WorkflowState.Started` is still consulted once, in `UpdateWorkflowStatusToDone` (`state.Started == 0` → `StatusSkipped`). If that signal became an explicit `Skipped` bool on `WorkflowState`, both `Started` fields could go entirely — the caveat is proto compatibility (an older agent talking to a newer server would need the old field still honored, or a capability/version gate). Happy to do that cleanup in this PR or as a follow-up, whichever you prefer.

6543 (changes requested):
this is a bit delecate can we do solve this issue without ai it is not an loww hanging fruite ...

6543:
and as per my comment it does make sense to have agent propvided times, we should A. use also start times from agent and B sanitize it on the server side ... e.g. start times can not be before pipeline creation time etc...

6543:
> this is a bit delecate can we do solve this issue without ai it is not an loww hanging fruite ...

i looked at the patch and it mostly looks fine tbh... if we have decided on https://github.com/woodpecker-ci/woodpecker/issues/6808#issuecomment-4936444543 and the final answer is server side ...

this is a good base.

but currently we "interpret" what the state is based on the "metadata" of the agent on the serverside ... e.g. `server/pipeline/step_status.go`

a better way would be to exactly send over the intend by the agent e.g. "mark as starete", "it failed", "it was cancled", ...
this refactor should also go into that and remove timestamps we dont use anymore then ... but you can only remove those if refactor mentioned is done ...

lafriks (dismissed):

TowyTowy:
Thanks for the careful look, and I hear the concern about AI on something this delicate — that's fair, and I'd rather get it right than push it. Being transparent: this was AI-assisted and human-reviewed on my end.

On direction: this PR does the pure server-authoritative version (ignore agent times entirely). From your #6808 comment it sounds like you've since landed on keeping agent-provided times but sanitizing them server-side (start can't precede pipeline creation, end can't precede start, etc.), which is a different shape than what's here. I'm happy to rework it that way if it's useful as the base you mentioned — that part is bounded.

The bigger piece you describe — having the agent send explicit intent ("started" / "failed" / "cancelled") instead of the server interpreting metadata, then dropping the timestamps that become unused — is a real state-machine refactor, and given the delicacy that feels like something the team may prefer to own directly rather than have come in as an AI-assisted PR. So whichever you'd rather: I can (a) trim this toward the server-side sanitization approach as a starting point, or (b) close it and leave the whole thing with you. No attachment either way — your call.

qwerty287:
As I commented in #6808 already I also think the best way would be only use agent times, so this PR is the completely wrong direction now... Sorry for that

TowyTowy:
Thanks @qwerty287, @6543 - that settles the direction. Since the consensus is to keep agent-provided timestamps and sanitize them server-side (start >= pipeline-creation, end >= start), this PR's server-authoritative approach is the wrong shape, so I'll close it rather than push it the wrong way.

One byproduct that may still be useful: `TestStepStatusSkewedAgentClock` reproduces #6808 concretely - it starts a step, then finishes it with a badly skewed agent clock, and it fails on `main` today (negative duration, `Finished=1000 < Started`). Happy to leave it as a reference for whoever picks up the sanitize-on-server work, or drop it - your call.

Appreciate the careful review.

PR #6830, opened 2026-07-10
Migrate modernize encryption service

6543:
fix https://ci.woodpecker-ci.org/repos/3780/pipeline/35857/24 by remove unmaintained `golang.org/x/crypto`

use new sha3 for keyID derivation


qwerty287 (approved):
Fine if we can remove the crypto package, but untested (but it's disabled and not working anyways)

But as bcrypt is a "proper" password hashing function that would be better than sha3 I think.

qwerty287:
Looks like there exists a maintained go library for argon2: https://github.com/matthewhartstonge/argon2

That could also be an alternative.

PR #6839, opened 2026-07-12
feat(server): add unix socket permission option

### LINKED ISSUE #6805: Allow setting Unix socket permissions

In https://github.com/woodpecker-ci/woodpecker/pull/6721, support for exposing woodpecker to a reverse proxy via Unix sockets was introduced.
When this is used, woodpecker-server creates a new socket at a location defined in the config, and the reverse proxy then tries to forward requests to that socket. To do so, it needs to have access to the socket.
If the reverse proxy is under another user that does not share groups (for security), this means that the script starting woodpecker (or the administrator) needs to manually `chmod` the socket before the proxy can access it.

### Suggested solution

Another go application with a web interface that can be exposed over a Unix socket is Gitea, which provides the `UNIX_SOCKET_PERMISSION` variable that users can configure to set the permissions.
Adding something similar to woodpecker would make using the Unix socket support easier.

### Alternative

It is probably possible to use some systemd feature to set the permissions, it is certainly possible to use a shell script to do so, and putting both services under the same user "solves" this issue at the cost of security.
If using some systemd feature or a script is intended, documenting this might be helpful, since I am probably not the only person unaware of that.

### Additional context

_No response_

### Validations

cyphercodes:
## Summary

Adds `WOODPECKER_UNIX_SOCKET_PERMISSION` / `--unix-socket-permission` for HTTP listeners using `unix://`, so admins can configure the created socket mode directly.

Closes #6805

## Tests

- `GOMAXPROCS=2 go test -tags external_web -p 1 ./cmd/server -run '^TestParseUnixSocketPermission$' -count=1 -v`
- `GOMAXPROCS=2 go test -tags external_web -p 1 ./cmd/server -count=1`
- `git diff --check`

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -269,13 +269,7 @@ func run(ctx context.Context, c *cli.Command) error {
 				return err
 			}
 			if network == "unix" && c.String("unix-socket-permission") != "" {
-				mode, err := parseUnixSocketPermission(c.String("unix-socket-permission"))
-				if err != nil {
-					stopServerFunc(err)
-					return err
-				}
-				if err := os.Chmod(addr, mode); err != nil {
-					err = fmt.Errorf("could not set unix socket permission: %w", err)
+				if err := setUnixSocketPermission(addr, c.String("unix-socket-permission")); err != nil {
 					stopServerFunc(err)
 					return err
 				}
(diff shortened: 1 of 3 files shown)


lafriks (approved):

cmd/server/server.go
```diff
@@ -267,6 +268,18 @@ func run(ctx context.Context, c *cli.Command) error {
 				stopServerFunc(err)
 				return err
 			}
+			if network == "unix" && c.String("unix-socket-permission") != "" {
+				mode, err := parseUnixSocketPermission(c.String("unix-socket-permission"))
```
6543: can you move os.Chmod into parseUnixSocketPermission helper func and rename parseUnixSocketPermission acordinlgy

6543 (changes requested):

cyphercodes:
Addressed the review request by moving the chmod call into the unix socket permission helper and renaming it to `setUnixSocketPermission`.

Verification:
- `git diff --check`
- `go test -tags external_web ./cmd/server -run TestSetUnixSocketPermission`
- `go test -tags external_web ./cmd/server`

6543 (approved):

PR #6840, opened 2026-07-13
WIP: Move secrets out of queue

6543:
## Test claude fable5 taggle secret in queue

so we can estimate how much work it currently  would be and if we could start upstreaming some things that would make sense for us


6543:
cherry-picked a commit -> https://github.com/woodpecker-ci/woodpecker/pull/6847

6543:
also mock generation: https://github.com/woodpecker-ci/woodpecker/pull/6848

PR #6842, opened 2026-07-13
WebUI: unregister scaffold Tab on unmount and stable Tab order

6543:
Tabs register themselves into the Scaffold's tabs list on mount but never remove their entry on unmount. Since param-only navigation (e.g. the superseded-by link, repo-to-repo links) reuses the wrapper and its Scaffold, a conditionally rendered tab like the pipeline *Errors* tab or the repo *Pull requests* tab stayed visible after its condition turned false, linking to an empty page.

Fix: remove the entry in `onBeforeUnmount`. The mount-time dedup guarantees at most one entry per route, so removal by route is safe. Includes lifecycle tests (register, dedup, unregister, re-register).

Found during review of #6577

### DIFF
--- a/web/src/components/layout/scaffold/Tab.test.ts
+++ b/web/src/components/layout/scaffold/Tab.test.ts
@@ -1,0 +1,91 @@
+import { mount } from '@vue/test-utils';
+import { describe, expect, it } from 'vitest';
+import { defineComponent, h, nextTick, ref } from 'vue';
+
+import type { Tab as TabType } from '~/compositions/useTabs';
+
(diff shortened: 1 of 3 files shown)


qwerty287 (approved):
Not tested and I'm also not really into this vue and js/ts logic, so maybe it would be good if someone else would also check it out...

6543:
well UI tests always look clustered ... i would not wana write them myselfe but ai generated tests have no real downside - it wont ship in prod and worstcase you just delete them and have less coverage ...

PR #6843, opened 2026-07-13
WebUI: Fixes three bugs around component unmount and reset in compositions

6543:
Fixes three bugs around component unmount and reset in the web UI compositions.

| Composition | Symptom | Fix |
|---|---|---|
| `useInterval` | Unmount during the initial fetch leaked the interval: endless API polling and state writes on a dead component | Track unmount state, skip interval registration after unmount |
| `usePagination` | Fast filter/branch switching showed duplicated or wrong list entries because a stale in-flight response resolved after `resetPage()` | Generation token; stale responses return without touching state |
| `useFavicon` | Error/pending favicon could stick globally because resetting was left to callers | Auto-reset via `onScopeDispose` once the last consumer unmounts |

All fixes were developed test-first; 14 new unit tests cover the races, including unmount-mid-fetch, rapid remounts, reset-while-loading and overlapping favicon consumers.

### DIFF
--- a/web/src/compositions/useFavicon.test.ts
+++ b/web/src/compositions/useFavicon.test.ts
@@ -1,0 +1,111 @@
+import { shallowMount } from '@vue/test-utils';
+import { beforeAll, beforeEach, describe, expect, it } from 'vitest';
+import { nextTick } from 'vue';
+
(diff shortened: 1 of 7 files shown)


qwerty287 (approved):
also seems fine to me but I didn't review the tests

PR #6844, opened 2026-07-14
Explizite ConfigFetchExtension Error

6543:
make https://github.com/woodpecker-ci/woodpecker/discussions/6841 be explicit and nice formatted in webui ...


server/pipeline/create.go
qwerty287, line 99: We have this case branch below - it won't work anymore, right?

server/services/config/http.go
```diff
@@ -76,6 +77,17 @@ func (h *httpService) Fetch(ctx context.Context, forge forge.Forge, user *model.
...
 	status, err := h.client.Send(ctx, http.MethodPost, h.endpoint, body, response)
+
+	// handle 422 - extension explicitly rejected the pipeline, response body is the error message
+	if status == http.StatusUnprocessableEntity {
+		message := ""
+		var statusErr *utils.HTTPStatusError
+		if errors.As(err, &statusErr) {
+			message = statusErr.Body
+		}
+		return nil, &ErrConfigExtension{Message: message}
```
qwerty287, line 88: Couldn't we treat all bad codes as `ErrConfigExtension`? With a proper error message that includes the HTTP status code

6543: no as the config extension could itselfe have an issue etc ... this is adedicated channel to allow the extension to forward an own error to the webui.

In my opinion to differentiate between the two issues makes a huge UX difference

qwerty287: ~~But in any caye it shows an error with the response body, right?~~

Ah no I see it's different, sorry

qwerty287:
Read the code again, and I thought I understood it, but I don't get it.

The only difference between 422 and e.g. 500 status code is that 422 is treated as config extension error and 500 as generic error? I don't like this tbh. Any bad response code should be a config extension error. If there's a different reason (e.g. config extension is not reachable cause you have a typo in your URL), that should be a generic error, yes, but any error that's sent back by the config extension should be config extension type.

6543:
ok, will look into Pro/Cons again and adress code

PR #6845, opened 2026-07-14
Make grpc secret autogenerated

6543:
Currently an admin has to read the docs carefully and understand what an grpc sercret is ...
... this flips it, by just generate a random one if not set and make agents re-auth quicker if jwt does not work anymore ...


qwerty287 (approved):
also looks fine, but didn't check the tests and also didn't test

6543:
# test prove of agent recovery after secret rotation:
## Before
image

## Now
image
image

PR #6851, opened 2026-07-15
Speed up and harden pipeline creation: concurrent deduped status posting and single-request config fetch

slayerjain:
Three performance/resilience improvements to the pipeline creation path, measured and validated on a production instance (keploy org, pipelines with 60+ workflows):

## 1. Post per-workflow commit statuses concurrently

`updatePipelineStatus` posted one status per workflow sequentially. For a pipeline with 63 workflows that alone takes tens of seconds. Statuses are now posted with a bounded concurrency of 10. This also fixes an inconsistency: the old loop aborted on the first error, leaving all remaining workflows without a status; now every status is attempted and failures are logged individually.

## 2. Post forge statuses only once before start

Pipeline creation posted the full set of per-workflow statuses **three times** before the pipeline started (after compile, on the pending transition and again in `start()`), all with identical pending state — approval and restart flows had the same double-posting. The pubsub events are kept for the UI, but forge statuses are now posted once, in `start()` (blocked pipelines post their round in the blocked branch, since they never reach `start()`).

Combined effect of 1+2 on our instance (63-workflow pipeline, webhook to first workflow start): **~91s median → 2s**.

This commit also hardens the `start()` failure path: previously a failure left the pipeline pending forever with pending commit statuses; now the pipeline *and its workflows* transition to error and terminal statuses are posted.

## 3. Fetch config directories with a single GraphQL request

`Dir()` fetched a directory listing plus one REST request per file. Besides the latency, this makes config fetching fragile: with ~60 config files, a single transient forge error fails the whole fetch — during a recent period of intermittent GitHub 502s, virtually every pipeline creation on our instance failed with "fallback did not find config", because the probability of at least one of 60 requests failing was near 1.

`Dir()` now fetches the whole directory (names + contents) with one GraphQL tree query, falling back to the per-file REST path for symlinked (GraphQL returns the link target instead of the linked content), truncated or binary entries and on any transport/query error. The GraphQL endpoint is derived for both github.com and GHES, and works with the same bearer tokens as the REST calls.

Behavior notes:
- Subdirectories inside a config directory are now skipped consistently on both paths. Previously the REST path errored on them ("is a folder not a file"), failing config fetching for repos with nested directories.
- The phase-timing log added in the first commit is at debug level.

All changes are covered by tests (fixture-served GraphQL responses incl. symlink/fallback/not-found cases, race-enabled).


slayerjain:
I do use llm/agents to automated repeated work.

6543:
can we split things up in topic dedicated pulls please?

slayerjain:
Split as requested, and rebased onto current `main` (this branch had fallen 124 commits behind):

- #7024 — `perf(pipeline)`: post per-workflow commit statuses once and concurrently
- #7025 — `perf(github)`: fetch config directories with a single GraphQL request

The two share no files, so they can be reviewed and merged independently. Closing this one in favour of them.

While rebasing I also fixed a few things that were wrong in the original:

- the `start()` failure path errored the workflows but left their steps pending forever — it now transitions the steps too, matching `decline.go`
- status posting used an `errgroup` whose callback always returned nil, so the `Wait()` was dead code and errors were swallowed entirely; it is now a plain `WaitGroup` + semaphore that respects ctx cancellation and logs a summary when statuses fail
- concurrency dropped from 10 to 4, since forges rate limit concurrent writes from a single user much more aggressively than sequential ones
- the GraphQL path returned `ErrConfigNotFound` for any response without a repository object, including malformed ones — since that error makes `Create` delete the pipeline, a bad response could silently drop a webhook. Only a response that actually resolved the repository can report not-found now; anything else falls back to REST. Regression test added.

PR #6853, opened 2026-07-15
Make sure K8s labels are not too long

### LINKED ISSUE #6827: kubernetes: step with long name won't start
### Component

agent

### Describe the bug

In our generated workflow config, we have pretty long step names. This makes workflow fail on kubernetes with 
```
Pod "wp-01kx3jmh7tg6xps2wgnhnh7nvy" is invalid: metadata.labels: Invalid value: "rtos-jot-release-arm-gnu-toolchain-14.2.rel1-x86-64-arm-none-eabi": must be no more than 63 characters
``` 

### Steps to reproduce

1. Install woodpecker 3.16.0 with kubernetes backend
2. Run the following pipeline
```
when:
  - event: [push]

steps:
  - name: this is fine
    image: ubuntu:latest   
    commands:
      - echo "hi"

  - name: but this is way too long to be converted into the kubernetes label which has to follow dns naming semantics
    image: ubuntu:latest
    commands:
      - echo "this never runs"

```
3. git and `this is fine` steps will be fine, but the last one fails

### Expected behavior

_No response_

### System Info

```shell
helm chart 3.6.5
```

### Additional context

_No response_

### Validations

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/6827

cut off if too long

### DIFF
--- a/pipeline/backend/kubernetes/utils.go
+++ b/pipeline/backend/kubernetes/utils.go
@@ -46,10 +46,6 @@ func getHostnameOrEmpty(name string) string {
 	}
 	clean = strings.ReplaceAll(clean, ".", "-")
 
-	if len(clean) > maxDNSLabelLen {
-		clean = clean[:maxDNSLabelLen]
-	}
-
 	clean = strings.Trim(clean, "-")
 
 	if dnsLabelPattern.MatchString(clean) {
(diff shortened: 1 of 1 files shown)


xoxys:
The entire package is a bit bloated IMO. We should make use of https://pkg.go.dev/k8s.io/apimachinery@v0.36.2/pkg/api/validate/content, and we should also not just truncate. Instead, I would suggest something like this:

```go
func LabelValue(name string) string {
	if len(name) <= DNS1035LabelMaxLength {
		return name
	}

	h := fnv.New32a()
	h.Write([]byte(name))

	hashStr := fmt.Sprintf("%08x", h.Sum32())

	reservedLength := 1 + len(hashStr)
	maxBaseLength := DNS1035LabelMaxLength - reservedLength

	truncated := name[:maxBaseLength]
	truncated = strings.TrimRight(truncated, "-")

	return fmt.Sprintf("%s-%s", truncated, hashStr)
}
```

Why this approach:

- prevent duplicates
- ensures the truncated name does not end up with an invalid char 
- same objects will reliably generate the same label value

However, there are even more issues. The utils are mixing up DNS name (Kubernetes resource names) requirements and label value requirements (which are not the same)

qwerty287:
Ok, I'm not into k8s at all. Could you do a proper refactoring/fix of this?

xoxys:
Yes, can take a look.

PR #6855, opened 2026-07-15
refactor: rework k8s utils to handle resource names and labels

### LINKED ISSUE #6827: kubernetes: step with long name won't start

xoxys:
Fixes: https://github.com/woodpecker-ci/woodpecker/issues/6827

Refactored Kubernetes backend to properly handle DNS names and label values with distinct validation rules, replacing the old regex-based validation with upstream Kubernetes validation functions.

- Uses upstream Kubernetes validation instead of custom regex patterns
- New sanitization for label values (max 63 chars, allows underscores/dots)
- Long names are truncated with SHA-256 hash suffix to prevent collisions
- `toDNSName()` for DNS-1123 subdomains (max 253 chars), `toLabelValue()` for label values (max 63 chars)
- Pod labels and secret labels now use `toLabelValue()` instead of `toDNSName()`

However, this in fact also introduces a breaking behavior:

Label values now preserve underscores (e.g., `1_0` stays `1_0` instead of becoming `1-0`). This is correct behavior per Kubernetes spec but may cause label selector mismatches during rolling upgrades.

### DIFF
--- a/pipeline/backend/kubernetes/pod.go
+++ b/pipeline/backend/kubernetes/pod.go
@@ -89,7 +89,7 @@ func stepToPodName(step *types.Step) (name string, err error) {
 }
 
 func podName(step *types.Step) (string, error) {
-	return dnsName(podPrefix + step.UUID)
+	return toDNSName(podPrefix + step.UUID)
 }
 
 func podMeta(step *types.Step, config *config, options BackendOptions, podName, taskUUID string) (kube_meta_v1.ObjectMeta, error) {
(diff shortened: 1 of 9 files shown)


qwerty287 (approved):
Code looks good, the ci fail is unrelated

6543 (approved):

PR #6857, opened 2026-07-16
ReFetch secrets on pipeline restart

### LINKED ISSUE #2982: Restarting a pipeline that failed due to missing secrets is not enough
### Component

server

### Describe the bug

Restarting a pipeline that failed due to missing secrets is not enough.
I have seen it in a tag event.
```
secret "***" not found or not allowed to be used
```
Then I added the secret and restarted the pipeline and the error changed to
```
[generic]
pipeline definition not found
```

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"next-5a7e314f5a"}
```

### Additional context

Might be related to https://github.com/woodpecker-ci/woodpecker/issues/1678

### Validations

ivanvc:
Pipeline configs are persisted to the store BEFORE createPipelineItems is called. This ensures that retries of failed pipelines can still access the config.

Fixes #2982.


qwerty287 (approved):
Thanks, seems working. There are some other issues related to restarting errored pipelines though… Restarting e.g. a pipeline after removing a secret gives you a 500 but creates the pipeline with errors…

server/pipeline/create_test.go
```diff
@@ -0,0 +1,64 @@
...
+	forgeYamls := []*forge_types.FileMeta{
+		{Name: ".woodpecker.yaml", Data: []byte("steps:\n  test:\n    image: alpine")},
+	}
+
+	persistedConfig := &model.Config{ID: 1, RepoID: 10, Name: "woodpecker"}
+	mockStore.On("ConfigPersist", &model.Config{
+		RepoID: int64(10),
+		Name:   "woodpecker",
+		Data:   forgeYamls[0].Data,
+	}).Return(persistedConfig, nil)
```
6543, line 45: can you also add a test that checks what you try to fix and not only internal code path ....
e.g.: create pipeline with workflow with secret usage -> let it fail, update secret, let it restart

6543 (changes requested):

server/pipeline/create_test.go
6543: PS: I would suggest to extend `e2e/scenarios/restart_test.go` ... as this should ensure the functionality even if we refactor code a lot

ivanvc: Done, @6543. Added an e2e test case in 2f675ad.

6543 (approved):
thanks

PR #6859, opened 2026-07-17
Fix CreatePipeline retry

### LINKED ISSUE #3884: Likely concurrency issue pq: 'duplicate key value violates unique constraint \"UQE_pipelines_s\"'

nicolasbatistoni:
## The bug

The retry added in #6067 to work around the pipeline-number race **never runs**.

`CreatePipeline` inserts go through `wrapInsert()`, which normalizes every driver-specific unique constraint violation into a sentinel:

```go
// server/store/types/errors.go
ErrInsertDuplicateDetected = errors.New("on insert duplicate based on constraints was detected")
```

`isUniqueConstraintError()` then string-matches the **raw driver** messages. The sentinel's text matches **none** of them — note it says `constraints`, not `unique constraint`:

| pattern checked | matches the sentinel? |
|---|---|
| `duplicate key` | ❌ |
| `Duplicate entry` | ❌ |
| `UNIQUE constraint failed` | ❌ |
| `unique constraint` | ❌ |
| `UNIQUE violation` | ❌ |

So it always falls through to `backoff.Permanent(err)`: `CreatePipeline` gives up on the **first** collision, the webhook gets a **500**, and the pipeline is **never created**. GitHub does not retry webhooks, so the push is silently lost — no red build, nothing to look at.

## How it regressed

- #6067 (2026-02-05) added the retry and closed #3884.
- #6259 (2026-03-19) introduced `wrapInsert` and applied it inside `CreatePipeline`, silently defeating it.

Affected: **v3.14.0 → `main`** (verified per tag). v3.13.0 had no retry at all, so there is no version to downgrade to either.

## Measured in production

Self-hosted, Postgres, kubernetes backend, v3.16.0: **17 × HTTP 500 in 24 h across 10 repos**, each one a lost pipeline.

A PR merge fires two push webhooks ~130 ms apart (merge commit + branch deletion), which hits the race reliably:

```
13:08:29.348  push → 200
13:08:29.477  push → 500
```

Server log:

```json
{"level":"error","error":"backoff: permanent error (last error: on insert duplicate based on constraints was detected)","message":"failed to save pipeline for x/y"}
{"level":"error","error":"Error #01: failed to save pipeline for x/y","method":"POST","path":"/api/hook","status":500}
```

That `backoff: permanent error` line is the proof: it is the sentinel, verbatim, on the code path that was supposed to retry.

## The change

Recognize the sentinel with `errors.Is()`, **keeping** the raw string checks — `isUniqueConstraintError` is also reachable with errors that did not go through `wrapInsert`.

## Scope

This only revives the **mitigation**. The underlying race is still there (`CreatePipeline` does `SELECT MAX(number)` + `INSERT` without a lock) and **#5534 is the real fix** — this is not a substitute for it, just a restoration of the safety net that was silently lost.

## Tests

Adds `TestIsUniqueConstraintError`, verified in **both** directions:

```
against current main:  --- FAIL: TestIsUniqueConstraintError
                           Should be true
                           the sentinel returned by wrapInsert must be retryable

with this change:      ok   .../server/store/datastore   0.004s
```

`go build` and `go test ./server/store/datastore/` pass; `gofmt` clean.

Refs #3884, #6067, #6259, #5534

### DIFF
--- a/server/store/datastore/pipeline.go
+++ b/server/store/datastore/pipeline.go
@@ -16,6 +16,7 @@ package datastore
 
 import (
 	"context"
+	"errors"
 	"strings"
 	"time"
 
@@ -24,6 +25,7 @@ import (
 	"xorm.io/xorm"
 
 	"go.woodpecker-ci.org/woodpecker/v3/server/model"
+	"go.woodpecker-ci.org/woodpecker/v3/server/store/types"
 )
 
 func (s storage) GetPipeline(id int64) (*model.Pipeline, error) {
@@ -196,6 +198,16 @@ func (s storage) CreatePipeline(pipeline *model.Pipeline, stepList ...*model.Ste
 func isUniqueConstraintError(err error) bool {
 	if err == nil {
 		return false
+	}
+
+	// The inserts in CreatePipeline go through wrapInsert, which normalizes driver-specific unique
+	// constraint violations into types.ErrInsertDuplicateDetected. Its message ("on insert duplicate
+	// based on constraints was detected") matches none of the raw driver patterns below, so without
+	// this check the retry added in #6067 never triggers and CreatePipeline fails permanently on the
+	// first collision. Keep the raw string checks too: this function is also reachable with errors
+	// that did not go through wrapInsert.
+	if errors.Is(err, types.ErrInsertDuplicateDetected) {
+		return true
 	}
 
 	errStr := err.Error()
--- a/server/store/datastore/pipeline.go
+++ b/server/store/datastore/pipeline.go
@@ -17,7 +17,6 @@ package datastore
 import (
 	"context"
 	"errors"
-	"strings"
 	"time"
 
 	"github.com/cenkalti/backoff/v7"
@@ -171,7 +170,7 @@ func (s storage) CreatePipeline(pipeline *model.Pipeline, stepList ...*model.Ste
 		pipeline.Created = time.Now().UTC().Unix()
 		// only Insert set auto created ID back to object
 		if err := wrapInsert(sess.Insert(pipeline)); err != nil {
-			if isUniqueConstraintError(err) {
+			if errors.Is(err, types.ErrInsertDuplicateDetected) {
 				return struct{}{}, err
 			}
 			return struct{}{}, backoff.Permanent(err)
@@ -181,7 +180,7 @@ func (s storage) CreatePipeline(pipeline *model.Pipeline, stepList ...*model.Ste
 			stepList[i].PipelineID = pipeline.ID
 			// only Insert set auto created ID back to object
 			if err := wrapInsert(sess.Insert(stepList[i])); err != nil {
-				if isUniqueConstraintError(err) {
+				if errors.Is(err, types.ErrInsertDuplicateDetected) {
 					return struct{}{}, err
 				}
 				return struct{}{}, backoff.Permanent(err)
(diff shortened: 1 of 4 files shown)


server/store/datastore/pipeline.go
```diff
@@ -198,6 +200,16 @@ func isUniqueConstraintError(err error) bool {
 		return false
 	}
 
+	// The inserts in CreatePipeline go through wrapInsert, which normalizes driver-specific unique
+	// constraint violations into types.ErrInsertDuplicateDetected. Its message ("on insert duplicate
+	// based on constraints was detected") matches none of the raw driver patterns below, so without
+	// this check the retry added in #6067 never triggers and CreatePipeline fails permanently on the
+	// first collision. Keep the raw string checks too: this function is also reachable with errors
+	// that did not go through wrapInsert.
```
lafriks: I don't think such a lengthy comment is needed here. "Check wrapInsert normalized error" is more than enough

lafriks (changes requested):

nicolasbatistoni:
Thanks — shortened to `// Check wrapInsert normalized error.` as suggested, and trimmed the test comment in the same spirit. The long-form explanation lives in the PR description, which is a better home for it anyway.

`gofmt` clean, `go test ./server/store/datastore/` passing.

server/store/datastore/pipeline.go
```diff
@@ -198,6 +200,11 @@ func isUniqueConstraintError(err error) bool {
...
 	}
 
+	// Check wrapInsert normalized error.
+	if errors.Is(err, types.ErrInsertDuplicateDetected) {
+		return true
+	}
+
 	errStr := err.Error()
 	// Check for common unique constraint error patterns across different databases
 	return strings.Contains(errStr, "duplicate key") ||
```
6543: can we move that check here into the ErrInsertDuplicateDetected logic ...

and instead of a smal isUniqueConstraintError we can use the error.Is directly

server/store/datastore/pipeline_test.go
```diff
@@ -281,3 +282,16 @@ func TestDeletePipeline(t *testing.T) {
...
 	assert.EqualValues(t, 1, count)
 }
+
+func TestIsUniqueConstraintError(t *testing.T) {
+	// Without this, the retry in CreatePipeline is dead code (#6067).
+	assert.True(t, isUniqueConstraintError(types.ErrInsertDuplicateDetected))
+
+	// Raw driver errors, for callers that don't go through wrapInsert.
+	assert.True(t, isUniqueConstraintError(errors.New(`pq: duplicate key value violates unique constraint "UQE_pipelines_s"`)))
+	assert.True(t, isUniqueConstraintError(errors.New("UNIQUE constraint failed: pipelines.repo_id")))
```
6543: can we move that also to wrapInsert logic ... and complete the tests there

nicolasbatistoni:
Done — good call, this is better than what I sent.

There were **two** places that knew driver error strings, with **different sets**: `wrapInsert` had 3 patterns (`HasPrefix`), `isUniqueConstraintError` had 5 (`Contains`). My original patch just added a third check on top. Now:

- `wrapInsert` is the **only** place that knows driver wording, with the union of both sets.
- `isUniqueConstraintError` is **gone**; `CreatePipeline` uses `errors.Is(err, types.ErrInsertDuplicateDetected)` directly (both call sites).
- Tests **merged into the existing `TestWrapInsert`** rather than a new one: the store there only exercises the sqlite wording, so the postgres/mysql patterns — which can't be reproduced against a real sqlite store — are now covered right next to it.

Net **-39 / +23** lines: the fix ends up removing more than it adds.

`gofmt` clean, `go vet` clean, `go test ./server/store/datastore/` passing.

6543 (approved):

lafriks (approved):

6543:
@nicolasbatistoni thanks!

PR #6874, opened 2026-07-20
feat: allow configuring allowed orgs per forge

### LINKED ISSUE #6852: Support setting allowed orgs for each forge separately

Our woodpecker instance is planned to be connected to at least two forges, on both we want to cherry-pick which organizations are allowed to use it.

Currently we can only set a global list of allowed orgs (`WOODPECKER_ORGS`), which is [applied to all forges](https://github.com/woodpecker-ci/woodpecker/blob/fce6fb820c28cbfbe2833fcc266eda77e89a5367/server/api/login.go#L140). Users can't be expected to squat all allowed organizations' names on every forge, so the current global org setting could be considered a vulnerability - a malicious actor could create an organization on a second connected forge and use the allowed name for it to gain access.

### Suggested solution

Have a configurable list of allowed orgs for each forge.
Do not apply `WOODPECKER_ORGS` to secondary forges.

### Validations

gerbsen:
`WOODPECKER_ORGS` applies to every connected forge. Since organization names are only unique within a single forge, someone could create an organization with an approved name on a secondary forge to gain access.

Add an `orgs` field to the forge model, editable per forge in the admin UI. When set, it replaces the global `WOODPECKER_ORGS` list for logins using that forge; forges without their own list keep using the global one, so existing setups are unaffected.

Closes #6852


gerbsen:
Hey @qwerty287 I've now rebased this feature branch to latest main. I also tested this now locally with the following test-cases in addition to the go-tests. I've also added some general docs regarding multi-forge setup since I thought this might be helpful to others as well. I also ran until a bug where the login never finished and noticed a [missing pagination break](https://github.com/woodpecker-ci/woodpecker/pull/6874/changes#diff-2ccf2dbcd2765a72f3032d57a18b01c7e767b5db730001948969e7f75aa1ad4dR168) which probably makes problems with rate limits at GL/GH. Please let me know if this works for you or where you see problems/change requests.

| #   | `WOODPECKER_ORGS` | GitHub forge `orgs` | GitLab forge `orgs`   | Log in via | Expect                                               | ✔   |
| --- | ----------------- | ------------------- | --------------------- | ---------- | ---------------------------------------------------- | --- |
| 1   | `my-gh-org`       | *(empty)*           | *(empty)*             | GitHub     | ✅ allowed — global fallback                         | ✔   |
| 2   | `org-i-am-not-in` | *(empty)*           | *(empty)*             | GitHub     | ❌ `org_access_denied` — fallback denies             | ✔   |
| 3   | `my-gh-org`       | `woodpecker-test`   | *(empty)*             | GitHub     | ✅ allowed — forge list grants on its own            | ✔   |
| 4   | **`my-gh-org`**   | **`my-gl-org`**     | *(empty)*             | GitHub     | ❌ **denied** — the key test                         | ✔   |
| 5   | `my-gh-org`       | *(empty)*           | `my-gl-org`           | GitLab     | ✅ allowed — GitLab uses its own list                | ✔   |
| 6   | `my-gh-org`       | *(empty)*           | *(empty)*             | GitLab     | ❌ denied — GitLab falls back to a *GitHub* org name | ✔   |
| 7   | `my-gh-org`       | *(empty)*           | `my/full/path/org`    | GitLab     | ✅ allowed — full path                               | ✔   |
| 8   | `my-gh-org`       | *(empty)*           | `my-missing-path-org` | GitLab     | ❌ denied — path segment alone doesn't match         | ✔   |

qwerty287:
Thanks!

Haven't checked the code in detail, but 2 points:
1. we have some kind of design for the UI to configure such lists. E.g. images list for secrets. You can find that here: https://github.com/woodpecker-ci/woodpecker/blob/main/web/src/components/secrets/SecretEdit.vue#L24-L42 (it would be good to refactor that into its own component, we use that at multiple places…)
2. I personally would say what you configure in `WOODPECKER_ORGS` should always be applied to all forges, and what you configure for the forges themselves is allowed additionally. I see that this would be less flexible, but otherwise I would rename the env var to something like `WOODPECKER_DEFAULT_ORGS`, because that's what it's basically doing then - it's the default if none are added.

gerbsen:
> Thanks!
> 
> Haven't checked the code in detail, but 2 points:
> 
> 1. we have some kind of design for the UI to configure such lists. E.g. images list for secrets. You can find that here: https://github.com/woodpecker-ci/woodpecker/blob/main/web/src/components/secrets/SecretEdit.vue#L24-L42 (it would be good to refactor that into its own component, we use that at multiple places…)

I don't understand this comment here. Could you explain this more? The ORGs can be configured in the [UI here](https://github.com/woodpecker-ci/woodpecker/pull/6874/changes#diff-d6b1595c06c11b81ea3778e8c1ef9e793c779a47abf4890a2f892939caa08c84): 

Screenshot 2026-08-06 at 09 52 50

> 2. I personally would say what you configure in `WOODPECKER_ORGS` should always be applied to all forges, and what you configure for the forges themselves is allowed additionally. 
> I see that this would be less flexible, but otherwise I would rename the env var to something like `WOODPECKER_DEFAULT_ORGS`, because that's what it's basically doing then - it's the default if none are added.

okay, that makes sense to me. If you want to avoid the issue addressed in the PR you can simple not set the value. I will apply this today hopefully. 

Have you seen the error from the pipeline before for `make fuzz`? It looks like some random error which I can't reproduce on my local machine.

qwerty287:
> I don't understand this comment here. Could you explain this more? The ORGs can be configured in the [UI here](https://github.com/woodpecker-ci/woodpecker/pull/6874/changes#diff-d6b1595c06c11b81ea3778e8c1ef9e793c779a47abf4890a2f892939caa08c84):

Yes, but right now we have a single text field with comma-separated value.

To configure simple lists in the UI, we are using a more "verbose" UI, for example (that's the code I linked above) when configuring the list of images usable by a secret:
grafik

When configuring the allowed orgs for a forge, there should be a similar UI I think.

> Have you seen the error from the pipeline before for make fuzz? It looks like some random error which I can't reproduce on my local machine.

Yes, looks like some timeout. Just ignore it I would say...

gerbsen:
hey @qwerty287 not 100% sure how I managed to add a commit with 0 changes through the GH UI, but I guess you squash together everything anyways right? Would be great if you could have a look. IF there is still something left, just tell me. :) In any case, have a great weekend. I will probably not have the time to do anything WP related. 🖖

docs/docs/30-administration/10-configuration/12-forges/40-gitlab.md
```diff
@@ -21,6 +21,14 @@ Please use `http://woodpecker.mycompany.com/authorize` as the Authorization call
 
 If you run the Woodpecker CI server on a private IP (RFC1918) or use a non standard TLD (e.g. `.local`, `.intern`) with your GitLab instance, you might also need to allow local connections in GitLab, otherwise API requests will fail. In GitLab, navigate to the Admin dashboard, then go to `Settings > Network > Outbound requests` and enable `Allow requests to the local network from web hooks and services`.
 
+## Allowed organizations
+
+If you restrict logins using [`WOODPECKER_ORGS`](../10-server.md#orgs) or the [allowed organizations of a forge](11-overview.md#restricting-who-can-log-in), GitLab groups are matched by their [`full_path`](https://docs.gitlab.com/api/groups/) and not by their display name. For a subgroup you have to use the complete path, e.g. `my-group/my-subgroup`; the name of a single path segment does not match.
```
qwerty287, line 26: It shouldn't be the "display name" for all forges, but the "username" (the one that's in the URL - this is unique)

```suggestion
If you restrict logins using [`WOODPECKER_ORGS`](../10-server.md#orgs) or the [allowed organizations of a forge](11-overview.md#restricting-who-can-log-in), GitLab groups are matched by their [`full_path`](https://docs.gitlab.com/api/groups/) and not by their name. For a subgroup you have to use the complete path, e.g. `my-group/my-subgroup`; the name of a single path segment does not match.
```

server/api/login.go
```diff
@@ -322,6 +335,17 @@ func HandleAuth(c *gin.Context) {
 	c.Redirect(http.StatusSeeOther, server.Config.Server.RootPath+"/")
 }
 
+// allowedOrgsForForge returns the organizations a user has to be a member of to
+// be allowed to log in using the given forge. The global WOODPECKER_ORGS setting
+// applies to every forge, orgs configured on a forge itself are allowed in
+// addition to it.
+func allowedOrgsForForge(forgeModel *model.Forge) *permissions.Orgs {
```
qwerty287, line 342: As you can just call the `With()` func with an empty slice, I would just remove that function and always use `server.Config.Permissions.Orgs.With(forgeModel.Orgs)` in line 148

web/src/assets/locales/en.json
```diff
@@ -613,6 +613,9 @@
   "save": "Save",
   "add": "Add",
   "delete": "Delete",
+  "allowed_orgs": "Allowed organizations",
+  "allowed_orgs_desc": "Organizations whose members may log in using this forge, in addition to the global WOODPECKER_ORGS setting.",
+  "allowed_orgs_placeholder": "org-1",
```
qwerty287, line 618: Would put a description here
```suggestion
  "allowed_orgs_placeholder": "Organization name",
```

docs/docs/30-administration/10-configuration/12-forges/11-overview.md
```diff
@@ -16,3 +16,26 @@
...
 
 In addition to this, Woodpecker supports [addon forges](../100-addons.md) if the forge you are using does not meet the [Woodpecker requirements](../../../92-development/02-core-ideas.md#forges) or your setup is too specific to be included in the Woodpecker core.
+
+## Multiple forges
+
+:::danger
+Support for connecting multiple forges is not finished yet and has to be considered experimental. Please do not use it in public or otherwise untrusted environments. See [known limitations](#known-limitations) below.
+:::
+
+Only **one** forge can be configured using environment variables, for example [GitHub](20-github.md) or [GitLab](40-gitlab.md). Enabling several forge drivers at once does not create several forges. All other forges have to be added by an admin under `Settings` -> `Forges` and exist in the database only. The forge from the environment is written back on every server start, so changing its driver replaces it instead of adding a new one.
```
qwerty287, line 26: No examples necessary here.
```suggestion
Only **one** forge can be configured using environment variables. Enabling several forge drivers at once does not create several forges. All other forges have to be added by an admin under `Settings` -> `Forges` and exist in the database only. The forge from the environment is written back on every server start, so changing its driver replaces it instead of adding a new one.
```

docs/docs/30-administration/10-configuration/12-forges/11-overview.md
```diff
@@ -16,3 +16,26 @@
...
+
+As organization names are only unique within a single forge, someone could create an organization with a name from `WOODPECKER_ORGS` on another connected forge to gain access. If you connected more than one forge, keep `WOODPECKER_ORGS` empty and configure the allowed organizations of each forge instead, including the one from the environment.
+
+The forge configured through environment variables has no own environment setting for this, its list is edited in the admin UI like the one of any other forge. Only the options listed in its docs are written back on server start, the allowed organizations are kept.
+
+Use the organization name, for GitLab the group's [`full_path`](https://docs.gitlab.com/api/groups/) like `my-group/my-subgroup`. Matching is case-insensitive and only happens while logging in. Note that a forge can hide organizations from Woodpecker although the user is a member of them, see GitHub and [GitLab](40-gitlab.md#allowed-organizations).
```
qwerty287, line 36: > Note that a forge can hide organizations from Woodpecker although the user is a member of them, see GitHub and [GitLab](40-gitlab.md#allowed-organizations).

This sentence lacks information. There's no info about this on the Gitlab page, and no link for Github at all.

qwerty287 (commented):
Looks good now, just some small nits, mostly about docs.

Yes, commits are squashed.

docs/docs/30-administration/10-configuration/12-forges/11-overview.md
gerbsen: I meant here that you need to be at least developer to login. But I think it's better in general to remove this sentence.

server/api/login.go
gerbsen: I'm always not sure who's supposed to mark these threads as closed. Leaving them open for now.

qwerty287: It's fine if you just close them.

qwerty287:
Could you run `make generate-openapi`? https://ci.woodpecker-ci.org/repos/3780/pipeline/36751/37

gerbsen:
I'm currently looking into: https://ci.woodpecker-ci.org/repos/3780/pipeline/36753/23

qwerty287:
That's not related to your code

(It's a security vulnerability in one of our docs dependencies, so no problem, that's likely just a build-time dependency for docs, but the `image-size` lib isn't maintained anymore… I hate nodejs ecosystem :laughing: )

gerbsen:
> That's not related to your code

I know, but I still wanna fix it. :)

qwerty287:
You can't I think, the `image-size` library isn't maintained anymore (https://github.com/image-size/image-size)… That's just node.js ecosystem right…

You would have to check which of our docs dependencies is using `image-size` and then you have to replace it in those package I think

qwerty287 (approved):

gerbsen:
🎉🙏

Issue #6875, opened 2026-07-20
Virtualize pipeline log viewer and add full-log search


6543:
The pipeline log viewer renders every displayed log line inside one large CSS Grid.

Firefox clamps very large implicit grids at roughly 10,000 rows. Lines beyond that boundary can overlap or render in the wrong location. Chromium currently renders the same log correctly.

The configurable `WOODPECKER_MAX_PIPELINE_LOG_LINE_COUNT` only limits how often the problem appears. Increasing it exposes the Firefox failure again. It also does not prevent the complete log from being downloaded, decoded, and processed by the browser.

## Expected behavior

- Large logs render correctly in Firefox, Chromium, and other supported browsers.
- The number of mounted DOM elements remains bounded, independent of the number of loaded log lines.
- All currently loaded log lines remain searchable.
- Pressing Ctrl+F or Cmd+F while using the log viewer opens a log-specific search.
- Search covers all loaded lines, not only lines currently mounted or visible.
- Selecting a search result scrolls to and highlights its line.
- Search results inside collapsed command groups remain reachable.
- Existing log behavior remains available:
  - ANSI colors
  - wrapped lines
  - timestamps
  - command grouping and collapsing
  - line links such as `#L9987`
  - live-log auto-scroll
  - complete log download

## Proposed solution

Replace the current unbounded log layout with a virtualized log viewer that only renders the visible portion of the log.

Keep the complete loaded log model separate from the rendered DOM so searching, line navigation, grouping, and downloads do not depend on which lines are currently visible.

Add an integrated log search interface that handles Ctrl+F and Cmd+F while the log viewer is active.

Backend pagination can be added separately to avoid initially transferring and processing complete logs. It should complement the virtualized viewer rather than block the frontend fix.

## Acceptance criteria

- A log containing at least 50,000 lines renders without overlapping lines in Firefox and Chromium.
- The mounted log-line DOM count stays bounded while scrolling through the complete log.
- Ctrl+F and Cmd+F search every currently loaded line.
- Search supports next result, previous result, match count, and closing with Escape.
- Navigating to a result mounts and scrolls to the corresponding line.
- Navigating to `#L<number>` works even when that line was not mounted.
- Search or hash navigation expands the containing command group when necessary.
- Live logs follow new output only while the viewer is already at the bottom.
- Scrolling, resizing, wrapping, grouping, and collapsing do not cause significant position jumps.
- Full-log download remains independent of the mounted DOM.
[...]

PR #6876, opened 2026-07-20
Add pipeline log download endpoint

6543:
prepare for https://github.com/woodpecker-ci/woodpecker/issues/6875

followup todo: move logs related code into own file `logs.go`


server/api/pipeline.go
```diff
@@ -295,42 +297,102 @@ func GetPipelineLastByBranch(c *gin.Context) {
...
-	logs, err := server.Config.Services.LogStore.LogFind(step)
-	if err != nil {
-		handleDBError(c, err)
-		return
-	}
+	return pl, step, true
+}
 
-	c.JSON(http.StatusOK, logs)
+func safeDownloadFilename(filename string) string {
```
6543: do we have similar helpers?
or libs that has such?

server/api/pipeline_test.go
```diff
@@ -125,6 +126,47 @@ func TestGetPipelines(t *testing.T) {
...
+	logs := []*model.LogEntry{
+		{StepID: step.ID, Line: 0, Data: []byte("first line")},
+		{StepID: step.ID, Line: 1, Data: []byte("\x1b[31msecond line\x1b[0m")},
+	}
+
+	mockStore := store_mocks.NewMockStore(t)
+	mockStore.On("GetPipelineNumber", repo, int64(42)).Return(pipeline, nil)
+	mockStore.On("StepLoad", pipeline.ID, int64(3)).Return(step, nil)
+
+	mockLogStore := log_service_mocks.NewMockService(t)
```
6543:
```suggestion
	mockLogStore := log_mocks.NewMockService(t)
```

server/api/pipeline_test.go
```diff
@@ -35,6 +35,7 @@ import (
 	queue_mocks "go.woodpecker-ci.org/woodpecker/v3/server/queue/mocks"
 	"go.woodpecker-ci.org/woodpecker/v3/server/scheduler"
 	config_service_mocks "go.woodpecker-ci.org/woodpecker/v3/server/services/config/mocks"
+	log_service_mocks "go.woodpecker-ci.org/woodpecker/v3/server/services/log/mocks"
```
6543:
```suggestion
	log_mocks "go.woodpecker-ci.org/woodpecker/v3/server/services/log/mocks"
```

server/api/pipeline.go
```diff
@@ -295,42 +297,104 @@ func GetPipelineLastByBranch(c *gin.Context) {
...
+			}
+		}
+		if _, err = c.Writer.Write(entry.Data); err != nil {
+			_ = c.Error(err)
+			return
+		}
+	}
+}
+
+func pipelineStepFromParams(c *gin.Context) (*model.Pipeline, *model.Step, bool) {
```
qwerty287: This is likely used in many more api endpoints. Would fit better to the middleware like `session.Repo()` in my opinion

web/src/components/repo/pipeline/PipelineLog.vue
```diff
@@ -439,21 +439,17 @@ async function download() {
...
     return;
   } finally {
     downloadInProgress.value = false;
   }
-  const fileURL = window.URL.createObjectURL(
-    new Blob([logs.map((line) => decode(line.data ?? '')).join('\n')], {
-      type: 'text/plain',
-    }),
-  );
+  const fileURL = window.URL.createObjectURL(logs);
```
qwerty287, line 452: Just wondering, is it possible to just put a link to the download endpoint here with `<a href="…/download">`? Instead of downloading the logs in JS? But I don't know whether this would work… But it would be cleaner I think

6543: If we do that a log you download would always have the same name: `download`

if we restructure the api and parse params manually by file name we can do that. but this would cluster our api backend code

qwerty287: According to https://stackoverflow.com/a/13308094 you can set a filename in the header... Not sure if it works though

server/api/pipeline.go
6543: I was going to keep refactoring out of this pull ... but sure can do so

6543: done ...

6543:
:point_up: i started writing the patches myselfe but just of curiosity i let claude write it too ... as i was 50% finished i looked at the patch and it was 1:1 the same code :laughing: so i just applyed that ones ;d

web/src/components/repo/pipeline/PipelineLog.vue
6543: @qwerty287 we already set this header and let the browser go to the download will create the file as of now - so i was wrong i thought it was not set jet ...

anyway - changing the UI to respect the header is either: do a redirect to the url witch will unmount ui parts ... or have another tab opened ... or parse the header in our webui - witch adds a lot of more complex code 

so as far as i can tell this is the simplest solution

qwerty287: @6543 I tested this and it is working as expected for me - opening the API download URL saves the logs with the correct filename (e.g. `qwerty287-woodpecker-149-clone.log`).

It also works if I modify the web UI to directly use the URL and remove the TS logic - try the diff: [href.diff](https://github.com/user-attachments/files/30613191/href.diff.txt)

6543: well yours is better :D

web/src/lib/api/client.ts
```diff
@@ -79,6 +89,10 @@ export default class ApiClient {
     return this._request('GET', path);
   }
 
+  async _getBlob(path: string): Promise<Blob> {
```
qwerty287: We don't need that anymore, right?

web/src/lib/api/index.ts
```diff
@@ -159,6 +159,10 @@ export default class WoodpeckerClient extends ApiClient {
     return this._get(`/api/repos/${repoId}/logs/${pipeline}/${step}`) as Promise<PipelineLog[]>;
   }
 
+  async downloadLogs(repoId: number, pipeline: number, step: number): Promise<Blob> {
```
qwerty287: same here

qwerty287 (approved):
besides the leftovers lgtm (I also tested this, works)

and the pipelines fail.

PR #6877, opened 2026-07-20
CLI exec sanitize secrets too

6543:
the `cli exec` should mimic server-agent execution, and share as much code as possible ...

... this addresses secrets masking and refactor it a bit to make it more modular


qwerty287:
This looks good to me - but do we really need a separate `SanitizeWriter`? Can't we just directly have `SecretReplaceWriter`? Yes, I know that's less modular, but it's very unlikely we need to sanitize something else, right?

6543:
I was looking into getting rid of sendingval secrets to agents to sanitize things

6543:
still if you you want to i'll remove it and stick those changes into anotherbpull

qwerty287:
The comment was not about the structure of the PR, but the code only.

You add a `SanitizeWriter` that is built to be modular - but we don't need that. We can just build a `SecretsReplaceWriter` with a fixed sanitizing function (if you don't get what I mean, I can send you a diff)

Issue #6277, opened 2026-03-20
Rich input types for manual pipeline parameters


Fedezzab:
### Clear and concise description of the problem

When triggering a pipeline manually via the UI, all parameters are presented as free-text input fields. Users must type values from memory, which is error-prone and inconvenient — especially when the valid options are a known, finite set (e.g. build variants, configurations, feature flags).

Other CI tools like GitLab CI and Jenkins allow defining parameters with predefined selectable values, making the manual trigger UI much more user-friendly.

### Suggested solution

Allow defining parameter metadata so that the "Run pipeline" UI can render appropriate input widgets instead of plain text fields. Specifically:

- **Dropdown menus** for parameters with a predefined list of values
- **Checkboxes** for boolean parameters (e.g. enable/disable a flag)
- **Free-text fallback** for parameters that may accept custom values alongside predefined ones

These definitions should be supported in **two places**:

1. **In the `.woodpecker.yml` pipeline file** — so parameter definitions live with the code and are version-controlled
2. **In the Woodpecker UI / repo settings** — for cases where parameters need to be managed without code changes

### Validations

- [x] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/versions]
- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

qwerty287:
There's an open (but mostly stale) PR to fix this: https://github.com/woodpecker-ci/woodpecker/pull/5632

qwerty287:
And I found a duplicate... https://github.com/woodpecker-ci/woodpecker/issues/3326

PR #6878, opened 2026-07-21
feat: add console-defined typed parameters for manual pipeline runs

### LINKED ISSUE #3326: Templates for manual workflows

simonmeyerrr:
## Summary

Adds typed, predefined **parameters** for manual pipeline runs. Repo admins can define parameters in the repo settings UI; when a pipeline is triggered manually, the run form renders proper input widgets (dropdown, checkbox, number, text) pre-filled with defaults instead of only a free-text key/value box. Chosen values are passed to the pipeline as environment variables.

Closes #3326
(also addresses the request in #6277, which was closed as a duplicate of #3326)

## Motivation

Typing manual-run variables from memory is error-prone, especially when valid values are a known set (build variants, deploy targets, feature flags). Other CI tools (GitLab, Jenkins, GitHub Actions `workflow_dispatch`) let you predefine selectable parameters.

## Design decisions

- **Console/DB-defined, per-repo (for now).** Parameters are stored in the DB and managed in repo settings. This intentionally avoids having to pre-fetch the forge and parse workflow YAML before the run form can render (a concern raised on the earlier attempts #4861 and #5632) and it also works for repos whose config is produced by the external HTTP config service, where no static `.woodpecker.yml` exists.
- **No changes to the execution layer.** Parameter values are injected through the existing `Pipeline.AdditionalVariables` → env var path. Definitions are just typed metadata + validation.
- **Backward compatible.** With no parameters defined, the manual run form behaves exactly as before (free-text variables only). Ad-hoc variables remain supported alongside parameters.
- **Server-side validation.** Required parameters and choice options are enforced on pipeline creation, not just in the UI; missing values fall back to defaults.
- **Forward-compatible.** A `source` field (`repo_config` today) and a `ref`-aware list endpoint leave room for a future "parameters defined in the workflow YAML" source without reworking the data model, API, or run form. Intended precedence: workflow-defined definitions authoritative for type/options, repo-config as defaults/fallback.

## What's included

- Model + migration + store (CRUD) with tests
- REST API `/api/repos/{repo_id}/parameters` (list/get/create/update/delete), gated by push perms
- Repo settings UI to manage parameters
- Manual run form renders typed widgets, keeps the free-text editor, and supports URL query prefill (e.g. `?branch=main&DEPLOY_TARGET=production`)
- Parameter types: string, number, boolean, choice

## Testing

I manually tested the different kind of parameters by setting up a local environment with gitea and dummy workflows:
image
image
image
image


simonmeyerrr:
(the force push was about removing a commit where I tried to fix the security issue in pnpm, but ended up removing it as it's an existing and unrelated issue on main)

qwerty287:
Thanks for the PR. Please expect some time till someone is able to review this. Is this built upon https://github.com/woodpecker-ci/woodpecker/pull/5632?

simonmeyerrr:
> Thanks for the PR. Please expect some time till someone is able to review this. Is this built upon #5632?

Thanks! It's a fresh implementation from scratch that follows the same idea discussed there and in #3326 though: typed parameters feed into the existing additional pipeline variables for manual runs, so the execution layer stays untouched.

PR #6879, opened 2026-07-21
docs: document the scope of built-in environment variables

### LINKED ISSUE #6795: Document and Define scope of built-in environment variables
add to the list if they can be used ad config evaluation time and/or at runtime

youdie006:
## What

Adds a **Scope** column to the built-in environment variables table (`docs/docs/20-usage/50-environment.md`), marking each variable as usable at config-evaluation time and/or at runtime, per #6795. A short note above the table explains the two scopes.

## How each scope was derived (from code, not guessed)

- **`config`** = the variable is present in the map returned by `(*Metadata).Environ()` (`pipeline/frontend/metadata/environment.go`). That map is what feeds `${...}` substitution and `when` evaluation (`pipeline/frontend/builder/builder.go`, `compiler` seeds `c.env` from it).
- **`runtime`** = the variable is present in the step container's environment when it executes (`compiler/convert.go` copies `c.env` into each step; the agent adds a few more in `pipeline/runtime/step.go`).

The split hinges on the server building the compile-time metadata with an **empty `Step`** (`server/pipeline/metadata/metadata.go`), so step-scoped variables are absent at config-eval time and only supplied by the agent at runtime. Most variables are `config, runtime`; the runtime-only ones are the step-specific variables and `CI_PIPELINE_STATUS` (recomputed per step, whereas `CI_PREV_PIPELINE_STATUS` comes from `Environ()` and is `config, runtime`).

## One thing worth your eyes (possible separate bug)

I labeled `CI_STEP_NUMBER` and `CI_STEP_URL` as `runtime` (they are step-scoped, so definitely not `config`). But I could not find code that actually populates them at runtime: they are only produced by `Environ()` when `Step.Number` is set, yet the server constructs metadata with an empty `Step`, and the agent's `setStepEnv` (`pipeline/runtime/step.go`) sets only `CI_STEP_NAME` / `_TYPE` / `_STARTED` — not `_NUMBER` / `_URL`. So their `runtime` label reflects documented intent; their real availability looks like a latent gap. That seems like a separate issue rather than a docs one — flagging it here rather than silently documenting them as working.

## Format

The issue didn't specify a presentation, so I went with a `Scope` column (values `config, runtime` / `runtime`). Happy to switch to inline notes or a different format if you'd prefer.

## Verification

Prettier (the docs format gate) passes; `markdownlint-cli2` with the repo's `.markdownlint.yaml` reports 0 issues; the table stays 4 columns across all rows.

Closes #6795.

---

*AI disclosure: this PR was drafted with Claude Code (AI-assisted). Each variable's scope was grounded in the Go sources cited above; verified with prettier + markdownlint.*


6543:
# Stats

- 99 table rows: 88 variables and 11 section headings
- No descriptions or examples changed
- The only table-content change is the new `Scope` column
- 77 variables are marked `config, runtime`
- 11 variables are marked `runtime`
- Seven explanatory lines were added above the table

## Variables marked runtime-only

- `CI_PIPELINE_STATUS`
- `CI_STEP_NAME`
- `CI_STEP_TYPE`
- `CI_STEP_NUMBER`
- `CI_STEP_STARTED`
- `CI_STEP_URL`
- `CI_WORKSPACE`
- `CI_SCRIPT`
- `CI_NETRC_USERNAME`
- `CI_NETRC_PASSWORD`
- `CI_NETRC_MACHINE`

6543 (approved):
I think a simple table is UX whies not as nice anymore ... but this is for now ok

PR #6880, opened 2026-07-21
Accept registry addresses with ports

### LINKED ISSUE #6849: Registry credentials for `host:port` registries can never match: UI validation forces a scheme, credential matching requires the bare address
### Component

web-ui, server

### Describe the bug

For a private registry addressed as bare `host:port` (e.g. an insecure/plain-HTTP registry on a LAN, `10.0.1.32:5000`), it is impossible to create a working pull credential through the UI:

1. Adding a registry with address `10.0.1.32:5000` is rejected by the UI/API with:
   `Error inserting registry. parse "10.0.1.32:5000": first path segment in URL cannot contain colon`
   (Go's `url.Parse` requires a scheme when the host contains a port.)
2. The only accepted form is `http://10.0.1.32:5000` — but registry-credential matching compares the stored address against the **image reference's registry domain**, which is always the bare `10.0.1.32:5000` (image refs never carry a scheme). The scheme-prefixed entry therefore never matches.

Result: the credential is silently never applied; pulls of step images from that registry go out anonymous, the registry returns 401, and the step fails.

The two behaviors are individually defensible but mutually exclusive — the validation only accepts a form the matcher can never use, so `host:port` registries end up with credentials that are dead on arrival, with no error pointing at the mismatch.

### Compounding observability issue
When the (anonymous) pull is rejected, the step fails after ~2–3 s with **no step logs at all**; the server log shows only:

```
{"level":"error","error":"stream: not found","message":"done: cannot close log stream for step NNN"}
```

Nothing surfaces "image pull failed: authentication required" to the user, which makes the root cause very hard to find. (In our case it was hidden further because the image was already in the agent's local cache for weeks — the credential's brokenness only surfaced on the first genuinely new tag.)

### Steps to reproduce

1. Run a private registry with basic auth on plain HTTP at `host:port` (e.g. zot/registry:2 with htpasswd), no anonymous pull.
2. In repo Settings → Registries, try to add address `10.0.1.32:5000` → rejected with the `url.Parse` error above.
3. Add it as `http://10.0.1.32:5000` (accepted).
4. Run a pipeline whose step uses `image: 10.0.1.32:5000/some/image:tag` (not already present on the agent).
5. Step fails in seconds with no logs; registry access log shows an unauthenticated pull.

### Expected behavior

Either accept the bare `host:port` form (it's the canonical form for image references), or normalize the stored address (strip the scheme) before matching, or match scheme-insensitively. Additionally, an image-pull failure should surface as a visible step/pipeline error message rather than a logless failed step.

### Workaround
Insert the row directly in the server database with a scheme-less address (`UPDATE`/`INSERT` on the `registries` table, e.g. `address = '10.0.1.32:5000'`) — after which the credential matches and pulls authenticate correctly. Verified working on v3.16.0.

### System Info

```shell
- Woodpecker server/agent v3.16.0, docker backend, single agent (linux/arm64)
- Registry: zot (OCI dist-spec 1.1.0), plain HTTP on a private network, htpasswd basic auth
```

### Additional context

_No response_

### Validations

AnayGarodia:
## Summary

Registry credentials are matched against the registry domain from an OCI image reference. This changes registry-address validation to use the same `distribution/reference` grammar, so canonical bare addresses such as `10.0.1.32:5000` are accepted while schemes, paths, credentials, and malformed ports are rejected.

The legacy `index.docker.io` spelling remains accepted and continues to match Docker Hub. Regression coverage includes domain, IPv4, IPv6, and localhost addresses with ports, plus an explicit matcher case.

The separate log-stream warning described in the issue is not changed here; current `main` already carries step-start errors through the runtime tracer, while this PR fixes the credential mismatch itself.

## Testing

- `go test -race -timeout 30s ./server/model ./pipeline/frontend/yaml/utils`
- `go test -race -tags test -timeout 60s ./server/api`
- `golangci-lint run` (with the generated web embed path present)
- `git diff --check`

Fixes #6849


server/model/registry.go
```diff
@@ -67,8 +68,18 @@ func (r *Registry) Validate() error {
 		return errRegistryPasswordInvalid
 	}
 
-	_, err := url.Parse(r.Address)
-	return err
+	address := r.Address
+	if address == "index.docker.io" {
```
6543: why is this needed ?

6543:
also the code reads more like a workaround ... :thinking: I'm not completly against it. but before ack such code we should make sure it is maintainable and this looks just like some code compy from stackoverflow with no reason why - through my feelings can be wrong

server/model/registry.go
AnayGarodia: It was a workaround for the OCI parser requiring a repository path before it will expose a registry domain. I removed it in 4e8a20a6: validation now anchors the exported `reference.DomainRegexp` and explicitly rejects single-label names (except `localhost`) because those are interpreted as Docker Hub repository paths, not registry domains.

AnayGarodia:
Refactored the validation in 4e8a20a6 to address the maintainability concern. The dummy repository path and Docker Hub rewrite are gone; the code now directly uses the OCI library’s documented domain[:port] grammar, anchors it to the whole input, and documents the one extra image-reference rule for bare single-label names. Added coverage for `localhost` and the single-label rejection. The branch is also current with `main`; `go test -race -timeout 30s ./server/model ./pipeline/frontend/yaml/utils` passes.

AnayGarodia:
Closing this — the workaround framing (and the bot flag on the account) isn't a good foundation to build trust on in this repo. Appreciate the review time.

PR #6887, opened 2026-07-22
feat(backend): add Tenki sandbox execution backend

michelfortes-luxor:
## What

Adds a new `tenki` execution backend. Each workflow runs inside a [Tenki](https://tenki.cloud) sandbox (an ephemeral cloud microVM): one sandbox is created per workflow and each step runs as a command inside it, so steps share the workspace filesystem — mirroring the Docker backend's shared volume.

## Why

Running each pipeline step in an isolated microVM gives stronger isolation for untrusted PR pipelines than a shared Docker daemon, while keeping the agent lightweight. The backend uses only stable sandbox primitives (create / exec / destroy).

## How it works

- Implements the `Backend` interface in `pipeline/backend/tenki/`, wired into the agent and `woodpecker exec`.
- One sandbox per workflow (`SetupWorkflow`), one exec per step (`StartStep`); steps share the sandbox filesystem.
- Step scripts are built with the shared `common.GenerateContainerConf` (run under `/bin/sh -e` with per-command tracing).
- `step.Image` is ignored — steps run on Tenki's standard base image (no templates/snapshots).
- Configured via `WOODPECKER_BACKEND_TENKI_*` (API key, endpoint, project/workspace, allow-outbound, timeouts); project/workspace auto-resolve from the API key identity.

## Limitations (initial version)

Image-based plugins, services, and the default image-based clone step are not supported yet — set `skip_clone: true` and clone with explicit `git` commands (documented on the backend page).

## Testing

- Unit tests including lifecycle paths (create-failure cleanup, cancellation, kill, teardown failure, exit-status mapping), run with `-race`; `golangci-lint` clean.
- Validated end to end via `woodpecker exec --backend-engine tenki`: single/multi-step pipelines (shared workspace), exit-code propagation, and clean teardown with no orphan sandboxes.

## Docs

New backend page (`docs/.../11-backends/40-tenki.md`), agent env-var reference, and supported-platforms updates.


michelfortes-luxor:
A couple of CI notes for reviewers:

- **`static`** was failing on `lint-editorconfig` — my `ensureWorkspaceDir` shell-script literal had space-indented continuation lines, which the checker flags in a `.go` file. Fixed in 556d95a by inlining the script into a single `||` expression; behavior is unchanged.
- **`securityscan`** is red because Trivy reports pre-existing CVEs in `docs/pnpm-lock.yaml` (Docusaurus/npm deps: body-parser, brace-expansion, dompurify, fast-uri, fast-xml-parser, svgo). These are unrelated to this backend change, reproduce on `main`, and `securityscan` is not a required check.

The remaining required checks are just pending CI approval.

qwerty287:
Thanks, but I'm not sure if the main repo is the right place for this. I don't think there are many people using this, and probably no maintainer. I would suggest using https://woodpecker-ci.org/docs/administration/configuration/backends/custom

PR #6888, opened 2026-07-22
Document woodpecker-cli configuration files

### LINKED ISSUE #6885: Document woodpecker-cli configuration

Right now, the only documentation I am able to find regarding configuration files for woodpecker-cli is that https://woodpecker-ci.org/docs/cli lists an optional `--config`/`-c` flag for the “path to config file”.

I can’t find anywhere where it states what path(s) it looks at by default (ie. without `--config` flag), what format (JSON, YAML, TOML, INI-ish, …) it should be in, or what values can be set in it.

I assume `woodpecker setup` is supposed to create a configuration file, but that doesn’t help if you’re [not able to run `woodpecker setup`](https://github.com/woodpecker-ci/woodpecker/issues/5453).

There might also be other reasons to want to manually create a configuration file, over, say, using environment variables or other approaches to configure the CLI.

### Suggested solution

A page or section in the CLI documentation about the configuration file, containing information on default lookup/search path(s) (possibly also a note on inheritance, or lack thereof, in case a configuration file is found in multiple places), the format of the file, and what values can be set in it.

### Alternative

A `man 5` page describing the file or documentation of some other form.

### Additional context

_No response_

### Validations

AnayGarodia:
The generated CLI reference now explains:

- platform-specific context and legacy config paths
- both JSON formats and the values stored in each
- token storage in the OS keyring
- config, flag, and environment-variable precedence
- config-free use with `WOODPECKER_SERVER` and `WOODPECKER_TOKEN`

The text lives in the CLI documentation generator so regeneration preserves it.

Verification:
- `go generate ./cmd/cli`
- `go test ./cmd/cli ./cli/internal/config`
- `go vet ./cmd/cli ./cli/internal/config`

Closes #6885

### DIFF
--- a/cmd/cli/docs.go
+++ b/cmd/cli/docs.go
@@ -22,6 +22,35 @@ import (
 	docs "github.com/urfave/cli-docs/v3"
 )
 
+const configurationDocs = `# CONFIGURATION
+
+The CLI normally uses contexts created by ` + "`woodpecker-cli setup`" + `. Context metadata is JSON in ` + "`woodpecker/contexts.json`" + ` below the platform configuration directory:
+
+- Linux: ` + "`$XDG_CONFIG_HOME/woodpecker/contexts.json`" + ` (defaults to ` + "`~/.config/woodpecker/contexts.json`" + `)
+- macOS: ` + "`~/Library/Application Support/woodpecker/contexts.json`" + `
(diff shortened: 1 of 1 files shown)


6543:
ci is failing

AnayGarodia:
Closing this one too — not going to keep pushing PRs on this repo right now.

PR #6890, opened 2026-07-23
Use toLabelValue instead of toDNSName in Kubernetes step label

ivanvc:
Follow up on #6855 / #6827, as the `test` and `woodpecker-ci.org/step` labels still use `toDNSName()` instead of `toLabelValue()`.

cc @xoxys.


xoxys:
Thanks! I would suggest we drop the no-op wrapper `stepLabel()` at all? WDYT?

ivanvc:
Uhm, I just realized that removing that function and its tests removes the new test I added.

ivanvc:
Alright, I had a brainfart for a moment :joy:

Now it's using `toLabelValue`, and I included tests.

6543 (approved):

PR #6906, opened 2026-07-27
Fix pipeline queue default template data

### LINKED ISSUE #6905: CLI pipeline queue default format fails on Feed
## Component

cli

## Describe the bug

`woodpecker-cli pipeline queue` fails with the built-in default format when the server has queued or running pipelines. The default template references `.FullName`, but the command renders each item returned by `PipelineQueue()`, which is a `woodpecker.Feed` value without that field.

The command exits before printing the queue:

```text
{"level":"fatal","error":"template: _:1:8: executing \"_\" at <.FullName>: can't evaluate field FullName in type *woodpecker.Feed","time":"2026-07-27T22:10:44+04:00","message":"error running cli"}
```

An explicit format that only uses existing feed fields works:

```sh
woodpecker-cli pipeline queue --format '{{.Number}} {{.Status}}'
```

I intend to submit a PR with a regression test for the default template.

## Steps to reproduce

1. Connect `woodpecker-cli` to a server that has at least one pending or running pipeline.
2. Run:

```sh
woodpecker-cli pipeline queue
```

3. The command fails with the template error above.

## Expected behavior

The default `pipeline queue` command should render queued/running pipelines without requiring `--format`.

## System Info

```shell
Reported with the CLI update/version path showing 3.16.0 as available.
Source check: woodpecker-ci/woodpecker main at d3b39ff860186dba34810cd59bc7ed8023b2f0ee still has the default template mismatch.
Local source command: go run ./cmd/cli --version => woodpecker-cli version dev
```

## Additional context

The default template currently starts with `{{ .FullName }} #{{ .Number }}`, while `PipelineQueue()` decodes `/api/pipelines` into `[]*woodpecker.Feed`.

Komzpa:
## Summary

Fix `woodpecker-cli pipeline queue` so its default template has the data it expects. The command renders queue entries as `woodpecker.Feed`, but the default format starts with `.FullName`; the feed API and SDK now include `full_name` from the joined repository row.

Add a CLI regression test that renders the default queue template against a real `woodpecker.Feed`, and extend the datastore test to prove queued feed rows carry the repository full name.

Closes woodpecker-ci/woodpecker#6905

## Testing

```sh
go test ./cli/pipeline -run TestPipelineQueueDefaultTemplateRendersFeed -count=1
go test ./server/store/datastore -run TestGetPipelineQueue -count=1
make generate-openapi
make test-cli
make test-server-datastore
mkdir -p web/dist && printf 'test' > web/dist/index.html && PATH="$(go env GOPATH)/bin:$PATH" make lint
git diff --check
```

### DIFF
--- a/cli/pipeline/queue_test.go
+++ b/cli/pipeline/queue_test.go
@@ -1,0 +1,48 @@
+// Copyright 2026 Woodpecker Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pipeline
+
+import (
+	"bytes"
+	"testing"
+	"text/template"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.woodpecker-ci.org/woodpecker/v3/woodpecker-go/woodpecker"
+)
+
+func TestPipelineQueueDefaultTemplateRendersFeed(t *testing.T) {
+	tmpl, err := template.New("_").Parse(tmplPipelineQueue)
+	require.NoError(t, err)
+
+	var out bytes.Buffer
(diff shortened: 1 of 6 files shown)


6543:
first thought it would work without extending the feed api, but this is the best KISS solution

6543 (approved):

Komzpa:
> We noticed a mix of typical human activity and some patterns often associated with automation or AI agents.

Hi, I'm Darafei, team member of PostGIS. I'm currently torturing our install of woodpecker (http://woodie.osgeo.org/) trying to migrate "everything we had before" into it using AI (running windows tests under wine is fun!).

6543:
> running windows tests under wine is fun!

oh my ... well :D

PR #6907, opened 2026-07-27
feat: add AtomGit forge driver

jetsung:
## Add AtomGit forge driver

This pull request adds support for [AtomGit](https://atomgit.com) as a forge provider in Woodpecker CI.

### Changes
- **model**: new `ForgeTypeAtomGit` type and expose `OAuthHost` in `PublicCopy`
- **forge/setup**: register and configure the atomgit forge
- **services/setup**: `atomgit` flag with `https://api.atomgit.com` and OAuth host defaults
- **cmd/server/flags**: `WOODPECKER_ATOMGIT*` env vars and enable flag
- **api/repo**: lazily backfill `forge_url` and other fields for stale repos
- **server/forge/atomgit**: new driver package (auth, repo lookup, conversion, webhook parsing) with tests and fixtures
- **web**: add atomgit icon

### Notes
- AtomGit's API host (`api.atomgit.com`) returns correct clone URLs, while its OAuth endpoints live on the web host (`atomgit.com`); the driver points OAuth there unless overridden.
- Repos enabled before the forge populated `forge_url` are self-healed on first `GET /api/repos/{id}` request (forge contacted at most once).

### DIFF
--- a/cmd/server/flags.go
+++ b/cmd/server/flags.go
@@ -478,7 +478,7 @@ var flags = append([]cli.Flag{
 	&cli.StringFlag{
 		Name:    "forge-url",
 		Usage:   "url of the forge",
-		Sources: cli.EnvVars("WOODPECKER_FORGE_URL", "WOODPECKER_GITHUB_URL", "WOODPECKER_GITLAB_URL", "WOODPECKER_GITEA_URL", "WOODPECKER_FORGEJO_URL", "WOODPECKER_BITBUCKET_URL", "WOODPECKER_BITBUCKET_DC_URL"),
+		Sources: cli.EnvVars("WOODPECKER_FORGE_URL", "WOODPECKER_GITHUB_URL", "WOODPECKER_GITLAB_URL", "WOODPECKER_GITEA_URL", "WOODPECKER_FORGEJO_URL", "WOODPECKER_BITBUCKET_URL", "WOODPECKER_BITBUCKET_DC_URL", "WOODPECKER_ATOMGIT_URL"),
 	},
 	&cli.StringFlag{
 		Sources: cli.NewValueSourceChain(
@@ -490,6 +490,7 @@ var flags = append([]cli.Flag{
 				"WOODPECKER_FORGEJO_CLIENT_FILE",
 				"WOODPECKER_BITBUCKET_CLIENT_FILE",
 				"WOODPECKER_BITBUCKET_DC_CLIENT_ID_FILE",
+				"WOODPECKER_ATOMGIT_CLIENT_FILE",
 			)),
 			cli.EnvVar("WOODPECKER_FORGE_CLIENT"),
 			cli.EnvVar("WOODPECKER_GITHUB_CLIENT"),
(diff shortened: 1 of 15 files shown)


cmd/server/flags.go
```diff
@@ -478,7 +478,7 @@ var flags = append([]cli.Flag{
 	&cli.StringFlag{
 		Name:    "forge-url",
 		Usage:   "url of the forge",
-		Sources: cli.EnvVars("WOODPECKER_FORGE_URL", "WOODPECKER_GITHUB_URL", "WOODPECKER_GITLAB_URL", "WOODPECKER_GITEA_URL", "WOODPECKER_FORGEJO_URL", "WOODPECKER_BITBUCKET_URL", "WOODPECKER_BITBUCKET_DC_URL"),
+		Sources: cli.EnvVars("WOODPECKER_FORGE_URL", "WOODPECKER_GITHUB_URL", "WOODPECKER_GITLAB_URL", "WOODPECKER_GITEA_URL", "WOODPECKER_FORGEJO_URL", "WOODPECKER_BITBUCKET_URL", "WOODPECKER_BITBUCKET_DC_URL", "WOODPECKER_ATOMGIT_URL"),
```
6543, line 481: we should work with newline here too now ...

server/api/repo.go
```diff
@@ -379,7 +379,51 @@ func LookupRepo(c *gin.Context) {
 //	@Param		Authorization	header	string	true	"Insert your personal access token"	default(Bearer <personal access token>)
 //	@Param		repo_id			path	int		true	"the repository id"
 func GetRepo(c *gin.Context) {
-	c.JSON(http.StatusOK, session.Repo(c))
+	repo := session.Repo(c)
+
+	// Lazily backfill forge-derived fields (forge_url, pr_enabled, ...) that
+	// may be missing on repos enabled before the forge driver populated them.
```
6543, line 385: can we not do that and use the repair mechanism/1/

why would be a forge_url efen missing

server/forge/atomgit/atomgit.go
```diff
@@ -0,0 +1,653 @@
...
+		return nil, err
+	}
+	req.Header.Set("Authorization", "Bearer "+token)
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Content-Type", "application/json")
+	client := httputil.WrapClient(c.httpClient(), "forge-atomgit")
+	return client.Do(req)
+}
+
+func (c *AtomGit) get(ctx context.Context, token, apiURL string, out any) error {
```
6543, line 577: this is a full rest client embeded into a single file -> you should move it into a dedicated file

6543:
I'm still not sure if we should pull this driver into woodpecker core and not have it as addon on your side. if atomgit would be an opensource forge that would be all fine but it is a proletarian platform and not really relevant. even for china gitee would be a better call ... :thinking:

jetsung:
> I'm still not sure if we should pull this driver into woodpecker core and not have it as addon on your side. if atomgit would be an opensource forge that would be all fine but it is a proletarian platform and not really relevant. even for china gitee would be a better call ... 🤔

AtomGit is not an open-source forge. I don’t use Gitee for other reasons. Official AtomGit staff have said they plan to provide pipeline support as well, but I’ve already been waiting a long time for that feature and it still hasn’t launched. That’s why I implemented the functionality with Woodpecker. It’s fine if this PR isn’t merged — I can just use it on my own side.

6543:
@jetsung have you looked into https://woodpecker-ci.org/docs/administration/configuration/addons ?

6543:
adding you to the awesome list in our docs to your forge addon is what we can always do if you want to :)

Issue #6493, opened 2026-04-24
Yaml parser errors out on deeply nested pipeline yaml file


mmeier86:
### Component

server

### Describe the bug

When using a pipeline yaml file with deep nesting, Woodpecker shows a `max depth reached` error. Here is an example pipeline.yaml which fails reliably with this error:

```yaml
image: hello-world
when:
  - branch:
    - tester
steps:
  test:
    image: golang
    commands:
      - go install
      - go test
    backend_options:
      kubernetes:
        affinity:
          nodeAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              nodeSelectorTerms:
                - matchExpressions:
                    - key: accelerator
                      operator: In
                      values:
                        - nvidia-tesla-v100
```
The `backend_options` is taken directly from the [docs](https://woodpecker-ci.org/docs/administration/configuration/backends/kubernetes#example-node-affinity-for-gpu-workloads).

To me, it looks like the issue comes from the [xyaml lib](https://codeberg.org/6543/xyaml/src/branch/main/merge.go).

### Steps to reproduce

I've added a UT for the `ParseString` function: https://github.com/woodpecker-ci/woodpecker/blob/db70e65d21f8daa84808c6303d9608e4f72908bd/pipeline/frontend/yaml/parse.go#L34

See this commit in my fork: https://github.com/woodpecker-ci/woodpecker/commit/ef741be81eb51664b5ae0e26aaac4d9dc179b877

It runs the example Yaml I've shown above. The UT is executed when running `make test-lib` and fails with this error:
```
--- FAIL: TestParse (0.01s)
    --- FAIL: TestParse/Should_handle_deeply_nested_yaml (0.00s)
        parse_test.go:79: 
                Error Trace:    /home/michael/projects/woodpecker/pipeline/frontend/yaml/parse_test.go:79
                Error:          Received unexpected error:
                                max depth reached
                Test:           TestParse/Should_handle_deeply_nested_yaml
FAIL
coverage: 100.0% of statements
FAIL    go.woodpecker-ci.org/woodpecker/v3/pipeline/frontend/yaml       0.035s
```

### Expected behavior

I expect parsing of the example Yaml file to not fail.

### System Info

```shell
My version is 3.13.0:

{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.13.0"}
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
[...]

PR #6911, opened 2026-07-28
all yaml unmarshalling should use maxDepth

zmully:
- standardizes YAML unmarshalling so that `maxDepth` is used consistently when processing pipeline configurations

For v3.16.0, #6493 updated the xyaml version and increased the parsing depth of the yaml configuration, but my pipelines are still failing validation on my server with a `max depth reached` error if kubernetes node affinities in the pipelines are specified. 

I think this is because the frontend yaml parser was updated, but the `lintDeprections` and `lintBadHabits` validations were not, they're still using the xyaml default depth of 10. I took a go at running the CLI through Delve with my configuration, setting breakpoints on the two validations and the xyaml method and the maxDepth is never updated to `15` when these two linters are executed.

This PR introduces a shared yaml unmarshall for consistency in parsing. 

I'm not well versed in golang, so this PR was created with help from opencode/gpt-5.6. I didn't see any contributor guidelines about AI use, but want to disclaim its use here in case I missed it.


qwerty287:
Thanks! The package renaming to `woodpeckerYaml` doesn't comply with our naming standards. I would just import it as `yaml`

qwerty287:
There's an issue somewhere in the code https://ci.woodpecker-ci.org/repos/3780/pipeline/36397/31#L6

zmully:
Apologies, my local golang dev environment was hosed and prevented me from running the tests locally. The redeclaration should be corrected and tests should be passing now.

qwerty287:
Thanks, the linter still fails. https://ci.woodpecker-ci.org/repos/3780/pipeline/36404/31

Please remove the explicit `yaml` naming of the packages (shouldn't be necessary).

`goYaml` still isn't our convention. I would go with `go_yaml`, and you have to add that to `.golangci.yaml`

6543 (approved):

zmully:
@6543 thank you very much for shepherding this through, I much appreciate the opportunity to learn from it.

Issue #6064, opened 2026-02-04
Cross-file `depends_on` step references fail validation


Partysun:
### Component

cli, server

### Describe the bug

**Woodpecker Version**: v3.13.0 (server & agent)

**Issue**: Cross-file `depends_on` step references fail validation

**Description**: 
When using multiple workflow YAML files (.build.yaml, .test.yaml, .deploy.yaml), 
steps in one file cannot depend on steps in another file. Validation fails with:
"One or more of the specified dependencies do not exist"

**Reproduction**:
- .woodpecker/.build.yaml: step "build-ci"
- .woodpecker/.test.yaml: step "test-ci" with `depends_on: [build-ci]` ❌ fails
- .woodpecker/.deploy.yaml: step "deploy-ci" with `depends_on: [build-ci, test-ci]` ❌ fails

**Workaround**: 
Combine all steps into single .woodpecker.yaml file - works fine ✅

**Expected**: 
Steps should be able to depend on steps from other workflow files in .woodpecker/ directory

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.13.0"}
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

qwerty287:
> Steps should be able to depend on steps from other workflow files in .woodpecker/ directory

No, what you're observing is expected.

You can use `depends_on` at workflow level to let workflows depend on another one.

Partysun:
[docs](https://woodpecker-ci.org/docs/usage/workflows)

In docs I can see this:

.woodpecker/
├── build.yaml
├── deploy.yaml
├── lint.yaml
└── test.yaml

```yaml
.woodpecker/test.yaml
steps:
  - name: test
    image: debian:stable-slim
    commands:
      - echo testing
      - sleep 5

depends_on:
  - build
```

I have the same structure. Strange — maybe I'm doing something 'wrong,'... Anyway I've gotten my work done about workflows for our projects."

6543:
still if it is not clear we need better docs ...

qwerty287:
No @Partysun, this is exactly what I mean.

[...]

PR #6913, opened 2026-07-28
Allow step depends_on to reference other workflows

### LINKED ISSUE #2685: Prevent workflow depends_on dead-lock
In case two workflows depend on each another we currently get a dead-lock with tasks in queue that will never disappear again.

LaGrunge:
Currently the only way to synchronize workflows is workflow-level `depends_on`, which serializes entire workflows. In practice this is often too coarse: a long compile step could happily run in parallel with a dependency-building workflow, and only one or two later steps actually need its results. Today the usual workaround is polling an external store (S3, a registry, ...) from inside a step, plus some hand-rolled failure beacon so consumers don't burn their whole poll timeout when the producer fails. I ran into exactly this in my own pipelines, and #6064 shows other people expecting cross-file step dependencies to just work.

This PR lets a **step** depend on another workflow of the same pipeline, or on a single step of it:

```yaml
steps:
  - name: build            # runs in parallel with the 'auxiliaries' workflow
    image: golang
    commands: [go build]

  - name: pin-deps
    image: golang
    depends_on:
      - build              # regular local dependency
      - workflow: auxiliaries
        step: resolve-pins # optional: gate on a single step instead of the whole workflow
        optional: true     # optional: ignore if the target isn't part of the pipeline
    commands: [./use-pins.sh]
```

Only the declaring step is gated; everything else overlaps with the referenced workflow. Failure semantics mirror local step dependencies: if the target finishes with anything but success (including being skipped because its own deps failed), the gated step is skipped, downstream steps cascade, `when: status: failure` steps still run, and the workflow is marked failed with a descriptive error.

## How it works

- The yaml `depends_on` item grammar gains a `workflow` key (with optional `step`), reusing the existing `name`/`optional` object form. Workflow-level `depends_on` rejects it.
- The compiler keeps external deps out of the stage DAG and carries them as a new `wait_for` field on the backend step; only local deps determine stage placement.
- The pipeline builder — the only place that sees all workflows — validates targets: unknown or filtered-out required targets fail at pipeline creation, optional ones are dropped, matrix targets must contain the step in every instance. Self-references and cycles are rejected on the combined graph of workflow-level and step-level edges, projected to workflow→workflow. The projection is deliberately conservative to avoid agent deadlocks.
- At runtime the agent blocks the gated step (it stays `pending` in the UI) on a new `WaitWorkflow` unary RPC. The server resolves the target from the *calling* workflow's pipeline — so the existing ownership check is the entire authorization story — and polls the store until the target is terminal. Individual calls are capped at one minute; the agent client re-polls, and the regular `Extend` heartbeat keeps the lease alive. Store polling is used because queue state only covers currently-running tasks, while the target may not have started yet or may already be done.
- The gRPC protocol version is bumped, which also fences old agents off servers that produce configs with `wait_for`.
- `woodpecker-cli exec` ignores cross-workflow dependencies with a warning.

## Known limitation (documented)

A waiting step occupies an agent slot, so with very tight agent capacity the referenced workflow may not get scheduled until the wait times out via the workflow timeout. Waking waiters via pubsub instead of store polling would be a possible follow-up optimization.

## Testing

- Unit tests across all layers: yaml grammar and schema, linter, compiler, builder validation (incl. matrix and cycle cases), the server RPC (mocked store) and the runtime with a stub waiter.
- New e2e scenarios (fixtures 18–20): happy path with whole-workflow and per-step targets, failure cascade, missing optional target. Plus a timestamp-based e2e test proving the overlap property: the consumer's early steps run while the producer still works, the per-step gate opens exactly when its target step finishes, and the whole-workflow gate waits for everything.
- Docs added to the workflow-syntax and workflows pages.

This also rejects mutual and self `depends_on` cycles between workflows at pipeline
creation instead of enqueueing tasks that can never run (both scenarios are covered
by builder tests).

Closes #2685

Related: #6064, and discussion #5436 where moving `depends_on` to the step level
was the suggested workaround but fails validation — which is exactly the gap this
PR fills.


6543 (changes requested):
can we just not do that ... this is way to much complexity

LaGrunge:
> can we just not do that ... this dds way to much complexity

This is exactly what I need to simulate in my pipelines with extra fragile tools. Looks I'm not the only one here

qwerty287:
> Looks I'm not the only one here

Could you explain that? I've never seen somebody requesting this, and there's no open issue about it as well

LaGrunge:
> > Looks I'm not the only one here
> 
> Could you explain that? I've never seen somebody requesting this, and there's no open issue about it as well

I mentioned relevant discussion and the relevant issue in the description.

This feature is required for the pipelines, that work long time. 
I have the C++ code with heavy tests in different build configurations. The tests require the build of many dependencies that are common. So I have aux pipeline, building the common deps, and variety of pipelines building the C++ code differently. I cannot move tests out of C++ building pipelines, it depends on the state of worker and shared volume, so I need aux pipeline to work in parallel with regular ones and need the tests steps wait for this aux pipeline.

qwerty287:
The mentioned issue seems to be only misunderstanding, there wasn't a request in adding cross-file `depends_on` there.

PR #6933, opened 2026-07-31
Update frontend schema.json for documented backend options

zmully:
- Updates the frontend linter `schema.json` with all documented options
- Removes the `step_backend_kubernetes_service_account`, this seems to be a dead piece of schema, maybe it did reference the correct schema 3 years ago when introduced, but I cannot find any reference to in the current codebase, so I removed it
- Adds documentation links to schema descriptions where possible
- Adds a more fully specified test pipeline for a Kubernetes backed which assumes Kubernetes 1.36 (though I don't think this really matters). I wanted to make sure that fully specified nodeSelectors and affinities and such would pass the linter since I had issues with the linter in #6911 

I want to update the schema since we're using this schema internally to generate Typescript interfaces for our [projen](https://projen.io) based workflows. Right now we have to maintain a custom schema overlay because of these missing fields. 

This PR was created with assistance from opencode/gpt-5.6-terra.


6543 (approved):
looks good not tested

Issue #775, opened 2022-02-10
agents shuld picks up oldest jobs (by creation date)


6543:
just wait ... the job scheduler is a bit dumb atm ... (it just picks up a proc ..., we need to sort them - but that's another issue)

_Originally posted by @6543 in https://github.com/woodpecker-ci/woodpecker/issues/769#issuecomment-1034997824_

PR #6941, opened 2026-08-03
Add configurable queue priority rules

Komzpa:
## Summary

- add configurable server-side queue priority rules for queued tasks
- match rules on repo, event, branch, ref, author, sender, PR labels, event reason, and rerun count
- make the FIFO queue pick the highest-priority runnable task first, while preserving FIFO ordering for ties and concurrency reservations
- expose the persisted task priority in queue API/OpenAPI output
- read optional YAML rules from `.woodpecker/queue-priority.yml` on the repository's default branch
- reprioritize existing pending/dependency-waiting tasks for the same repo when the rules are loaded again

## Motivation

This is a small implementation path for https://github.com/woodpecker-ci/woodpecker/discussions/2261, which asks for fairer queue scheduling and mentions prioritizing main/dev branch pipelines over arbitrary pull requests.

The priority is kept outside the pipeline YAML so individual pipeline runs cannot raise their own place in the queue. Woodpecker reads the rules from the repository's default branch, which lets maintainers prefer default/stable branch builds, deprioritize bot senders, or push PR update/rebase events later without a larger scheduler framework. Reloading the config during pipeline creation also updates older queued tasks for that repository, so a long queue can adapt after maintainers push new rules.

Example:

```yaml
# .woodpecker/queue-priority.yml
rules:
  - priority: 100
    event: push
    branch: master
  - priority: 80
    event: push
    branch: stable-*
  - priority: -10
    event: pull_request
  - priority: -20
    event: pull_request
    event_reason: synchroniz*
  - priority: -30
    sender: renovate*
```

Related but not duplicate: woodpecker-ci/woodpecker#775 asks agents to pick the oldest jobs, and woodpecker-ci/woodpecker#6754 is a draft scheduler/store refactor.

This is intentionally compatible with woodpecker-ci/woodpecker#6754 rather than a replacement for that refactor: the queue priority reprioritization method lives in its own scheduler file, and merge-tree checks succeed in both merge directions.

## Testing

```sh
go test -race -timeout 180s -tags test go.woodpecker-ci.org/woodpecker/v3/server/model go.woodpecker-ci.org/woodpecker/v3/server/pipeline go.woodpecker-ci.org/woodpecker/v3/server/queue go.woodpecker-ci.org/woodpecker/v3/server/api go.woodpecker-ci.org/woodpecker/v3/server/store/datastore go.woodpecker-ci.org/woodpecker/v3/server/scheduler go.woodpecker-ci.org/woodpecker/v3/server/forge/gitea go.woodpecker-ci.org/woodpecker/v3/server/forge/forgejo go.woodpecker-ci.org/woodpecker/v3/server/forge/github
make generate-openapi
git merge-tree HEAD origin/pr-6754
git merge-tree origin/pr-6754 HEAD
```

### DIFF
--- a/cmd/server/flags.go
+++ b/cmd/server/flags.go
@@ -228,14 +228,6 @@ var flags = append([]cli.Flag{
 			TrimSpace: true,
 		},
 	},
-	&cli.StringSliceFlag{
-		Sources: cli.EnvVars("WOODPECKER_QUEUE_PRIORITY_RULES"),
-		Name:    "queue-priority-rules",
-		Usage:   "Rules that add or subtract queue priority for matching pipelines. Example: priority=100 event=push branch=main",
-		Config: cli.StringConfig{
-			TrimSpace: true,
-		},
-	},
 	&cli.DurationFlag{
 		Sources: cli.EnvVars("WOODPECKER_SESSION_EXPIRES"),
 		Name:    "session-expires",
--- a/cmd/server/flags.go
+++ b/cmd/server/flags.go
@@ -228,6 +228,14 @@ var flags = append([]cli.Flag{
 			TrimSpace: true,
 		},
 	},
+	&cli.StringSliceFlag{
+		Sources: cli.EnvVars("WOODPECKER_QUEUE_PRIORITY_RULES"),
+		Name:    "queue-priority-rules",
+		Usage:   "Rules that add or subtract queue priority for matching pipelines. Example: priority=100 event=push branch=main",
+		Config: cli.StringConfig{
+			TrimSpace: true,
+		},
+	},
 	&cli.DurationFlag{
 		Sources: cli.EnvVars("WOODPECKER_SESSION_EXPIRES"),
 		Name:    "session-expires",
--- a/cmd/server/openapi/docs.go
+++ b/cmd/server/openapi/docs.go
@@ -5801,6 +5801,10 @@ const docTemplate = `{
                 "pipeline_id": {
                     "type": "integer"
                 },
+                "priority": {
+                    "description": "Priority controls queue ordering. Higher priority tasks run before lower\npriority tasks, with Created and Name preserving FIFO behavior for ties.",
+                    "type": "integer"
+                },
                 "repo_id": {
                     "type": "integer"
                 },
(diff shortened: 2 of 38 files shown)


qwerty287:
Thank you for the PR, but before you continue, we maybe should discuss whether such a config file is a good way. Cause I personally do not really like it but I also don't see an easy alternative right now.

@woodpecker-ci/maintainers

Komzpa:
Sure. @qwerty287 is this a good place to discuss or should I join some matrix/IRC?

I'm happy to redo this into any other form of config if carrying it in repo does not look good. Some kind of config is probably needed. 

The problem I'm resolving is that OSGeo Woodpecker instance where we have a lot of hanging PRs that get mass-rebased after merges to main branch and that causes multi-hour wait for developers who wish to get interactive feedback, or be sure that they didn't break main.

qwerty287:
I think it's fine to discuss here. Please have some patience though as there's holiday time ;)

qwerty287:
I just had an idea how this could be designed more "integrated", but then the priority would only be set for a single workflow, not the whole pipeline.

Similar to the `runs_on` directive, on the workflow-level `when` we could add the `priority` key.

You would then have something like
```yaml
when:
    - branch: main
      event: push
      priority: 10
    - event: tag
      priority: 100
# ...

steps:
# ...
```
and then use the priority from the matching `when`. Here a `push` to `main` would get 10, a `tag` 100. And if multiple `when` branches match, just use the max configured priority.

That would be a solution I would like from the configuration design, but as I wrote, this would be per-workflow, not per-pipeline. If you have many workflows, you would have to set that everywhere.
On the other hand this could also be used to set different priorities within one pipeline. I.e. you have two workflow and one has higher priority, then the more important one can run and the less important one only runs if there is a free agent "seat". This is not possible with `depends_on` currently because in that case the less important workflow does not run even if there would an agent that has free resources to run it.

What do you think? Would that cover you usecase?

Komzpa:
@qwerty287 something like this? https://github.com/woodpecker-ci/woodpecker/pull/6986

qwerty287:
Yes, exactly. Already checked that out :)

6543 (changes requested):
I dont thing we should add fuzzy scheduler logic into config

6543:
and same answer as https://github.com/woodpecker-ci/woodpecker/pull/6986#issuecomment-5310360552 ...

> this is not a complete block jet as all maintainers need to reach a consensus if we fully wana block or allow it

lafriks:
If such feature is added this imho should be out of pipeline definition and be manageable using some kind of rules  by server admin

PR #6943, opened 2026-08-03
k8s backend: add default option to disable host userns

tehlordvortex:
Builds on top of the work in https://github.com/woodpecker-ci/woodpecker/pull/6710

Adds a (disabled by default) option to set `hostUsers: false` by default for all pods.
Additionally, this relaxes the "non-root" security context requirement to allow overriding the global setting if running in a custom user namespace, since root != true root in that case.


qwerty287:
Thanks for the PR.

@xoxys could you also check this? :upside_down_face: thanks

pipeline/backend/kubernetes/flags.go
```diff
@@ -114,6 +114,11 @@ var Flags = []cli.Flag{
 		Name:    "backend-k8s-secctx-nonroot",
 		Usage:   "`run as non root` Kubernetes security context option",
 	},
+	&cli.BoolFlag{
+		Sources: cli.EnvVars("WOODPECKER_BACKEND_K8S_DISABLE_HOST_USERNS"), // cspell:words userns
```
xoxys: I would prefer a positive name semantically e.g. `WOODPECKER_BACKEND_K8S_USER_NAMESPACES`. 

Two negatives collide in the middle:

- DisableHostUserns = true means "host userns is disabled".
- That maps to hostUsers = false, which also means "host users off".

So to figure out whether the new feature is active, you read `DISABLE…=true` and `hostUsers=false` and mentally invert both. The field whose value is `false` is the one that turns the feature **on**.

pipeline/backend/kubernetes/pod.go
```diff
@@ -620,8 +628,8 @@ func podSecurityContext(sc *SecurityContext, secCtxConf SecurityContextConfig, s
 			fsGroup = sc.FSGroup
 		}
 
-		// only allow to set nonRoot if it's not set globally already
-		if nonRoot == nil && sc.RunAsNonRoot != nil {
+		// only allow to set nonRoot if it's not set globally already or using user namespaces
+		if (nonRoot == nil || usingUserns) && sc.RunAsNonRoot != nil {
```
xoxys: I'm not that happy with the "relaxes the non-root security context requirement" part of the implementation because it silently converts an explicit admin security guarantee into an assumption, and that assumption is verified nowhere in the code path.

With this change we would override an admin policy, not just a pod field. `WOODPECKER_BACKEND_K8S_SECCTX_NONROOT=true` is a hardening control the admin set deliberately. `(nonRoot == nil || usingUserns) && sc.RunAsNonRoot != nil` and `stepPrivileged || usingUserns` now let a step-supplied value override that global. A pipeline author can defeat the admin's non-root policy by setting `runAsNonRoot: false` now.

I would suggest we decouple the two decisions into two explicit flags, and only relax root when both hold:

- Placement flag (`DISABLE_HOST_USERNS`) controls only the `hostUsers` default. No effect on root policy
- Trust flag (`ALLOW_ROOT_IN_USERNS`, default false) is used as explicit, documented statement that root-in-userns is acceptable for this cluster

I know it might be a bit inconvenient, however the current approach is IMO not ideal.

xoxys (changes requested):

pipeline/backend/kubernetes/flags.go
tehlordvortex: yep, that's a good point, i'll update the name. thanks!

pipeline/backend/kubernetes/pod.go
tehlordvortex: so, the reason i chose to implement it this way is that k8s does the same thing: https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/#integration-with-pod-security-admission-checks

when `hostUsers: false`, `runAsNonRoot` is not constrained, even with the "restricted" policy, as the container can never be "true" root.
As such, this doesn't affect the security posture/exposure of the cluster.

xoxys: Right, Kubernetes does relax `runAsNonRoot`/`runAsUser` for userns pods under restricted/baseline, so root-in-userns isn't true host root - I don't dispute that.

However, I think the analogy is a bit off, because the two mechanisms do different things:

- PSA is a validator, relaxing it only means it stops rejecting a pod that sets `runAsNonRoot: false`. It injects nothing, so a pod that already sets `runAsNonRoot: true` is unaffected
- `SECCTX_NONROOT` is an injection, the admin explicitly asked woodpecker to force `runAsNonRoot: true` onto every pod

So "k8s relaxes runAsNonRoot under userns" only justifies not blocking a userns pod, it doesn't justify turning off a hardening control the admin deliberately enabled. With this PR, enabling `DISABLE_HOST_USERNS` silently undoes `SECCTX_NONROOT` and there's no way to get both. That's why I'd decouple them.

tehlordvortex: hello! apologies for the delay on getting to this.

i see the point you're making about suddenly changing the behaviour of something an admin has configured tbh.
i've updated the name (to `WOODPECKER_BACKEND_K8S_USER_NAMESPACES`) and also introduced another flag for the override: `WOODPECKER_BACKEND_K8S_USER_NAMESPACES_OVERRIDE_SECCTX_NONROOT`.

PR #6950, opened 2026-08-04
Change group identifier to full_path for GitLab

gerbsen:
Gitlab group names are not unique. In order to secure the login process, the WOODPECKER_ORGS for Gitlab now compare the full_path value returned by GitLab's API to check membership.

This might break existing configurations.

### DIFF
--- a/docs/docs/30-administration/10-configuration/10-server.md
+++ b/docs/docs/30-administration/10-configuration/10-server.md
@@ -18,7 +18,7 @@ You can also restrict the registration:
   WOODPECKER_ADMIN=john.smith,jane_doe
   ```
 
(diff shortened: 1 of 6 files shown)


qwerty287:
Thanks!
One question: If you are member of a group `group` and in `WOODPECKER_ORGS` you allow `group/subgroup`, can the user access WP then?
I'm personally open about both ways of handling this, but it should be documented.

gerbsen:
No, it is an exact match currently. I would prefer if it stays this way, if you dont mind. Where do you think it should be documented?

qwerty287:
Add it here: https://woodpecker-ci.org/docs/administration/configuration/server#forge-and-user-configuration ?

qwerty287 (approved):

PR #6957, opened 2026-08-05
Make the Kubernetes cluster domain configurable

### LINKED ISSUE #6956: Kubernetes backend: services are unreachable by hostname on clusters with a custom cluster domain
## Component

agent

## Describe the bug

On a Kubernetes cluster that uses a custom cluster domain (anything other than
`cluster.local`), a step cannot reach a service declared in a `services:` block by its
hostname. The name simply does not resolve.

The agent creates a headless service per workflow and gives the pods the matching
subdomain, which is correct. It then adds a DNS search entry so that the short name
works, and that entry has `cluster.local` written into it:

```go
// pipeline/backend/kubernetes/pod.go
func dnsConfig(namespace, subdomain string) *kube_core_v1.PodDNSConfig {
	return &kube_core_v1.PodDNSConfig{
		Searches: []string{fmt.Sprintf("%s.%s.svc.cluster.local", subdomain, namespace)},
	}
}
```

On our cluster the kubelet entries use the real domain and only that one does not, so it
can never match:

```
search woodpecker.svc.k8s.example.io svc.k8s.example.io k8s.example.io \
       default.svc.k8s.example.io wp-hsvc-145741.woodpecker.svc.cluster.local
nameserver 169.254.25.10
options ndots:5
```

We could not find a `WOODPECKER_BACKEND_K8S_*` setting for this, which is why we are
reporting it rather than fixing our configuration.

What made us open the issue is the blast radius. This is not specific to one project or
one image: on a cluster with a custom domain, the mechanism that connects steps to
services is broken for every pipeline on the instance.

## Steps to reproduce

1. Run a cluster with a custom cluster domain. With kubeadm that is `--cluster-domain`,
   with kubespray it is `dns_domain` / `cluster_name`. Ours is `k8s.example.io`.
2. Run the agent with `WOODPECKER_BACKEND=kubernetes`.
3. Add a service and a step that talks to it:

   ```yaml
   services:
     - name: minio
       image: minio/minio
       ports: [9000]

   steps:
     - name: test
       image: alpine
       commands:
         - nc -z minio 9000
   ```

4. The step fails on name resolution (`bad address 'minio'` with busybox).
   The service container itself is up, its logs are fine.
5. `cat /etc/resolv.conf` inside the step shows the search entry above: the four kubelet
   entries carry the cluster domain, the one added by the agent carries `cluster.local`.
6. The full record does exist. Querying
   `minio.wp-hsvc-<id>.woodpecker.svc.k8s.example.io` returns the right address.

## Expected behavior

A service is reachable by its hostname, as the documentation says, whatever the cluster
domain is.

Two ways we can see:

- a setting, for example `WOODPECKER_BACKEND_K8S_CLUSTER_DOMAIN`, defaulting to
  `cluster.local` so nothing changes for anyone;
- or derive it from the agent's own `/etc/resolv.conf`, which already contains the right
  domain, and keep the constant only as a fallback.

The first one is a three line change and we are happy to send a PR if you would rather
have it that way. Tell us which you prefer and we will open it.

## System Info

```shell

server: woodpeckerci/woodpecker-server:v3.16.0
agent:  woodpeckerci/woodpecker-agent:v3.16.0  (WOODPECKER_BACKEND=kubernetes)
Kubernetes: v1.31.8, cluster domain k8s.example.io, CoreDNS 1.11.3 behind nodelocaldns
```

## Additional context

The line is still the same on `main` today, so this is not fixed in `next`.

The `dns_search` step option added in #5959 cannot be used as a workaround here. When a
step declares `dns` or `dns_search`, the backend replaces the whole `DNSConfig`:

```go
if len(step.DNS) != 0 || len(step.DNSSearch) != 0 {
    spec.DNSConfig = &kube_core_v1.PodDNSConfig{}
    ...
}
```

so the `wp-hsvc-<id>` entry is dropped along with the rest, and the workflow id changes on
every run anyway, so there is nothing stable to put in the pipeline file.

What we do in the meantime is rebuild the FQDN at runtime from `resolv.conf`, which works
but is not something we would like to keep in every pipeline:

```sh
SUB=$(tr ' ' '\n' < /etc/resolv.conf | grep -m1 '^wp-hsvc-' | cut -d. -f1,2)
DOM=$(tr ' ' '\n' < /etc/resolv.conf | grep -m1 '^svc\.')
# -> minio.wp-hsvc-145741.woodpecker.svc.k8s.example.io
```

We have also fixed it on our side by making the cluster answer on `cluster.local` as well,
which is fine for us but is not something every operator can do.

## Validations

- [x] Read the docs
- [x] Checked that there isn't already an issue reporting the same bug
- [x] Checked that the bug isn't fixed in the `next` version already

romain-rossi:
Closes #6956

On a cluster installed with a custom cluster domain, a step cannot reach a service declared in a `services:` block by its hostname: the DNS search entry the agent adds to the step pod has `cluster.local` written into it, so it can never match. There was no setting to change it.

This adds `WOODPECKER_BACKEND_K8S_CLUSTER_DOMAIN`, defaulting to `cluster.local`, and uses it to build that entry. Nothing changes for anyone who does not set it.

`dnsConfig` falls back to `cluster.local` when the value is empty, so the existing tests — which build the config as a literal — keep passing untouched, and an embedder building the config programmatically keeps the current behaviour.

On our cluster, whose domain is not `cluster.local`, we verified the mechanism by hand: a step pod given the search entry this patch produces does resolve the service by its short hostname, where the current one does not. We have not run a patched agent yet.


qwerty287:
Thanks.

@xoxys could you check whether this is fine?

xoxys:
Sure but Im currently sick. Can check next week.

qwerty287:
Sure, no problem, get well soon

docs/docs/30-administration/10-configuration/11-backends/20-kubernetes.md
```diff
@@ -517,6 +517,19 @@ Enables namespace isolation per Woodpecker organization. When enabled, each orga
...
+---
+
+### BACKEND_K8S_CLUSTER_DOMAIN
+
+- Name: `WOODPECKER_BACKEND_K8S_CLUSTER_DOMAIN`
+- Default: `cluster.local`
+
+The DNS domain of the Kubernetes cluster. It is used to build the DNS search entry that lets a step reach a service by its hostname.
+
+Set it if your cluster was installed with a custom cluster domain (`--cluster-domain` with kubeadm, `dns_domain` with kubespray). Leaving the default on such a cluster makes services unreachable by hostname.
```
xoxys, line 529:
```suggestion
Set this value if your cluster was installed with a custom cluster domain.
```

pipeline/backend/kubernetes/flags.go
```diff
@@ -31,6 +31,12 @@ var Flags = []cli.Flag{
 		Usage:   "Whether to enable namespace segregation per organization feature. When enabled, Woodpecker will create the Kubernetes resources to separated Kubernetes namespaces per Woodpecker organization.",
 		Value:   false,
 	},
+	&cli.StringFlag{
+		Sources: cli.EnvVars("WOODPECKER_BACKEND_K8S_CLUSTER_DOMAIN"),
+		Name:    "backend-k8s-cluster-domain",
+		Usage:   "backend k8s cluster domain, used to build the DNS search entry that makes services reachable by hostname",
```
xoxys, line 37: Same here, the value could be used for other scenarios in the future, keep description more generic

```suggestion
		Usage:   "backend k8s cluster domain, used to build the DNS search entry",
```

pipeline/backend/kubernetes/flags.go
```diff
@@ -31,6 +31,12 @@ var Flags = []cli.Flag{
...
+		Value:   "cluster.local",
```
xoxys, line 38:
```suggestion
		Value:   defaultClusterDomain,
```

.cspell.json
```diff
@@ -112,6 +112,8 @@
     "Jetpack",
     "Kaniko",
     "Keyfunc",
+    "kubeadm",
```
xoxys, line 115: Can be reverted after we remove it from the docs.

xoxys (commented):
Just a few nits. Code looks good to me. Thanks!

.cspell.json
```diff
@@ -112,6 +112,8 @@
...
+    "kubespray",
```
qwerty287:
```suggestion
```

qwerty287 (approved):

PR #6962, opened 2026-08-06
feat(web): extract ListEditor form component

gerbsen:
As you have mentioned yesterday [here](https://github.com/woodpecker-ci/woodpecker/pull/6874#issuecomment-5192877931) here is a feature branch with a **list component**. I would then rebase my changes in the other PR once this has been merged. But I have to make a disclaimer here: I'm not really into vue (I did angular for a long time). So this was all generated by Claude. I validated this on my local machine, but I'm not 100% sure if I've covered everything. 

--- 

The same "list of strings with add/remove rows" block was duplicated three times: once in SecretEdit for plugin images and twice in the repo General settings for trusted clone plugins and approval-allowed users.

Extract it as `ListEditor`, a sibling of the existing `KeyValueEditor`, and use it at all three call sites.

Behaviour differences to the inline copies:

- entries are trimmed, and blank or already listed values are rejected. The duplicate case previously produced colliding `:key`s and a delete that removed every matching row.
- each input gets a unique id instead of all of them reusing the id handed down by `InputField`.
- the component emits a new array rather than mutating the one it was given.

SecretEdit committed text left unconfirmed in the input when the form was submitted; `commitPendingItem()` is exposed so it still can.

Also clean up the naming while touching these call sites. The i18n key `repo.settings.general.netrc_only_trusted` was named after a boolean setting that #4352 replaced with a list and migration 020 dropped from the schema; every other layer has since said `netrc_trusted`. Rename the key to match. Only en.json is changed, per the translation policy in the UI development docs, so translators will see the label as a new string in Weblate and non-English users get the English fallback until it is translated.

SecretEdit also borrowed the repo-settings placeholder key for its images field; it now has its own `secrets.plugins.placeholder`.

### DIFF
--- a/docs/docs/92-development/03-ui.md
+++ b/docs/docs/92-development/03-ui.md
@@ -41,6 +41,67 @@ The following list contains some tools and frameworks used by the Woodpecker UI.
 - [Volar & vue-tsc](https://github.com/johnsoncodehk/volar/) for type-checking in .vue file
   - use the take-over mode of Volar as described by [this guide](https://github.com/johnsoncodehk/volar/discussions/471)
 
+## Form components
+
+Reusable form controls live in `web/src/components/form/`. They are all built to be placed inside an `InputField`, which renders the label, the optional description and a docs link, and passes down the `id` the control has to attach to its input:
+
+```vue
+<InputField :label="$t('some.label')">
+  <template #default="{ id }">
+    <TextField :id="id" v-model="value" />
+  </template>
+  <template #description>
+    {{ $t('some.description') }}
+  </template>
+</InputField>
+```
+
+Besides the single-value controls (`TextField`, `NumberField`, `Checkbox`, `SelectField`, `RadioField`) there are two editors for collections:
+
+- `ListEditor` for a list of strings, such as plugin images or usernames
+- `KeyValueEditor` for a `Record<string, string>`, such as environment variables
+
+### ListEditor
+
+`ListEditor` renders one read-only row per entry with a delete button, plus an input and an add button for new entries. Entries are not editable in place — they are removed and re-added instead. Use it with `v-model` on a `string[]`:
+
+```vue
+<InputField :label="$t('repo.settings.general.netrc_trusted.netrc_trusted')">
+  <template #default="{ id }">
+    <ListEditor
+      :id="id"
+      v-model="repoSettings.netrc_trusted"
+      :placeholder="$t('repo.settings.general.netrc_trusted.placeholder')"
+    />
+  </template>
+</InputField>
+```
+
+| Prop          | Type        | Description                                                          |
+| ------------- | ----------- | -------------------------------------------------------------------- |
+| `modelValue`  | `string[]?` | The current entries. An undefined value is treated as an empty list. |
+| `id`          | `string?`   | Id for the new-entry input, so the `InputField` label points at it.  |
+| `placeholder` | `string?`   | Placeholder of the new-entry input.                                  |
+| `addTitle`    | `string?`   | Tooltip of the add button.                                           |
+| `deleteTitle` | `string?`   | Tooltip of the delete buttons.                                       |
+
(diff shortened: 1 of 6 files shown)


web/src/assets/locales/en.json
```diff
@@ -119,8 +119,8 @@
           "allow": "Allow Deployments",
           "desc": "Allow deployments for successful pipelines. All users with push permissions can trigger these, so use with caution."
         },
-        "netrc_only_trusted": {
-          "netrc_only_trusted": "Custom trusted clone plugins",
+        "netrc_trusted": {
```
qwerty287: Please keep the old naming, otherwise all translations need to be re-translated.

web/src/assets/locales/en.json
```diff
@@ -507,7 +507,8 @@
     "saved": "Secret saved",
     "plugins": {
       "images": "Available only for the following plugins",
-      "desc": "List of plugin images where this secret is available. Leave empty to allow all plugins and normal steps."
+      "desc": "List of plugin images where this secret is available. Leave empty to allow all plugins and normal steps.",
+      "placeholder": "Plugin image"
```
qwerty287: I would just reuse the `netrc_only_trusted` string here to avoid duplicates. The translations need a rework at some point anyways, it's quite mixed up already.

web/src/views/repo/settings/General.vue
qwerty287, line 1: Can you add something like `commitPendingItem()` here as well, just like in `SecretEdit`?

docs/docs/92-development/03-ui.md
```diff
@@ -41,6 +41,67 @@ The following list contains some tools and frameworks used by the Woodpecker UI.
...
+  </template>
+</InputField>
+```
+
+Besides the single-value controls (`TextField`, `NumberField`, `Checkbox`, `SelectField`, `RadioField`) there are two editors for collections:
+
+- `ListEditor` for a list of strings, such as plugin images or usernames
+- `KeyValueEditor` for a `Record<string, string>`, such as environment variables
+
+### ListEditor
```
qwerty287: I don't think we need this specific docs. It's good to have some general ones (the paragraph above), but we also don't have that for all the other components.

(If we add something like this, it should be added for all components.)

qwerty287 (commented):
Thanks!

Wasn't asking you to do this, this was more a wish at my side 😆

docs/docs/92-development/03-ui.md
gerbsen: Just let me know if this is okay for you now.

web/src/assets/locales/en.json
gerbsen: reverted

docs/docs/92-development/03-ui.md
qwerty287: Yes, this looks good to me, thanks!

qwerty287:
Thanks, the code looks good to me now, I'll give it a try later :+1:

web/src/components/form/ListEditor.vue
```diff
@@ -0,0 +1,65 @@
...
+    placeholder?: string;
+    addTitle?: string;
+    deleteTitle?: string;
+  }>(),
+  {
+    modelValue: () => [],
+    id: undefined,
+    placeholder: undefined,
+    addTitle: undefined,
+    deleteTitle: undefined,
```
qwerty287: Ah sorry just saw these two options. They're unused if I'm not mistaken.

While I think it's a good idea to have titles for the buttons, I think they can be fixed with `Delete` and `Add`. Could you remove those two options and insert fixed `Delete` and `Add` strings as titles? (Check whether we have those strings in `en.json` already, if not add new)

gerbsen: well spotted! fixed in 429e2511892e0d6a785af9e989511933ab6ef4dc

qwerty287 (approved):

PR #6963, opened 2026-08-06
Add branch, tag, and commit selection for manual pipelines

### LINKED ISSUE #6810: Support selecting a tag when manually triggering a pipeline via the Web UI
## Description

Currently, when manually triggering a pipeline from the Woodpecker Web UI (clicking the "Run pipeline" button), the interface only provides a **branch selector**. There is no way to select a **tag**.

## Problem

Tags are widely used to mark release versions, and many teams have dedicated pipelines that should only run on tags (e.g., release builds, publishing artifacts, changelog generation). The inability to manually trigger a tag-based pipeline from the UI means users are forced to either:

1. Push a new tag via Git just to trigger the pipeline — inconvenient and pollutes the tag history.
2. Use the CLI (`woodpecker-cli deploy`) — works but less accessible for team members who prefer the UI.
3. Temporarily modify the workflow's `when.event` filter — hacky and error-prone.

### Suggested solution

The manual trigger dialog should allow users to choose between **branch** and **tag** (similar to GitLab CI's "Run pipeline" dialog), and populate the corresponding list based on the selection.

Mockup of the expected UI:

```
┌─ Run Pipeline  ──────────────────────────┐
│                                          │
│  Source:  ○ Branch   ● Tag               │
│                                          │
│  Tag:     [ v2.0.0          ▾ ]          │
│                                          │
│  [ Cancel ]              [ Run ]         │
└──────────────────────────────────────────┘
```

When a tag is selected, the pipeline should run with `CI_COMMIT_TAG` set and `CI_COMMIT_BRANCH` empty, matching the behavior of a tag push event.

### Alternative

1. **Use the CLI**: `woodpecker-cli deploy <repo> <build> <env> --event=tag` already works, but requires CLI setup and is less accessible than the UI.
2. **Push a tag via Git**: `git tag foo && git push origin foo` triggers the pipeline naturally, but creates unnecessary tags just to test or re-run a pipeline.

### Additional context

_No response_

### Validations

simonmeyerrr:
Closes #6810

## Summary

- Allow manually triggering a pipeline from a **branch**, **tag**, or **commit SHA** (mutually exclusive) via UI, API, CLI, and woodpecker-go.
- Add forge `Tags` / `TagHead` / `Commit` support (GitHub, GitLab, Gitea, Forgejo, Bitbucket Cloud & DC) and `GET /repos/:repo_id/tags`.
- UI: searchable combobox with separate search input, default-branch badge, and clearer docs that the event stays `manual` (choosing a tag does not select `when.event: tag` workflows).

## Screenshots

image
image
image
image
image


server/model/pipeline_options_test.go
qwerty287, line 1: Please name this `pipeline_test.go` only

server/model/repo_tag.go
qwerty287, line 1: We are only using the name of the tag, right? Would it be possible to return just a string array when asking for tags, just like for branches?

server/api/pipeline.go
```diff
@@ -88,24 +107,33 @@ func CreatePipeline(c *gin.Context) {
...
 	}
+	switch {
+	case opts.Tag != "":
+		pipeline.Ref = "refs/tags/" + opts.Tag
+		pipeline.TagTitle = opts.Tag
+	case opts.SHA != "":
+		pipeline.Ref = commit.SHA
+	default:
+		pipeline.Branch = opts.Branch
+		pipeline.Ref = opts.Branch
```
qwerty287, line 134: Shouldn't the ref be `refs/heads/<branch>` here?

server/api/pipeline.go
```diff
@@ -65,15 +65,34 @@ func CreatePipeline(c *gin.Context) {
 		return
 	}
 
+	if err := opts.Validate(); err != nil {
+		c.String(http.StatusBadRequest, "%s", err.Error())
```
qwerty287, line 69:
```suggestion
		c.String(http.StatusBadRequest, err.Error())
```

server/forge/addon/server.go
```diff
@@ -195,6 +195,48 @@ func (s *RPCServer) BranchHead(args []byte, resp *[]byte) error {
 	return err
 }
 
+func (s *RPCServer) Tags(args []byte, resp *[]byte) error {
+	var a argumentsBranchesPullRequests
```
qwerty287, line 199: Would rename this struct to something like `argumentsListOptions` to make more clear were this can be used. Adding `Tags` results in `argumentsBranchesPullRequestsTags` which is too long.

server/forge/addon/args.go
```diff
@@ -62,6 +62,18 @@ type argumentsBranchHead struct {
 	Branch string     `json:"branch"`
 }
 
+type argumentsTagHead struct {
```
qwerty287, line 65: Would merge that (and also `argumentsCommit`) with `argumentsBranchHead`, `argumentsActivateDeactivate` and rename to something like `argumentsString`

qwerty287 (commented):
Thanks!

Did a first review, but did not check everything yet.

server/model/repo_tag.go
simonmeyerrr: The idea behind this type is that some forge (not all of them) provide commit and timestamp information related to the tag in their API. These data could be useful for displaying more information about the tag in the UI or allow sorting based on date when the information is available.
So instead of going with just string array, I wanted to go with an object that would allow more room for new features and usage in the future without breaking changes rather than locking to the tag string itself.

server/api/pipeline.go
simonmeyerrr: The code that was in place before right above was:
```
Ref:                 opts.Branch,
```

qwerty287: Ok, then it was incorrect before as well I think. Would be better to change it in my opinion

server/model/repo_tag.go
qwerty287: Yes, but for branches we also don't have that yet, and I would prefer adding the complexity when we need it.

Also, I don't think we should make it too complex. Woodpecker is not the forge - it's not a missing feature in my opinion if we don't display a tag's creation date and commit SHA, you can just use the forge to get that information. But not sure what others think about this.

server/api/pipeline.go
simonmeyerrr: Do you want me to change it in this PR or would you prefer a dedicated PR ?
It might be a "breaking change" for user relying on `CI_COMMIT_REF` as being just the branch (even if they should have used `CI_COMMIT_BRANCH`)

server/model/repo_tag.go
simonmeyerrr: It's true that in the current PR, I'm not using it to display it. But in the following PR that add a "Tags" tab, I could display it and use the date as a sorting value when it's available.
But if you think it's not the purpose of Woodpecker I can remove it.

server/api/pipeline.go
qwerty287: This is not a breaking change because it is documented as the ref - as you, they should have used the branch var.

Yes, as that's a one-line change I would just change it in this PR

server/model/repo_tag.go
qwerty287: I personally think that we should not add something like this.
I would just not display dates and also just keep the default sorting (which should be sorted by date anyways I think when you fetch them from the forge)

But we can also wait for other maintainer's opinions, then you have to be patient though as there's holiday time ;)

server/api/pipeline.go
simonmeyerrr: I updated it 😃

server/model/repo_tag.go
simonmeyerrr: As, I'm not currently using it, I finally did go like you suggested it so we don't have to deal with stuff we don't actually use. So it's back to being just a string array.

server/forge/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -402,6 +402,79 @@ func (c *client) BranchHead(ctx context.Context, u *model.User, r *model.Repo, b
...
+	for {
+		var tags bitbucketTagList
+		resp, err := bc.GetPaged(ctx, "api", fmt.Sprintf("projects/%s/repos/%s/tags", r.Owner, r.Name), &tags, &opts)
+		if err != nil {
+			return nil, fmt.Errorf("unable to list tags: %w", err)
+		}
+		for _, tag := range tags.Values {
+			all = append(all, tag.DisplayID)
+		}
+		if !p.All || resp.LastPage {
```
qwerty287: I know you just copied this from other BB DC methods, but I think we should not support `p.All` in forges. I'll refactor this later and introduce a custom type for DB pagination (there we need `All`) and forge pagination. Could you remove that? Then we also get rid of the `for` loop

qwerty287 (commented):
Still have to review and test the UI, but otherwise the code looks good.

Have you tested this with different forges?

simonmeyerrr:
> Have you tested this with different forges?

I only tested with Gitea, not the other forges

qwerty287:
Ok, I think it would be good to test this with as many forges as possible in a real-world-scenario, but I also have only gitea set up in my test env... As this only touches manual pipelines (i.e. no hooks required) it shouldn't be necessary to have a publicly accessible server to test this with e.g. github. Still this seems to be quite some effort :laughing:

qwerty287:
@simonmeyerrr could you solve the conflicts? And then I would say we can just go on. The code looks good to me, and we can also test the other forges that on the next version. I tried testing with public forges but I wasn't successful in setting this up on my local environment...

simonmeyerrr:
@qwerty287 I resolved conflicts.
I didn't managed to have the time to test it with other forges, but as you said the changes should be ok to ship like this and fix afterward if there is any issue on a specific forge

qwerty287:
Thanks, could you check the formatting? https://ci.woodpecker-ci.org/repos/3780/pipeline/37620/49#L7

qwerty287:
Thanks.

I checked it out again and found an issue with the ComboBox component

grafik

I cannot scoll down there and see all branches, only the first few (this repo has 7 branches). When I select tags, I can scroll down, but still there are two tags I cannot see in the list (I have 11 tags in the repo).

PR #6969, opened 2026-08-07
Make command execution more robust on windows [partially reverted]

hasufell:
`iex` is a poor mechanism to execute arbitrary commands, because it requires great care wrt escaping.

In turn, '-encodedcommand' doesn't require any escaping. We can pass a base64 encoded UTF-16LE string directly to it.


pipeline/backend/common/script_test.go
```diff
@@ -26,11 +26,11 @@ const (
 )
 
 func TestGenerateContainerConf(t *testing.T) {
-	gotEnv, gotEntry := GenerateContainerConf([]string{"echo hello world"}, "windows", "/woodpecker/some")
+	gotEnv, gotEntry, _ := GenerateContainerConf([]string{"echo hello world"}, "windows", "/woodpecker/some")
 	assert.Equal(t, windowsScriptBase64, gotEnv["CI_SCRIPT"])
 	assert.Equal(t, "powershell.exe", gotEnv["SHELL"])
 	assert.Equal(t, []string{"powershell", "-noprofile", "-noninteractive", "-command", "[System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($Env:CI_SCRIPT)) | iex"}, gotEntry)
-	gotEnv, gotEntry = GenerateContainerConf([]string{"echo hello world"}, "linux", "/woodpecker/some")
+	gotEnv, gotEntry, _ = GenerateContainerConf([]string{"echo hello world"}, "linux", "/woodpecker/some")
```
6543: the test has to be adjusted

pipeline/backend/common/script.go
```diff
@@ -16,20 +16,28 @@ package common
...
+		utf16leBytes, err := encoder.Bytes([]byte(generateScriptWindows(commands, workDir)))
+		if err != nil {
+			return nil, nil, err
+		}
+
+		env["CI_SCRIPT"] = base64.StdEncoding.EncodeToString(utf16leBytes)
 		env["SHELL"] = "powershell.exe"
 		// cspell:disable-next-line
-		entry = []string{"powershell", "-noprofile", "-noninteractive", "-command", "[System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($Env:CI_SCRIPT)) | iex"}
+		entry = []string{"powershell", "-noprofile", "-noninteractive", "-encodedcommand", env["CI_SCRIPT"]}
```
6543, line 35: looks like Windows 7 and Windows Server 2008 had this option first added ...

... did never know! 👍

6543 (approved):

hasufell:
I noticed that I'm sometimes hitting `The command line is too long.`... so I think the only way around that is to write the cmd into a file and execute that.

What are your thoughts?

hasufell:
I think this is a regression... the old command had a bit more headroom, but may essentially also run into it.

hasufell:
Another option is to use stdin. I have no insights how well that would work for docker or kubernetes.

The "write cmd into script file first" approach on the other hand would require us to inject the file into the environment:

1. for local backend we just write to a file
2. for ssh, we could use `scp`
3. for docker, we could probably just mount it 
4. for kubernetes, maybe same as docker? (I have no insights here)

6543:
the new encoding needs more chars ... that's why long commands now could trigger that ...

https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

-> 32,767 UTF-16 characters

so that's not so much

6543:
writing to a script could solve one problem but docker and k8s don't have a nice interface to do so ...

hasufell:
I made it work with stdin and my custom libvirt backend that's still WIP. But that's via an ssh connection. I have no idea how to test the windows code with docker. I guess I would need kubernetes? But I'm not really interested in kubernetes anyway.

We could:

- revert to the previous code
- open a ticket to avoid the `iex` pattern and fix the maximum command length limitation

For libvirt, I use a different codepath anyway.

6543:
to test windows container you need windows and docker desktop ... switch to windows container mode and build some images to work with ... https://github.com/GECO-IT/woodpecker-windows ...

6543:
so now the question is what can carry more bits - env vars or arguments ...

6543:
@hasufell hey ... we maintainers decided to partially revert it till we have a better solution to not have regressions ...

... can you re-submit the reverted revert and then we can work on a better solution?

6543:
-> #7065

hasufell:
@6543 I'm not sure what you mean with revert the reverted patch :smile: 

I opened a ticket for discussion: https://github.com/woodpecker-ci/woodpecker/issues/7066

6543:
even better :)

PR #6973, opened 2026-08-07
Allow to add custom message to manual pipelines

### LINKED ISSUE #6732: Manual Pipeline "reason" parameter

When using manual pipeline, I want the log to display the "reason" why it was deployed instead of just MANUAL PIPELINE

<details>
<summary>MANUAL PIPELINE example</summary>

Image
</details>

### Suggested solution

For POST [/repos/{repo_id}/pipelines](https://woodpecker-ci.org/api#tag/Pipelines/paths/~1repos~1%7Brepo_id%7D~1pipelines/post), include an additional optional "options.reason" parameter
```json
{
  "branch": "master",
  "reason": "user triggered build"
}
```
This would then be displayed in the build reason:
Image

### Alternative

It can continue to display the branch
```
Manual: user triggered build @ master
```

### Additional context

I am integrating Woodpecker using the manual pipeline to build at specific times, but it's difficult to tell why a build was triggered.

### Validations

qwerty287:
closes https://github.com/woodpecker-ci/woodpecker/issues/6732


server/api/pipeline.go
```diff
@@ -106,6 +106,12 @@ func createTmpPipeline(event model.WebhookEvent, commit *model.Commit, user *mod
 
 		ForgeURL: commit.ForgeURL,
 	}
+
+	if opts.Message == "" {
```
lafriks: `opts.Message != ""`? or there is no point concatenating it below

lafriks (changes requested):

lafriks (approved):

PR #6975, opened 2026-08-07
agent: retry connect on DeadlineExceeded, not just Unavailable

### LINKED ISSUE #6942: agent: WOODPECKER_CONNECT_RETRY_* doesn't cover auth RPC DeadlineExceeded, only Unavailable
### Describe the bug

`WOODPECKER_CONNECT_RETRY_COUNT` / `WOODPECKER_CONNECT_RETRY_DELAY` (added in #6470) only retry the agent's initial connect loop when the returned gRPC status is `codes.Unavailable`. The agent's own auth RPC (`agent/rpc/auth_client_grpc.go`, `authClientTimeout = 5s`) frequently fails with `codes.DeadlineExceeded` instead — e.g. when the path to the server is slow/lossy rather than immediately refusing the connection. `codes.DeadlineExceeded` is not covered by the `status.Code(err) == codes.Unavailable` check in `runWithRetry()` (`cmd/agent/core/agent.go`), so the retry budget is skipped entirely and the agent calls `log.Fatal()` on the very first failed attempt — even with a large configured retry count/delay.

In production we saw an agent crash-loop 150+ times in ~17h (10s per cycle: start → 5s blocked auth RPC → fatal exit) while a sibling agent with identical config, reaching the server over a shorter/local network path, only restarted 5 times in the same window. The configured `WOODPECKER_CONNECT_RETRY_COUNT=60` / `_DELAY=5s` (a 5-minute budget, specifically set to smooth over exactly this kind of blip) never engaged because the failure mode was `DeadlineExceeded`, not `Unavailable`.

### System Info

Agent: dev build `next-0812d5e790` on top of v3.15.0 (Kubernetes backend), server v3.15.0.
Traced against `main` at commit `c1aea9c4ec` — the relevant code (`cmd/agent/core/agent.go` `runWithRetry()`, `agent/rpc/auth_client_grpc.go` `authClientTimeout`) is unchanged as of that commit.

### Steps to reproduce

1. Configure `WOODPECKER_CONNECT_RETRY_COUNT=60`, `WOODPECKER_CONNECT_RETRY_DELAY=5s` on an agent.
2. Make the network path to the server slow enough that the auth RPC exceeds its hardcoded 5s timeout (`authClientTimeout` in `agent/rpc/auth_client_grpc.go`) without an immediate TCP-level refusal.
3. Observe: agent logs `agent could not auth: rpc error: code = DeadlineExceeded desc = context deadline exceeded` and exits immediately via `log.Fatal()`, without ever entering the configured retry loop.

### Expected behavior

The connect-retry budget (`WOODPECKER_CONNECT_RETRY_COUNT`/`_DELAY`) should also cover `codes.DeadlineExceeded` on the initial connect/auth path, since it represents the same class of "couldn't reach the server in time" failure that `codes.Unavailable` already covers. Suggested minimal fix in `cmd/agent/core/agent.go`, `runWithRetry()`:

```diff
 		for range retryCount {
-			if err = run(ctx, c, backendEngines); status.Code(err) == codes.Unavailable {
+			err = run(ctx, c, backendEngines)
+			if code := status.Code(err); code == codes.Unavailable || code == codes.DeadlineExceeded {
 				log.Warn().Err(err).Msg(fmt.Sprintf("cannot connect to %s, retrying in %v", c.String("server"), retryDelay))
 				time.Sleep(retryDelay)
 			} else {
 				break
 			}
 		}
```

Happy to open a PR with this change if maintainers agree with the approach.

retif:
## What

`runWithRetry` only retries when the error returned from `run()` carries `codes.Unavailable`, so `WOODPECKER_CONNECT_RETRY_COUNT`/`WOODPECKER_CONNECT_RETRY_DELAY` never engage for a different, common failure mode.

## Why

`AuthClient.Auth` wraps every Login attempt in a hardcoded 5s timeout (`authClientTimeout`, `agent/rpc/auth_client_grpc.go`). If the server address is slow to resolve (a DNS hiccup, a resolver that's briefly unreachable, etc.) that timeout fires before the RPC completes, and the error surfaces as `codes.DeadlineExceeded` rather than `codes.Unavailable`.

`runWithRetry`'s retry gate only checks for `codes.Unavailable`, so this class of failure is structurally excluded — the agent fatal-exits after a single attempt regardless of the configured retry policy, turning a transient DNS blip at startup into a hard crash (and repeated re-registration with the server, since a fresh pod restart has no persisted agent identity yet).

Observed in practice: an agent pod on a node with an unstable path to the server logged
```
{"level":"fatal","error":"... rpc error: code = DeadlineExceeded desc = context deadline exceeded","message":"error running agent"}
```
roughly 5 seconds after start — matching `authClientTimeout` exactly — with zero retries despite `WOODPECKER_CONNECT_RETRY_COUNT=60` / `WOODPECKER_CONNECT_RETRY_DELAY=5s` configured.

## Change

Extracts the retry decision into `isRetryableConnectError` and also retries on `codes.DeadlineExceeded`, alongside a table test covering both retryable codes plus the cases that should NOT retry (a real auth failure via `codes.PermissionDenied`, a nil error, a non-status error) so the distinction stays enforced.

## Testing

```
go test -run TestIsRetryableConnectError -v ./cmd/agent/core/
go test -run TestStringSliceAddToMap -v ./cmd/agent/core/
```
Both pass; the second confirms the pre-existing test in the same file is unaffected.

Closes #6942


qwerty287:
Thanks. I would not add a separate function for this, just do something like what I did in #6972 - I actually opened a PR with the same fix ;) but we can also merge yours.

retif:
@qwerty287 done — dropped the separate `isRetryableConnectError` function and inlined the `DeadlineExceeded`/`Unavailable` check directly in the retry loop, matching #6972. Also removed the now-dangling dedicated unit test for the extracted function. Pushed in 779930d52.

qwerty287 (commented):
thanks!

qwerty287 (approved):

PR #6985, opened 2026-08-10
Add libvirt backend

hasufell:
Fixes https://github.com/woodpecker-ci/woodpecker/discussions/2302

Depends on:

* https://github.com/woodpecker-ci/woodpecker/pull/6968
* https://github.com/woodpecker-ci/woodpecker/pull/6969

Relevant libvirt tickets (mostly affecting bhyve backend):

* https://gitlab.com/libvirt/libvirt/-/work_items/900
* https://gitlab.com/libvirt/libvirt/-/work_items/901
* https://gitlab.com/libvirt/libvirt/-/work_items/899

Notes on AI use:

* AI was **not** used for code generation

TODO:

* [x] documentation
* [x] test multiple VM workflow on linux
* [x] fix SIGTERM propagation


hasufell:
There's several caveats:

## How to connect to the machine

There's at least 3 methods:

1. connect to a serial console: awful, because you're sending bytes to a console, not actually executing a command... so you have to deal with escaping hell
2. by talking to a qemu-guest agent running inside the VM: this limits us to qemu and is discouraged by libvirt upstream as a "backdoor"
3. via ssh

I picked option 3... which means the user has to set up working ssh inside the VM. There's two ways to do the ssh connection: by inspecting a network interface inside the VM and getting the IP... or by using the [libvirt NSS](https://libvirt.org/nss.html) method, which sets up domain names equivalent to the VM name (more elegant)

The user also has to configure the following in `sshd_config`:

```
AcceptEnv *
```

## Shared disk

If we want to execute multiple steps and retain the working directory/checkout, we need to have a disk that is shared across VMs. Unfortunately, mounting a host folder doesn't work consistently across platforms and hypervisors, so the best method is to require the user to provide a disk and a configuration.

However, the guest operating system needs to know how to mount the disk, so we also need to assign e.g. a serial number.

### Windows prerequisites for mounting

To be able to mount the shared disk on windows, we need the following.

Disable UAC remote token filtering:

```powershell
New-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\System" `
    -Name "LocalAccountTokenFilterPolicy" -Value 1 -PropertyType DWord -Force
```

Set your network to private:

```powershell
Set-NetConnectionProfile -InterfaceIndex <index> -NetworkCategory Private
```

Start WinRM service:

```powershell
winrm quickconfig
```


## Ephemeral VMs

This supports fully ephemeral VMs: the disks of the "base VMs" are copied. For qemu disks, it is zero-cost copy thanks to qcow2. For raw disks we need to copy the whole file.

hasufell:
Example workflow:

```

when:
  - platform: freebsd/amd64
  - event: push
  - event: manual

skip_clone: true

steps:
 - name: build
   image: win10
   commands:
     - Write-Host "Hello, World!"
     - Write-Host "Hello, World2"
     - Set-Content -Path ".\hello.txt" -Value "hello"
     - Get-ChildItem
   backend_options:
     libvirt:
       ephemeral: true
       shared_disk: &shared_disk
         uuid: "BHYVE-ABCDEF-CAFE"
         disk_config: |
           <disk type='file'>
             <serial>BHYVE-ABCDEF-CAFE</serial>
             <driver name='file' type='raw'/>
             <source file='/var/lib/libvirt/images/win10-shared.img'/>
             <target dev='sdb' bus='sata'/>
           </disk>
       ssh_config:
         username: root
         password: root
         timeout: 2m
         hostname: win10
 - name: build
   image: win11
   commands:
     - Write-Host "Hello, World2!"
     - Get-Content -Path ".\hello.txt"
     - Get-ChildItem
   backend_options:
     libvirt:
       ephemeral: true
       shared_disk: *shared_disk
       ssh_config:
         username: root
         password: root
         timeout: 2m
         hostname: win11
```

qwerty287:
Thanks!
Your two windows PRs will hopefully get reviewed soon, I cannot really help there as I never used woodpecker on windows.

Just a question towards this new backend: As there were quite some people asking for this, I think it's a good idea to get that into the core repo. However, I do not think any of us maintainers will use it, so we depend on contributions from libvirt users like you. This also means if other people open PRs for this backend we will ping you and ask to review - is that fine for you?

hasufell:
> This also means if other people open PRs for this backend we will ping you and ask to review - is that fine for you?

Yes.

pipeline/backend/libvirt/libvirt.go
```diff
@@ -0,0 +1,877 @@
...
+			cmd := "mount"
+			var args []string
+			if options.SharedDisk.UUID != "" {
+				args = []string{"-o", "X-mount.mkdir", fmt.Sprintf("UUID=%s", options.SharedDisk.UUID), step.WorkspaceBase}
+			} else if options.SharedDisk.Label != "" {
+				args = []string{"-o", "X-mount.mkdir", fmt.Sprintf("LABEL=%s", options.SharedDisk.Label), step.WorkspaceBase}
+			} else {
+				return fmt.Errorf("Neither UUID nor Label defined [%s]", taskUUID)
+			}
+			err = AdhocSSH(ctx, client, cmd, args, taskUUID, step.UUID)
```
hasufell: This is not very portable:

1. we assume `X-mount.mkdir` works (might not for busybox binutils)
2. we assume the user can run `mount`, which is probably only true for root

pipeline/backend/libvirt/backend_options.go
```diff
@@ -0,0 +1,38 @@
...
+// BackendOptions defines all the advanced options for the libvirt backend.
+type BackendOptions struct {
+	SSHConfig  SSHConfig  `mapstructure:"ssh_config"`
+	SharedDisk SharedDisk `mapstructure:"shared_disk"`
+	Ephemeral  bool       `mapstructure:"ephemeral"`
+}
+
+type SSHConfig struct {
+	User           string `mapstructure:"username"`
+	Password       string `mapstructure:"password"`
```
hasufell: I'm not sure we can use secrets/env variables in the backend options?

qwerty287: That's not supported

6543:
your pull got conflicts - just merge main in :)

and https://ci.woodpecker-ci.org/repos/3780/pipeline/37299 ... still i know it is WIP ... just some notes i saw

hasufell:
I think this is now ready for review... documentation is still missing, but the implementation now works.

I'm not sure how to assist reviewers here... it requires a lot of knowledge of:

1. libvirt and its portability issues
2. ssh and its portability issues
3. windows and its quirks

If any decision/pattern feels odd, then please raise it, so I can explain and add an inline comment. It's also possible I forgot some of the decisions along the way when experimenting with different approaches. Most things that look weird are most likely there for portability reasons (e.g. `AcceptEnv *` in ssh not working reliably).

I tried to keep the changes to the other backends to a minimum, but the script generation did change.

hasufell:
Also, I rebased, so it's possible I messed something up.

pipeline/backend/common/script_posix.go
```diff
@@ -62,13 +76,21 @@ unset CI_NETRC_PASSWORD
 unset CI_SCRIPT
 mkdir -p "{{.WorkDir}}"
 cd "{{.WorkDir}}"
+if [ -n "$CI_WRITE_PID" ]; then
+{
+	printf "%%s" "$$" > "${TMPDIR:-/tmp}/woodpecker_%s.pid" || true
+}
+fi
+%%s
```
hasufell: There's a templating bug I need to resolve here... because the `printf` also uses `%s`.

qwerty287:
Thanks, is there a reason to downgrade so many dependencies?

hasufell:
> Thanks, is there a reason to downgrade so many dependencies?

Not at all. That's probably just me struggling with go tooling. It's been 8 years since I did go.

hasufell:
One thing I'm not sure about is how all this will interact with `depends_on` or rather: parallel execution.

When we have sequential steps, then the state of each step (as in: the filesystem of the workdir folder) will be visible in the next step. How does this work with two steps that both depend on the same root?

Currently I create a shared disk that is identified by the workflow id and then mounted in for each step. This seems correct for the sequential ordering, but what about potentially parallel steps?

hasufell:
https://woodpecker-ci.org/docs/usage/workflow-syntax#file-changes-are-incremental also doesn't give much information on how filesystem state is resolved wrt parallelization

hasufell:
Is there a way to disable parallelized steps for the libvirt backend for the time being? It's not a trivial problem to solve.

qwerty287:
Parallel steps should work on the *same* volumes. I.e. if one of the steps changes a file, this change should also be visible to another step that's running in parallell. So they share the workflow "state".

> Is there a way to disable parallelized steps for the libvirt backend for the time being? It's not a trivial problem to solve.

That's not that easy. If two steps should run in parallel, the runtime will just call the `StartStep` multiple times without waiting till the previous step has finished.

I'm against an option that allows to change this, it would increase the complexity of the runtime quite a lot.

hasufell:
> Parallel steps should work on the same volumes. I.e. if one of the steps changes a file, this change should also be visible to another step that's running in parallell. So they share the workflow "state".

I'm afraid this is not possible with libvirt. Mounting the same disk into two different VMs will cause filesystem corruption, because they're not the same kernel.

The alternative would be setting up samba/NFS or doing some complicated cluster filesystem setup. All of those will have major caveats (e.g. performance penalties) and increase complexity for libvirt beyond what I think is reasonable.

Edit: mounting host folders into the VM (e.g. via virtiofs) would be another possibility, but it's also a source of problems and less portable, requires installing drivers and starting services (inside the guest) etc.

qwerty287:
Ok, then we need to find a different solution for this.

A simple but rather hacky solution would be to just fail with an error (in the backend - that should be forwarded to the frontend as well) in case you try to run two steps in parallel. Then we can keep all the code changes in the libvirt backend.

Other opinions? @woodpecker-ci/maintainers

hasufell:
Yes, I'm fine with an error.

VMs after all are not as lightweight as containers. Spinning up a VM can take minutes. So step parallelization is not really an interesting feature, I'd say. Users can just parallelize in their build system (e.g. Makefile).

hasufell:
Actually... I think I can just detect this in the backend itself and implement a step-lock for libvirt. Then whichever step was first, will make all others wait.

hasufell:
> Actually... I think I can just detect this in the backend itself and implement a step-lock for libvirt. Then whichever step was first, will make all others wait.

I tried this, but I'm hitting rather strange behavior. Say we have

```
    |<-depends_on- B
X --|
    |<-depends_on- C
```

B is first and `StartStep` runs through, acquires the lock.
C is now waiting in `StartStep` until B's `DestroyStep` function is called (where we lift the lock).

However... if B fails abnormally... it appears `DestroyStep` is not called, unless `C` has also finished one way or another. Now we have a deadlock.

hasufell:
It appears it works when I unlock in `WaitStep` instead.

qwerty287:
Can you maybe extract the change to the Makefile into its own PR? This PR is really big already, would be good to keep it as small as possible (I know that the change to the Makefile is quite small, still we prefer it this way)

hasufell:
> Can you maybe extract the change to the Makefile into its own PR? This PR is really big already, would be good to keep it as small as possible (I know that the change to the Makefile is quite small, still we prefer it this way)

The change is tied to the new libvirt backend, it doesn't make much sense without it. We only need `CGO_ENABLED` for these additional targets because of libvirt.

pipeline/backend/libvirt/domain.go
```diff
@@ -0,0 +1,393 @@
...
+		return nil, "", "", "", err
+	}
+
+	// get the XML from the loaded domain
+	xml, err := domain.GetXMLDesc(virt.DOMAIN_XML_INACTIVE)
+	if err != nil {
+		return nil, "", "", "", err
+	}
+	// parse the XML into etree
+	doc := etree.NewDocument()
```
berrange: FWIW, you'll probably have a happier time long term if you use the official 'libvirtxml' Go API we provide that can deserialize any libvirt XML into a Go struct, and vica-verca, rather than directly manipulating raw XML:

  https://pkg.go.dev/libvirt.org/go/libvirtxml

Especially when formatting XML, it'll help ensure you don't miss any data escaping, which could risk accidentally opening up security flaws .

hasufell: Thanks, should be fixed. Quite a lot of nil pointer checks, but ok.

hasufell:
Hi, it's been quite a while. Is there anything I can do to speed up the review process?

PR #6986, opened 2026-08-10
Add workflow-level queue priority

Komzpa:
## Summary

This is a smaller alternative to woodpecker-ci/woodpecker#6941 based on the review suggestion to express queue priority directly on matching workflow `when` conditions.

It adds workflow-level `when.priority`, persists the selected task priority, and orders pending queue assignment by higher priority first while preserving the existing Created/Name FIFO ordering for ties. Conditions without `priority` do not affect queue ordering; if multiple matching conditions set priority, the highest matching value wins.

Instances that do not want workflow-level priority can disable it with `WOODPECKER_DISABLE_WORKFLOW_PRIORITY`.

## Tests

- `make generate-openapi`
- `git diff --check upstream/main...HEAD`
- `go test -race -timeout 180s -tags test go.woodpecker-ci.org/woodpecker/v3/pipeline/frontend/yaml/linter go.woodpecker-ci.org/woodpecker/v3/pipeline/frontend/builder`
- `go test -race -timeout 180s -tags test go.woodpecker-ci.org/woodpecker/v3/server/model go.woodpecker-ci.org/woodpecker/v3/server/queue go.woodpecker-ci.org/woodpecker/v3/server/pipeline go.woodpecker-ci.org/woodpecker/v3/pipeline/frontend/yaml/constraint go.woodpecker-ci.org/woodpecker/v3/pipeline/frontend/yaml/linter/schema`
- `go test -race -timeout 180s -tags test go.woodpecker-ci.org/woodpecker/v3/cmd/server go.woodpecker-ci.org/woodpecker/v3/server/pipeline` (fails before tests because this checkout has no embedded `web/dist/*`; `server/pipeline` passed)


docs/versioned_docs/version-3.17/20-usage/20-workflow-syntax.md
```diff
@@ -779,6 +779,17 @@ Example conditional execution by branch:
 
 The workflow now triggers on `main`, but also if the target branch of a pull request is `main`.
 
+You can add `priority` to workflow-level `when` conditions to change queue ordering for matching workflows. Higher priority workflows run before lower priority workflows. If multiple matching conditions define a priority, Woodpecker uses the highest matching priority. Conditions without `priority` do not affect queue ordering.
```
qwerty287: This feature won't be usable in 3.17, please remove from this page

qwerty287 (commented):
Thanks!

Code looks very good to me already.

Something we need to think about is the default value if nothing is set. Right now you just set 0 as default which is fine in general, but I'm not sure whether this could have an impact on partially public instances like codeberg - you could "abuse" the priority feature to always get your pipelines run first.
But probably there isn't an easy solution for this situation. Maybe it's helpful to allow to disable the feature completely? Or add a repo-level option that is only changeable by admins? Anyways such a change can get implemented in another PR

Komzpa:
@qwerty287 PTAL

qwerty287:
Thanks, but I actually wrote that adding a per-repo option to disable is something for a follow-up PR.
I also do not think that the privileged option is good to re-use here, this should get its own option instead.

6543 (changes requested):
I dont thing we should add fuzzy scheduler logic into config

6543:
@Komzpa hi

the idea is novel ... and i get the usecase ...

but before introducing new config for the core engine ... the scheduler stuff should be server specific and i want to make it way more inteligent e.g. allow for quotas, move tasks to agents that did the job best based on historycal runs, also have priorities etc...

we currently refactor the server/scheduler part and once that is ready we should look into this again.

I would recommend to instead if you have them ... separate based on labels and multible agents ...

and write this idea of "queue priority / dynamic queue rules" into an issue to keep track of it :)

6543:
> this is not a complete block jet as all maintainers need to reach a consensus if we fully wana block or allow it

qwerty287:
As I commented on matrix already, I think this is a good feature that's worth to add.

> I dont thing we should add fuzzy scheduler logic into config

I'm not sure if I get what you mean here, but by fuzzy you mean that it's possible to put any number there? 

I see that this is not optimal (esp. as it's also a "global" priority, which is why I requested to be able to disable this on large public instances), but would it be "non-fuzzy" if we would have something like `depends_on`, i.e. a "named prioritization":

If you have two workflows "high-prio" and "low-prio"
```yaml
# low-prio.yaml
prioritize: # or however you would call that key
    - high-prio
```
which then means high-prio has higher priority than low-prio.

This would be probably way more complex to implement but it then is just affecting a "local" environment (only the current pipeline) instead of all pipelines over all repos.

PR #6991, opened 2026-08-11
fix(gitlab): avoid N+1 member lookups in Repos()

### LINKED ISSUE #6882: GitLab forge: Repos() N+1 API calls break login for admin users
# GitLab forge: `Repos()` N+1 API calls break login for admin users

## Bug Description

The GitLab forge's `Repos()` function makes an additional `GetInheritedProjectMember` API call for **every project** returned by `ListProjects`. This causes two problems:

1. **Login completely broken for GitLab admin users** — `GetInheritedProjectMember` returns 404 for projects the admin can access through admin privileges but isn't an explicit member of (e.g., personal namespace projects from other users). The first 404 aborts the entire login flow.

2. **Login timeout for users with many repos** — With 50 projects per page and thousands of visible projects, the synchronous repo permission sync during first login can take 100+ seconds, exceeding typical gateway/proxy timeouts.

## Steps to Reproduce

1. Configure Woodpecker with a GitLab forge
2. Log in with a user that is a GitLab instance admin
3. Ensure other users have personal namespace projects on the GitLab instance

**Expected:** Login succeeds  
**Actual:** Login fails with `{"level":"error","error":"404 Not Found","message":"cannot update repo permissions for user <login>"}`

Even without the 404, users with 500+ visible projects will hit gateway timeouts because `updateRepoPermissions` runs synchronously on first login (when `noStoredRepositories == true`), regardless of the `WOODPECKER_ASYNC_REPOSITORY_UPDATE` setting.

## Root Cause

In `server/forge/gitlab/gitlab.go`, the `Repos()` function:

```go
projects, _, err := client.Projects.ListProjects(opts, gitlab.WithContext(ctx))
// ...
for i := range projects {
    projectMember, _, err := client.ProjectMembers.GetInheritedProjectMember(projects[i].ID, int64(intUserID), gitlab.WithContext(ctx))
    if err != nil {
        return nil, err  // ← 404 here kills the entire login
    }
    // ...
}
```

This calls `GET /api/v4/projects/:id/members/all/:user_id` for every project. For GitLab admins, `ListProjects?min_access_level=30` returns projects accessible via admin privileges where the user has no explicit membership — and the members endpoint returns 404 for those.

## The Fix

The `GetInheritedProjectMember` call is **entirely redundant**. The `gitlab.Project` struct returned by `ListProjects` already includes a `Permissions` field:

```go
type Project struct {
    // ...
    Permissions *Permissions `json:"permissions"`
}

type Permissions struct {
    ProjectAccess *ProjectAccess `json:"project_access"`
    GroupAccess   *GroupAccess   `json:"group_access"`
}

type ProjectAccess struct {
    AccessLevel AccessLevelValue `json:"access_level"`
    // ...
}
```

This provides the user's access level (via direct membership and group inheritance) without an additional API call. The fix eliminates O(n) API calls per page and resolves the admin 404 issue.

## Environment

- Woodpecker Server: v3.16.0
- GitLab: self-hosted (recent version)
- go-gitlab library: `gitlab.com/gitlab-org/api/client-go/v2` v2.43.0
- Helm chart: v3.6.5

spam-n-eggs:
Closes #6882

Repos() made one GET /api/v4/projects/:id/members/all/:user_id call per project — an N+1 that broke login for GitLab admins (404 on projects accessible via admin rights but without explicit membership, aborting the whole sync) and caused timeouts for users with many repos.

The projects list API already reports the current user's access level via the permissions attribute (project_access/group_access), so we now use that. A per-project membership lookup only happens when the embedded permissions are missing (e.g. access inherited from a nested parent group — GitLab's known gap), preserving prior behavior where the data is available.

Adds TestGitLabReposUsesEmbeddedPermissions asserting exactly one membership call (no N+1) and correct Pull/Push/Admin mapping.

### DIFF
--- a/server/forge/gitlab/gitlab.go
+++ b/server/forge/gitlab/gitlab.go
@@ -309,12 +309,24 @@ func (g *GitLab) Repos(ctx context.Context, user *model.User, p *model.ListOptio
 	repos := make([]*model.Repo, 0, len(projects))
 
 	for i := range projects {
-		projectMember, _, err := client.ProjectMembers.GetInheritedProjectMember(projects[i].ID, int64(intUserID), gitlab.WithContext(ctx))
-		if err != nil {
-			return nil, err
+		project := projects[i]
+
+		// The projects list API already reports the current user's access
+		// level via the `permissions` attribute, so in the common case we can
+		// skip the per-project membership lookup. GitLab may however omit it
+		// (e.g. when access is inherited from a nested parent group), in
+		// which case we fall back to a single explicit lookup per affected
+		// project.
+		var projectMember *gitlab.ProjectMember
+		if embeddedAccessLevel(project) == gitlab.NoPermissions {
+			var err error
+			projectMember, _, err = client.ProjectMembers.GetInheritedProjectMember(project.ID, int64(intUserID), gitlab.WithContext(ctx))
+			if err != nil {
+				return nil, err
+			}
 		}
 
-		repo, err := g.convertGitLabRepo(projects[i], projectMember)
+		repo, err := g.convertGitLabRepo(project, projectMember)
 		if err != nil {
 			return nil, err
 		}
--- a/server/forge/gitlab/gitlab.go
+++ b/server/forge/gitlab/gitlab.go
@@ -311,12 +311,7 @@ func (g *GitLab) Repos(ctx context.Context, user *model.User, p *model.ListOptio
 	for i := range projects {
 		project := projects[i]
 
-		// The projects list API already reports the current user's access
-		// level via the `permissions` attribute, so in the common case we can
-		// skip the per-project membership lookup. GitLab may however omit it
-		// (e.g. when access is inherited from a nested parent group), in
-		// which case we fall back to a single explicit lookup per affected
-		// project.
+		// The projects list API already reports the current user's access level
 		var projectMember *gitlab.ProjectMember
 		if embeddedAccessLevel(project) == gitlab.NoPermissions {
 			var err error
--- a/server/forge/gitlab/helper.go
+++ b/server/forge/gitlab/helper.go
@@ -48,17 +48,49 @@ func newClient(url, accessToken string, skipVerify bool) (*gitlab.Client, error)
 // isRead is a helper function that returns true if the
 // user has Read-only access to the repository.
 func isRead(proj *gitlab.Project, projectMember *gitlab.ProjectMember) bool {
-	return proj.Visibility == gitlab.InternalVisibility || proj.Visibility == gitlab.PrivateVisibility || projectMember != nil && projectMember.AccessLevel >= gitlab.ReporterPermissions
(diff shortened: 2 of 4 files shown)


server/forge/gitlab/gitlab.go
```diff
@@ -309,12 +309,24 @@ func (g *GitLab) Repos(ctx context.Context, user *model.User, p *model.ListOptio
 	repos := make([]*model.Repo, 0, len(projects))
 
 	for i := range projects {
-		projectMember, _, err := client.ProjectMembers.GetInheritedProjectMember(projects[i].ID, int64(intUserID), gitlab.WithContext(ctx))
-		if err != nil {
-			return nil, err
+		project := projects[i]
+
+		// The projects list API already reports the current user's access
```
qwerty287: Please shorten that comment. It's way too long.

server/forge/gitlab/helper.go
```diff
@@ -48,17 +48,49 @@ func newClient(url, accessToken string, skipVerify bool) (*gitlab.Client, error)
...
 
 // isAdmin is a helper function that returns true if the
 // user has Admin access to the repository.
-func isAdmin(projectMember *gitlab.ProjectMember) bool {
-	return projectMember != nil && projectMember.AccessLevel >= gitlab.MaintainerPermissions
+func isAdmin(proj *gitlab.Project, projectMember *gitlab.ProjectMember) bool {
+	return userAccessLevel(proj, projectMember) >= gitlab.MaintainerPermissions
+}
+
+// userAccessLevel returns the effective access level of the current user for
```
qwerty287: Same here

server/forge/gitlab/helper.go
```diff
@@ -48,17 +48,49 @@ func newClient(url, accessToken string, skipVerify bool) (*gitlab.Client, error)
...
+// in the project response is used. If neither provides a level,
+// gitlab.NoPermissions is returned.
+func userAccessLevel(proj *gitlab.Project, projectMember *gitlab.ProjectMember) gitlab.AccessLevelValue {
+	if projectMember != nil {
+		return projectMember.AccessLevel
+	}
+	return embeddedAccessLevel(proj)
+}
+
+// embeddedAccessLevel returns the access level of the current user as reported
```
qwerty287: and here

qwerty287 (commented):
Thanks!

qwerty287:
@chrisalbright do you want to test this?

spam-n-eggs:
@qwerty287 if we don't hear back from @chrisalbright I can share video of testing this with gitlab.  We're in a transitional phase at @dominion-solutions where we're moving off of GitLab, but we haven't got our self-hosted forgejo set up yet.

qwerty287:
Let's wait one or two more days and then we can merge it without response from @chrisalbright

qwerty287 (approved):
As there's no response and the code looks good to me I'll go on and merge this

chrisalbright:
Sorry @qwerty287, @spam-n-eggs I had surgery last week and am still recovering. @spam-n-eggs thank you for fixing this :)

6543:
@chrisalbright uh! get well, health is more important :)

PR #6995, opened 2026-08-12
fix: update star history chart link

Dessalines39394:
The star history chart in the README is currently broken due to GitHub stargazer API restrictions. This updates the chart to use a working source so it displays correctly again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ It combines both plugins by the Woodpecker core team and community-maintained on
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


qwerty287:
Thanks, the website says:

> this fork keeps star history alive from a different data source

What does this mean? From which source is the data if it's not from github?

Dessalines39394:
It is github but in the form of snapshots. The data was collected before the shutdown, and is now kept up-to-date using the stargazer count part of the repository metadata

qwerty287:
We decided to drop the chart instead https://github.com/woodpecker-ci/woodpecker/pull/6997

PR #6999, opened 2026-08-13
Return step logs strictly in line order

stevapple:
`LogFind` returned a step's log entries sorted by the auto-increment primary key. That works as long as auto-increment values arrive in insertion order, which holds on a single-node engine.

Distributed engines allocate them from per-node caches. On TiDB, consecutive `LogAppend` calls land on different nodes and receive id ranges that interleave, so a batch written later can carry lower ids than one written earlier. Sorting by `id` then reorders the log at batch granularity, and since the UI renders rows as the API returns them, the line numbers jump backwards mid-log: 572, then 107, then 1. We confirmed that over a half of historical runner steps have such interleaved logs.

log-order

`line` is what the agent assigns in `LineWriter` as it produces the output. It already carries the order, and it is populated for existing rows, so this is a query change with no backfill.

## Tests

`TestLogFindOrdersByLine` assigns ids explicitly so the first batch takes 4000/4001 and the second takes 30/31, reproducing the interleave. It fails against the old behaviour, returning `[third fourth first second]`.

The existing log store tests asserted row counts alone, which is why CI stayed green through both of these.

## Future directions

**A composite index on `(step_id, line)`** would let this sort be served straight from the index. It was left out of this PR deliberately. Declaring it `UNIQUE(s)` would match how `steps` and `workflows` declare `(pipeline_id, pid)`, but that constraint does not hold on real data today, so it would need a dedup migration first. This is requested per review and will be covered by a follow-up PR.


qwerty287:
Thanks.

> We found (step_id, line) is not unique in practice. Checking our instance turned up 182 duplicate pairs out of ~56k steps. Sampling them, each is a pair of byte-identical rows, which points at an agent re-sending a batch after a reconnect rather than at two genuinely different lines colliding, so it's generally lossless to dedup, but worth further investigation.

I never saw this. But I'm not sure if your solution is the right way to go here – I would prefer making `(step_id, line)` unique (and handle errors on insert due to duplicates properly) and then we can skip the deduplication. Would that be a possible fix instead of your current solution?

stevapple:
> I would prefer making `(step_id, line)` unique (and handle errors on insert due to duplicates properly) and then we can skip the deduplication.

I would prefer this solution too. The worry is from migration side — if we already ensure conflicting lines is unlikely to happen, then we should deduplicate once during migration, and make the `(step_id, line)` truly unique.

qwerty287:
I would just delete duplicated lines in the migration. I never saw this issue before, seems to only happen in very special setups (and as you said the DB data is the same for the log lines).

stevapple:
Both migration and unique index added. Here's our findings on how the duplication may happen:

The agent delivers logs at least once. In `agent/rpc/client_grpc.go`, `sendLogs` builds the request **once, outside** the retry:

```go
func (c *client) sendLogs(ctx context.Context, entries []*proto.LogEntry) error {
	req := &proto.LogRequest{LogEntries: entries}

	_, err := retryRPC(ctx, c, "log", func() (*proto.Empty, error) {
		r, err := c.client.Log(ctx, req)
		return r, classifyRPCErr(ctx, err)
	})
	return err
}
```

`retryRPC` runs that closure through `backoff.Retry`, so every attempt sends the same `req`. The server's `Log` handler appends whatever arrives, with no idempotency key. So if the server commits a batch and the response does not make it back — connection dropped, or a timeout that fires after the write landed — the agent retries and the same entries are stored a second time.

qwerty287:
Thanks, could you split that out into 2 PRs? I.e. here only returning in line order and another PR the unique migration

stevapple:
> Thanks, could you split that out into 2 PRs? I.e. here only returning in line order and another PR the unique migration

Done with #7004

qwerty287 (approved):

PR #7002, opened 2026-08-13
perf(schema): cache compiled YAML validation schema

### LINKED ISSUE #7000: YAML schema linter recompiles the schema on every lint operation
## Problem

The YAML schema linter recompiles the JSON schema every time `schema.Lint` is called.

The schema definition is constant, so compiling it for every document introduces repeated work when multiple pipelines are processed.

## Evidence

Profiling the schema tests showed that schema compilation was the dominant cost:

- `schema.Lint`: approximately 1.32 seconds cumulative
- `gojsonschema.NewSchema` / `SchemaLoader.Compile`: approximately 1.18 seconds cumulative
- Profile duration: approximately 1.47 seconds
- Total CPU samples: approximately 1.86 seconds

The relevant call path was:

```text
schema.Lint
└── gojsonschema.Validate
    └── gojsonschema.NewSchema
        └── SchemaLoader.Compile
```

## Proposed improvement

Compile the schema once and reuse the compiled `*gojsonschema.Schema` for subsequent validation calls.

The initialization should be protected with `sync.Once` because multiple pipelines may be linted concurrently.

## Expected result

- The validation behavior should remain unchanged, while repeated schema compilation is removed from subsequent `Lint` calls.

## Verification

The change should be verified with:

- [x] The existing schema tests
- [x] The race detector
- [x] Before-and-after CPU profiles

AhmadNajiKam:
## Problem

`schema.Lint` used `gojsonschema.Validate` with a schema loader on every call.

This caused the JSON schema to be parsed and compiled repeatedly, even though the schema definition is constant. Profiling showed that schema compilation was the main performance bottleneck.

## Solution

Compile the schema once and reuse the compiled schema for subsequent validation calls.

The initialization is guarded with `sync.Once` so that concurrent pipeline processing cannot initialize the cached schema more than once.

## Key changes

- Added cached storage for the compiled `*gojsonschema.Schema`.
- Added `sync.Once` for thread-safe one-time initialization.
- Replaced repeated schema compilation with validation through the cached schema.
- Preserved the existing schema validation behavior.

## Performance results

The same `TestSchema` workload was profiled 20 times before and after the change.

Before Flame graph:
screenshot-2026-08-14_00-12-04

After Flame graph:
screenshot-2026-08-14_00-12-17

The profile duration decreased by approximately 86.2%, while total CPU samples decreased by approximately 88%.

After the change, the remaining significant work is primarily YAML-to-JSON conversion and actual schema validation.

## Behavior changes

| Area | Before | After |
|---|---|---|
| Schema initialization | Per `Lint` call | Once |
| Concurrent initialization | Repeated compilation possible | Protected by `sync.Once` |
| Document validation | Uses a newly compiled schema | Uses the cached compiled schema |
| Validation rules | Existing rules | Unchanged |

## Verification

- Completed before-and-after CPU profile comparison.
- Confirmed that schema compilation no longer appears as a main hotspot.
- Confirmed that document validation still uses the JSON schema.
- The focused schema test workload completed successfully.
- The race-detector and full end-to-end test runs should be completed before merging.

Closes #7000

### DIFF
--- a/pipeline/frontend/yaml/linter/schema/schema.go
+++ b/pipeline/frontend/yaml/linter/schema/schema.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"sync"
 
 	"codeberg.org/6543/go-yaml2json/v2"
 	"github.com/xeipuuv/gojsonschema"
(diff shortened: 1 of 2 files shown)


qwerty287 (approved):
Thanks!

PR #7004, opened 2026-08-14
Enforce one log entry per line with a unique index

stevapple:
Nothing stops the same line number being stored twice for a step, and the log view then prints that line twice. Split out of #6999 as suggested in review.

The agent delivers logs at least once. `sendLogs` builds the request **once, outside** the retry, so `backoff.Retry` resends the identical batch; if the server committed it but the response was lost, the entries are stored again. Our instance holds **182 duplicated `(step_id, line)` pairs**, each a pair of byte-identical rows in contiguous runs — whole batches, the shape a resend produces.

## Fix

- **`log_entries` gains `UNIQUE(step_id, line)`**, matching how `steps` and `workflows` declare `(pipeline_id, pid)`.
- **A migration clears the way**, keeping the lowest id of every line number and dropping the rest — restricted to line numbers that actually repeat, so the work is proportional to the duplicates rather than to the table.
- **A resend now fails rather than passing unnoticed**, which loses nothing: the retried batch is identical and each chunk is one atomic `INSERT`, so a chunk that trips the constraint holds nothing new. It is surfaced rather than swallowed because that property lives in the agent, not in the store.

## Tests

- `TestLogAppendRejectsResentEntries` — a resent batch is refused and the stored entries are left as they were; passes without the index, since nothing rejects the second copy.
- `TestDeduplicateLogEntries` — the migration on its own, including a line number stored three times and identical line numbers under a second step that must survive.
- `TestMigrateRemovesDuplicateLogEntries` — the full `Migrate()` over an existing database holding a duplicate, so the cleanup running before index creation is exercised rather than assumed.

### DIFF
--- a/server/model/log.go
+++ b/server/model/log.go
@@ -27,9 +27,9 @@ const (
 
 type LogEntry struct {
 	ID      int64        `json:"id"       xorm:"pk autoincr 'id'"`
-	StepID  int64        `json:"step_id"  xorm:"INDEX 'step_id'"`
+	StepID  int64        `json:"step_id"  xorm:"UNIQUE(s) INDEX 'step_id'"`
 	Time    int64        `json:"time"     xorm:"'time'"`
-	Line    int          `json:"line"     xorm:"'line'"`
+	Line    int          `json:"line"     xorm:"UNIQUE(s) 'line'"`
 	Data    []byte       `json:"data"     xorm:"LONGBLOB"`
 	Created int64        `json:"-"        xorm:"created"`
 	Type    LogEntryType `json:"type"     xorm:"'type'"`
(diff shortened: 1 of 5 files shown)


qwerty287:
It seems there's a failing mysql test?

https://ci.woodpecker-ci.org/repos/3780/pipeline/36853/38#L13

stevapple:
> It seems there's a failing mysql test?
> 
> https://ci.woodpecker-ci.org/repos/3780/pipeline/36853/38#L13

Reproduced and addressed. It's caused by test fixture conflict.

qwerty287 (approved):

6543:
image

this is on our instance :/

PR #7005, opened 2026-08-14
Order the user feed by creation time instead of id

stevapple:
`UserFeed` was sorted by `pipelines.id`, treating the auto-increment primary key as a record of creation order. That holds on a single-node engine; distributed engines allocate ids from per-node caches, so a pipeline created later can hold a lower id.

The effect is not a visibly jumbled feed. `web/src/store/pipelines.ts` re-sorts whatever arrives by `created`:

```js
function comparePipelines(a, b) { return (b.created || -1) - (a.created || -1); }
```

So the display order is already right, and the server and the client simply disagree about what "most recent" means. What the client cannot fix is membership: the query is capped at `perPage`, so when `id` order diverges from `created` order the server returns the wrong 50 rows, and pipelines that belong in the feed are absent from it.

## Fix

Sort by `created`, which the UI already uses as the ordering key, with `id` as a tie-breaker — `created` has second resolution, so pipelines started together share a value and would have uncertain order otherwise.

## Test

`TestUserFeedOrdersByCreated` stores three pipelines whose creation times ascend while their ids descend. Against the previous behavior it fails with the feed exactly reversed, oldest first.

### DIFF
--- a/server/store/datastore/feed.go
+++ b/server/store/datastore/feed.go
@@ -64,7 +64,7 @@ func (s storage) UserFeed(user *model.User) ([]*model.Feed, error) {
 		Join("INNER", "perms", "repos.id = perms.repo_id").
 		Join("INNER", "pipelines", "repos.id = pipelines.repo_id").
 		Where(userPushOrAdminCondition(user.ID)).
-		Desc("pipelines.id").
+		Desc("pipelines.created", "pipelines.id").
 		Limit(perPage).
 		Find(&feed)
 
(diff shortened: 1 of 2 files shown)


qwerty287:
How many PRs of this type do you plan to open?

stevapple:
> How many PRs of this type do you plan to open?

All the discovered are now opened as PRs, so the 2 open (excluding the deduplication one) and 1 merged.

Sorry if you regard this as some kind of spamming. I decided not to mix them because although the cause is similar, the solutions are different and some needs to be discussed and agreed on.

qwerty287 (approved):

PR #7015, opened 2026-08-16
server: drop stale queued tasks for terminal workflows

DEATIN:
## What does this PR do?

Prevents stale tasks from being handed to agents when their workflow is already
in a terminal state or no longer exists.

The persistent queue can contain a task that is still present in the task store
while the corresponding workflow has already been completed through another
path. In that case, the task may be polled again, `RPC.Init` rejects it because
the workflow is already finished, and the task can later be resubmitted after
its lease expires.

This change makes the persistent queue verify the workflow before returning a
task to an agent.

It also ensures that `persistentQueue.Done()` removes the task from the
persistent task store.

## Changes

- drop queued tasks whose workflow no longer exists;
- drop queued tasks whose workflow is already in a terminal state;
- remove completed tasks from the persistent task store in `Done()`;
- keep active/pending workflow tasks working as before;
- add regression tests for:
  - missing workflows;
  - success;
  - failure;
  - killed;
  - canceled;
  - skipped;
  - error;
  - declined;
  - persistent task cleanup on `Done()`.

## Why?

This prevents stale queue entries from being repeatedly assigned to agents and
avoids a loop where an already-finished workflow is polled, rejected by
`RPC.Init`, and later resubmitted.

This complements #6765, which handles the opposite stale-queue case where the
task is already missing from the persistent store but is still present in the
in-memory queue.

## Testing

Added unit tests in `server/queue/persistent_test.go` covering terminal workflow
states, missing workflows, normal active workflow polling, and task-store
cleanup after `Done()`.

### DIFF
--- a/server/queue/persistent.go
+++ b/server/queue/persistent.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 
 	"github.com/rs/zerolog/log"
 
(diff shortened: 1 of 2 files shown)


6543 (approved):

6543:
note: the whole persistant queue logic etc will be refactored in fafour of a scheduler ... but for now this fix is a good short-term solution

PR #7016, opened 2026-08-16
k8s backend: fix step example using incorrect "detached" property

Palleas:
I was following the kubernetes backend documentation and noticed that my "service" (docker:dind) wasn't actually running in the background. I realized that's because the examples uses `detached: true` instead of **detach** (I've verified [here](https://github.com/woodpecker-ci/woodpecker/blob/f1ae0817bd282bde23cebf44f14cf382b15b22cf/pipeline/backend/types/step.go#L25))


6543 (approved):

6543:
thx!

PR #7023, opened 2026-08-17
Add hidden `decode-base64` agent command for Windows local backend

confusedsushi:
# Add hidden `decode-base64` agent command for Windows local backend

## Summary

Add a hidden internal `decode-base64` subcommand to `woodpecker-agent` and use it when generating Windows `cmd.exe` scripts in the local backend.

## Motivation

The local Windows backend currently needs to print each command with a leading `+ ` before executing it.

Escaping arbitrary command lines for `cmd.exe` is difficult and error-prone. The existing implementation uses POSIX shell escaping and additional replacements for characters such as `&` and `|`, which does not accurately reflect Windows command-line syntax and can alter the displayed command.

Instead, this change Base64-encodes the trace output when generating the batch file and lets the already-running Woodpecker agent decode it again:

```text
woodpecker-agent decode-base64 <encoded command>
<actual command>
```

This keeps the batch file itself free of arbitrary command text in the tracing statement, while preserving the original command exactly in the output.

## Changes

* Add a hidden `woodpecker-agent decode-base64` subcommand.
* The subcommand decodes a single Base64 argument and writes the decoded bytes directly to stdout.
* Use `os.Executable()` in the local Windows backend to invoke the same agent executable that generated the batch file.
* Encode `+ <command>\n` using standard Base64 before adding the trace line to the generated batch file.
* Remove the previous `shellescape`-based workaround for Windows command tracing.

The command is intentionally hidden because it is an implementation detail of the local backend rather than part of the public agent CLI.


6543 (approved):
lgtm from my side ... but i could be biased ...

6543:
@qwerty287 I had discussions internally where i am employed and this was the best solution for the cmd problems ...

... but i want to get the bias and potential interest conflict away just look at it

cmd/agent/core/run.go
```diff
@@ -39,6 +55,12 @@ func GenApp(backends []backend_types.Backend) *cli.Command {
 			Usage:  "ping the agent",
 			Action: pinger,
 		},
+		{
```
6543, line 43: uh! sorry did forgot that `cli exec` needs this too ... so the base64 decoder func should move into shared package (/cmd/shared/tools.go) and add the subcommand to cli too ...

6543 (changes requested):
sorry

qwerty287:
I don't use windows at all...

cmd/agent/core/run.go
```diff
@@ -39,6 +40,12 @@ func GenApp(backends []backend_types.Backend) *cli.Command {
...
+			Name:   "decode-base64",
+			Usage:  "decodes a base64 string",
+			Hidden: true,
```
6543:
```suggestion
			Hidden: true, // internal helper for local backend and cmd
```

cmd/cli/app.go
```diff
@@ -49,6 +50,12 @@ func newApp() *cli.Command {
 	app.Commands = []*cli.Command{
 		admin.Command,
 		context.Command,
+		{
+			Name:   "decode-base64",
+			Usage:  "decodes a base64 string",
+			Hidden: true,
```
6543:
```suggestion
			Hidden: true, // internal helper for local backend and cmd
```

6543 (approved):

PR #7029, opened 2026-08-17
Reject invalid started and done workflow states

lafriks:
Add additional checks to prevent incompatible agents reporting invalid workflow state to create strange results in database/UI that can not be traced back to what is happening and why workflows are failing (when they actually are not).


6543 (approved):
this is more about hardening the interface but I'm all for it :)

PR #6998, opened 2026-08-13
fix(runtime): filter out cancelled workflows from runtime errors

zsxh1990:
## Problem

Cancelled pipelines were being reported as workflow errors. The error check in `workflow.go` only filtered out step failures (non-zero exit codes, OOM kills) via `IsStepFailure()`, but the `ErrCancel` signal from context cancellation is an expected termination that should not surface as a runtime error.

This caused cancelled steps to create workflow error entries in the database.

## Fix

Added `errors.Is(err, pipeline_errors.ErrCancel)` to the filter alongside `IsStepFailure()` so that cancelled workflows terminate cleanly without creating error entries.

Closes #6935

## References

- Maintainer comment: https://github.com/woodpecker-ci/woodpecker/issues/6935#issuecomment-5149752173
- Related PR: https://github.com/woodpecker-ci/woodpecker/pull/6904

PR #7031, opened 2026-08-17
fix runtime to ignore cancel as runtime error

### LINKED ISSUE #6935: Cancelled steps create workflow error
### Component

agent

### Describe the bug

https://ci.woodpecker-ci.org/repos/3780/pipeline/36458/errors

https://github.com/woodpecker-ci/woodpecker/pull/6904#issuecomment-5143600786

cc @6543 

### Steps to reproduce

cancel a running pipeline

### Expected behavior

_No response_

### System Info

```shell
ci.woodpecker-ci.org
```

### Additional context

_No response_

### Validations

6543:
Once the workflow context is done every backend call is aborted mid flight, and the error it returns describes that abort rather than a step failure. The shape of that error is backend specific: the local and kubernetes backends surface context.Canceled, but an HTTP based backend surfaces context.Cause, so a canceled docker step ended up with the raw transport error as its step error, which the server stores and the UI shows.

Decide on the workflow context instead of on the error shape, and log the original error rather than reporting it. The step start failure paths did not normalize at all and now do too.

Timed out workflows take the same path, which matches how the workflow itself already reports a deadline as a cancellation.

close #6935
replace https://github.com/woodpecker-ci/woodpecker/pull/6998


pipeline/runtime/workflow.go
```diff
@@ -85,7 +85,9 @@ func (r *Runtime) Run(runnerCtx context.Context) error {
 	// A failing step is a successfully executed task from the runtime's point of
 	// view. It is reported through the step states, which make the workflow fail,
 	// so it must not surface as a runtime error. Use Err() to get it anyway.
-	if err := r.err.Get(); !pipeline_errors.IsStepFailure(err) {
+	// Also errors caused by cancellation are none workflow-wise.
+	if err := r.err.Get(); !pipeline_errors.IsStepFailure(err) ||
```
6543:
```suggestion
	if err := r.err.Get(); !pipeline_errors.IsStepFailure(err) &&
```

6543:
## before
image

## with pull
image

qwerty287 (approved):
untested

PR #7032, opened 2026-08-17
Enforce unique name for services and steps if dag is used

6543:
Currently on duplicated names we have undefined behaviour and depending on the backend this will fail or be silently ignored and produce random outcome.

This gates against all of it by returning an error with explicit message instead.

### DIFF
--- a/pipeline/frontend/yaml/linter/linter_test.go
+++ b/pipeline/frontend/yaml/linter/linter_test.go
@@ -217,6 +217,14 @@ func TestLintErrors(t *testing.T) {
 			from: "steps: { build: { image: golang }, publish: { image: golang, depends_on: [ binary ] } }",
 			want: "One or more of the specified dependencies do not exist",
 		},
+		{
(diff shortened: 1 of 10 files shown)


pipeline/frontend/yaml/linter/linter_test.go
```diff
@@ -217,6 +217,14 @@ func TestLintErrors(t *testing.T) {
...
 			want: "One or more of the specified dependencies do not exist",
 		},
+		{
+			from: "{steps: { build: { image: golang } }, services: [ { name: database, image: mysql }, { name: database, image: postgres } ] }",
+			want: "Service names must be unique, `database` is used more than once",
+		},
+		{
+			from: "steps:\n  build:\n    image: golang\nservices:\n  database:\n    image: mysql\n  database:\n    image: postgres\n",
+			want: "Service names must be unique, `database` is used more than once",
+		},
```
6543:
```suggestion
```

qwerty287 (approved):
untested

lafriks (approved):

PR #7036, opened 2026-08-18
Don't inject matrix env vars into plugins

qwerty287:
This allows to work around the plugin isolation currently.


lafriks:
can you describe it a bit more in detail why would plugins have different env than non-plugin steps?

qwerty287:
> can you describe it a bit more in detail why would plugins have different env than non-plugin steps?

You can possibly influence how the plugin behaves if you set custom env vars and then you may be able to steal credentials. It should only be possible to use the options defined by plugins in the `PLUGIN_*` env var namespace set via `settings`.

That's also the reason why the `environment` key is not allowed on plugins – but using matrix env vars you can work around this right now (see https://woodpecker-ci.org/docs/usage/plugins/overview#plugin-isolation).

lafriks:
Ok, thanks for describing problem

lafriks (approved):

6543 (approved):

PR #7043, opened 2026-08-20
fix(ci): serialize datastore package tests

6543:
Migration tests reset the shared database schema while Go runs package tests concurrently. Serial package execution prevents unrelated datastore tests from losing tables mid-run.

fix https://ci.woodpecker-ci.org/repos/3780/pipeline/37137/37
image


qwerty287:
But don't they run serially anyways because we have two commands there, one only runs migration test and the other one only runs the non-migration tests?

We have

> go test -p 1 -timeout 300s -tags 'test $(TAGS)' -run TestMigrate go.woodpecker-ci.org/woodpecker/v3/server/store/...

first, only migration

> go test -p 1 -race -timeout 120s -tags 'test $(TAGS)' -skip TestMigrate go.woodpecker-ci.org/woodpecker/v3/server/store/...

second, only non-migration

So they should not run in parallel anyways, right? Or what do I miss?

6543:
go tooling does run them in parallel esp. if it detects multible cores!

yes the two different test scope dont run in paralell but the subtests can do.

qwerty287 (approved):
ok if that's the issue

PR #7044, opened 2026-08-20
Introduce token for scoped access

6543:
This adds tokens to woodpecker for scoped access.

currently this is only used for badges to gate private repos ... but more could come.

dRAFT because only spec drafted, ai coded and heavy reviewed ... but no real execution done jet

### DIFF
--- a/shared/utils/random.go
+++ b/shared/utils/random.go
@@ -1,0 +1,49 @@
+// Copyright 2026 Woodpecker Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import (
+	"encoding/base32"
+	"math"
+
+	"github.com/tink-crypto/tink-go/v2/subtle/random"
+)
+
+// base32NoPadding keeps generated strings free of the '=' padding, so they can
+// be used in URLs and headers without escaping.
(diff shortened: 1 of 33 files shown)


shared/utils/random.go
```diff
@@ -0,0 +1,51 @@
...
+	bitsPerBase32Char     = 5
+	bitsPerByte           = 8
+	maxRandomStringLength = uint64(math.MaxUint32) * bitsPerByte / bitsPerBase32Char
+)
+
+// RandomString returns a cryptographically secure random string of exactly
+// length characters, drawn from the base32 alphabet. It is the single place
+// tokens, secrets and hashes are generated, so their strength does not depend
+// on every call site getting the encoding right.
+func RandomString(length int) string {
```
6543, line 39: this helper we should switch ussage in extra pull of e.g. agent secret generation etc... too it

qwerty287:
Thanks, it's a good addition in general (we also have #665), but it does not really fix the issue I mentioned on matrix. We already have an API token (which is non-scoped) so we can just use this token to protect the cctray endpoint

6543:
the token we have grans full access to the account it belongs to ... that's a bit to much to ask for embeding in cctray or badge i would say

qwerty287:
I'm not sure whether badge should get a token at all. Do you think it really matters if I can link between last pipeline status and repo ID? It's a bit user-unfriendly to always require a token when inserting a badge into a readme I think.
CCTray is an app that you usually install locally on your PC, so I think having a full-API token there is ok generally...

6543:
the token is optional as long as the repo is public ... and if wanted we can also just not have one for public repos

qwerty287:
That's not my point.

I'm totally in favor of merging this PR if it's finished.

But I think we could already add authentication to the cc.xml endpoint (this would be breaking though I think...). Yes, we only have the full-api token, but the CC endpoint is used for this CCTray application which is a local app. You can just put your token in there, it's as safe as using our CLI.

But as this would be breaking change it's probably better to change this at a later point.

6543:
this pull was made to have different scoped tokens later on.

why not add user tokens ... 

... we need tl add different indexes and checks in code to ensure some tokens exist only per repo while others can have multible ...

PR #6804, opened 2026-07-01
fix(web): allow dashboard header layout to wrap gracefully on smaller…

mayankdev-oss:
This PR fixes a layout bug on the repository dashboard where the action buttons overlap and clip into the search bar text field on narrower viewport sizes

By adding md:flex-wrap and switching to separate layout spacing variables md:gap-x-10 md:gap-y-2 on the main header row inside Header.vue the layout elements can now wrap into multiple rows safely if screen real estate runs tight. This keeps the UI completely accessible and looking great on smaller laptop or tablet screens while the team continues discussing long term button locations
Related Issue

Fixes #6739
Checklist

    [x] Added responsive flex wrapping utility classes

    [x] Separated horizontal and vertical wrap spacing boundaries

    [x] Verified layout scaling behavior on responsive breakpoints

PR #7047, opened 2026-08-20
Prevent repository buttons from overlapping search bar

### LINKED ISSUE #6739: Button overlap search bar
Image

Version: "next-b8dd9854a5"

xoxys:
Minimal fix for: https://github.com/woodpecker-ci/woodpecker/issues/6739
Supersedes: https://github.com/woodpecker-ci/woodpecker/pull/6804

image

image

image


qwerty287:
Not sure, but this could break in translations. E.g. German "Hinzufügen" is way longer than "Add" (yes we could also say "Neues" but I guess there are languages without good alternative)

xoxys:
OK lets go with the icon button approach from you, then? I have updated the screenshot and code.

qwerty287 (approved):
Thanks!

PR #7049, opened 2026-08-20
Docker: give SQLite a writable temp dir

### LINKED ISSUE #7020: Migration `deduplicate-log-entries` broken on scratch-image using sqlite
Image

see https://github.com/woodpecker-ci/woodpecker/pull/7004#issuecomment-5311021892

6543:
Migration deduplicate-log-entries (#7004) runs a
DELETE whose plan needs a temp b-tree, which spills to a temp file. SQLite looks for one in SQLITE_TMPDIR, TMPDIR, /var/tmp, /usr/tmp, /tmp, then CWD. The server image is built FROM scratch, sets no WORKDIR and runs as uid 1000, so none of those exist or are writable and startup fails with "disk I/O error: permission denied".

Point SQLite at the data volume. It is writable by definition, since the DB and its rollback journal already live there.

close https://github.com/woodpecker-ci/woodpecker/issues/7020

### DIFF
--- a/docker/Dockerfile.server.alpine.multiarch.rootless
+++ b/docker/Dockerfile.server.alpine.multiarch.rootless
@@ -11,6 +11,7 @@ ENV GODEBUG=netdns=go
 ENV WOODPECKER_IN_CONTAINER=true
 ENV XDG_CACHE_HOME=/var/lib/woodpecker
 ENV XDG_DATA_HOME=/var/lib/woodpecker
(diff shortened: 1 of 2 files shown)


6543:
tested on our infra ... after convirmation i reverted our db back to commit state ... to make sure once this is merged we get propper convirmation ...

qwerty287 (approved):
untested

PR #7067, opened 2026-08-24
Makefile enhancement

6543:
if no docker is installed ... ignore it
image

and use `go run ...` where possible to not spam ~/go/bin and have version conflicts


6543:
now:
image
image

### also usefull inside docker:
image

6543:
image

qwerty287:
I personally do not really like that this recompiles all the tools like gofumpt etc. all the time (yes there's the build cache, but you need to build it once).
I have those tools installed globally on my system and would like to use the global installations…

6543:
Ok in this case i will have a if exist use else use run

qwerty287 (approved):

PR #7069, opened 2026-08-25
feat: Add pipeline token to authenticate against Woodpecker APIs

GioPan04:
This PR is currently more of a POC and it's not ready to be merged. I'm seeking suggestions and comments, if this feature is accepted by the maintainers I will refine it.

## Implementation

I like the concept of the `GITHUB_TOKEN` access token of GitHub's CI. This implementation is inspired by it.

Currently, in this implementation, whenever a pipeline is started, the server generates a secret (named `woodpecker_token`) and adds it along side of the other secrets. Both the user-id and the pipeline-id are included into the JWT claims.
The token is valid as long as the pipeline is running, but as a safeguard it also has an expiration of "timestamp of when it's started + repo timeout".
The scope of the token is the same of a normal user/session token.

## Why

I needed a way of seamlessly calling woodpecker APIs to start a deployment after a pipeline has finished executing. The [trigger](https://woodpecker-ci.org/plugins/trigger) plugin exists, but you still need to embed a user token.
With this PR the plugin would work out of the box, after the user has enabled the token in the repo settings.

### DIFF
--- a/server/pipeline/items.go
+++ b/server/pipeline/items.go
@@ -70,17 +70,27 @@ func parsePipeline(ctx context.Context, forge forge.Forge, store store.Store, cu
 		})
 	}
 
-	t := token.New(token.PipelineToken)
-	t.Set("user-id", strconv.FormatInt(user.ID, 10))
-	t.Set("pipeline-id", strconv.FormatInt(currentPipeline.ID, 10))
-	// TODO: If a pipeline is restarted, Timestamp is equal to the timestamp of start of the last pipeline.
-	tokenString, err := t.SignExpires(user.Hash, currentPipeline.Timestamp+repo.Timeout*60)
+	if len(repo.ExposeAccessTokenEvents) > 0 {
+		t := token.New(token.PipelineToken)
+		t.Set("user-id", strconv.FormatInt(user.ID, 10))
+		t.Set("pipeline-id", strconv.FormatInt(currentPipeline.ID, 10))
+		// TODO: If a pipeline is restarted, Timestamp is equal to the timestamp of start of the last pipeline.
+		tokenString, err := t.SignExpires(user.Hash, currentPipeline.Timestamp+repo.Timeout*60)
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate pipeline token: %w", err)
+		}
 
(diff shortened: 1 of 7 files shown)


server/pipeline/items.go
```diff
@@ -68,6 +70,28 @@ func parsePipeline(ctx context.Context, forge forge.Forge, store store.Store, cu
 		})
 	}
 
+	if len(repo.ExposeAccessTokenEvents) > 0 {
+		t := token.New(token.PipelineToken)
+		t.Set("user-id", strconv.FormatInt(user.ID, 10))
+		t.Set("pipeline-id", strconv.FormatInt(currentPipeline.ID, 10))
+		// TODO: If a pipeline is restarted, Timestamp is equal to the timestamp of start of the last pipeline.
```
GioPan04, line 77: This is the only issue that I'm facing that wasn't able to solve. I could just get the current timestamp, but I don't really like it

qwerty287:
Thanks, but personally I'm against adding this.

This is a full-API token for one specific user. If the user is an admin in WP, you would be able to do anything in WP.

Why can't you add your token from the user settings as secret to the repo?

GioPan04:
Hi, thanks for the reply.
My Woodpecker instance is used by different people, and asking them to put their own token inside of their settings is kind of counterintuitive.
On a general level, this would enable plugins like "trigger" to work out of the box.

Maybe I'm missing something, but I don't see how this could be exploited. The scope of the token is limited to the user that is starting the pipeline, so even if they want to do something bad, they can just do it with their own token.
The only issue would be if a pipeline prints out the token, but the same issue applies for any secret.

edit: I misread your last line, but the issue still stands. If I set my token into the repo settings, then everyone has access to the token, and I don't see how that would be better

6543:
https://github.com/woodpecker-ci/woodpecker/pull/7044 is written in a way to add more different types. i personally would go that route. still mentioned pull needs tweeking and some desissions are not jet made

GioPan04:
I didn't know about #7044. It would surely be a better starting point.
If I understand correctly that PR only implements the concept of a scoped token, but the feature of an automatically generated token for the pipeline is not in the scope of that PR.
When that PR will get merged I'll refactor this to use a scoped token.

Issue #6881, opened 2026-07-21
Add Agent specific runtime buildin env vars


6543:
example: CI_AGENT_LABELS , CI_AGENT_ID, CI_AGENT_NAME

as per https://github.com/woodpecker-ci/woodpecker/issues/6795#issuecomment-4822854809

they are only in runtime ...  (we documnet scope now: https://github.com/woodpecker-ci/woodpecker/pull/6879)

PR #7070, opened 2026-08-26
feat(agent): add CI_AGENT_ID and CI_AGENT_LABELS runtime env vars

ChrisJr404:
Adds two agent specific runtime environment variables to steps, as requested in #6881:

- `CI_AGENT_ID`, the server assigned ID of the agent running the step
- `CI_AGENT_LABELS`, the agent's labels rendered as a sorted, comma separated list of `key=value` pairs (e.g. `backend=docker,hostname=agent-1,platform=linux/amd64,repo=*`)

They are injected in the same spot as `CI_MACHINE` and `CI_SYSTEM_PLATFORM`, so they are runtime only and not available at config evaluation time. Both values are already known on the agent, so no extra server round trip is needed.

The issue also lists `CI_AGENT_NAME`. I left that one out for now because the agent name is assigned server side and is not sent back to the agent during registration (`RegisterAgentResponse` only returns the agent ID), so exposing it would need a proto and server change on top of this. Happy to do that as a follow up, or fold it into this PR if you'd prefer.

I sorted the labels so `CI_AGENT_LABELS` is stable across runs regardless of map ordering. If you'd rather have a different separator or format (JSON, space separated, etc.), just say the word and I'll adjust.

Added a unit test for the label formatting and updated the environment variables docs.


qwerty287:
Thanks and sorry for the late reply.

For other "complex" data, we usually use JSON (i.e. CI_PIPELINE_FILES or complex plugin settings data). What do you think about using JSON for the agent labels?

ChrisJr404:
JSON works for me and I agree it is more consistent with how CI_PIPELINE_FILES and the complex plugin settings are already exposed. I will change CI_AGENT_LABELS to emit a JSON object (the label map) instead of the flat key=value list, and keep CI_AGENT_ID as a plain scalar. Will push that shortly.

ChrisJr404:
Pushed. `CI_AGENT_LABELS` now emits a JSON object of the label map (keys sorted for determinism via `json.Marshal`), for example `{"backend":"docker","platform":"linux/amd64"}`, and an empty label set still renders as an empty string. Updated the env-var docs table and the unit test to match. `CI_AGENT_ID` stays a plain scalar.

qwerty287:
Can you check the linter? https://ci.woodpecker-ci.org/repos/3780/pipeline/37414/36#L4

qwerty287 (approved):

PR #7076, opened 2026-08-27
fix(server): include timezone data in container images

### LINKED ISSUE #6688: Cron schedules fail with “unknown time zone” in server container

luken:
Fixes #6688.

Both published server image variants currently lack IANA timezone data, so named Woodpecker cron timezones are rejected.

This keeps the change in container packaging only: both scratch and Alpine server images copy the same Go-distributed `zoneinfo.zip` and set `ZONEINFO` to it. That makes the cron scheduler's named `time.LoadLocation` values available consistently without a host mount or runtime package install.

Validation:
- `go test -tags external_web ./cmd/server -run 'Test.*Cron' -count=1`\n- inspected the actual `golang:1.27` archive: 598 zones, including `America/Denver`, `Europe/Berlin`, `US/Eastern`, and `Asia/Calcutta`\n- `git diff --check`\n\nClaude Code review completed in a focused loop; its accepted feedback is reflected here. Scheduler, `TZ`/local-time semantics, and broader CI changes were deliberately left out of this cron parsing fix.

### DIFF
--- a/docker/Dockerfile.server.alpine.multiarch.rootless
+++ b/docker/Dockerfile.server.alpine.multiarch.rootless
@@ -1,3 +1,5 @@
+FROM --platform=$BUILDPLATFORM docker.io/golang:1.27 AS timezone-data
+
 FROM docker.io/alpine:3.24
 
 ARG TARGETOS TARGETARCH
@@ -12,8 +14,11 @@ ENV WOODPECKER_IN_CONTAINER=true
 ENV XDG_CACHE_HOME=/var/lib/woodpecker
 ENV XDG_DATA_HOME=/var/lib/woodpecker
 ENV SQLITE_TMPDIR=/var/lib/woodpecker
+# Keep the .zip suffix: Go uses this IANA archive for named cron timezones.
+ENV ZONEINFO=/usr/share/zoneinfo.zip
 EXPOSE 8000 9000 80 443
 
+COPY --from=timezone-data /usr/local/go/lib/time/zoneinfo.zip /usr/share/zoneinfo.zip
 COPY dist/server/${TARGETOS}_${TARGETARCH}/woodpecker-server /bin/
 
 USER woodpecker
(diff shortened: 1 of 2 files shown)


qwerty287:
@lafriks is that the way you suggested how to do this?

6543:
tested it ...

emulated scratch/alpine (no `/usr/share/zoneinfo`, no `/etc/localtime`) with a go1.27 binary:

| | `LoadLocation("Europe/Berlin")` | `time.Local` with `TZ=Europe/Berlin` |
|---|---|---|
| current images | `unknown time zone Europe/Berlin` | UTC |
| this pull | ok, CEST +0200 | UTC |
| unziped tree at `/usr/share/zoneinfo` | ok | Europe/Berlin |

so the cron part works, zip is fine (go1.27 ships tzdata 2026c, 598 zones, stored uncompressed as go's zip loader needs).

but `$ZONEINFO` is only read by `LoadLocation`, not by `initLocal` ... so `TZ=Europe/Berlin` still gives UTC and that's half of #6688 still open.

so i would:
- alpine: just `apk add --no-cache tzdata` -> 182 KiB download, tzdata 2026c too, and it fixes `TZ` ... pulling the whole golang image for a 400 KB zip is a bit much
- scratch: copy the unziped tree to `/usr/share/zoneinfo` instead ... that's one of go's default search paths, fixes `TZ` as well and we can drop the env var

also `go test ./cmd/server -run 'Test.*Cron'` from the description matches no tests at all, there are none in `cmd/server` ... please dont list checks that did not run

6543 (changes requested):
please adjust

qwerty287:
#7093

PR #7078, opened 2026-08-28
fix(forge/gitlab): match group path when checking GitLab org membership

YouCD:
## Summary

GitLab groups have both a display `name` (e.g. `灰度研判检测中心`) and a URL `path` (e.g. `gray-eye`). Woodpecker stores the `path` as the org name, but `OrgMembership` only compared against the `name` field. When these differ, the membership check always fails with 403, even if the user is an Owner of the group.

## Changes

- `server/forge/gitlab/gitlab.go`: in `OrgMembership`, match against both `group.Name` and `group.Path` when looking up the group by owner name.

## Test Plan

- Verify that a user who is a member of a GitLab group where `name != path` can now access the org in Woodpecker (no more 403 on `/api/orgs/{id}`).

### DIFF
--- a/server/forge/gitlab/gitlab.go
+++ b/server/forge/gitlab/gitlab.go
@@ -706,7 +706,7 @@ func (g *GitLab) OrgMembership(ctx context.Context, u *model.User, owner string)
 	}
 	var gid int64
 	for _, group := range groups {
-		if group.Name == owner {
+		if group.Path == owner {
 			gid = group.ID
 			break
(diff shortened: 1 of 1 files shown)


qwerty287:
Thanks.
I think this should only compare about the `path`, not the display name. We only use the paths internally, using `name` is quite fragile as you can have multiple orgs with the same name.

YouCD:
Closing as discussed

Issue #4355, opened 2024-11-11
when.status:failure doesn't work


zc-devs:
### Component

server

### Describe the bug

Step with `when.status: [failure]` doesn't run on failed pipeline, instead runs step with `when.status: [success]`.

### Steps to reproduce

1. Woodpecker + some Forge
2. Make pipeline

```yaml
# build.yaml
skip_clone: true
steps:
  build:
    image: alpine
    commands:
      - echo 'Building the app'
      - exit 1 # 0 - success, 1 - fail
```
```yaml
# notifications.yaml
skip_clone: true
depends_on: [build]
runs_on: [success, failure]
steps:
  fail-notification:
    when:
      - status: [failure]
    image: alpine
    commands:
      - echo 'Build failed'
  success-notification:
    when:
      - status: [success]
    image: alpine
    commands:
      - echo 'Build succeed'
```
3. Run manually and get wrong step execution
![Screenshot 2024-11-11 1](https://github.com/user-attachments/assets/77cad464-54ff-4aef-996a-a18d6ff5ad33)

### Expected behavior

1. `success-notification` step runs
2. `fail-notification` skips

### System Info

```shell
WP `next-f87e80381b`, `2.7.1`, Gitea, Postgres 16
```

### Additional context

https://github.com/woodpecker-ci/woodpecker/issues/4337#issuecomment-2468498847

### Validations

- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]

pat-s:
It works within a pipeline but not across, hence the title should likely be updated to include this condition/detail.

PR #7091, opened 2026-09-02
Fix workflow-level when.status so a failure-only workflow does not run on success

dylanpulver:
`genItemForWorkflow` guards both appends with `IncludesStatusFailure`:

```go
if !slices.Contains(item.RunsOn, "failure") && parsed.When.IncludesStatusFailure(...) {
	item.RunsOn = append(item.RunsOn, "failure")
}
if !slices.Contains(item.RunsOn, "success") && parsed.When.IncludesStatusFailure(...) {   // <- success
	item.RunsOn = append(item.RunsOn, "success")
}
```

So `when: status: [ failure ]` on a workflow yields `RunsOn = [failure success]`, and `Task.ShouldRun` takes its first branch — `runsOnFailure() && runsOnSuccess()` returns true unconditionally — so the workflow is scheduled after a **green** dependency as well. A failure-only notification workflow fires on every successful pipeline.

Measured here, driving `Build()` and then the real `model.Task.ShouldRun()`:

```
before: when.status:[failure]  -> RunsOn=[failure success]  after success=true   after failure=true
after:  when.status:[failure]  -> RunsOn=[failure]          after success=false  after failure=true
```

The step-level path pairs them correctly (`compiler/convert.go:158-160`); the workflow-level analogue got the wrong half of the pair.

**The one-word swap is not the right fix, and this is the part worth reviewing.** `IncludesStatusSuccess` returns true for a constraint that sets no status at all (its own comment says so). With only the call changed, a workflow with the deprecated `runs_on: [ failure ]` next to a plain `when: event: push` gains `"success"` and starts running on green — the same bug, moved. Measured:

```
one-word swap: deprecated runs_on:[failure] -> RunsOn=[failure success]   <- regressed
this patch:    deprecated runs_on:[failure] -> RunsOn=[failure]
```

So `when.status` is only consulted when `runs_on` was not set. That also makes the two `slices.Contains` guards redundant, and `slices` is no longer imported.

Why nothing caught it: `TestRunsOn` is the only test that ever looks at `item.RunsOn`, and its fixture is `status: [ success, failure ]` — the single input where `IncludesStatusFailure` is true, so both appends fire and the assertion passes. `constraint_test.go` covers `status: [failure]` correctly; only the call site was untested. The new table adds failure-only, success-only, no-status, and the deprecated `runs_on` case.

Origin: both lines were added by `39a7ebda` (#6183), which `compare` puts behind v3.13.0 and inside v3.14.0.

Deliberately not claiming an issue: #4355 is a related cross-workflow `when.status` report from 2024, but it describes the opposite symptom (a failure step not running) and I have not checked whether this changes it. I did not find an existing report of this direction.

Ran `go test ./pipeline/... ./server/model/... ./server/queue/...`: 21 packages ok. The single failure is `pipeline/backend/dummy [build failed]`, which is build-constrained and fails identically on an unmodified worktree at the same commit.

Written with Claude Code (Claude Opus 5, `claude-opus-5`); every measurement above was run before opening this. Not verified end to end against a live server and agent — the chain is exercised in-process from `Build()` through `Task.ShouldRun()`, which is the only consumer of that predicate.


pipeline/frontend/builder/builder.go
```diff
@@ -187,11 +186,17 @@ func (b *PipelineBuilder) genItemForWorkflow(workflow *Workflow, axis matrix.Axi
...
 
-	if !slices.Contains(item.RunsOn, "failure") && parsed.When.IncludesStatusFailure(workflowMetadata, true, environ) {
-		item.RunsOn = append(item.RunsOn, "failure")
-	}
-	if !slices.Contains(item.RunsOn, "success") && parsed.When.IncludesStatusFailure(workflowMetadata, true, environ) {
-		item.RunsOn = append(item.RunsOn, "success")
+	// Derive from when.status only when the deprecated runs_on was not set,
+	// so that runs_on keeps deciding on its own. IncludesStatusSuccess is true
+	// for a constraint that sets no status at all, which would otherwise add
+	// "success" to an explicit runs_on: [failure].
```
qwerty287:
```suggestion
```

qwerty287 (commented):
Thanks, not sure how this could happen 🤔 

can you remove that comment though?

qwerty287 (approved):

PR #7093, opened 2026-09-03
Add timezone data to docker images

### LINKED ISSUE #6688: Cron schedules fail with “unknown time zone” in server container

lafriks:
Also optimize layers in final image

Closes #7076 Fixes #6688


qwerty287:
Thanks, but we only need that data in the server container, right? The agent doesn't need it

lafriks:
I added in all so you can set TZ also for agents and cli for example logging timestamps etc and so we don't step on problems in other places where timezone could be needed in the future

qwerty287 (approved):
ok yes that's a good idea

untested from my side

6543 (approved):

6543:
one note: legacy timezones are only present in alpine image now ... golang does not have them installed. we would have to do: `apt install -y tzdata-legacy` in the buildstage for scratch images

6543:
@lafriks 

```
# add backward compatible timezone names (US/Eastern, Asia/Calcutta, ...) to match alpine images
RUN apt update && \
  apt install -y --no-install-recommends tzdata-legacy && \
  rm -rf /usr/share/zoneinfo/right
```

Issue #4558, opened 2024-12-12
git reset fails sometimes when connecting to gitea forge


lunedis:
### Component

server, agent

### Describe the bug

I have been using woodpecker-CI for multiple years now in our company, and we rarely had any issues.

Ever since two days ago, pipelines _sometimes_ fail with an error connecting to the gitea forge:

```
+ git init --object-format sha1 -b main
Initialized empty Git repository in /woodpecker/src/GITEA_URL/ORGANIZATION/REPOSITORY/.git/
+ git config --global --replace-all safe.directory /woodpecker/src/GITEA_URL/ORGANIZATION/REPOSITORY
+ git remote add origin https://GITEA_URL/ORGANIZATION/REPOSITORY.git
+ git fetch --no-tags --depth=1 --filter=tree:0 origin +caf9334a3d53b3[2](https://CI_URL/repos/73/pipeline/471#L2)f4c382044ca4ff5d5982d4e58:
From https://GITEA_URL/ORGANIZATION/REPOSITORY
 * branch            caf9[3](https://CI_URL/repos/73/pipeline/471#L3)34a3d53b32f4c382044ca4ff5d5982d4e58 -> FETCH_HEAD
+ git reset --hard -q caf9334a3d53b32f4c382044ca4ff5d5982d4e58
fatal: unable to access 'https://GITEA_URL/ORGANIZATION/REPOSITORY.git/': Failed to connect to GITEA_URL port [4](https://CI_URL/repos/73/pipeline/471#L4)43 after 2002 ms: Could not connect to server
fatal: could not fetch 374dd4d87[5](https://CI_URL/repos/73/pipeline/471#L5)f23bb4a2e8b631762d933d12d78d27 from promisor remote
exit status 128
```

Restarting the pipeline often fixes the issue.

I am failing to find any reasons to why this is happening. It seems to connect to gitea normally before (it loads the pipeline, runs the first fetch, and it works on restart). I have updated to woodpecker-next and the issue stays the same. I have not updated neither woodpecker or gitea in the past few days (just a few weeks ago).

I had a similiar issue half a year ago, when my renovate-bot would open many pull requests simultaneously and it would fail with a "Cannot read username" message.

Is this a problem with woodpecker? Am I running into a gitea rate limit? I am not running fail2ban on my host.

### Steps to reproduce

Hard to reproduce sadly :(

### System Info

```shell
woodpeckerci/plugin-git: 2.6.0
[...]

PR #7104, opened 2026-09-04
Make oauth2 refresh time configurable

johanvdw:
Since oauth tokens are still fetched at the start of a pipeline (until #2851 ) because tokens run out, see eg #4558.

Setting the token refresh expiry time in the forge is insufficient: it will indeed decrease the chance that time is running out, but still relies on a hardcode refresh time.


johanvdw:
Note, I have marked this as a draft, because I noticed that even with this patch disabled, tokens for a gitea service could still stop after 15 minutes, which means that the minimum 30 minutes time is already not working correctly there (perhaps old tokens are rejected by gitea?).

Issue #4446, opened 2024-11-23
Agent stops taking jobs after server throws 5XX errors


aaronriedel:
### Component

agent

### Describe the bug

When the server (running in kubernetes) restarts my docker agent refuses to take new jobs until restarted. In the agent logs I can see several 5XX Errors while the server reboots. After that the agent shows as online in the UI but does not take jobs.

Agent logs: See below

### Steps to reproduce

1. Install Woodpecker server in Kubernetes
2. Install agent in seperate server using docker
3. Kill the server so that it recreates
4. Trigger pipeline that would use the docker agent
5. See it pending

### Expected behavior

The agent should properly reconnect to the Server via gRPC after the server restarts.

### System Info

```shell
Server:
`{"source":"https://github.com/woodpecker-ci/woodpecker","version":"2.7.3"}`

Helm values:

---
server:
  ingress:
    # -- Enable the ingress for the server component
    enabled: true
    # -- Add annotations to the ingress
    annotations:
      # kubernetes.io/ingress.class: nginx
      kubernetes.io/tls-acme: "true"
    hosts:
      - host: woodpecker.example.com
        paths:
          - path: /
            backend:
              serviceName: woodpecker-svc
              servicePort: 80
    tls:
      - hosts:
          - woodpecker.example.com
        secretName: woodpecker-tls-key
  statefulSet:
    replicaCount: 1
  env:
    WOODPECKER_ADMIN: 'aaron'
    WOODPECKER_HOST: 'https://woodpecker.example.com'
    WOODPECKER_OPEN: true
    WOODPECKER_FORGEJO: true
    WOODPECKER_FORGEJO_URL: 'https://git.example.com'
    WOODPECKER_LOG_LEVEL: "error"
  extraSecretNamesForEnvFrom:
    - woodpecker-forgejo
```

gRPC Ingress:
```
---
apiVersion: v1
kind: Service
metadata:
  name: woodpecker-grpc
  namespace: woodpecker
  annotations:
    traefik.ingress.kubernetes.io/service.serversscheme: h2c
spec:
  selector:
    app.kubernetes.io/instance: woodpecker
    app.kubernetes.io/name: server
  ports:
    - name: grpc
      protocol: TCP
      port: 9000
      targetPort: grpc
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/tls-acme: "true"
[...]

PR #6515, opened 2026-04-28
Refactor agent rpc retry

6543:
address https://github.com/woodpecker-ci/woodpecker/pull/6470#issuecomment-4313889128

PR #7111, opened 2026-09-06
fix(agent): count the retry window from the first failure, not from the start of a long poll

### LINKED ISSUE #7110: Agent retry window counts from the start of a long poll, so the first error after an idle Next() or a running Wait() gives up without a retry
### Component

agent

### Describe the bug

The window is passed to `backoff.WithMaxElapsedTime`, and `backoff.Retry` measures it from the moment it was called. `Next()` and `Wait()` are server-side long polls (`queue.Poll` / `queue.Wait` block until work arrives or the workflow finishes), so for an agent that has been idle, or a workflow that has been running, for longer than the window, the very first transport error already exceeds it. The agent then logs

```
{"level":"error","error":"backoff: maximum elapsed time exceeded (last error: rpc error: code = Unavailable desc = closing transport due to: connection error: desc = \"error reading from server: EOF\", received prior goaway: ...)","message":"grpc error: next(): code: Unavailable"}
{"level":"error","error":"backoff: maximum elapsed time exceeded (last error: ...)","message":"runner error, retrying..."}
```

without a single retry.

* `Next()`: in multi-workflow mode the runner falls into its 5-second "runner error, retrying..." loop and eventually reconnects, so the configured window is simply never applied. In single-workflow mode the runner exits with "runner done with error".
* `Wait()`: `runner.go` treats a `Wait()` error as fatal and cancels the workflow context, so a server restart during a step that has been running for longer than `WOODPECKER_RETRY_TIMEOUT` aborts the pipeline instead of retrying the RPC.
* Whether the agent shuts down afterwards depends on which error happened to be the last one: only `errNotConnected` from the `IsConnected()` guard is translated into `ErrConnectionLost`, a transport-level `Unavailable` at the end of the window is not. Two agents hitting the same server restart took different paths here (one shut down and was restarted by Kubernetes, the other stayed in the runner retry loop).

**Root cause.** `agent/rpc/client_grpc.go`, `retryOpts`: `backoff.WithMaxElapsedTime(c.connectionRetryTimeout)`; `backoff.Retry` (cenkalti/backoff v7) checks `time.Since(startedAt)+next > MaxElapsedTime` with `startedAt` taken before the first operation call. The retry window should start at the first failure. I have a patch with tests ready and will open a PR referencing this issue.

### Steps to reproduce

1. Server and agent v3.18.0 (Kubernetes backend, one agent with `WOODPECKER_MAX_WORKFLOWS=1`, defaults otherwise).
2. Leave the agent idle for more than 2 minutes.
3. Restart the server pod.
4. Agent log shows the two lines above at the moment the server drops the connection.

For the `Wait()` case start a workflow with a step that sleeps for 5 minutes and restart the server after 3 minutes.

### Expected behavior

After a server restart the agent keeps retrying `Next()` / `Wait()` for `WOODPECKER_RETRY_TIMEOUT` (default 2m) and reconnects, or shuts down with "connection to server lost" once that window has elapsed. A running workflow is not affected by a server restart that is shorter than the window.

### System Info

```
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.18.0"}
```

Helm chart 3.7.3, Kubernetes backend, k3s on arm64.

### Additional context

Related: #4446 (agent stops taking jobs after a server restart), #6414 (introduced the retry window).

### Validations

- [x] Read the docs.
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already (agent/rpc/client_grpc.go is identical on main and v3.18.0).

thomas-baumeister:
### Problem

`WOODPECKER_RETRY_TIMEOUT` (default 2m) is passed to `backoff.WithMaxElapsedTime`, and `backoff.Retry` measures that time from the moment it was called. `Next()` and `Wait()` are server-side long polls (`queue.Poll` / `queue.Wait` block until work arrives or the workflow finishes), so the operation itself legitimately runs for far longer than the window before anything fails. The first transport error after such a poll, typically a server restart, therefore exceeds the window immediately and gives up **without a single retry**.

Reproduced with v3.18.0 (Gitea forge, one agent with the docker backend, defaults otherwise): agent idle for 4 minutes, then the server is stopped for 20 seconds:

```
{"level":"error","error":"backoff: maximum elapsed time exceeded (last error: rpc error: code = Unavailable desc = closing transport due to: connection error: desc = \"error reading from server: EOF\", received prior goaway: code: NO_ERROR, debug data: \"graceful_stop\")","message":"grpc error: next(): code: Unavailable"}
{"level":"error","error":"backoff: maximum elapsed time exceeded (last error: ...)","message":"runner error, retrying..."}
```

* `Next()`: a spurious error and a 5-second detour through the runner's "runner error, retrying..." loop (multi-workflow mode); in single-workflow mode the runner exits with "runner done with error". The configured retry window is never applied.
* `Wait()`: `runner.go` treats a `Wait()` error as fatal and cancels the workflow context. In the reproduction below a 5-minute step whose server was restarted after 150 seconds was killed within 15 seconds of the restart.

### Fix

`failureWindowBackOff` wraps the exponential policy and starts the clock at the **first failure of a streak**; once `WOODPECKER_RETRY_TIMEOUT` has elapsed since then it returns `backoff.Stop`. An attempt that outlived the window (a long poll that reconnected and broke again later) starts a new streak instead of extending the old one, so a workflow that survives one server restart is not caught by the next. `backoff.WithMaxElapsedTime(0)` disables the library's own start-of-call limit (its default would otherwise be 15 minutes).

Deliberately unchanged: which errors are treated as "connection lost" (only the `IsConnected()` guard leads to `ErrConnectionLost`), the give-up cause (`backoff.ErrMaxElapsedTime`, so log readers and `errors.Is` callers see the same thing as before), `WOODPECKER_RETRY_TIMEOUT=0` meaning infinite, and the runner's handling of `ErrConnectionLost` (#6414).

### Tests

* `TestFailureWindowBackOff` (fake clock, no sleeps): the window counts from the first failure, `Reset` starts over, an attempt that outlived the window starts a new streak that is itself bounded, a zero window never stops.
* `TestRetryRPCWindowStartsAtFirstFailure`: a first attempt that blocks longer than the whole window and then fails is retried and succeeds.
* `TestRetryRPCLongAttemptStartsNewFailureStreak`: failure, then an attempt that blocks longer than the window and fails, then success: three attempts, no give-up.
* `TestRetryRPCGivesUpAfterWindow`: continuous `errNotConnected` ends in `ErrConnectionLost`; continuous `Unavailable` / `Unauthenticated` end in `ErrMaxElapsedTime`, each after at least one retry and close to the window.

`make lint` (golangci-lint v2.12.2) and `make test-agent` pass (race detector, 10 repetitions of the package).

### End-to-end check

Same stack as above, once with the unpatched v3.18.0 agent and once with this branch built into the agent image:

| scenario | unpatched v3.18.0 agent | this branch |
|---|---|---|
| A: agent idle for 150 s, server stopped for 20 s | `backoff: maximum elapsed time exceeded` on `next()`, `runner error, retrying...`; reconnects via the runner loop, takes work afterwards | no error line, only the `waiting for server connection` warnings; reconnects, takes work afterwards |
| B: 5-minute step, server stopped for 20 s after 150 s | `backoff: maximum elapsed time exceeded` on `wait()`, `server returned unexpected err while waiting for workflow to finish run`, pipeline **killed** 15 s after the restart | `wait()` retries through the restart, pipeline **success** after the full 300 s |

One observation from B that is independent of this change: after the restart the server no longer knows the running task (`extend(): queue: task not found`, logged every lease renewal), yet `done()` is accepted and the pipeline finishes with the correct state.

Fixes #7110.

Related: #4446 (agent stops taking jobs after a server restart), #6414 (introduced the retry window).


qwerty287:
Thanks, but I'm not sure whether this is a bug at all.

The retry timeout was only intended for failures on startup I think and not for failures once it's up and running...

In any case, I would like to avoid writing our own backoff logic. Maybe it would be possible to implement the feature in the backoff library?

thomas-baumeister:
Thanks for looking at it. Two things on the "is this a bug" question.

**The documented contract is the run-time one.** The flag usage in `cmd/agent/core/flags.go` and `docs/docs/30-administration/10-configuration/30-agent.md` both describe `WOODPECKER_RETRY_TIMEOUT` as how long the agent keeps retrying to reconnect *after the gRPC connection is lost*. There is also no separate start-up path in the client: `retryRPC` is the same loop for `next`, `wait`, `init`, `done`, `extend`, `update`, `log` and `report_health`.

**And that is how it was implemented when the timeout was introduced.** #6414 tracked `connectionLostAt` and gave up only after that much *continuous* disconnection ("Start timeout window when connection is lost", "Exit after 2 minutes of continuous disconnection"). The refactor in #6515 replaced it with `backoff.WithMaxElapsedTime(connectionRetryTimeout)`, which the library measures from the moment `backoff.Retry` is called. For the short RPCs the two are almost the same, so nothing looks different there. For the two long polls it inverts the behaviour: if `Next()` has been idle longer than the window, or `Wait()` has been sitting on a running step longer than the window, the first transport error is already past the deadline and `Retry` returns without a single retry. So this PR restores the semantics of #6414 rather than adding a new feature.

The practical effect today is that the outcome of a server restart depends on how long the agent had been polling, not on how long the server is away: an agent that polled for one minute survives a 20 second restart, an agent that polled for three minutes does not. And in the `Wait()` case `runner.go` turns that error into `cancelWorkflowCtx(err)`, so a workflow that is running fine gets killed while the server is already coming back.

On the backoff library: the wrapper does not implement any backoff maths. Every interval still comes from `backoff.NewExponentialBackOff()`; the wrapper only decides when to return `backoff.Stop`. That is what the `BackOff` interface exists for, and `WithMaxElapsedTime` is explicitly documented as counting "from when Retry is called", so the library does not consider this a defect on its side. I am happy to open an issue or PR at cenkalti/backoff for a first-failure window (either an option or a decorator shipped with the library), and once that lands the wrapper here becomes a one-liner. It would need a library release though, so I would rather not block the agent fix on it.

If you would still rather see this solved somewhere else than in a wrapper here, tell me which shape you want and I will rework the PR.

qwerty287:
I wasn't saying that there's a bug in the library, but this could be implemented as a feature there as well. I think there are more usecases were this could be helpful. (I don't know of course whether they would accept this feature, but without asking we won't know)

So I prefer adding this in the cenkalti/backoff library.

thomas-baumeister:
Opened https://github.com/cenkalti/backoff/issues/188 for the library side, with a reproducer against v7.0.0 and the possible shapes a first-failure window could take there.

Whatever comes of it needs a release of the library, so I would still like to fix the agent here in the meantime, in whichever form you prefer.

qwerty287:
Let's wait a few days for the maintainer of the lib.

We don't need to wait for a release, we can also use a version from a commit. So this only needs to get implemented if the maintainer accepts it.

PR #7117, opened 2026-09-07
Log an expected log-stream close failure at debug instead of error

mattwilkinsonn:
`logger.Close` returns `logging.ErrNotFound` for a step that never opened a log
stream, which is the case for any step that was killed or never started. The
RPC done handler logs that at error level (`server/rpc/rpc.go:413`), so
finishing a workflow that contains such a step writes
"done: cannot close log stream for step N" with "stream: not found" to the
server log.

This checks for `ErrNotFound` and logs it at debug. Other close failures still
log at error.

Tests are in `server/logging/log_test.go`:
`TestCloseUnopenedStreamReturnsErrNotFound` covers the case the new branch
handles, `TestCloseOpenStreamSucceeds` covers the normal path, and
`TestTailUnopenedStreamReturnsErrNotFound` covers the neighbouring lookup.
They assert the sentinel is reachable through `errors.Is`, which is what the
change depends on.

`Close` has only the two outcomes, so the tests are exhaustive for sequential
use. It does drop the lock between the stream lookup and the map delete, so
concurrent `Close` calls on the same step are a separate question that this
change does not touch.


qwerty287 (approved):
Thanks!

qwerty287:
Sorry I just saw something I would like to change:
Your tests are in server/logging/close_test.go but it would be nice to have them in the already existing file server/logging/log_test.go

mattwilkinsonn:
> Sorry I just saw something I would like to change: Your tests are in server/logging/close_test.go but it would be nice to have them in the already existing file server/logging/log_test.go

Moved, and deleted close_test.go

6543 (approved):

PR #7118, opened 2026-09-07
Report an in-setup pipeline status as pending on all forges

mattwilkinsonn:
Every forge maps a Woodpecker status to its own commit-status state through a
small switch, and none of them handled `model.StatusCreated`. It fell through
the default case in all six, which reports a failure. A status report sent
while a pipeline is still being set up therefore posts red for a pipeline that
has not failed.

| Forge                 | Function                                                          | Default reported         |
| --------------------- | ----------------------------------------------------------------- | ------------------------ |
| GitHub                | `convertStatus`, `server/forge/github/convert.go:50`              | `statusError`            |
| Bitbucket             | `convertStatus`, `server/forge/bitbucket/convert.go:38`           | `statusFailure`          |
| Bitbucket Data Center | `convertStatus`, `server/forge/bitbucketdatacenter/convert.go:29` | `BuildStatusStateFailed` |
| Gitea                 | `getStatus`, `server/forge/gitea/gitea.go:618`                    | `StatusFailure`          |
| Forgejo               | `getStatus`, `server/forge/forgejo/forgejo.go:616`                | `StatusFailure`          |
| GitLab                | `getStatus`, `server/forge/gitlab/status.go:24`                   | `Failed`                 |

Each now lists `StatusCreated` in its pending case. `StatusCreated` is
internal-only and never terminal, so pending is the accurate report.

On GitHub, `StatusKilled` and `StatusError` were also reaching the default;
they are now explicit cases returning error, which is what they already
produced.

Every default stays where it was. A value outside the enum is a bug, and
reporting a bug as pending is worse than reporting it as a failure: on a
required check, a state that never resolves blocks the pull request with
nothing to show for it, where a failure is at least visible and actionable.
GitHub's default now logs the unhandled value before returning error, so the
bug is traceable rather than silent.

`GetPipelineStatusDescription` in `server/forge/common/status.go` had the same
gap, and `StatusCanceled` and `StatusSkipped` were falling through to
"unknown status" there. Canceled and skipped get their own text, since gitea's
`getStatus` sends both to `StatusFailure` and describing them as pending would
pair a red failure mark with "Pipeline is pending". `StatusCreated` shares the
pending case, which is what it is.

Tests:

- `Test_convertStatus` in `server/forge/github/convert_test.go` covers every
  enum member plus an out-of-enum value asserting the default still reports
  error.
- `server/forge/common/utils_test.go` covers the description for each status.
- `server/forge/gitea/status_test.go` and
  `server/forge/forgejo/status_test.go` are new. Neither forge had any test
  for its conversion; each covers the full enum plus an out-of-enum value.
- The existing tables in `server/forge/bitbucket/convert_test.go`,
  `server/forge/bitbucketdatacenter/convert_test.go` and
  `server/forge/gitlab/status_test.go` gained the `StatusCreated` case.

Reverting `StatusCreated` in any one of the six turns that forge's test red.


qwerty287:
Thanks!

This affects other forges as well I think - would you mind to change that as well so we fix it for all?

mattwilkinsonn:
> Thanks!
> 
> This affects other forges as well I think - would you mind to change that as well so we fix it for all?

Yep - just extended to all of them.

server/forge/common/status.go
```diff
@@ -71,6 +71,13 @@ func GetPipelineStatusDescription(status model.StatusValue) string {
 		return "Pipeline is pending approval"
 	case model.StatusDeclined:
 		return "Pipeline was rejected"
+	case model.StatusCanceled:
+		return "Pipeline was canceled"
+	case model.StatusSkipped:
+		return "Pipeline was skipped"
+	// StatusCreated is internal-only and never terminal.
+	case model.StatusCreated:
```
6543: can you add it to the statusPending line L60 ...

server/forge/github/convert.go
```diff
@@ -49,13 +49,17 @@ const (
...
+	case model.StatusPending, model.StatusRunning, model.StatusBlocked, model.StatusSkipped, model.StatusCanceled, model.StatusCreated:
 		return statusPending
 	case model.StatusFailure, model.StatusDeclined:
 		return statusFailure
+	case model.StatusKilled, model.StatusError:
+		return statusError
 	case model.StatusSuccess:
 		return statusSuccess
 	default:
+		// A value outside the enum is a bug; report it loudly.
```
6543: well should we log it then!?!

server/forge/forgejo/forgejo.go
qwerty287, line 632: This also should log an error then

server/forge/forgejo/status_test.go
qwerty287, line 1: Please put this in forgejo_test.go (to have it in the test file linked to the actual source file)

server/forge/gitea/gitea.go
qwerty287, line 634: Also log here

server/forge/gitea/status_test.go
qwerty287, line 1: Same here, put in gitea_test.go

server/forge/gitlab/status.go
qwerty287, line 39: also log here

qwerty287 (commented):
Besides these it looks good to me now.

server/forge/forgejo/status_test.go
mattwilkinsonn: Moved

server/forge/gitea/status_test.go
mattwilkinsonn: Moved

qwerty287 (approved):

PR #7119, opened 2026-09-07
Fix crash rendering a pipeline with a stepless workflow

### LINKED ISSUE #7114: Pipeline view crashes when a workflow has zero steps
### Component

web-ui, server

### Describe the bug

A workflow can be persisted with zero steps. The API then omits its `children`
field, because `model.Workflow.Children` is tagged `json:"children,omitempty"`
(`server/model/workflow.go:32`):

```go
Children   []*Step           `json:"children,omitempty"   xorm:"-"`
```

The web UI declares that field as non-optional (`web/src/lib/api/types/pipeline.ts:126`):

```ts
children: PipelineStep[];
```

So several call sites dereference it without a guard and throw
`TypeError: Cannot read properties of undefined`. The task list fails to
render, taking the workflows that do have steps down with it. The surrounding
page chrome (pipeline header, tab bar) still renders, so the symptom is an
empty task area rather than a blank page.

The unguarded sites on current `main`:

- `web/src/components/repo/pipeline/PipelineStepList.vue:111` — `workflow.children[0].pid`
- `web/src/components/repo/pipeline/PipelineStepList.vue:160` — `workflow.children.some(...)`
- `web/src/components/repo/pipeline/PipelineLog.vue:491` — `workflow.children.reduce(...)` inside `findStep` (declared at `:488`)
- `web/src/views/repo/pipeline/Pipeline.vue:94` and `:111` — index `children[0]` without optional chaining on the intermediate

### How a workflow ends up with zero steps

A services stage is appended unconditionally, but each service container is
`when`-filtered inside the loop (`pipeline/frontend/yaml/compiler/compiler.go:209-227`):

```go
// add services steps
if len(conf.Services.ContainerList) != 0 {
    stage := new(backend_types.Stage)

    for _, container := range conf.Services.ContainerList {
        if match, err := container.When.Match(c.metadata, false, whenEnv); !match && err == nil {
            continue
        }
        ...
        stage.Steps = append(stage.Steps, step)
    }
    config.Stages = append(config.Stages, stage)   // appended even if 0 steps survived
}
```

If every service *and* every step is filtered out, `config.Stages` is still
non-empty (it holds that empty services stage), so the emptiness check in
`pipeline/frontend/builder/builder.go:169` does not fire:

```go
if len(ir.Stages) == 0 {
    return nil, nil
}
```

The item is returned and persisted. `workflowsFromPipelineBuilder` only appends
a `model.Step` per `stage.Steps` entry (`server/pipeline/items.go:274-294`), so
`workflow.Children` stays `nil`, and `workflowsCreate` inserts the workflow row
with no step rows (`server/store/datastore/workflow.go:57`).

Reading it back, `stepListWorkflow` returns an empty slice
(`server/store/datastore/step.go:48-55`), so `omitempty` then drops the field
from the JSON.

### Steps to reproduce

1. In a repo with a single workflow file, declare a `services` block and a
   `steps` block where every container is excluded by its own `when` condition
   for the event you are about to trigger. For example:

   ```yaml
   services:
     db:
       image: postgres
       when:
         event: tag

   steps:
     build:
       image: alpine
       commands: ['echo hi']
       when:
         event: tag
   ```

2. Trigger a `push` (not a `tag`), so both the service and the step are filtered
   but the workflow itself is not.
3. Open that pipeline in the web UI.
4. The task area renders empty, with no workflows listed at all. The browser
   console shows `TypeError: Cannot read properties of undefined (reading
   'some')` at `PipelineStepList.vue` (or `'pid'` / `'reduce'`, depending on
   which site is hit first).

### Expected behavior

The pipeline view renders. A workflow with no steps should display as an empty
or skipped workflow rather than throwing, and it should not prevent sibling
workflows from rendering.

### System Info

```shell
v3.17.0
```

### Additional context

I have a patch for this and will open a PR: make `children` optional in the
frontend type to match the API, guard the four deref sites, and extract
`findStep` out of `PipelineLog.vue` so the empty case can be unit-tested.

Whether the compiler should also avoid appending an empty services stage is a
separate question — the frontend should not crash on it either way, so I have
kept the fix to the UI.

### Validations

- [x] Read the [[docs](https://woodpecker-ci.org/docs/intro)](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [[already an issue](https://github.com/woodpecker-ci/woodpecker/issues)](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.

mattwilkinsonn:
Closes #7114

A workflow with no children serializes `children` as absent, since the field is
`json:"children,omitempty"` (`server/model/workflow.go:32`). The frontend type
declares it as a non-optional `PipelineStep[]`
(`web/src/lib/api/types/pipeline.ts:126`), so several call sites dereference it
without a guard and throw "Cannot read properties of undefined" when such a
workflow is present. The task list then fails to render, taking the workflows
that do have steps down with it.

The unguarded sites are `workflow.children[0].pid` and
`workflow.children.some(...)` in `PipelineStepList.vue`, and
`workflow.children.reduce(...)` inside the `findStep` helper in
`PipelineLog.vue`. `Pipeline.vue` also indexes `workflows[0].children[0]`
without optional chaining on the intermediate.

This makes `children` optional in the type to match the API and guards each
deref. `findStep` moves out of `PipelineLog.vue` into `web/src/lib/findStep.ts`
so it can be tested directly, alongside the other pipeline helpers in
`web/src/lib/pipeline.ts`.

Tests are in `web/src/lib/findStep.test.ts` and
`web/src/components/repo/pipeline/PipelineStepList.test.ts`. They cover the
absent-key shape that `omitempty` actually produces, an explicit null, an empty
array, a stepless workflow before and after the target one, and a pipeline
where every workflow is stepless. Reverting either guard turns them red.

## UI

sidebyside

Before, the task area is empty and neither workflow renders, including `build`,
which has two steps. The console throws at `PipelineStepList.vue`. After, the
stepless `services` workflow renders as a collapsed row and `build` renders
below it, with a clean console.

Captured against a stub API serving the absent-`children` shape, so the exact
payload is pinned; the fixture and steps are reproducible without a server.

The issue describes how a workflow reaches zero steps (an all-filtered
`services` block leaves an empty stage, so the `len(ir.Stages) == 0` check does
not fire). Whether the compiler should also avoid appending an empty services
stage is a separate question; the frontend should not crash on it either way,
so this change is limited to the UI.


qwerty287:
Thanks, how does this look in the UI then? Can you post a screenshot?

mattwilkinsonn:
> Thanks, how does this look in the UI then? Can you post a screenshot?

Added a before/after, the before crashes the StepList entirely and doesn't render, the after renders the stepless workflow and any real workflows along with it.

6543:
fix itselfe looks resonable, the lib extraction i dont know ... as it only has one consumer ... can ot be reused elsewhere in our frontend code already? else i would stay with inline 🤔

qwerty287 (approved):

Issue #5970, opened 2026-01-14
pipeline definition not found on re-runs with config extension


wolffberg:
### Component

server

### Describe the bug

_We do not think this issue directly relates to the other `pipeline definition not found` issues._

When you run a pipeline with the config extension that fails it will always result in a `pipeline definition not found` if the pipeline is re-run, even if the problem is fixed.

We suspect that the pipeline config is not written to the database on failures which prevents any fixes from being applied.

If you trigger a new pipeline run with the fixes in a new commit it succeeds.

### Steps to reproduce

> **Results in failure**
> 1. Run a pipeline with the config extension that results in an error
> a. E.g. Reference a secret that doesn't exist
> 3. Fix the issue
> 4. Re-run the failed pipeline

> **Succeeds**
> 1. Run a pipeline with the config extension that results in an error
> a. E.g. Reference a secret that doesn't exist
> 3. Fix the issue
> 4. Trigger a new pipeline run

### Expected behavior

The pipeline should pick up the fixes and succeed.

### System Info

```shell
{"source":"https://github.com/woodpecker-ci/woodpecker","version":"3.12.0"}
```

### Validations

- [x] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [x] Check that there isn't [already an issue](https://github.com/woodpecker-ci/woodpecker/issues) that reports the same bug to avoid creating a duplicate.
- [x] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions]

qwerty287:
No, internally this is the same issue as https://github.com/woodpecker-ci/woodpecker/issues/2982. I'll close this as duplicate.

PR #7120, opened 2026-09-07
Restart a pipeline that errored before its config was persisted

### LINKED ISSUE #7115: Restart still fails with "pipeline definition not found" when the original pipeline errored before its config was persisted
### Component

server

### Describe the bug

Restarting a pipeline that errored before any config row was written still fails
with `pipeline definition not found`, even when the config is now valid and a
refetch would return it.

This is the residual case behind #2982, and #5970 (closed as a duplicate of it).
Both are closed, so this is a new report for the part that survives.

`Restart` reads the *old* pipeline's persisted configs, then refetches through
the config service, but the emptiness check is against the old rows rather than
the refetch result (`server/pipeline/restart.go`):

```go
configs, err := store.ConfigsForPipeline(lastPipeline.ID)   // :45  old rows
// ...
pipelineFiles, err := forge.Fetch(...)                       // :58  refetch
// ...
if len(configs) == 0 {                                       // :77  guards on the OLD rows
    return nil, &ErrNotFound{Msg: "pipeline definition not found"}
}
```

So a pipeline whose config was never persisted can never be restarted, however
valid the current definition is. A fresh commit works, which is what makes this
look like a config-extension problem rather than a restart problem.

### Why #6857 does not cover it

#6857 moved config persistence in `Create` ahead of `createPipelineItems`, so a
*parse* failure no longer leaves a pipeline without configs. Two paths still
reach an error-state pipeline with zero config rows:

- `server/pipeline/create.go:93` — the `configFetchErr` return sits above the
  persistence block at `:96-112`, so a fetch failure persists nothing.
- `updatePipelineWithErr` still stores the pipeline in error state, so the row
  exists and the UI offers a restart on it.

## Steps to reproduce

1. Make the config unfetchable for a pipeline: either reference a secret the
   config extension needs and has not got, or make the config service fail.
2. Trigger a pipeline. It ends in error state with no config rows persisted.
3. Fix the cause, so the definition now fetches cleanly.
4. Restart the errored pipeline.
5. It fails with `pipeline definition not found`. Triggering a new pipeline on a
   fresh commit succeeds.

### Steps to reproduce

1. Make the config unfetchable for a pipeline: either reference a secret the
   config extension needs and has not got, or make the config service fail.
2. Trigger a pipeline. It ends in error state with no config rows persisted.
3. Fix the cause, so the definition now fetches cleanly.
4. Restart the errored pipeline.
5. It fails with `pipeline definition not found`. Triggering a new pipeline on a
   fresh commit succeeds.

### Expected behavior

The restart uses the refetched definition and runs it. A pipeline that is
offered as restartable in the UI should be restartable.

### System Info

```shell
v3.17.0
```

### Additional context

I have a patch and will open a PR: guard on the refetched `pipelineFiles`
instead of the old rows, and persist and link the refetched config through
`findOrPersistPipelineConfig` the way `Create` does, so the restart's config of
record matches what it runs. `ConfigPersist` dedups on (repo, name, hash), so an
unchanged config reuses existing rows.

Related: #2982 (closed), #5970 (closed as duplicate of #2982), #6857, #2759,
#1678, #1496.

### Validations

mattwilkinsonn:
Closes #7115

`Restart` reads the old pipeline's persisted configs
(`server/pipeline/restart.go:45`), refetches through the config service
(`:58`), then fails with "pipeline definition not found" if the *old* rows were
empty (`:77`). A pipeline that errored before any config was persisted
therefore cannot be restarted even when the refetch returns a valid
definition.

#6857 moved config persistence in `Create` ahead of `createPipelineItems`, so
a parse failure no longer leaves a pipeline without configs. The
`configFetchErr` path in `Create` still returns at `create.go:93`, above the
persistence block at `:96`, and `updatePipelineWithErr` stores the pipeline in
error state, so a restartable pipeline with no config rows is still reachable.

This guards on the refetched `pipelineFiles` rather than the old rows, and
persists and links the refetched config through `findOrPersistPipelineConfig`
the way `Create` does, so the restart's config of record matches what it runs.
`ConfigPersist` dedups on (repo, name, hash), so an unchanged config reuses the
existing rows.

The refetch itself is existing behaviour and is pinned to the parent pipeline's
commit, so a restart still builds the revision it was started for.

Tests are in `server/api/pipeline_test.go`. `TestPostPipeline` covers the
restart succeeding when the refetch yields a config, asserting the workflows
are actually created and the config is persisted and linked, and covers the
refetch coming back empty still reporting "pipeline definition not found".

On the earlier reports: #2982 and #5970 both cover this symptom and both are
closed, #5970 as a duplicate of #2982. I opened #7115 rather than reopening
either, because what remains is narrower than what those describe: not the
parse-failure route #6857 fixed, but the config-fetch route that still returns
before persistence. If you read that as already-settled ground, I am happy to
close this and take it to the issue instead.

### DIFF
--- a/server/pipeline/restart.go
+++ b/server/pipeline/restart.go
@@ -41,15 +41,16 @@ func Restart(ctx context.Context, store store.Store, lastPipeline *model.Pipelin
 		return nil, &ErrBadRequest{Msg: "cannot restart a pipeline with status blocked"}
 	}
 
-	// fetch the old pipeline config from the database
-	configs, err := store.ConfigsForPipeline(lastPipeline.ID)
+	// Seed the refetch with the old pipeline's persisted config. The forge
+	// fetcher reuses this as-is on restart; a config extension may replace it.
+	oldConfigs, err := store.ConfigsForPipeline(lastPipeline.ID)
 	if err != nil {
 		log.Error().Err(err).Msgf("failure to get pipeline config for %s", repo.FullName)
 		return nil, &ErrNotFound{Msg: fmt.Sprintf("failure to get pipeline config for %s. %s", repo.FullName, err)}
 	}
 
 	var pipelineFiles []*forge_types.FileMeta
-	for _, y := range configs {
+	for _, y := range oldConfigs {
 		pipelineFiles = append(pipelineFiles, &forge_types.FileMeta{Data: y.Data, Name: y.Name})
 	}
 
@@ -74,7 +75,12 @@ func Restart(ctx context.Context, store store.Store, lastPipeline *model.Pipelin
 		return nil, errors.New(msg)
 	}
 
-	if len(configs) == 0 {
+	// Guard on the refetched config, not the old persisted rows. A pipeline that
+	// errored before persisting any config (e.g. a transient config-extension
+	// failure) has no old rows, yet a healthy refetch here can still yield a
+	// valid definition to restart from. Only a genuinely empty result — no old
+	// config and nothing served — is "definition not found".
+	if len(pipelineFiles) == 0 {
 		newPipeline, uErr := UpdateToStatusError(store, *newPipeline, errors.New("pipeline definition not found"))
 		if uErr != nil {
 			log.Debug().Err(uErr).Msg("failure to update pipeline status")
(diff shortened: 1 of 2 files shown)


qwerty287:
Thanks, please do not write 5-line comments in actual code. This is ok in tests or as a kind of docstring but not in this case

mattwilkinsonn:
> Thanks, please do not write 5-line comments in actual code. This is ok in tests or as a kind of docstring but not in this case

Got it, went through and reduced comments in all of the PRs, and followed up on #7116 with #7126.

qwerty287:
The actual fix here is the change in ll. 77-79, right? The rest is just refactoring?

mattwilkinsonn:
> The actual fix here is the change in ll. 77-79, right? The rest is just refactoring?

No, need to persist the refetched files as well or `linkPipelineConfigs` wouldn't link anything

server/pipeline/restart.go
```diff
@@ -83,6 +85,18 @@ func Restart(ctx context.Context, store store.Store, lastPipeline *model.Pipelin
 		}
 		return newPipeline, nil
 	}
+	// Persist and link the new config, as Create does. ConfigPersist
+	// deduplicates on (repo, name, hash).
```
qwerty287, line 89: This is enough I think

```suggestion
	// Persist and link the new config, as Create does.
```

mattwilkinsonn: Done

qwerty287 (approved):
Thanks!

PR #7122, opened 2026-09-08
Use more unique sign to detect commands

### LINKED ISSUE #6691: Group logs by command in step logs UI Fail at pnpm install

qwerty287:
grafik

closes https://github.com/woodpecker-ci/woodpecker/issues/6691

use ▶ and two spaces


lafriks (approved):
LGTM but can we also hide command line from showing duplicate again when opening section?

qwerty287:
This was intended to have a simple clickable link to this line (https://github.com/woodpecker-ci/woodpecker/pull/6398#issuecomment-4278284012)

6543:
all old logs wont be colapsable anymore ... we store the woodpecker version that did run the workflow ... can we not use this and switch between legacy and new method to detect?

e.g. version == '' -> legacy ... very old logs :)
       version <= 3.18 ... also legacy

else new

UI vue code logic only ...

qwerty287:
I could do so but I personally don't think it's a big issue if all old logs are not collapsible anymore. I mean, how often do you check logs that are older than a few days... And for anything new it works. So I would actually keep it how it is and break that feature for old logs

xoxys:
If its possible to have a fallback with a couple lines of code we should do ao at least for some time. If its getting more complex I also think its not worth it.

qwerty287:
Thought about this and I don't think this would be that easy. The issue is not the logs parsing, but the version parsing I think:
- we would have compare semvers
- and next versions is even harder as there's no less/greater comparing possible there (only equal/unequal)

An easy way would be to check the first char of the first line, if it's a + use the old parsing otherwise the new. But that is not really robust I think

6543:
most people do not run next so i would ignore cases like next

qwerty287:
This seems like a very bad implementation to me. If we keep it for one or two version that would be fine for me, but not forever.

6543:
hmm ok, at least we should document this symbol pattern in our docs and esp in the plugin section so plugin devs intentionally can create colapsed lines ... e.g. like docker-buildx plugin

qwerty287:
I added some docs. Not sure if the wording is good, feel free to suggest improvements

6543 (approved):

PR #6609, opened 2026-05-18
local backend: on linux / mac start commands in own process group and kill the group on cancel

6543:
# Issue

i got workflows canceled
image

looking at what has canceled it:
image
it was the `make` that the local agent executed!!!

# Reason

`exec.CommandContext` uses the same process group as the main one by default. and make kills it.
I also discovered that sub-processes from the command where not killed, so now we create a new group and kill it as a whole.

PR #7132, opened 2026-09-12
Fix agent panic in local backend cleanup when a workflow is canceled before a step started

### LINKED ISSUE #7131: Agent panics (nil pointer) in local backend DestroyWorkflow when a workflow is canceled before a step started
### Component

agent

### Describe the bug

The local backend installs a `cmd.Cancel` hook in `newCmd()` that dereferences `cmd.Process`:

https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/backend/local/cmd_unix.go#L31-L34

`DestroyWorkflow()` invokes that hook for **every** stored step state during cleanup:

https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/backend/local/local.go#L272-L284

A step's command is stored in `stepState` *before* `cmd.Start()` is called (`command.go`, `clone.go`, `plugin.go`), and `exec.Cmd.Start()` returns early with `ctx.Err()` **without setting `cmd.Process`** when the context is already canceled. So if a workflow gets canceled after the agent picked it up but before a step actually started, the stored command has `cmd.Process == nil`, the cancel hook dereferences nil, and the cleanup panics.

Because the cleanup runs inside the `sync.OnceFunc` in `pipeline/runtime/workflow.go`, the panic is recovered and repanicked, which kills the whole agent process. With `Restart=always` systemd brings the agent back, but the running workflow is lost.

This was introduced by #6609 (first released in v3.15.0) and is still present on `main`.

`DestroyStep()` (`local.go#L250-L253`) and `cmd_win.go` (`taskkill ... cmd.Process.Pid`) have the same defect.

Production panic (trimmed), `woodpecker-agent` v3.18.1, linux/amd64, `WOODPECKER_BACKEND=local`:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1d61759]
goroutine 59 [running]:
sync.OnceFunc.func1.1.1()
panic({0x3a9a958?, 0x3de31d0?})
go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/local.newCmd.func1()
        pipeline/backend/local/cmd_unix.go:33 +0x19
go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/local.(*local).DestroyWorkflow.func1({...}, {...})
        pipeline/backend/local/local.go:279 +0x95
internal/sync.(*HashTrieMap[...]).iter(...)
sync.(*Map).Range(...)
go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/local.(*local).DestroyWorkflow(...)
        pipeline/backend/local/local.go:272 +0xb3
go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime.(*Runtime).Run.func1()
        pipeline/runtime/workflow.go:49 +0x7f
```

This is the same class of bug as #6008 / #6399 (nil deref on a canceled step), but in the cancel hook instead of `WaitStep()`.

### Steps to reproduce

How it happened in production:

1. Run woodpecker-server 3.18.1 (docker) and woodpecker-agent 3.18.1 (release tarball, linux/amd64) with `WOODPECKER_BACKEND=local` and `WOODPECKER_MAX_WORKFLOWS=1`, agent as a systemd service with `Restart=always`.
2. Set the repo option to cancel previous pipelines: `cancel_previous_pipeline_events = ["push", "pull_request"]`.
3. Open a pull request, wait for the agent to pick up a workflow, then push another commit to the PR about 3 seconds later so the server cancels the older pipeline.
4. The older pipeline is canceled while the clone step's command was created but never started (the steps are stored as `canceled` with zero start/finish timestamps).
5. The agent panics with the trace above and systemd restarts it about 5 seconds later.

Deterministic reproduction as a unit test (`pipeline/backend/local/cmd_unix_test.go`, no server involved). It panics on v3.18.1 and on current `main`:

```go
//go:build !windows

package local

import (
	"context"
	"testing"

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"

	"go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/types"
)

func TestDestroyWorkflowOfNotStartedStep(t *testing.T) {
	backend, _ := New().(*local)
	backend.tempDir = t.TempDir()

	require.NoError(t, backend.SetupWorkflow(context.Background(), &types.Config{}, "task"))

	// the workflow got canceled before the step was started
	ctx, cancel := context.WithCancel(context.Background())
	cancel()

	step := &types.Step{
		UUID:     "step-canceled-before-start",
		Name:     "canceled-before-start",
		Type:     types.StepTypeCommands,
		Image:    "sh",
		Commands: []string{"echo never executed"},
	}
	require.Error(t, backend.StartStep(ctx, step, "task"))

	// panics: the stored command has no process
	assert.NoError(t, backend.DestroyWorkflow(context.Background(), &types.Config{}, "task"))
}
```

### Expected behavior

Canceling a workflow whose step command was never started should clean the workflow up and report the workflow as canceled. It should not take the agent process down.

### System Info

```shell
woodpecker-server 3.18.1 (docker image), sqlite
woodpecker-agent   3.18.1 (official release tarball), linux/amd64
WOODPECKER_BACKEND=local
WOODPECKER_MAX_WORKFLOWS=1
agent runs as a systemd service with Restart=always
repo setting: cancel_previous_pipeline_events = ["push", "pull_request"]
```

### Additional context

Side effects we observed after the agent was killed:

- systemd restarted the agent within about 5 seconds.
- The server kept a stale `running` workflow row for the canceled pipeline.
- The temporary workspace directory of the workflow was left behind, because the panic happened before `os.RemoveAll(state.baseDir)` in `DestroyWorkflow()`.

I intend to submit a PR for this (guard the cancel hook against a nil `cmd.Process` in `cmd_unix.go` and `cmd_win.go`, plus regression tests).

### Validations

- [X] Checked that the bug isn't fixed in the `next` version already [https://woodpecker-ci.org/versions] (`git diff v3.18.1..main -- pipeline/backend/local/` is empty, the code path is unchanged on `main`)

usiegj00:
Closes #7131

# Issue

The cancel hook the local backend installs in `newCmd()` dereferences `cmd.Process`, and `DestroyWorkflow()` calls that hook for every stored step command:

https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/backend/local/local.go#L272-L284

A step command is stored in `stepState` *before* `cmd.Start()` is called, and `exec.Cmd.Start()` returns early with `ctx.Err()` **without setting `cmd.Process`** if the context is already canceled. So a workflow that is canceled after the agent picked it up but before a step actually started leaves a stored command with `cmd.Process == nil`, and the cleanup dereferences nil.

The cleanup runs inside the `sync.OnceFunc` in `pipeline/runtime/workflow.go`, so the panic is recovered and repanicked, which kills the whole agent process. We hit this in production on v3.18.1 with `WOODPECKER_BACKEND=local` when a new push to a pull request canceled the older pipeline about 3 seconds after the agent picked up one of its workflows.

This was introduced by #6609, so it affects v3.15.0 and later.

# Fix

Guard the cancel hook against a command that was never started. `DestroyStep()` and `cmd_win.go` share the same code path, so both are covered: the guard sits in `newCmd()` for unix and windows.

A command that never started has no process and no process group, so there is nothing to kill and `nil` is the correct result.

# Tests

`cmd_unix_test.go` adds regression tests for both cleanup entry points, `DestroyWorkflow()` and `DestroyStep()`. They drive the real code path: `StartStep()` with an already canceled context stores the step state and then fails to start the command, exactly as it happens when a pipeline is canceled.

Without the fix on current `main`:

```
--- FAIL: TestDestroyWorkflowOfNotStartedStep (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x102e8d0a0]
...
go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/local.newCmd.func1()
	pipeline/backend/local/cmd_unix.go:33 +0x20
go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/local.(*local).DestroyWorkflow.func1(...)
	pipeline/backend/local/local.go:279 +0x8c
sync.(*Map).Range(...)
go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/local.(*local).DestroyWorkflow(...)
	pipeline/backend/local/local.go:272 +0x98
FAIL	go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/local	0.385s
```

With the fix:

```
=== RUN   TestDestroyWorkflowOfNotStartedStep
--- PASS: TestDestroyWorkflowOfNotStartedStep (0.00s)
=== RUN   TestDestroyStepOfNotStartedStep
--- PASS: TestDestroyStepOfNotStartedStep (0.00s)
PASS
ok  	go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/local	0.398s
```

`make test-lib` passes, and `go vet` / `go build` are clean for `GOOS=linux` and `GOOS=windows`.


qwerty287:
Thanks, could you checkout https://ci.woodpecker-ci.org/repos/3780/pipeline/37630/31#L4?
Just use `WithCancelCause` instead

usiegj00:
Thanks, fixed. The test now uses `context.WithCancelCause(context.Background())` with `cancel(nil)`, the same way `pipeline/runtime/step_test.go` pre-cancels a context. Only the test file changed, the fix itself is unchanged.

`make lint` reports `0 issues` locally now. The tests still pass with the fix and still panic without it.

qwerty287 (approved):

PR #7136, opened 2026-09-13
fix: skip short multiline-secret fragments when masking logs

### LINKED ISSUE #7103: Multiline secrets cause masking of partial matches
### Component

server

### Describe the bug

We recently migrated our CI to Woodpecker and moved all our secrets over. I noticed the logs were illegible after the migration. As I investigated further, it was due to partial matches in multi-line secrets.

### Steps to reproduce

Using 3.18.0 with a multi-line secret, such as:

```json
{
"foo": [
"bar"
]
}
```

And a log line such as:

```
[{description,"Run PropEr test suites"},{vsn,"0.12.1"},{registered,[]}]
```

The output gets masked to:

```
[********description,"Run PropEr test suites"********,********vsn,"0.12.1"********,********registered,[************************
```

### Expected behavior

I believe the right approach is not to mask the logs, since it's only a partial match of the secret. The existing unit tests with multi-line logs don't cover partial matches; adding a new partial-match test case makes the behavior clear.

So, I would expect that the log line should still be:

```
[{description,"Run PropEr test suites"},{vsn,"0.12.1"},{registered,[]}]
```

I added these two regression test cases:

```diff
diff --git a/pipeline/shared/replace_secrets_test.go b/pipeline/shared/replace_secrets_test.go
index bcf40999d..deaacd20c 100644
--- a/pipeline/shared/replace_secrets_test.go
+++ b/pipeline/shared/replace_secrets_test.go
@@ -51,6 +51,16 @@ func TestNewSecretsReplacer(t *testing.T) {
                log:     "start log\ndone\nnow\nan\nmulti line secret!! ;)\nwith\ntwo\n\nnewlines",
                secrets: []string{"an\nmulti line secret!!", "two\n\nnewlines"},
                expect:  "start log\ndone\nnow\n********\n******** ;)\nwith\n********\n\n********",
+       }, {
+               name:    "secret with multiple lines with partial match",
+               log:     "start with\ntwo",
+               secrets: []string{"an\nmulti line secret!!", "two\n\nnewlines"},
+               expect:  "start with\n********",
+       }, {
+               name:    "multiline JSON secret over-masks Erlang/Elixir style logs",
+               log:     `[{description,"Run PropEr test suites"},{vsn,"0.12.1"},{registered,[]}]`,
+               secrets: []string{"{\n\"foo\":[\n\"bar\"\n]\n}"},
+               expect:  `[********description,"Run PropEr test suites"********,********vsn,"0.12.1"********,********registered,[************************`,
        }}
 
        for _, c := range tc {
```

In the first one (as mentioned), it expands the current existing test cases. I believe the right behavior should be:

```diff
diff --git a/pipeline/shared/replace_secrets_test.go b/pipeline/shared/replace_secrets_test.go
index bcf40999d..3c90ae9ce 100644
--- a/pipeline/shared/replace_secrets_test.go
+++ b/pipeline/shared/replace_secrets_test.go
@@ -51,6 +51,16 @@ func TestNewSecretsReplacer(t *testing.T) {
                log:     "start log\ndone\nnow\nan\nmulti line secret!! ;)\nwith\ntwo\n\nnewlines",
                secrets: []string{"an\nmulti line secret!!", "two\n\nnewlines"},
                expect:  "start log\ndone\nnow\n********\n******** ;)\nwith\n********\n\n********",
+       }, {
+               name:    "secret with multiple lines with partial match",
+               log:     "start with\ntwo",
+               secrets: []string{"an\nmulti line secret!!", "two\n\nnewlines"},
+               expect:  "start with\ntwo",
        }}
 
        for _, c := range tc {
```

Which currently fails with:

```
=== RUN   TestNewSecretsReplacer/secret_with_multiple_lines_with_partial_match
    replace_secrets_test.go:70:
                Error Trace:    /home/ivan/Code/Devops/ci-config-server/woodpecker/pipeline/shared/replace_secrets_test.go:70
                Error:          Not equal:
                                expected: "start with\ntwo"
                                actual  : "start with\n********"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,2 +1,2 @@
                                 start with
                                -two
                                +********
                Test:           TestNewSecretsReplacer/secret_with_multiple_lines_with_partial_match
```

### System Info

```shell
source	"https://github.com/woodpecker-ci/woodpecker"
version	"3.18.0"
```

### Additional context

This is related to https://github.com/woodpecker-ci/woodpecker/issues/388. But the fix introduced this behavior.

The potential solution is to apply the minimum length (3 chars) check **per line after splitting**. Not once for the whole value. I can open the PR to address the issue.

### Validations

vulragrag-star:
Fixes #7103

`NewSecretsReplacer` already splits multiline secrets so each line can be
masked independently (the replacer runs per log line). The 3-character
minimum was only applied to the whole secret before that split, so short
fragments such as `two` from `two\n\nnewlines`, or `{` / `]` lines from a
JSON secret, still entered the replacer and over-masked unrelated log
output.

This applies the same minimum length to each line after splitting. Real
secret material longer than three characters is still redacted; lone short
fragments no longer false-positive.

## Testing

```bash
go test ./pipeline/shared/ -count=1
```

Added regression cases for a partial multiline match and for a multiline
JSON secret against Erlang/Elixir-style log lines (the report in #7103).


pipeline/shared/replace_secrets.go
```diff
@@ -33,9 +33,11 @@ func NewSecretsReplacer(secrets []string) *strings.Replacer {
 		if len(old) <= minStringLength {
 			continue
 		}
```
qwerty287: We can remove this part now

```suggestion
```

qwerty287 (commented):
Thanks!

pipeline/shared/replace_secrets.go
vulragrag-star: Removed in 655823f6 — the per-line min-length check after the split already covers short secrets, so the outer guard was redundant.

PR #7139, opened 2026-09-13
Fix data race in SSE stream handlers when the client disconnects early

### LINKED ISSUE #6766: race detected ...
https://ci.woodpecker-ci.org/repos/3780/pipeline/35412/38

now we need to look if it's because of the test or actuall prod code issue

KR-Ravindra:
## Problem

CI on `main` fails intermittently with `race detected during execution of test` in `TestEventStreamSSEConcurrentDisconnect` (pipelines 35412 and 35413, #6766).

To answer the question in the issue: the racing *write* only exists in tests (`t.Cleanup` resets `server.Config.Services.Scheduler`; production sets the config once at startup). The racing *read* is production code shape, though: both SSE handlers start a goroutine that reads `server.Config.Services.*` and return without waiting for it, so that goroutine can outlive the request.

## Root cause

`server/api/stream.go:99-108`: `EventStreamSSE` runs `server.Config.Services.Scheduler.Subscribe(...)` in a `go func()` and returns as soon as the request context is done. Nothing orders that goroutine's read before code that runs after the handler returned (in the CI trace the goroutine had already finished, but without a happens-before edge). `LogStreamSSE` has the same shape at `stream.go:212-242` with `server.Config.Services.Logs.Tail`.

## Fix

Both handlers now resolve the service on the handler goroutine before spawning, and the deferred cleanup waits for the goroutine after cancelling its context (`Subscribe` and `Tail` return once the context is cancelled). No goroutine outlives the handler, so anything that happens after the handler returned is ordered after every access the handler made.

## How tested

New `TestEventStreamSSEDisconnectedClient` and `TestLogStreamSSEDisconnectedClient` call the handlers with an already-cancelled request and swap the service around each call (the stress tests do that once, in `t.Cleanup`).

Before, 8 of 8 runs of `go test -race -tags test -run 'DisconnectedClient$' ./server/api/` fail:

```
WARNING: DATA RACE
Write at 0x0000032d9ae0 by goroutine 9:
  go.woodpecker-ci.org/woodpecker/v3/server/api.TestEventStreamSSEDisconnectedClient()
      server/api/stream_test.go:184
Previous read at 0x0000032d9ae0 by goroutine 57:
  go.woodpecker-ci.org/woodpecker/v3/server/api.EventStreamSSE.func2()
      server/api/stream.go:100
--- FAIL: TestEventStreamSSEDisconnectedClient (0.00s)
    testing.go:1712: race detected during execution of test
```

After: the new tests pass with `-count=30`, the existing `ConcurrentDisconnect` stress tests with `-count=50`, and the whole `server/api` package, `go vet`, `gofumpt` and `golangci-lint` are clean for the changed files.

## Links

- Closes https://github.com/woodpecker-ci/woodpecker/issues/6766

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.

Re-ran on top of current main (69432b9):

1. Unpatched main plus the new tests: \`-race\` reports the read at stream.go:100 (goroutine started at :99) against the test's assignment at stream_test.go:172/186, as described.
2. This branch: \`go test -race -tags test ./server/api/\` passes, and \`-count=20\` of the new tests plus both ConcurrentDisconnect stress tests passes.
3. \`Wait()\` cannot hang: the memory pubsub \`Subscribe\` returns on \`<-c.Done()\` (pub.go:67-92) and \`logger.Tail\` selects on \`c.Done()\`/\`s.done\` (log.go:111-134); \`cancel(nil)\` runs before \`Wait()\` in both defers, and the LogStreamSSE early returns happen before \`Add(1)\`.

To the open question in #6766: the racing write is test-only (the \`t.Cleanup\` reset), but the unsynchronised read is production shape, a handler goroutine outliving the request. Marking ready; \`bug\` and \`server\` labels would fit.

server/api/stream.go
```diff
@@ -197,19 +206,27 @@ func LogStreamSSE(c *gin.Context) {
...
 
-	err = server.Config.Services.Logs.Open(ctx, step.ID)
+	err = _logs.Open(ctx, step.ID)
 	if err != nil {
 		log.Error().Err(err).Msg("log stream: open failed")
 		logWriteStringErr(io.WriteString(rw, "event: error\ndata: can't open stream\n\n"))
 		return
 	}
 
+	tailDone.Add(1)
```
6543: tailDone.Go should be prevered

https://pkg.go.dev/sync#WaitGroup.Go

KR-Ravindra:
@6543 done in 4b69b74: both new WaitGroups in `server/api/stream.go` now use `WaitGroup.Go` instead of `Add(1)`/`defer Done()` (the pre-existing `innerDone` is untouched). Rebased onto current main; `go test -race -tags test ./server/api/` and `go vet` pass locally. Could you take another look?

Drafted with AI assistance and checked against the code.

6543 (approved):

PR #7141, opened 2026-09-14
fix(server): accept host:port registry addresses so credentials can match

### LINKED ISSUE #6849: Registry credentials for `host:port` registries can never match: UI validation forces a scheme, credential matching requires the bare address

Jorge-Polanco-Roque:
Closes #6849.

Registry address validation and credential matching used two parsers that disagreed:
- **Validation** (`Registry.Validate`) used `url.Parse`, which rejects a bare `host:port` like `10.0.1.32:5000` (`first path segment in URL cannot contain colon`) — so you're forced to save `http://10.0.1.32:5000`.
- **Matching** (`MatchHostname`) compares against `reference.Domain()` of the image ref, which is always the bare `host:port` (image refs never carry a scheme). So the scheme-prefixed entry the validator forced can never match → `host:port` registries can't authenticate.

Fix (the direction suggested in the issue): validate the address through `distribution/reference` — the same library the matcher uses — requiring `reference.Domain(parsed) == address`. Now validation agrees with matching by construction: accepts `host:port`, `docker.io`, `localhost[:port]`, IPv4/IPv6; rejects schemes, paths, credentials, malformed ports, and bare single-label names. `index.docker.io` is short-circuited to mirror the matcher's existing special case.

**Tests:** `TestRegistryValidate` + `TestRegistryValidateCredentials`, plus a `host:port` matcher case; verified they fail against the old `url.Parse` validation. `go test`, `go vet`, `gofmt`/`gofumpt`, and golangci-lint (repo-pinned) are clean. Server-side only — the Vue form has no address validation beyond `required`.

Out of scope: the secondary logless-failed-step observability note in the issue (current `main` already carries step-start errors through the tracer).

---
_Disclosure: I use AI assistance for code review and drafting; I review and verify every change myself._


server/model/registry_test.go
```diff
@@ -0,0 +1,81 @@
...
+		{name: "Docker Hub legacy domain", address: "index.docker.io"},
+		{name: "domain", address: "registry.example.com"},
+		{name: "domain with port", address: "registry.example.com:5000"},
+		{name: "IPv4 with port", address: "10.0.1.32:5000"},
+		{name: "IPv6 with port", address: "[2001:db8::1]:5000"},
+		{name: "localhost", address: "localhost"},
+		{name: "localhost with port", address: "localhost:5000"},
+		{name: "empty", wantErr: errRegistryAddressInvalid},
+		{name: "single-label repository path", address: "registry", wantErr: errRegistryAddressInvalid},
+		{name: "scheme", address: "http://registry.example.com", wantErr: errRegistryAddressInvalid},
```
6543, line 39: this was valid before ... i myselfe are unsure if it could have worked for private http registryes or if it never worked at all ... anyway

the validator let it pass before and now we have to make sure that this will still be valid ->IF<- it had worked for private registyes before

PR #7146, opened 2026-09-14
fixes #6145 reload on reconnect

### LINKED ISSUE #6145: Pipeline feed spinner never stops after pipeline completes — requires F5 to clear
### Component

web-ui

### Describe the bug

Pipeline feed spinner (top right) never stops after pipeline completes. UI shows pipeline as still running when server considers it finished.

### Steps to reproduce

1. Trigger a pipeline manually or via push to https://codeberg.org
2. Watch pipeline complete with green tick on all steps
3. Observe pipeline feed icon (top right border animation) continues indefinitely
4. Press F5 — spinner clears and pipeline shows correct completed state

### Expected behavior

Pipeline feed spinner stops automatically when pipeline completes.
**Actual behaviour**
Spinner runs forever. Tapping cancel returns Bad Request: Cannot cancel a non-running or non-pending or non-blocked pipeline, confirming server knows pipeline is finished but UI does not.

### System Info

```shell
Woodpecker version: 3.13.0
Server: Docker container woodpeckerci/woodpecker-server:v3.13.0
Agent: Docker container woodpeckerci/woodpecker-agent:v3.13.0
Platform: ARM64 Raspberry Pi 5, Raspberry Pi OS Bookworm
Browser: tested on desktop and mobile
```

### Additional context

Workaround is F5 hard refresh. Occurs on every pipeline run without exception.

**Validations**
 Confirmed server and agent are same version 3.13.0
 Confirmed pipeline completes successfully (green tick on all steps)
 Confirmed cancel button returns bad request error on visually-running pipeline
 Confirmed F5 resolves the UI state

### Validations

muayyad-alsadi:
Fixes #6145 

basically reloading the pipeline feed whenever an SSE connection is re-established.


qwerty287:
Thanks!

While I think this is a good idea, I'm not sure whether the linked issue is related, cause I think there the connection was not lost.

muayyad-alsadi:
they said f5 (refresh the page) fixed it. so we can assume that it's a stall connection.
in other projects I noticed that when connection is lost, websocket events will be sent on the one end but not the other rather than throw some error.

do you suggest different approach?

PR #7150, opened 2026-09-15
Fix workspace volume mismatch when using CLI exec with the Kubernetes backend

### LINKED ISSUE #7148: Mismatch between created PVC and its reference in the job pod
### Component

cli

### Describe the bug

Executing a test pipeline through CLI and k8s backend generates mismatched resource names. For instance the below command
```
KUBECONFIG=~/.kube/myconfig woodpecker-cli exec --log-level debug --pipeline-event push --commit-branch temp --commit-sha "$(git rev-parse HEAD)" --plugins-privileged woodpeckerci/plugin-docker-buildx --backend-engine kubernetes --backend-k8s-storage-class my-rwx-sc --backend-k8s-storage-rwx=true --backend-k8s-volume-size=1Gi ./.woodpecker/build.yaml
```
gets stuck at
```
6:24AM INF go.woodpecker-ci.org/woodpecker/v3/cli/exec/exec.go:109 > auto detected workflow in multi workflow setup, parent directory is used as workspace
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/cli/exec/exec.go:116 > auto selected workspace folder at "/home/myuser/myproject"
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/cli/common/hooks.go:52 > checking for updates ...
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/cli/update/updater.go:39 > current version: 3.18.1
# build
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime/workflow.go:113 > executing 1 stages, in order of: CLI=exec
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime/workflow.go:122 > stage CLI=exec StagePos=0 Steps=test
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime/step.go:36 > prepare CLI=exec step=test
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/pipeline/runtime/step.go:54 > executing CLI=exec step=test
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/pipeline/backend/kubernetes/secrets.go:59 > Secret names were defined in backend options, but secret access is disallowed by instance configuration.
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/cli/update/updater.go:78 > no new version available
6:24AM DBG go.woodpecker-ci.org/woodpecker/v3/cli/common/hooks.go:63 > no update required
```
and is blocked from scheduling due to
```
  Warning  FailedScheduling   63s   default-scheduler   0/2 nodes are available: persistentvolumeclaim "wp-01m2hhg5xxfm4tm5kykttn7mtm-default" not found. preemption: 0/2 nodes are available: 2 Preemption is not helpful for scheduling., ,
  Normal   NotTriggerScaleUp  40s   cluster-autoscaler  pod didn't trigger scale-up (it wouldn't fit if a new node is added):
```
but a successful provision of a PVC named `wp-01m2hhg5y2gxv7115yd1n6tdhy-0-default` happens without issues.

I'm testing a simple workflow exec
```
when:
- event: push
  branch: temp

steps:
- name: test
  image: debian
  commands:
  - env
  backend_options:
    kubernetes:
      secrets:
      - name: mysecret
        key: mykey
        target:
          env: MYKEY
```

### Steps to reproduce

1. do a cli exec with k8s backend of the supplied test workflow
2. mismatched PVC names between actually-provisioned PVC and what is referenced in the pod manifest

### Expected behavior

The referenced PVC in the job pod should match the created PVC. This issue is reproducible consistently at my end.

### System Info

```shell
Helm chart: 3.7.3
WP server: 3.18.1
WP agent: 3.18.1
CLI: 3.18.1
CLI installed through homebrew on Linux arm64
```

### Additional context

_No response_

### Validations

tunglambk:
`woodpecker-cli exec` compiles the workspace volume with a CLI-generated name, while the Kubernetes backend creates the PVC for the pipeline's own workflow volume. The steps therefore referenced `wp_<cli-ulid>_default` (and, because `--local` defaults to `true`, the local repository path as another PVC) while only `wp_<builder-ulid>_0_default` existed, so the pod stayed Pending on `persistentvolumeclaim "..." not found`.

The fix now lives entirely in the Kubernetes backend. `SetupWorkflow` normalizes the steps' volume list before the workflow volume is created: the volume mounted at the step's workspace base is rewritten to the workflow volume the backend creates, and local filesystem paths mounted inside the workspace are dropped because the workspace volume already covers that path. The compiler and CLI are unchanged, so the backend handles any caller that names the workspace volume differently, regardless of which engine the caller thought it was targeting.

Volumes outside the workspace are left alone: an ordinary volume such as `cache:/cache` still references its own PVC, and only local host paths inside the workspace are dropped. A local host path mounted outside the workspace still has no Kubernetes equivalent, which is outside the scope of this fix.

Tested with:
- `go test -tags test -race ./pipeline/backend/kubernetes/...` — the new `TestWorkspaceVolumeUsesWorkflowPVC` fails before this change (the pod references `wp-cli-prefix-default` and `home-user-repo` instead of the created PVC) and passes after
- `make test-lib`, `make test-cli`, `make test-server`
- `go vet ./pipeline/backend/kubernetes/...`
- `gofmt` and `gofumpt -extra` clean

Fixes #7148


qwerty287:
Thanks, but please do not put K8s-specific code into the general files.
Would be nice to rewrite the volume name the directly in the K8s backend, is this possible?

tunglambk:
Done. The fix now lives in `pipeline/backend/kubernetes/volume.go`: `SetupWorkflow` rewrites the step volume mounted at the workspace base to the workflow volume the backend creates, and drops local host paths mounted inside the workspace because the workspace volume already covers that path. The changes to `pipeline/frontend/yaml/compiler/convert.go` and `cli/exec/exec.go` are reverted, so no Kubernetes-specific code is left in the shared files.

Added `TestWorkspaceVolumeUsesWorkflowPVC` in the Kubernetes backend package. It builds the pod via `SetupWorkflow`/`StartStep` and checks that the workspace mount references the created PVC instead of the caller-named volume, that the local repository path is not referenced as a PVC, and that `cache:/cache` and an unrelated host path keep their own references. It fails before the change and passes now.

Ran `go test -tags test -race ./pipeline/backend/kubernetes/...`, `make test-lib`, `make test-cli`, `make test-server`, `go vet ./pipeline/backend/kubernetes/...`, and `gofmt`/`gofumpt -extra`.

qwerty287:
Thanks, please shorten the comments. We don't need multi-line comments explaining everything in the code.

tunglambk:
Comments shortened in the Kubernetes backend; tests still pass.

qwerty287:
Thanks.

@xoxys could you check this out?

xoxys:
Already on my list :)

PR #7151, opened 2026-09-15
fix(pipeline): guard restart on refetched configs and persist them

### LINKED ISSUE #7115: Restart still fails with "pipeline definition not found" when the original pipeline errored before its config was persisted

Tyagiquamar:
### What does this PR do?

Restarting a pipeline that errored before any config row was persisted (e.g. during an initial fetch failure) previously failed with `pipeline definition not found`. This happened because `Restart()` checked for empty configs against the *old* database rows (`len(configs) == 0`) and linked the old `configs` slice rather than the refetched `pipelineFiles`.

This PR:
1. Guards the empty config check against the refetched `pipelineFiles` instead of the old database rows.
2. Persists and links the refetched configs using `findOrPersistPipelineConfig` (matching `Create()`), ensuring the restart's config of record matches what it executes.

Fixes #7115.

### Testing

Added `TestRestartRefetchedConfigPersisted` in `server/pipeline/create_test.go` and verified all tests pass via `go test ./server/pipeline/...`.


qwerty287:
there is already #7120

Issue #7063, opened 2026-08-24
FIFO expire resubmits a running task without restoring the backup row; Poll then drop-stale Error()s live work (3.17.0)


caoer:
On 3.17.0 with the persistent queue (`WithTaskStore`), `persistentQueue.Poll` deletes the `tasks` backup row when work is assigned (`TaskDelete`, `server/queue/persistent.go:69`). That is by design for crash-restore (running work should not be re-queued on server restart).

`fifo.resubmitExpiredPipelines` (`constant.TaskTimeout`, hard-coded `time.Minute` — `shared/constant/constant.go:41`) later moves that same ID from `running` to `pending` **without** `TaskInsert` (`fifo.go:421-431` — no store write; fifo holds no store reference at all). The next `Poll` of that ID hits `TaskDelete` → `sql: no rows in result set` → PR #6765's path (`pull queue item: %s: not found in backup, dropping stale task`, `persistent.go:75`) which then `Queue.Error`s the ID (`persistent.go:76`). The agent that still holds the containers is now desynced: it may run for many more minutes, then `queue.Done` logs `queue: task not found` / `cannot ack workflow`. The workflow is stored `killed`/`Canceled` with empty `cancel_info`.

PR #6765 assumed a missing backup row meant the workflow was already finished or canceled. That is false when the row is missing because of the first Poll plus an expire-resubmit.

### Minimal repro

1. Woodpecker server 3.17.0, sqlite, `WOODPECKER_LOG_LEVEL=warn` is enough (drop-stale and `queue: task expired` are error-level).
2. One agent, a workflow that runs >2 minutes (e.g. `sleep 600`).
3. After assign, block agent→server gRPC for >1 minute while the container keeps running (iptables, or pause the agent process).
4. Server journal: `queue: task expired` / `while waiting the queue returned an unexpected error`, then later `not found in backup, dropping stale task` if another Poll occurs.
5. Workflow ends `killed`/`Canceled`, `cancel_info` null. Agent may still be executing.

### Version matrix

- 3.17.0: affected (#6765 drop path in tree; `TaskTimeout=1m` since #4114).
- 3.18.0 / PR #7015: **does not cover** this. #7015's new terminal-workflow check sits only in the successful-delete `else` branch; the `ErrRecordNotExist` drop-stale path is untouched, and expire-while-agent-still-running has a non-terminal workflow at expire time — the row already gone from the first Poll never reaches the new check.
[...]

PR #7152, opened 2026-09-15
Fix persistent queue recovery after lease expiry

ruslan-shaydullin:
## Problem and change

Refs #7063 (expiry/requeue behavior), reported and diagnosed by @caoer.

The first persistent Poll deletes the task backup. FIFO lease expiry requeues the task in memory, but its next Poll treated the missing backup as proof that it was stale and dropped a live workflow.

Validate workflow state after every backup-delete outcome, preserving the terminal/missing-workflow filtering from #6765 and #7015. Carry the selected FIFO entry through validation and check that exact lease before returning it, so cancellation or reassignment during storage I/O cannot return the old assignment. Store calls stay outside the queue mutex.

Recovery uses a new lease; old waiters still receive `ErrTaskExpired` (#5823). Expiry still requeues only in memory: a restart while that task is pending retains the existing loss-of-recovery limitation. This does not change the keepalive flags or agent-resume protocol.

## Validation

Local validation is complete at `f075d1d4`. This PR remains Draft while the six required Woodpecker CI groups await maintainer approval: [pipeline 37729](https://ci.woodpecker-ci.org/repos/3780/pipeline/37729) is blocked and no steps have run. pre-commit.ci passed; the pending upstream jobs are not included in the successful local checks below.

Native macOS arm64, Go 1.26.3 (upstream CI uses Go 1.27):

- New test-only regressions fail on `c05de048` and pass on this branch, including real SQLite Pending/Running recovery and cancellation during polling.
- `go test -v -race -count=1 -timeout 180s -tags test ./server/queue` (final commit, including all four real SQLite test functions).
- `go test -race -count=1 -timeout 180s -tags test ./server/scheduler ./server/rpc`
- `make test-server`
- `WOODPECKER_DATABASE_DRIVER=sqlite3 WOODPECKER_DATABASE_DATASOURCE=:memory: make test-server-datastore`
- Pinned gofumpt, `make lint`, `make generate-openapi` (empty diff), `git diff --check`.

The generic Queue partial-task-plus-error compatibility test passes on both base and head. Tests also cover repeated expiry, old waiters, competing pollers, same-agent reassignment, terminal/missing workflows, store errors, dependencies/concurrency reservations, and restart behavior. Linux network interruption and PostgreSQL/MySQL runtime checks were not run locally.

AI assistance was used for implementation, tests, and independent automated review.


ruslan-shaydullin:
Thanks @6543 for approving CI. Could you rerun the failed `static` and `test` groups in [pipeline 37729](https://ci.woodpecker-ci.org/repos/3780/pipeline/37729)?

The [static failure](https://ci.woodpecker-ci.org/repos/3780/pipeline/37729/4) is an EOF from agentscan's account API request. In the [test group](https://ci.woodpecker-ci.org/repos/3780/pipeline/37729/5), golangci-lint printed `0 issues.` before timing out, and `TestLogStreamSSEConcurrentDisconnect` reported a data race in `server/api/stream_test.go:123` / `server/api/stream.go:234`. Those files are identical at base `c05de048` and head `f075d1d4`; I have not reproduced the race on base.

Queue, RPC, scheduler, and the SQLite/PostgreSQL/MySQL steps passed. I've updated the PR description with the actual CI results and kept it Draft while the failed checks remain unresolved.

PR #7153, opened 2026-09-15
Golang.1.27

6543:
adopt golang v1.27 with all it's goodies

TODOs:
 - [ ] wait for golangcli linter
 - [ ] add step to enforce `go fix ./...`


shared/optional/option_test.go
```diff
@@ -82,8 +82,9 @@ func TestExtractValue(t *testing.T) {
 	assert.Nil(t, val)
 }
 
+//go:fix inline
```
6543, line 85: rm func?

6543:
-> nix dev env wont work as tools need to be build with golang v1.27 too ... 

and blocked by our linter: https://github.com/golangci/golangci-lint/issues/6780

qwerty287:
The update itself is also part of https://github.com/woodpecker-ci/woodpecker/pull/7147

But it's blocked by https://github.com/woodpecker-ci/woodpecker/pull/7057 which is blocked till https://github.com/golangci/golangci-lint/issues/6764 is released

(See https://ci.woodpecker-ci.org/repos/3780/pipeline/37723/31#L4)

6543:
I would do golang bump manually ... as it contains many improvements and touches code files

PR #7155, opened 2026-09-16
Add AgentListWithOpts with pagination options

LaGrunge:
Every other list endpoint in the SDK takes list options since #4463; AgentList was left without them. It sends no page parameters at all, so the server answers with the first page only and caps it at 50 agents (maxPageSize), ordered by id ascending.

A caller that means "all agents" therefore silently loses the newest ones on an installation with more than 50 agent records, with no way to ask for the rest. The woodpecker autoscaler is one such caller: past that many agent records it cannot see the agents it has just created, decides their instances are dangling and tears them down before they ever connect.

Give AgentList an AgentListOptions argument like UserList has.


qwerty287:
Thanks!
I'm not sure whether we should just merge this because this is a breaking change.

A simple idea would be to make the parameter optional (with `func AgentList(opt AgentListOptions...)`) and we can then make it required when we publish the next major version

6543:
i would merge it once we have a majour release planed

qwerty287:
Well, we don't have a major release planned and it could take some months till we do one :laughing:

LaGrunge:
This is the prereq for https://github.com/woodpecker-ci/autoscaler/pull/738
Any advice appreciated

6543:
i would add a new method called the same with 'WithOpts' suffix and depricate the old one ... so both exist ... this way it is non breakning

LaGrunge:
> i would add a new method called the same with 'WithOpts' suffix and depricate the old one ... so both exist ... this way it is non breakning

Nice, will do that tomorrow

LaGrunge:
> i would add a new method called the same with 'WithOpts' suffix and depricate the old one ... so both exist ... this way it is non breakning

Done as you advised, https://github.com/woodpecker-ci/autoscaler/pull/738 updated accordingly

6543 (approved):

PR #7157, opened 2026-09-16
Prevent injection of matrix env vars into default clone step

qwerty287:
Before the fix, `container.Environment` wasn't empty anymore, thus `IsPlugin()` is `false` and the matrix env vars are injected.


pipeline/frontend/yaml/compiler/compiler.go
```diff
@@ -168,14 +168,13 @@ func (c *Compiler) Compile(conf *yaml_types.Workflow) (*backend_types.Config, er
...
 		}
-		for k, v := range c.cloneEnv {
-			container.Environment[k] = v
-		}
 		step, err := c.createProcess(container, conf, backend_types.StepTypeClone)
 		if err != nil {
 			return nil, err
 		}
 
+		maps.Copy(step.Environment, c.cloneEnv)
```
anbraten, line 176: Could you add a test for this?

anbraten (approved):

6543 (approved):

Issue #2763, opened 2023-11-06
Run pipeline: allow executing a single/specific pipeline if more than one pipeline definitions exist


pat-s:
### Clear and concise description of the problem

The "Run pipeline" button in the UI, which is triggered for `"event: manual"`, currently executes all existing pipeline definitions under `.woodpecker/`.

### Suggested solution

The button could allow invoking specific pipelines only instead of starting all existing ones.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

lukasz-zaroda:
This feature is very important for my use case. Basically, what I need to do is to have a different pipeline that would build the database for my app, and a different pipeline that would deploy the app by using this database. These steps actually don't depend on each other. The database is a derivative of another live database. The app can be deployed without building the database (in this case, the last one built will be used). The idea being that the database would be periodically built by using cron, and the app deployed only when code changes. Currently, it seems to be not possible to include these semi-dependent pipelines in a single repository. The only workaround that comes to my mind is to create a dedicated repository for the "database building" pipeline, which would work, but seems to be unnecessary. I would love to be able to handle everything related to the project, in this project's repository, and not creating utility repositories just to run other project related pipelines independently.

pat-s:
> The idea being that the database would be periodically built by using cron, and the app deployed only when code changes

If you set different `when` triggers in the individual pipeline YAMLS, this should work. It becomes more complicated if you want to trigger individual pipelines manually.

lukasz-zaroda:
[...]

Issue #1886, opened 2023-06-25
Restart only one workflow


gioele:
### Clear and concise description of the problem

Pressing the "Restart" button will restart all workflows. This can quite time-consuming. Often what one wants is just to restart one or two workflows and not wait (nor waste resources) while waiting for all other workflows to run.

### Suggested solution

It would be nice to have an additional button for "Restart this workflow".

### Additional context

This report is based on the version of Woodpecker next currently used by Codeberg CI.

### Validations

- [X] Checked that the feature isn't part of the `next` version already [https://woodpecker-ci.org/faq#which-version-of-woodpecker-should-i-use]
- [X] Read the [Contributing Guidelines](https://github.com/woodpecker-ci/woodpecker/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://woodpecker-ci.org/docs/intro).
- [X] Check that there isn't already an [issue](https://github.com/woodpecker-ci/woodpecker/issues) that request the same feature to avoid creating a duplicate.

6543:
would this skip depends_on filter etc ...?

I would argue to use normal restart and  #1731 instead

6543:
@gioele would that solve the issue :point_up:  ?

gioele:
@6543: Just by looking at the code and with no screenshots, it's hard to grasp what the UX implied by #1731 is. The way I understand it is that users will be able to manually select the tests that they want to skip, instead of selecting what they want to run again.

IMO that seems a quite convoluted way to achieve the same result.

For example, in case a test failed because of a temporary hiccup, I'd say that the desire of the user would be expressed as "run again the test that failed", not "run again all tests excluding those who have not failed".

pat-s:
I was just going to propose the same: "Rerun failed workflows" or similar. Especially when there is a large matrix going on and some builds failed for any reason (happens quite often in our use cases), it would be neat to be able to rerun just these without commenting parts of the yaml and then reverting again.

E.g. in my specific case all builds rely on some external config which can be adjusted async and then I would just need to hit "rerun failed" in WP. 

This would probably require a button which adds a filter based on workflow exist status. Then the question is, if the second+ run is grouped under the previous run index or just gets the next increment index value?

6543:
Well it is importand that no depends on checks get ignored as they can do some gatekeaping, witch could be a securety issue if we just not check souch cases

zc-devs:
While this issue and #2763 talk about workflow restarting, I would like to endorse

[...]

PR #7159, opened 2026-09-17
feat: allow manual and cron triggers to select workflows

cabezayunke:
Implements #2763 — pick which workflows a manual run or a cron job executes, instead of always running every config in the repo.

## Problem

`pipeline.Create` builds a workflow for every config the config service returns. The only way to narrow a run is the `when` block inside each config, which the repo author controls — not the person pressing "Run pipeline". Splitting unrelated work (build the database nightly, deploy the app on demand) currently means splitting the repo.

## Approach

The configs fetched from the forge are narrowed to the selection right after `configService.Fetch`, before anything is persisted. Filtering there rather than inside `PipelineBuilder.Build` means the pipeline is only ever linked to the configs it actually runs, so:

- `GET /pipelines/{n}/config` shows the real subset
- a restart replays the same selection instead of widening back out to everything

A name matches either the config path (`.woodpecker/deploy.yaml`) or the sanitized workflow name used by the UI and by `depends_on` (`deploy`). An empty list runs everything, so webhooks and existing crons are untouched.

An unknown name is rejected with a 400 listing the available names, rather than silently collapsing into an empty pipeline reported as "filtered" with no explanation.

### `depends_on` and a selection: runs standalone, not rejected

Earlier revisions of this PR also rejected a selection that left a required `depends_on` unsatisfied. [woodpecker-ci/woodpecker#2763](https://github.com/woodpecker-ci/woodpecker/issues/2763#issuecomment-3374977142) settled on a different design: qwerty287 (maintainer) — *"I would say it's ok to just ignore the depends_on in case the dependencies are not selected... If you run a workflow without its hard dependencies, it's fine that it fails then."*

So a selected workflow now runs without a dependency it left out of the selection, rather than being rejected or silently dropped:

- `PipelineBuilder.IgnoreMissingDependencies` (new, nil by default, additive — every other caller gets byte-identical behavior to before) tells the builder a missing dep was left out on purpose by the trigger's selection, as opposed to pruned by a `when` filter for this run, which still drops its dependents exactly as before.
- The UI shows each workflow's `depends_on` as its checkbox description, and a non-blocking warning names any selected workflow whose dependency isn't also selected — informational, not a gate. No more auto-ticking/auto-unticking dependencies for the user.

## Surfaces

- `PipelineOptions.workflows` on the manual trigger, `workflows` column on `crons` (migration `031`)
- `GET /repos/{repo_id}/workflows?branch=` so the UI can list what is selectable on a branch, each with the workflows it names in `depends_on`
- repeatable `--workflow` / `-w` on `pipeline create` and `repo cron add|update`
- checkbox list on the manual run form and the cron form, extracted into `WorkflowSelect.vue`, plus a workflow-count summary ("All workflows" / "1 workflow" / "{count} workflows") next to both forms' submit buttons so an empty selection reads as a choice, not an oversight

## Tests

- `server/pipeline/filter_test.go` — matcher table: path vs name, duplicates, blanks, unknown name
- `pipeline/frontend/builder/utils_test.go` — `IgnoreMissingDependencies` is additive: nil preserves the exact pre-existing drop behavior, a named exclusion is treated like an optional dep instead
- `server/api/pipeline_test.go` — `CreatePipeline`'s actual gin handler (not just `pipeline.Create` directly) accepts a selection with an unmet dependency
- `server/api/cron_test.go` — cron save-time validation, including error-status mapping (400 vs 500) and that a missing dependency no longer blocks the save
- `e2e/scenarios/workflow_selection_test.go` — full in-process stack with a real agent: default runs every workflow, a selection runs only its own, selection by forge path, restart replays the selection, unknown name rejected, a workflow missing its dependency runs standalone
- `web/src/components/repo/WorkflowSelect.test.ts`, `web/src/compositions/useWorkflowSummary.test.ts` — dependency description text, the non-blocking warning, and the workflow-count summary

Verified against a real local stack (server + agent + web + a local Gitea forge), not just the test suite: manual `pipeline create` with 0/1/several/unknown/dependency-missing selections, and `repo cron add|ls|update|show|rm`, all through the actual CLI against a running server, with the pipeline's persisted config table checked directly.

Several bugs turned up only from manual testing against a real local stack, not from the unit/e2e suite, all fixed here:
- a stuck-forever loading spinner on the manual-run form when `createPipeline` rejects (no `try/finally`, pre-existing, just never reachable before this PR's UI stopped guaranteeing a valid selection)
- `repo cron show|update|rm --id` always acting on cron `0` — `--id` was a `StringFlag` read back with `c.Int64("id")`, which `urfave/cli/v3` silently resolves to `0` for a flag registered under the wrong type. Pre-existing, unrelated to workflow selection, but made this PR's own `--workflow` flag on `cron update` untestable through the CLI until fixed.
- `repo cron update` could never clear a workflow selection back to "everything" — `Cron.Workflows` was tagged `omitempty`, so an intentionally empty list serialized the same as "field not sent." Added `--clear-workflows` to disambiguate.
- ticking every checkbox in the UI stored the full explicit name list rather than an empty selection, so a workflow added to the repo later silently wasn't included even though the form looked like "everything" when saved.
- a `depends_on` naming a config that doesn't exist at all (typo, or genuinely missing) was being treated the same as one left out of the selection on purpose, and tolerated instead of reported.
- two configs sanitizing to the same name (`.woodpecker/a.yml` and `.woodpecker/a.yaml` both become `a`) let a selection silently resolve to whichever the forge listed first; now rejected as ambiguous.

`go build ./...` and the full `go test ./...` (server, pipeline, cli, e2e) are green; `golangci-lint` clean on every touched package (one pre-existing `staticcheck` hit in `cli/internal/util.go` is untouched). UI `vue-tsc`, `eslint`, and `vitest` all pass. OpenAPI docs and the client mock are regenerated.

## Design confirmed on #2763

The `depends_on` question (reject a selection with an unmet dependency vs. let it run standalone) was raised in #2763. @qwerty287 confirmed on 2026-09-13: *"I would say it's ok to just ignore the depends_on in case the dependencies are not selected... If you run a workflow without its hard dependencies, it's fine that it fails then."* This PR implements that.

## Known overlap

#6963 and #6878 are both still open and both touch `server/api/pipeline.go` and `RepoManualPipeline.vue` (and #6963 also `cli/pipeline/create.go`). #6878's migration is currently `030_add_parameters_table.go`, already stale against `main`'s merged `030_deduplicate_log_entries.go` — it will likely renumber to `031` on its next rebase, which would collide with this PR's `031_add_cron_workflows.go`. Happy to rebase on whichever lands first.

#1886 (restart a single workflow) would be a small follow-up on top of this.


cabezayunke:
UI examples:

Screenshot 2026-09-15 at 10 47 12
Screenshot 2026-09-15 at 10 47 33
Screenshot 2026-09-15 at 10 47 41

CLI scenarios tested:
```

# 1. default — no --workflow flags, runs everything (regression check)
go run ./cmd/cli pipeline create --branch main <repo>

# 2. select a single workflow
go run ./cmd/cli pipeline create --branch main --workflow build <repo>

# 3. select several — repeat the flag
go run ./cmd/cli pipeline create --branch main -w build -w test <repo>

# 4. typo'd name — should 400, listing the real names
go run ./cmd/cli pipeline create --branch main --workflow nope <repo>

# 5. a workflow that depends on one you didn't select — should now succeed
#    and run standalone (this is the behavior we just changed)
go run ./cmd/cli pipeline create --branch main --workflow deploy <repo>

Cron, same idea:

# create a cron with a selection
go run ./cmd/cli repo cron add --repository <repo> --name nightly \
  --schedule "@every 5m" --workflow build

# list crons, confirm the selection stuck
go run ./cmd/cli repo cron ls <repo>

# update an existing cron's selection (needs the cron id from `ls`)
go run ./cmd/cli repo cron update --repository <repo> --id <cron-id> --workflow build --workflow test

# trigger it manually to see it run now instead of waiting for the schedule
go run ./cmd/cli repo cron run <repo> <cron-id>
```

qwerty287:
Thanks!
Please be patient till we can review this. Our review capabilities are quite limited…

woodpecker-go/woodpecker/types.go
```diff
@@ -266,12 +266,27 @@ type (
 		Created   int64  `json:"created"`
 		Branch    string `json:"branch"`
 		Enabled   bool   `json:"enabled"`
+		// Workflows optionally narrows the run to the named workflows.
```
6543, line 269: claude likes to write books ... pleas dont, if it is not fully clear a better var name is better and it is ok to have one line of comment if it is not obvious but default is none ... applys to the rest of this pull ...

@cabezayunke

6543:
also at least for cron jobs, we have an explizite `when.cron` filter why do we now have to store that info on the server with a different filter again -> I'm against it

PR #7160, opened 2026-09-17
docs: document manual Linux binary installation

joseph0531:
## Summary

Hi! This is my first contribution to Woodpecker. I noticed that the package installation page points readers to binary tarballs but doesn't explain how to set them up.

This PR adds a Linux example covering extraction, environment files, systemd services, and basic checks after startup. I kept the existing package installation instructions in place.

Related to #1292.

## Checks

- Prettier check passed for `30-packages.md`.
- `git diff --check` passed.

I haven't run the Linux service commands locally because I'm working on macOS. Feedback is welcome, especially on the installation steps.


qwerty287 (approved):
Thanks, looks good to me

PR #7162, opened 2026-09-18
Disable `cmd.exe` AutoRun commands for local pipelines on Windows

confusedsushi:
Pass the `/D` flag when invoking `cmd.exe` for local pipeline commands.

According to the `cmd.exe` help:

> If /D was NOT specified on the command line, then when CMD.EXE starts, it  
> looks for the following REG_SZ/REG_EXPAND_SZ registry variables, and if  
> either or both are present, they are executed first.
>
>     HKEY_LOCAL_MACHINE\Software\Microsoft\Command Processor\AutoRun
>
> and/or
>
>     HKEY_CURRENT_USER\Software\Microsoft\Command Processor\AutoRun

Commands configured in these `AutoRun` registry entries can modify the shell environment or otherwise interfere with pipeline execution before the generated script is run.

Passing `/D` disables this behavior and makes local Windows pipeline execution independent of machine- or user-specific `cmd.exe` AutoRun configuration.


6543 (approved):

6543:
@confusedsushi the tests will fail, because the generated script is now different just append a commit addressing this. I started the ci so you can see

PR #7165, opened 2026-09-19
Allow canceling individual workflows

### LINKED ISSUE #7156: Cancel Single Workflow

Hey everybody! 😄 

**Massive** Woodpecker fans here! We migrated from CircleCI and are all in on Woodpecker!! 🐦 

One thing that is really handy in Circle is the ability to cancel a single workflow in a pipeline

Especially if a bigger PR has **all** workflows working except for one - the one we are testing - and we have a max 6 concurrent workflow, 2 per AWS autoscaled instance with a max of 3 AWS instances concurrent

We are kinda sitting here waiting every time for the first 6 workflows to clear so we can get to test workflow #9

Is this a nice idea? Being able to cancel **not** the entire pipeline but just a **single** workflow?? 🙌 

### Suggested solution

A cancel button on the side of each workflow in the left-hand sidebar that allows you to cancel a single workflow! 🤩 

### Alternative

_No response_

### Additional context

_No response_

### Validations

AhmadNajiKam:
## Summary

* Add a cancel button beside pending and running workflows in the pipeline sidebar.
* Add a push-protected API endpoint for canceling one workflow without stopping the rest of the pipeline.
* Support cancellation of queued and running workflows while preserving dependency conditions.
* Prevent cancellation/start/finish races with conditional workflow updates and lifecycle locking.
* Handle cancellation before queue insertion by marking the workflow terminal when its task is not yet scheduled.
* Restore dependency states correctly after server restarts.
* Add API, pipeline, queue, datastore, frontend, and end-to-end test coverage.
* Document individual workflow cancellation and update the OpenAPI specification.

## Behavior

* Pending workflows become **canceled**.
* Running workflows receive a stop signal and become **killed** after cleanup.
* Independent workflows continue running normally.
* Dependent workflows follow their configured success/failure conditions.
* Only users with **push permission** can cancel workflows.
* Blocked or completed workflows cannot be canceled individually.

## Testing

```bash
go test ./server/pipeline ./server/queue
git diff --check
```

> [!NOTE]
> `go test ./server/api` is currently blocked by unrelated missing test helpers in `server/api/repo_test.go`.

Closes #7156


server/rpc/rpc.go
```diff
@@ -372,6 +387,7 @@ func (s *RPC) Done(c context.Context, strWorkflowID string, state rpc.WorkflowSt
 
 	if workflow, err = pipeline.UpdateWorkflowStatusToDone(s.store, *workflow, state); err != nil {
 		logger.Error().Err(err).Msgf("pipeline.UpdateWorkflowStatusToDone: cannot update workflow state: %s", err)
+		return err
```
6543, line 390: why?

server/pipeline/lock.go
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pipeline
+
+import "sync"
+
+// Serialize lifecycle transitions with cancellation, including parent aggregation.
+// A bounded lock table avoids retaining one lock per historical pipeline.
+var lifecycleLocks [256]sync.Mutex
```
6543, line 21: we wana get rid of gobal vars esp when they do lock mechansisms!!!

at some pint a server should be multi instance possible

this is cheep but will make more work in the future

6543: PS: only in sqlite case this would be fine

AhmadNajiKam: Indeed, the design needs improvement for the multiple servers case, I just read about distributed locks and thought about moving that responsibility to the database and making an atomic transaction to handle the switch, I think there are some caveats to that but if we ensured that no workflow can move to running state if it was previously cancelled that will be better, what do you think?

AhmadNajiKam: Also, do you think the feature is important? some people would argue that a pipeline should be a straightforward all or nothing, should I make it a priority rn?

